PR #127, opened 2021-01-11
Integration/0.2.8

ProofOfKeags:


appmgr/src/version/v0_2_8.rs
```diff
@@ -0,0 +1,36 @@
...
+                .arg("no")
+                .arg(&app_id)
+                .invoke("Docker")
+                .await?;
+        }
+        Ok(())
+    }
+    async fn down(&self) -> Result<(), Error> {
+        Ok(())
+    }
```
ProofOfKeags, line 35: is this down required by the structure of the file? Because this down isn't legit, but we also don't support downs in general right now.

ui/src/app/app-routing.module.ts
```diff
@@ -33,6 +33,11 @@ const routes: Routes = [
     canActivateChild: [AuthGuard],
     loadChildren: () => import('./pages/apps-routes/apps-routing.module').then(m => m.AppsRoutingModule),
   },
+  // {
+  //   path: 'drives',
+  //   canActivate: [AuthGuard],
+  //   loadChildren: () => import('./pages/server-routes/external-drives/external-drives.module').then( m => m.ExternalDrivesPageModule),
+  // },
```
ProofOfKeags, line 40: remove?

ui/src/app/app.component.ts
```diff
@@ -47,6 +49,11 @@ export class AppComponent {
       url: '/notifications',
       icon: 'notifications-outline',
     },
+    // {
+    //   title: 'Backup drives',
+    //   url: '/drives',
+    //   icon: 'albums-outline',
+    // },
```
ProofOfKeags, line 56: remove?

appmgr/src/version/v0_2_8.rs
dr-bonez: the down is valid. the up corrects a mistake that was made in 0.2.7, but if you were to revert to 0.2.7 you wouldn't want to undo it.

dr-bonez (approved):

PR #134, opened 2021-01-12
agent: fixes manifest parsing

afg419:


agent/src/Lib/External/AppManifest.hs
```diff
@@ -49,55 +47,49 @@ instance FromJSON AssetMapping where
...
+                    , appManifestMount :: FilePath
+                    , appManifestAssets :: [AssetMapping]
+                    , appManifestOnionVersion :: OnionVersion
+                    , appManifestDependencies :: HM.HashMap AppId VersionRange
+                    } -> AppManifest
 
-hasUi :: forall n. AppManifest n -> Bool
-hasUi AppManifestV0 {..} = isJust $ HM.lookup 80 appManifestV0PortMapping
+hasUi :: AppManifest -> Bool
+hasUi AppManifest {..} = isJust $ HM.lookup 80 appManifestPortMapping
```
ProofOfKeags, line 66: this is redundant with `uiAvailable` delete one of them, IDC which one, they seem equivalent both in perf and semantics

PR #150, opened 2021-01-15
Agent/feature/welcome ack

afg419:


agent/src/Handler/Types/V0/Base.hs
```diff
@@ -36,6 +36,7 @@ data ServerRes = ServerRes
     , serverSsh                    :: [SshKeyFingerprint]
     , serverAlternativeRegistryUrl :: Maybe Text
     , serverSpecs                  :: SpecsRes
+    , serverWelcomeAck             :: Bool
```
ProofOfKeags, line 39: Which one is true? Maybe new type for clarity?

ProofOfKeags (approved):
Only one comment. Not strictly necessary but go for it if you agree.

PR #152, opened 2021-01-16
UI/feature/auto check

MattDHill:


ui/src/app/services/sync.notifier.ts
```diff
@@ -71,12 +81,81 @@ export class SyncNotifier {
...
     await modal.present()
   }
+
+  private async handleUpdateCheck (server: Readonly<S9Server>) {
+    if (!server.autoCheckUpdates || this.checkedForUpdates) return
+
+    this.checkedForUpdates = true
+
+    if (server.versionLatest && this.emver.compare(server.versionInstalled, server.versionLatest) === -1) {
+      return this.presentAlertNewOS(server.versionLatest)
```
afg419, line 98: if this errors checkedForUpdates is still true

afg419: if someone decides not to upgrade the OS, will they ever see the alert for new apps should their be some?

ui/src/app/services/sync.notifier.ts
```diff
@@ -71,12 +81,81 @@ export class SyncNotifier {
...
+    } catch {
+      this.checkedForUpdates = false
+    }
+  }
+
+  private async presentAlertNewApps () {
+    const alert = await this.alertCtrl.create({
+      backdropDismiss: true,
+      header: 'Updates Available!',
+      message: 'New service updates are availbale in the Marketplace.',
```
afg419, line 115: available

ui/src/app/modals/os-welcome/os-welcome.page.html
```diff
@@ -6,27 +6,29 @@
       </ion-button>
     </ion-buttons>
     <ion-title >
-      <ion-label style="font-size: 20px;" class="ion-text-wrap">Welcome to {{ version }}</ion-label>
+      <ion-label style="font-size: 20px;" class="ion-text-wrap">Welcome to EmbassyOS {{ version }}!</ion-label>
```
afg419, line 9: Does this fit on all mobile screens? I think I tried this and it doesn't

ui/src/app/pages/server-routes/developer-routes/dev-ssh-keys/dev-ssh-keys.page.html
```diff
@@ -21,13 +21,6 @@
   </ion-item>
 
   <ion-item-group>
-    <ion-item-divider style="margin-top: 0px;">Description</ion-item-divider>
```
afg419, line 24: huh? why

ui/src/app/services/sync.notifier.ts
```diff
@@ -71,12 +81,81 @@ export class SyncNotifier {
...
+    await alert.present()
+  }
+
+  private async updateEmbassyOS (versionLatest: string) {
+    this.loader
+      .displayDuringAsync(async () => {
+        await this.apiService.updateAgent(versionLatest)
+        this.serverModel.update({ status: ServerStatus.UPDATING })
+      })
+      .catch(e => alert(e))
```
afg419, line 159: alert?

ui/src/app/modals/os-welcome/os-welcome.page.html
```diff
@@ -6,27 +6,29 @@
...
-      0.2.8 is a small but important update designed to enhance awareness around potential pitfalls of using certain services.
-      It introduces warnings for installing, uninstalling, backing up, and restoring backups of stateful services such as LND or c-lightning.
-      This release also draws a distinction between services that can be launched inside the browser and those that are designed to run in the background.
-    </p>
-  </div>
+  <h2>Highlights</h2>
+  <p>
+    0.2.8 is a small but important update designed to enhance awareness around potential pitfalls of using certain services.
+    It introduces warnings for installing, uninstalling, backing up, and restoring backups of stateful services such as LND or c-lightning.
+    0.2.8 introduces automatic checks for updates, a setting that can be enabled or disabled in your Embassy config, and it also draws a distinction between services that are designed to be launched inside the browser and those that are designed to run in the background.
```
afg419, line 19: You're not giving UI launching the attention it deserves in these notes. For most users, it will be the only change they notice.

ui/src/app/services/sync.notifier.ts
```diff
@@ -1,26 +1,35 @@
 import { Injectable } from '@angular/core'
```
afg419, line 1: Just to be clear, we are making a number of assumptions in this implementation:
- We get server updates only via the sync daemon
- We will only show a welcome message at most once in the lifetime of a UI
- We will only check for updates at most once in the lifetime of a UI
- If there is an opportunity to upgrade the OS, there is no need to indicate new apps available.

Each may be true in practice, now, but none are necessarily true. If any of the above assumptions later become false, it will be difficult to remember to change things here. It will also non self-documenting. It will be difficult for a new dev to figure out on their own why this is here/how this works without learning, from somewhere else, why the above assumptions are valid.

ui/src/app/services/server-config.service.ts
```diff
@@ -52,6 +52,17 @@ export class ServerConfigService {
...
       },
+      autoCheckUpdates: {
+        spec: {
+          type: 'boolean',
+          name: 'Auto Check for Updates',
+          description: 'On launch, EmabssyOS will automatically check for updates of itself and your installed services. Updating still requires user approval and action. No updates will ever be performed automatically.',
+          default: true,
+        },
+        saveFn: (val: string) => {
+          return this.apiService.patchServerConfig('autoCheckUpdates', val).then(() => this.serverModel.update({ name: val }))
```
afg419, line 63: X `.then(() => this.serverModel.update({ name: val }))`

ui/src/app/modals/os-welcome/os-welcome.page.html
```diff
@@ -6,27 +6,29 @@
...
-      If you have LND or c-lightning installed, it is critical that you update them to the latest version. 
-      An oversight in Start9’s USB backups system has created a situation where restoring a LND or c-lightning backup can result in permanent loss of channel funds.
-      To be clear, <ion-text style="font-weight: 'bold';">DO NOT</ion-text> attempt to restore a LND or c-lightning backup until you have updated these services to the latest versions.
+      If you have LND or c-lightning installed, please update them to the latest versions. 
+      An oversight in Start9’s USB backups system has created a situation where <b>restoring</b> a LND or c-lightning backup could potentially result in permanent loss of channel funds.
+      To be clear, <ion-text style="font-weight: 'bold';">DO NOT</ion-text> attempt to <b>restore</b> a LND or c-lightning backup until you have updated to the latest versions.
     </p>
   </div>
+  
+  <ion-button style="text-align: right; margin-top:12px;" fill="outline" (click)="dismiss()">
```
afg419, line 31: Let's center or right align this button

PR #176, opened 2021-01-28
implement base ui for LAN services

afg419:


ui/src/app/pages/apps-routes/app-installed-list/app-installed-list.page.ts
```diff
@@ -99,11 +99,17 @@ export class AppInstalledListPage extends Cleanup {
...
     event.preventDefault()
     event.stopPropagation()
-    address = address.startsWith('http') ? address : `http://${address}`
-    return window.open(address, '_blank')
+
+    let uiAddress: string
+    if (this.isTor) {
+      uiAddress = torAddress.startsWith('http') ? torAddress : `http://${torAddress}`
+    } else {
+      uiAddress = `https://${id}.${this.serverModel.peek().serverId}.local`
```
afg419, line 110: This is entirely fine here, probably preferable even. The general downside to accessing the contents of a behavior subject is that you cannot guarantee a split second after accessing the value that the value is still the same. Now, our serverId is not ever going to change, so there's no issue here. One approach I use (e.g on the torAddress above) is to access and pass the data directly back from the view. This doesn't help with the above guarantee, but, it does give a "what you see is what you get" guarantee to the user, assuming the data is displayed. That is, the data is consistent with what the user is seeing, even if it is no longer consistent with the actual sota for that data.

afg419: There is a LOT of peeking going on in the reactive store lib I am currently writing for 0.3.0. I have and will need to think more about race condition failure modes like the above.

PR #182, opened 2021-02-02
remove X char from product_key

plantimals:
Greetings Start9Labs,

In my attempt to build and run embassy-os from source, I uncovered a bug in the Makefile rule for product_key. The initial like echoing "X" into the file introduces a character which does not fit the charset for a product key. I removed that line, and turned the following one-liner from an append ">>" to a redirect ">". This is my first PR with your org, so apologies if I've botched the formatting in some way.

Thanks,
Rob

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,8 +23,7 @@ buster.img:
 	mv 2020-08-20-raspios-buster-armhf-lite.img buster.img
 
 product_key:
-	echo "X\c" > product_key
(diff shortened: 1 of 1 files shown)


ProofOfKeags:
After consulting with a colleague, it appears that this was intentional. There isn't actually anything in the code that will prevent product keys with a leading "X" from working, and it helps us tell at a glance who built it from source and who may have bought one from our various distribution mechanisms.

PR #189, opened 2021-02-05
agent: adds lan address to server specs

afg419:


agent/src/Handler/V0.hs
```diff
@@ -93,6 +93,7 @@ getSpecs settings = do
     specsDisk       <- fmap show . metricDiskSize <$> getDfMetrics
     specsNetworkId  <- runM $ injectFilesystemBaseFromContext settings getStart9AgentHostname
     specsTorAddress <- runM $ injectFilesystemBaseFromContext settings getAgentHiddenServiceUrl
+    specsLanAddress <- fmap (<> ".local") <$> runM $ injectFilesystemBaseFromContext settings getStart9AgentHostname
```
ProofOfKeags, line 96:
```suggestion
    specsLanAddress <- fmap (<> ".local") . runM $ injectFilesystemBaseFromContext settings getStart9AgentHostname
```

ProofOfKeags: This is going to take someone else an unreasonable amount of time to figure out that this is fmap for the function functor.

afg419: I didn't even know, the type error without it was THERE'S ANTHER FUNCTOR HERE BRO so I added the fmap

PR #193, opened 2021-02-12
UI/feature/revert-autoconfig

afg419:


ProofOfKeags:
Seems like this ought to be closed because when we implement this it will have to be based downstream of the FE rewrite.

Issue #188, opened 2021-02-04
Add a button to go to an installed service from its marketplace listing


dr-bonez:

adamgoth:
I was able to get this added pretty easily. I'm not sure if there was a specific location or style in mind but adding "View Service Details" down near "Other Versions" if the app is installed seemed like the simplest solution:

<img width="388" alt="Screen Shot 2021-02-08 at 9 30 19 PM" src="https://user-images.githubusercontent.com/5225766/107316892-d187c600-6a56-11eb-8845-4c2f04edb33c.png">

Below is another solution I explored but doesn't feel as clean:

<img width="462" alt="Screen Shot 2021-02-08 at 8 56 22 PM" src="https://user-images.githubusercontent.com/5225766/107316721-7c4bb480-6a56-11eb-9370-689a3789907a.png">

Let me know what you all think and I will put up a PR @dr-bonez @afg419 @MattDHill

ProofOfKeags:
I like the presentation of the former better but the latter placement is definitely more convenient. The trouble is going to be around whether you can compress the second solution into a space where the action is still clear.

adamgoth:
A couple other placement options/variations if you think it's worth the convenience of having it at the top of the view rather than the bottom:
1. <img width="394" alt="Screen Shot 2021-02-15 at 8 25 13 PM" src="https://user-images.githubusercontent.com/5225766/108014824-fedef180-6fcb-11eb-913d-375f3dc47ec9.png">
2. <img width="407" alt="Screen Shot 2021-02-15 at 7 56 07 PM" src="https://user-images.githubusercontent.com/5225766/108014832-02727880-6fcc-11eb-88c6-430ecc0b33bc.png">
3. <img width="409" alt="Screen Shot 2021-02-15 at 7 58 04 PM" src="https://user-images.githubusercontent.com/5225766/108014836-04d4d280-6fcc-11eb-9c34-775e2b01f513.png">
4. <img width="397" alt="Screen Shot 2021-02-15 at 7 58 24 PM" src="https://user-images.githubusercontent.com/5225766/108014838-07372c80-6fcc-11eb-994f-a92993a14a45.png">

"View Installed Service" may be a more clear button label than "View Service Details", any thoughts there?

dr-bonez:
@MattDHill

MattDHill:
Please use "Go to Service" and this image:

[...]

PR #204, opened 2021-02-19
ui: add go to service button on marketplace listing

adamgoth:
Adding a button for "Go to Service" from a marketplace listing.

Reference: #188 

Note: `[routerLink]` looked like the way routing over to the services page would be handled but if there's a more appropriate way, let me know.

### DIFF
--- a/ui/src/app/pages/apps-routes/app-available-show/app-available-show.page.html
+++ b/ui/src/app/pages/apps-routes/app-available-show/app-available-show.page.html
@@ -54,6 +54,9 @@
     </ion-button>
 
     <div *ngIf="vars.versionInstalled && vars.status !== 'INSTALLING' ">
+      <ion-button *ngIf="vars.versionInstalled" class="main-action-button" expand="block" fill="outline" [routerLink]="['/services', 'installed', vars.id]">
+        Go to Service
+      </ion-button>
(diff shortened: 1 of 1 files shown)


ui/src/app/pages/apps-routes/app-available-show/app-available-show.page.html
```diff
@@ -54,6 +54,9 @@ <h3>
     </ion-button>
 
     <div *ngIf="vars.versionInstalled && vars.status !== 'INSTALLING' ">
+      <ion-button *ngIf="vars.versionInstalled" class="main-action-button" expand="block" fill="outline" [routerLink]="['/services', 'installed', vars.id]">
```
MattDHill, line 57: This conditional is redundant. `vars.versionInstalled` already satisfied in outer scope.

MattDHill, line 57: This conditional is redundant. vars.versionInstalled already satisfied in outer scope.

MattDHill (changes requested):

ui/src/app/pages/apps-routes/app-available-show/app-available-show.page.html
adamgoth: Whoops, good catch - Think I had been moving the button around the page a bit and forgot about the conditional after settling on that location.

MattDHill (approved):

PR #232, opened 2021-03-06
adds availability to installed full

ProofOfKeags:


agent/src/Handler/Apps.hs
```diff
@@ -303,6 +303,8 @@ getInstalledAppByIdLogic appId = do
                 , appInstalledFullLastBackup             = backupTime
                 , appInstalledFullTorAddress             = Nothing
                 , appInstalledFullLanAddress             = Nothing
+                , appInstalledFullTorUi                  = False
+                , appInstalledFullLanUi                  = False
```
ProofOfKeags, line 307: These needed to be added to the full response. Not sure why they weren't there before, and also not sure why it worked, but it did.

dr-bonez (approved):

agent/src/Handler/Apps.hs
```diff
@@ -335,22 +337,30 @@ getInstalledAppByIdLogic appId = do
                             (HM.lookup depId installCache $> AppStatusTmp Installing)
                                 <|> (view _1 <$> HM.lookup depId remapped)
                     pure $ dependencyInfoToDependencyRequirement (AsInstalled STrue) (base, depStatus, depInfo)
-            manifest     <- lift $ LAsync.wait manifest'
+            manifest <- (lift $ LAsync.wait manifest') >>= \case
+                Nothing -> throwError $ NotFoundE "manifest" (show appId)
+                Just x  -> pure x
```
ProofOfKeags, line 342: The only situation where this should fail is because the app isn't installed so in theory the throw case should never be invoked but I can't guarantee that structurally.

agent/src/Handler/Apps.hs
```diff
@@ -335,22 +337,30 @@ getInstalledAppByIdLogic appId = do
...
             instructions <- lift $ LAsync.wait instructions'
             backupTime   <- lift $ LAsync.wait backupTime'
-            let lanAddress = LanAddress . (".onion" `Text.replace` ".local") . unTorAddress <$> infoResTorAddress
+            let lanAddress = do
+                    addrBase <- infoResTorAddress
+                    let lanConfs = mapMaybe AppManifest.portMapEntryLan $ AppManifest.appManifestPortMapping manifest
+                    guard (not . null $ lanConfs)
+                    pure $ LanAddress . (".onion" `Text.replace` ".local") . unTorAddress $ addrBase
```
ProofOfKeags, line 349: replicated from the preview

agent/src/Handler/Apps.hs
```diff
@@ -335,22 +337,30 @@ getInstalledAppByIdLogic appId = do
...
+                                  , appInstalledFullLanUi                  = AppManifest.lanUiAvailable manifest
                                   , appInstalledFullConfiguredRequirements = HM.elems requirements
-                                  , appInstalledFullUninstallAlert = manifest >>= AppManifest.appManifestUninstallAlert
-                                  , appInstalledFullRestoreAlert = manifest >>= AppManifest.appManifestRestoreAlert
-                                  , appInstalledFullStartAlert = manifest >>= AppManifest.appManifestStartAlert
-                                  , appInstalledFullActions = fromMaybe [] $ AppManifest.appManifestActions <$> manifest
+                                  , appInstalledFullUninstallAlert = AppManifest.appManifestUninstallAlert manifest
+                                  , appInstalledFullRestoreAlert = AppManifest.appManifestRestoreAlert manifest
+                                  , appInstalledFullStartAlert             = AppManifest.appManifestStartAlert manifest
+                                  , appInstalledFullActions                = AppManifest.appManifestActions manifest
```
ProofOfKeags, line 363: Simplify with the guaranteed manifest type, add availability.

PR #245, opened 2021-03-11
catches if either file doesn't exist and runs the sync if so

ProofOfKeags:


agent/src/Lib/Synchronizers.hs
```diff
@@ -602,9 +602,17 @@ syncDropCertificateUniqueness = SyncOp "Eliminate OpenSSL unique_subject=yes" ch
...
+                liftIO
+                $       (fmap Just . BS.readFile . toS $ (rootCaDirectory <> "index.txt.attr") `relativeTo` base)
+                `catch` \(e :: IOException) -> if isDoesNotExistError e then pure Nothing else throwIO e
+            contentsInt <-
+                liftIO
+                $ (fmap Just . BS.readFile . toS $ (intermediateCaDirectory <> "index.txt.attr") `relativeTo` base)
+                `catch` \(e :: IOException) -> if isDoesNotExistError e then pure Nothing else throwIO e
+            case (contentsRoot, contentsInt) of
+                (Just root, Just int) -> pure $ uni /= root || uni /= int
+                _                     -> pure True
```
ProofOfKeags, line 615: this indicates it to run the sync if either file doesn't exist.

elvece (approved):

PR #258, opened 2021-03-16
Appmgr/bugfix/nginx file too large migrations

chrisguida:
Refresh nginx confs when we upgrade appmgr


elvece (approved):

chrisguida:
Fixed by https://github.com/Start9Labs/embassy-os/pull/256 and https://github.com/Start9Labs/embassy-os/pull/258

PR #305, opened 2021-04-01
enable markdown in config property descriptions and warnings

chrisguida:
Fixes https://github.com/Start9Labs/embassy-os/issues/51

### DIFF
--- a/ui/src/app/components/config-header/config-header.component.html
+++ b/ui/src/app/components/config-header/config-header.component.html
@@ -3,22 +3,28 @@
   <ng-container *ngIf="error">
     <ion-item>
       <ion-icon slot="start" name="warning-outline" color="danger" size="small"></ion-icon>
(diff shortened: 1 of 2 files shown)


ui/src/app/components/config-header/config-header.component.html
```diff
@@ -11,14 +11,14 @@
   <ion-item *ngIf="spec.description">
     <ion-label class="ion-text-wrap">
       <p><ion-text color="dark">Description</ion-text></p>
-      <p>{{ spec.description }}</p>
+      <p *ngIf="spec.description" [innerHTML]="spec.description | markdown"></p>
```
MattDHill, line 14: These conditional's already exists on the outer ion-item elements.

MattDHill (changes requested):

MattDHill (approved):

PR #314, opened 2021-04-26
Add proper cabal support

hasufell:
This was created via `stack2cabal -p now`: https://github.com/hasufell/stack2cabal

`.cabal` files should always be commited to the repository. These are the only real build files as per the cabal spec and `package.yaml` is merely a 3rd party tool that is not supported by cabal.


ProofOfKeags (approved):

ProofOfKeags:
Thanks for the submission!

PR #344, opened 2021-06-23
Drew/updates

DrewAns:


ui/src/app/pages/apps-routes/release-notes/release-notes.module.ts
```diff
@@ -0,0 +1,26 @@
+import { NgModule } from '@angular/core'
```
MattDHill, line 1: You don't need this file at all

ui/src/app/pages/apps-routes/release-notes/release-notes.page.ts
```diff
@@ -0,0 +1,53 @@
...
+
+@Component({
+  selector: 'release-notes',
+  templateUrl: './release-notes.page.html',
+  styleUrls: ['./release-notes.page.scss'],
+})
+export class ReleaseNotes {
+  error = ''
+  pkgId: string
+  releaseNotes: { [version: string]: string}
```
MattDHill, line 14: Probably better to just read the release notes right from the release notes model. Why even have the local variable. Then, getReleaseNotes() could just set the value in the model.

MattDHill (changes requested):

PR #381, opened 2021-07-26
fixes post process build failure step

### LINKED ISSUE #370: Build Guide: Errors at step "Building the .img"
Hey,

I was following your guide as a noob with help from a friend (your guide really is not that noob-friendly btw)
https://github.com/Start9Labs/embassy-os/blob/master/BuildGuide.md
and pretty much everything went well until to the point where I was building the image ( Step 7 ii ). It did start the process, but then somewhere in the process the error shown in the image appeared. No idea how to go on from here.
end of make

Thanks for the help in advance,

ProofOfKeags:
fixes #370 

Turns out there was a breaking change in that package that was not remedied until 4.x. However we can't upgrade to 4.1.2 because it requires typescript 4.1.5. We can't upgrade to typescript 4.1.5 because Angular requires it to be less than 4.1.0. This is remedied in the 0.3.0 branch where we've upgraded Angular to 11.x.

This is the source of all of the build from source errors. The reason none of us caught it is due to cached node_modules directories on each of our machines, which prevented our dev machines for upgrading the dependency. If you wish to verify this, dollars to donuts you will not be able to build the UI code if you blow up your node_modules directory and don't merge in this change.

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -322,6 +322,12 @@
           "integrity": "sha512-CkCj6giN3S+n9qrYiBTX5gystlENnRW5jZeNLHpe6aue+SrHcG5VYwujhW9s4dY31mEGsxBDrHR6oI69fTXsaQ==",
           "dev": true
         },
+        "typescript": {
+          "version": "4.0.5",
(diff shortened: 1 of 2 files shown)


chrisguida (approved):
Builds on MacOS with M1

PR #382, opened 2021-07-27
Feature/ssh

ProofOfKeags:
This PR creates the entire set of functionality needed to be able to do ssh identical to the wire protocol that was used in 0.2.x. It additionally keeps track of when the keys were created so that people know how long those keys have been around. This is added as a new row to the response.

This PR lacks the startup sequence that uses the DB to initialize the authorized keys file, however it creates a function to synchronize the DB to that location by doing a wholesale replacement of the file.

As a consequence, manually editing this file will result in non-persistent changes on 1. restart, 2. key addition, and 3. key removal.

### DIFF
--- a/appmgr/src/ssh.rs
+++ b/appmgr/src/ssh.rs
@@ -1,0 +1,49 @@
+use rpc_toolkit::command;
+
+use crate::util::display_none;
+use crate::{context::EitherContext, Error};
+
+#[derive(serde::Deserialize, serde::Serialize)]
+pub struct PubKey(
+    #[serde(serialize_with = "crate::util::serialize_display")]
+    #[serde(deserialize_with = "crate::util::deserialize_from_str")]
+    openssh_keys::PublicKey,
+);
+impl std::str::FromStr for PubKey {
+    type Err = Error;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        s.parse().map(|pk| PubKey(pk)).map_err(|e| Error {
+            source: e.into(),
+            kind: crate::ErrorKind::ParseSshKey,
+            revision: None,
+        })
+    }
+}
+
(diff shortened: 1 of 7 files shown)


appmgr/src/ssh.rs
```diff
@@ -0,0 +1,140 @@
...
+    #[arg] fingerprint: String,
+) -> Result<(), Error> {
+    let pool = &ctx.as_rpc().unwrap().secret_store;
+    // check if fingerprint is in DB
+    // if in DB, remove it from DB
+    let n = sqlx::query!("DELETE FROM ssh_keys WHERE fingerprint = ?", fingerprint)
+        .execute(pool)
+        .await?
+        .rows_affected();
+    // if not in DB, Err404
```
ProofOfKeags, line 92: consider synchronizing the db and keys file regardless of success or failure here.

appmgr/src/ssh.rs
```diff
@@ -0,0 +1,140 @@
...
+        .is_none()
+    {
+        // if no duplicates, insert into DB
+        let raw_key = format!("{}", key.0);
+        sqlx::query!(
+            "INSERT INTO ssh_keys (fingerprint, openssh_pubkey, created_at) VALUES (?, ?, datetime('now'))"
+        , fp, raw_key).execute(pool).await?;
+        // insert into live key file, for now we actually do a wholesale replacement of the keys file, for maximum
+        // consistency
+        sync_keys_from_db(pool, Path::new(SSH_AUTHORIZED_KEYS_FILE)).await?;
```
ProofOfKeags, line 75: consider synchronizing even if the key is a duplicate.

kn0wmad (commented):
Just observing

dr-bonez (changes requested):

PR #440, opened 2021-09-03
Update compatibility image

elvece:


compat/src/config/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+        .map(|r| r.check(&config, &cfgs))
+        .bcollect::<Vec<_>>();
+    match rule_check {
+        Ok(_) => {
+            // create temp config file
+            let temp = std::fs::File::create("config_temp.yaml")?;
+            // copy new config that pass rule check into temp file
+            serde_yaml::to_writer(temp, &config)?;
+            std::fs::copy("config_temp.yaml", config_path)?;
+            std::fs::remove_file("config_temp.yaml")?;
```
dr-bonez, line 37:
```suggestion
            std::fs::rename("config_temp.yaml", config_path)?;
```

compat/src/config/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+    }
+}
+
+pub fn validate_dependency_configuration(
+    name: &str,
+    config: Config,
+    rules_path: &Path,
+) -> Result<(), anyhow::Error> {
+    let rules: Vec<ConfigRuleEntry> = serde_yaml::from_reader(std::fs::File::open(rules_path)?)?;
+    let mut cfgs = LinearMap::new();
```
dr-bonez, line 55: you may need to perform a config.get here just in case the config_rules refers to its own config

compat/src/config/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+    cfgs.insert(name, Cow::Owned(config.clone()));
+
+    let rule_check = rules
+        .into_iter()
+        .map(|r| r.apply(name, &mut config, &mut cfgs))
+        .bcollect::<Vec<_>>();
+
+    match rule_check {
+        Ok(_) => Ok(config),
+        Err(e) => Err(anyhow!("rule application failure for dependency check: {}", e))
```
dr-bonez, line 84:
```suggestion
        Err(e) => Err(anyhow!("{}", e))
```
these should be raw strings as that is how we want to present them to the user.

dr-bonez (changes requested):

PR #446, opened 2021-09-07
Feature/wifi sync

ProofOfKeags:
Creates a symlink at the system required location that points to a file in the actual system data directory so it persists across restarts. It is called from embassy-init so it will only run once per startup


appmgr/src/net/wifi.rs
```diff
@@ -517,6 +517,15 @@ pub fn country_code_parse(code: &str, _matches: &ArgMatches<'_>) -> Result<Count
     )))
 }
 
+pub async fn synchronize_wpa_supplicant_conf(main_datadir: &std::path::Path) -> Result<(), Error> {
+    let source = main_datadir.join("wpa_supplicant.conf");
+    if tokio::fs::metadata(&source).await.is_err() {
+        tokio::fs::write(&source, include_str!("wpa_supplicant.conf.template")).await?;
```
dr-bonez, line 523: does this need to be interpolated? if not, why .template?

appmgr/src/net/wifi.rs
```diff
@@ -517,6 +517,15 @@ pub fn country_code_parse(code: &str, _matches: &ArgMatches<'_>) -> Result<Count
...
+    }
+    tokio::fs::symlink(&source, "/etc/wpa_supplicant.conf").await?;
```
dr-bonez, line 525: this will error if /etc/wpa_supplicant.conf already exists. add a line before to delete it if it does

dr-bonez (changes requested):

appmgr/src/net/wifi.rs
ProofOfKeags: convention? 🤷🏼‍♂️ I can remove the .template but the point is that its a base config not the full thing.

appmgr/src/net/wifi.rs
ProofOfKeags: Is that actually what we want? In theory it shouldn't ever exist prior to this.

dr-bonez: the less we rely on a very specific host configuration, the better. especially when this gets ported to being a dpkg

ProofOfKeags: I just am not sure that silently deleting and overwriting it is the correct behavior here. Seems like erroring or abdicating WiFi responsibility is the right call in that scenario.

dr-bonez (approved):

PR #464, opened 2021-09-11
implements error log reporting

ProofOfKeags:


appmgr/src/bin/embassyd.rs
```diff
@@ -245,22 +249,23 @@ fn main() {
...
+    // initializes the bootstrap logger, this will be replaced with the EmbassyLogger later
+    let filter = match matches.occurrences_of("verbosity") {
+        0 => log::LevelFilter::Error,
+        1 => log::LevelFilter::Warn,
+        2 => log::LevelFilter::Info,
+        3 => log::LevelFilter::Debug,
         _ => log::LevelFilter::Trace,
-    });
+    };
+    simple_logging::log_to_stderr(filter);
```
ProofOfKeags, line 260: This is necessary in case there are error logs that happen in the methods that are used to bootstrap the logger, at least they won't be blackholed with this. It will be replaced during the `new` function for the `EmbassyLogger`

dr-bonez (approved):

PR #469, opened 2021-09-13
fixes error log reporting

ProofOfKeags:


chrisguida:
Can confirm, the error is gone

PR #503, opened 2021-09-21
implement package.action rpc/cli command

chrisguida:


appmgr/src/action/mod.rs
```diff
@@ -160,6 +176,44 @@ impl ActionImplementation {
     }
 }
 
+#[command(about = "Executes an action", display(display_serializable))]
```
dr-bonez, line 179: a  nicer display impl would be good.

dr-bonez (approved):
LGTM

PR #549, opened 2021-09-30
Remove time from Health Check Result

elvece:


dr-bonez (changes requested):
Rename `HealthCheckResultVariant` to `HealthCheckResult`

appmgr/src/dependencies.rs
```diff
@@ -303,7 +303,7 @@ impl std::fmt::Display for DependencyError {
                     } else {
                         write!(f, ", ")?;
                     }
-                    write!(f, "{} @ {} {}", check, res.time, res.result)?;
+                    write!(f, "{} @ {}", check, res)?;
```
dr-bonez, line 306:
```suggestion
                    write!(f, "{}: {}", check, res)?;
```

dr-bonez (changes requested):

PR #584, opened 2021-10-05
Feature/dry auto configure

elvece:


appmgr/src/dependencies.rs
```diff
@@ -274,6 +278,151 @@ impl DependencyConfig {
...
+        .auto_configure
+        .sandboxed(
+            &ctx,
+            &pkg_id,
+            &pkg_manifest.version,
+            &pkg_manifest.volumes,
+            Some(dep_manifest.config),
+        )
+        .await?
+        .map_err(|e| Error::new(anyhow!("{}", e.1), crate::ErrorKind::AutoConfigure));
```
dr-bonez, line 350: can probably delegate this to the dry function

appmgr/src/lib.rs
```diff
@@ -89,6 +89,7 @@ pub fn server() -> Result<(), RpcError> {
     control::stop,
     logs::logs,
     properties::properties,
+    dependencies::configure,
```
dr-bonez, line 92: this needs another parent command to fulfill the api. as it stands it would be package.configure.dry instead of package.dependency.configure.dry

dr-bonez (changes requested):

PR #738, opened 2021-10-28
Build README update

kn0wmad:
Further detail added to buildguide based on testing.  Probably will need some more refinement, but getting close

### DIFF
--- a/build/README.md
+++ b/build/README.md
@@ -30,17 +30,17 @@
 
 2. Clone the repo, move into it, and bring in required submodules
 
-```
-git clone --recursive https://github.com/Start9Labs/embassy-os.git
-cd embassy-os
-git submodule update --init --recursive
-```
+	```
+	git clone --recursive https://github.com/Start9Labs/embassy-os.git
+	cd embassy-os
+	git submodule update --init --recursive
+	```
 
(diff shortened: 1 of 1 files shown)


BitcoinMechanic (approved):

k0gen (changes requested):
You should use `sudo` with apt, avahi apt command is incomplete, when adding user to a docker group simply use current user name... check out  #743

build/README.md
```diff
@@ -1,31 +1,31 @@
 # Building Embassy OS
 
- The commands given assume a Debian or Ubuntu-based environment.  Variables that *MUST* be changed are denoted by `{THING}`. Replace this in its entirety; in other words, there will not be any `{}`s remaining after you enter your variable.
+ The commands given assume a Debian or Ubuntu-based environment.
```
ProofOfKeags, line 3: Add the warning or attention emoji to both the beginning and end of this line. I guarantee you no one will read this because people are notoriously bad at reading things that aren't in the direct bulleted list.

Alternatively make it part of the step by step instructions to "acquire a debian/ubuntu build environment"

ProofOfKeags (approved):

PR #808, opened 2021-11-17
Feature/s9pk sideload

ProofOfKeags:
Implements installs of s9pk files from the CLI.


appmgr/src/install/mod.rs
```diff
@@ -192,6 +204,173 @@ pub async fn install(
...
+        .result?;
+        tracing::debug!("package.sideload succeeded {:?}", guid);
+
+        // hit continuation api with guid that comes back
+        let file = tokio::fs::File::open(path).await?;
+        let content_length = file.metadata().await?.len();
+        let body = Body::wrap_stream(tokio_util::io::ReaderStream::new(file));
+        let client = reqwest::Client::new();
+        let res = client
+            .post(format!("http://localhost:5960/rest/rpc/{}", guid))
```
dr-bonez, line 349: need a better solution for this. can discuss if needed

appmgr/src/bin/embassyd.rs
```diff
@@ -153,10 +155,46 @@ async fn inner_main(cfg_path: Option<&str>) -> Result<Option<Shutdown>, Error> {
...
+                                                    .lock()
+                                                    .await
+                                                    .remove(&guid)
+                                                {
+                                                    None => Response::builder()
+                                                        .status(StatusCode::NOT_FOUND)
+                                                        .body(Body::empty()),
+                                                    Some(cont) => {
+                                                        let fuck_rust_so_god_damn_much = cont.handler;
+                                                        match fuck_rust_so_god_damn_much(req).await {
```
dr-bonez, line 185:
```suggestion
                                                        match (cont.handler)(req).await {
```

dr-bonez (changes requested):

appmgr/src/install/mod.rs
ProofOfKeags: I can't go through nginx right now due to complaints about body sizes being too large. If you have any ideas for dealing with that we can remove the port reference altogether. Though I remember something about you saying you want it to work remotely.

dr-bonez (approved):

PR #872, opened 2021-11-30
finalizing-install

DrewAns:
https://github.com/Start9Labs/embassy-os/issues/861


ProofOfKeags:
why was this closed instead of merged? I'm getting reports that the problem this was supposed to fix still isn't fixed.

PR #880, opened 2021-12-01
local mocks and default back buttons

MattDHill:


ui/src/app/services/api/embassy-mock-api.service.ts
```diff
@@ -653,10 +653,10 @@ export class MockApiService extends ApiService {
               op: PatchOp.REMOVE,
               path: `/package-data/${id}/install-progress`,
             },
-            {
-              op: PatchOp.REMOVE,
-              path: `/recovered-packages/${id}`,
-            },
+            // {
```
DrewAns, line 656: why is this commented out

MattDHill: Because it will throw an error if the package does not exist in recovered-packages. This should only be uncommented when testing that specific flow. ie installing a recovered package.

PR #954, opened 2021-12-14
sigterm duration and pipe

DrewAns:
NO TICKET.

This code uses the sigterm to display message while stopping on the UI.

### DIFF
--- a/ui/src/app/pipes/unit-conversion.pipe.ts
+++ b/ui/src/app/pipes/unit-conversion.pipe.ts
@@ -1,0 +1,25 @@
+import { Pipe, PipeTransform } from '@angular/core'
+
+@Pipe({
+  name: 'durationToSeconds',
+})
+export class DurationToSecondsPipe implements PipeTransform {
+  transform (duration: string): number {
+    if (!duration) return 0
+    const unit = units.filter(u => duration.includes(u)).reduce((a, b) => a.length > b.length ? a : b )
+    const num = duration.replace(unit, '')
+    return Number(num) * unitsToSeconds[unit]
+  }
+}
+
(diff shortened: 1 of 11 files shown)


ui/src/app/pipes/unit-conversion.pipe.ts
```diff
@@ -0,0 +1,25 @@
+import { Pipe, PipeTransform } from '@angular/core'
+
+@Pipe({
+  name: 'durationToSeconds',
+})
+export class DurationToSecondsPipe implements PipeTransform {
+  transform (duration: string): number {
+    if (!duration) return 0
+    const unit = units.filter(u => duration.includes(u)).reduce((a, b) => a.length > b.length ? a : b )
```
dr-bonez, line 9: I hope I'm not the only one that thinks this is crazy. I mean yes it will work but its wild.
Normally I would just use a regex match to split off the units, and match on them.
`^([0-9]*(\.[0-9]+)?)(ns|µs|ms|s|m|d)$` - your number is group 1, and your units are group 3.

dr-bonez (changes requested):

ui/src/app/pipes/unit-conversion.pipe.ts
```diff
@@ -0,0 +1,25 @@
+import { Pipe, PipeTransform } from '@angular/core'
+
+@Pipe({
+  name: 'durationToSeconds',
+})
+export class DurationToSecondsPipe implements PipeTransform {
+  transform (duration: string): number {
```
MattDHill, line 7: Deserves a `| undefined`

dr-bonez (approved):

PR #968, opened 2021-12-16
add set-marketplace rpc

### LINKED ISSUE #950: Add ability to change Registry
At minimum, it needs to be possible to change registry for alpha/beta testing, but it's also a big thesis of the company that the ability of someone else to host a Marketplace at least be available

Reported by: @kn0wmad

dr-bonez:
closes #950

### DIFF
--- a/appmgr/src/context/rpc.rs
+++ b/appmgr/src/context/rpc.rs
@@ -8,7 +8,7 @@ use std::time::Duration;
 
 use bollard::Docker;
 use patch_db::json_ptr::JsonPointer;
-use patch_db::{PatchDb, Revision};
+use patch_db::{DbHandle, PatchDb, Revision};
 use reqwest::Url;
 use rpc_toolkit::url::Host;
 use rpc_toolkit::Context;
@@ -16,6 +16,7 @@ use serde::Deserialize;
 use sqlx::sqlite::SqliteConnectOptions;
 use sqlx::SqlitePool;
 use tokio::fs::File;
+use tokio::process::Command;
 use tokio::sync::{broadcast, oneshot, Mutex, RwLock};
 use tracing::instrument;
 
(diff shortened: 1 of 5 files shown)


kn0wmad:
Awesome, thank you!  Can you roughly document the process to set?

dr-bonez:
`embassy-cli server set-marketplace <url>`
or
`embassy-cli package set-marketplace <url>` to only change the marketplace for fetching packages, but not os updates.

dr-bonez:
note: this cli option is not very defensive. if you put in an invalid url, it can break your embassy.

chrisguida:
Can I fix it by calling it again with a valid url?

kn0wmad:
Nice that you can exclude OS updates.  When you say "break," how bad is that?  Just cannot access Marketplace?

appmgr/src/context/rpc.rs
```diff
@@ -248,6 +249,38 @@ impl RpcContext {
...
+    #[instrument(skip(self, db))]
+    pub async fn set_nginx_conf<Db: DbHandle>(&self, db: &mut Db) -> Result<(), Error> {
+        tokio::fs::write("/etc/nginx/sites-available/default", {
+            let info = crate::db::DatabaseModel::new()
+                .server_info()
+                .get(db, true)
+                .await?;
+            format!(
+                include_str!("../nginx/main-ui.conf.template"),
+                lan_hostname = info.lan_address.host_str().unwrap(),
```
Blu-J, line 261: Thanks for using a named variables for this template format.

Blu-J:
Code Looks Good to me

dr-bonez:
> When you say "break," how bad is that? Just cannot access Marketplace?

Nginx will die. Technically it won't commit the change if this happens, so if you restart the embassy it will be fine.

PR #974, opened 2021-12-18
change sounds away from mario, replace their use sites

### LINKED ISSUE #814: Change all sounds that come from mario
Come up with new sounds for the sound lib that don't risk copyright infringement of litigious organizations.

ProofOfKeags:
Fixes #814

### DIFF
--- a/appmgr/src/update/mod.rs
+++ b/appmgr/src/update/mod.rs
@@ -30,6 +30,7 @@ use crate::disk::mount::filesystem::FileSystem;
 use crate::disk::mount::guard::TmpMountGuard;
 use crate::disk::BOOT_RW_PATH;
 use crate::notifications::NotificationLevel;
+use crate::sound::{BEP, UPDATE_FAILED_1, UPDATE_FAILED_2, UPDATE_FAILED_3, UPDATE_FAILED_4};
 use crate::update::latest_information::LatestInformation;
 use crate::util::Invoke;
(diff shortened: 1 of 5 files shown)


BitcoinMechanic (changes requested):
All the pitches and times look correct. It looks like COIN is being replaced with SHUTDOWN and DEATH with CHIME which wouldn't be right? Am I reading that wrong?

ProofOfKeags:
I don't know how you'd read that. I'd look at the use sites to understand what is being used for which actions. We renamed Coin to Chime, it's the same idea though. Nintendo can't copyright a sound that short.

appmgr/src/update/mod.rs
```diff
@@ -209,7 +213,24 @@ async fn maybe_do_update(ctx: RpcContext) -> Result<Option<Arc<Revision>>, Error
...
+                    .await
+                    .expect("could not play song: update failed 2");
+                UPDATE_FAILED_3
+                    .play()
+                    .await
+                    .expect("could not play song: update failed 3");
+                UPDATE_FAILED_4
+                    .play()
+                    .await
+                    .expect("could not play song: update failed 4");
```
dr-bonez, line 233: why are we playing 4 songs here?

ProofOfKeags: the song abstraction doesn't handle multitempo well

PR #984, opened 2021-12-20
eos update bugs

DrewAns:
https://github.com/Start9Labs/embassy-os/issues/978


ui/src/app/pages/marketplace-routes/marketplace.service.ts
```diff
@@ -22,6 +22,10 @@ export class MarketplaceService {
     private readonly patch: PatchDbService,
   ) { }
 
+  get eosUpdateAvailable () {
+    return this.eos.version !== this.patch.data['server-info'].version
```
MattDHill, line 26: Might want to explicitly look for a higher version in case we ever need to yank a version from prod for some reason.

PR #996, opened 2021-12-23
Fix  remove swap

### LINKED ISSUE #982: determine if swap space is causing cpu and i/o problems
cut a branch that disables swap space by removing relevant code in [appmgr/src/disk/main.rs](https://github.com/Start9Labs/embassy-os/blob/integration/0.3.0/appmgr/src/disk/main.rs) and work with @kn0wmad and @BitcoinMechanic to determine if this resolves problems with processes getting stuck taking a high percentage of the cpu. Also look into whether this is related to IBD rolling back.

related: #901

Blu-J:
Fixes #982


Blu-J:
Proof of valid
image

dr-bonez:
Why is patch-db being updated here?

PR #1007, opened 2021-12-29
prepend /marketplace

DrewAns:
https://github.com/Start9Labs/embassy-os/issues/959


elvece (approved):
Assuming that modal handles the views for license and instructions, looks good afaict

PR #1008, opened 2021-12-29
Bugfix/metrics leak

DrewAns:
https://github.com/Start9Labs/embassy-os/issues/977


MattDHill:
Might be nice to have some kind of indicator that this page is polling. Like a small spinner whenever the loop begins that displays for a minimum of 250ms and dismisses when the new data renders. I think it might give the page a feeling a liveness that it currently lacks, especially with at least a 4s interval between requests.

PR #1015, opened 2021-12-31
feat: Wifi swapped to NMCli

Blu-J:
Why
---
So we need the ability to have error messages come the frontend, instead of the black hole. NMCli will do this instead of use watching the wpa-suplicant
Also, there are bugs no knowing the nearby wifi's and their respective signal strengths

Backend: 
* [x] Route for list of wifis
* [x] Signal Strength with NMcli
* [x] Add a connection
* [x] Save
* [x] Forget a connection
* [x] Save/ Reload Country through boot

Frontend:
* [x] Listing available wifis
* [x] Signal strength per known wifi

Test
* [x] Add interface
* [x] Remove interface
* [x] Restart Device, keep connected + remembered
* [x] Connect to Wifi Listed
* [x] Connect to Wifi other
* [ ] Testing Hidden
* [x] Other Countries

Thanks
----- 
@DrewAns for the UI work

### DIFF
--- a/appmgr/src/init.rs
+++ b/appmgr/src/init.rs
@@ -59,12 +59,26 @@ pub async fn init(cfg: &RpcContextConfig) -> Result<(), Error> {
 
     crate::ssh::sync_keys_from_db(&secret_store, "/root/.ssh/authorized_keys").await?;
     tracing::info!("Synced SSH Keys");
-
-    crate::net::wifi::synchronize_wpa_supplicant_conf(&cfg.datadir().join("main")).await?;
-    tracing::info!("Synchronized wpa_supplicant.conf");
-
     let db = cfg.db(&secret_store).await?;
+
     let mut handle = db.handle();
+
+    crate::net::wifi::synchronize_wpa_supplicant_conf(
+        &cfg.datadir().join("main"),
+        &*crate::db::DatabaseModel::new()
+            .server_info()
+            .last_wifi_region()
+            .get(&mut handle, false)
+            .await
+            .map_err(|_e| {
+                Error::new(
+                    color_eyre::eyre::eyre!("Could not find the last wifi region"),
+                    crate::ErrorKind::NotFound,
+                )
+            })?,
+    )
+    .await?;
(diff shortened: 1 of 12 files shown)


Blu-J:
image
Thanks @DrewAns for the gui for the list of wifi's

MattDHill:
This is cool, but we will still need the ability to add a network by ssid. Please make sure that is not removed.

Blu-J:
> This is cool, but we will still need the ability to add a network by ssid, even if that network is not visible or totally out of range. Please make sure that is not removed.

So, the choice was left at the bottom as other, let's you type them in manually like before.

Blu-J:
image
Okay, nice to have the signal strengths

appmgr/src/nginx/main-ui.conf.template
```diff
@@ -19,7 +21,8 @@ server {{
         gzip on;
         gzip_vary on;
         gzip_min_length 1024;
-        gzip_types text/plain text/css text/xml text/javascript application/x-javascript application/xml;
+        gzip_types text/plain text/css text/xml text/javascript application/javascript image/svg+xml font/tts font/otf font/eot font/openttype application/x-javascript application/xml;
```
Blu-J, line 24: image
The js and the xml where not compressed

appmgr/src/init.rs
```diff
@@ -59,12 +59,26 @@ pub async fn init(cfg: &RpcContextConfig) -> Result<(), Error> {
...
+    crate::net::wifi::synchronize_wpa_supplicant_conf(
+        &cfg.datadir().join("main"),
+        &*crate::db::DatabaseModel::new()
+            .server_info()
+            .last_wifi_region()
+            .get(&mut handle, false)
+            .await
+            .map_err(|_e| {
+                Error::new(
+                    color_eyre::eyre::eyre!("Could not find the last wifi region?"),
```
dr-bonez, line 75:
```suggestion
                    color_eyre::eyre::eyre!("Could not find the last wifi region"),
```

dr-bonez (approved):

PR #1017, opened 2022-01-01
feat: New get

Blu-J:


appmgr/src/net/wifi.rs
```diff
@@ -213,31 +212,31 @@ fn display_wifi_info(info: WiFiInfo, matches: &ArgMatches<'_>) {
...
-                    }))
-                })
-                .collect::<()>()
-        }
+    table_ssids.add_row(row![bc => "SSID", "STRENGTH"]);
+    for (ssid, signal_strength) in &info.ssids {
+        let mut row = row![&ssid.0, format!("{}", signal_strength.0)];
+        row.iter_mut()
+            .map(|c| {
+                c.style(Attr::ForegroundColor(match &signal_strength.0 {
```
Blu-J, line 232: Match something like
image
Also, all the ssid's would not share the same strength so the previous implementation was wrong.

appmgr/src/net/wifi.rs
```diff
@@ -175,23 +175,22 @@ pub async fn delete(#[context] ctx: RpcContext, #[arg] ssid: String) -> Result<(
 #[derive(serde::Serialize, serde::Deserialize)]
 #[serde(rename_all = "kebab-case")]
 pub struct WiFiInfo {
-    ssids: Vec<String>,
-    connected: Option<String>,
```
Blu-J, line 179: More descriptive, also pair each ssid with the strength instead of for all. The signal strength was valid for just the connected.

appmgr/src/net/wifi.rs
Blu-J: Now this will look like 
image

Blu-J: Maybe all the networks should be like the frontend?
image

PR #1030, opened 2022-01-05
adds support for no product key builds as well as chosen product key …

ProofOfKeags:
…builds

This stemmed from the observation that we need to support no product keys on the device for the 02x update recovery path. You may no build eos as so:

`make KEY=thekeyiwant`
OR
`make NO_KEY=1`

the former syntax is chosen key, the latter is no key.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ clean:
 
 eos.img: $(EMBASSY_SRC) system-images/compat/compat.tar system-images/utils/utils.tar
 	! test -f eos.img || rm eos.img
-	./build/make-image.sh
+	if [ $(NO_KEY) -eq 1 ]; then NO_KEY=1 ./build/make-image.sh; else ./build/make-image.sh; fi
 
 system-images/compat/compat.tar: $(COMPAT_SRC)
 	cd system-images/compat && ./build.sh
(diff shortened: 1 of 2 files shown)


BitcoinMechanic:
Currently this results in being asked a product key. Expected behaviour is what would normally come after entering a product key with an image that is supposed to have one - asking if you want to start fresh or recover.

ProofOfKeags:
> Currently this results in being asked a product key. Expected behaviour is what would normally come after entering a product key with an image that is supposed to have one - asking if you want to start fresh or recover.

This is outside the scope of this feature and should not be considered when evaluating the code for merging.

Makefile
```diff
@@ -35,7 +35,7 @@ clean:
 
 eos.img: $(EMBASSY_SRC) system-images/compat/compat.tar system-images/utils/utils.tar
 	! test -f eos.img || rm eos.img
-	./build/make-image.sh
+	if [ $(NO_KEY) -eq 1 ]; then NO_KEY=1 ./build/make-image.sh; else ./build/make-image.sh; fi
```
dr-bonez, line 38: what??

ProofOfKeags: what's the issue?

dr-bonez (approved):

PR #1049, opened 2022-01-10
removed broken packages entry in patchdb

ProofOfKeags:
No ticket.
This is to address an instance of non-canonical lock ordering. We realized that this value, while written to, is never read and therefore not needed.

### DIFF
--- a/appmgr/src/install/mod.rs
+++ b/appmgr/src/install/mod.rs
@@ -546,7 +546,6 @@ pub async fn install_s9pk_or_cleanup<R: AsyncRead + AsyncSeek + Unpin>(
         let mut tx = handle.begin().await?;
 
         if let Err(e) = cleanup_failed(&ctx, &mut tx, pkg_id).await {
-            let mut tx = handle.begin().await?;
             tracing::error!(
                 "Failed to clean up {}@{}: {}: Adding to broken packages",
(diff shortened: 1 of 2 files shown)


appmgr/src/install/mod.rs
```diff
@@ -546,21 +546,13 @@ pub async fn install_s9pk_or_cleanup<R: AsyncRead + AsyncSeek + Unpin>(
...
                 e
             );
             tracing::debug!("{:?}", e);
-            let mut broken = crate::db::DatabaseModel::new()
-                .broken_packages()
-                .get_mut(&mut tx)
-                .await?;
-            broken.push(pkg_id.clone());
-            broken.save(&mut tx).await?;
-            tx.commit(None).await?;
```
dr-bonez, line 563: don't delete this `tx.commit`...

dr-bonez (changes requested):

appmgr/src/install/mod.rs
ProofOfKeags: no

dr-bonez (approved):

appmgr/src/install/mod.rs
dr-bonez: yeah nvm you're right

PR #1053, opened 2022-01-11
reconfigures dependents with live pointers when package is installed …

ProofOfKeags:
…or removed


Blu-J (approved):
Paired with, code still looks good to me. 
Would be nice to double check that this actually fixed the issue, but good to go after that.

dr-bonez (approved):

PR #1060, opened 2022-01-13
POC: To ts-matches

Blu-J:


ui/src/app/util/properties.util.ts
```diff
@@ -118,65 +117,61 @@ function parsePropertiesV1Permissive (properties: unknown, errorCallback: (err:
...
-          errorCallback(new Error(`/data/${idx}${err.dataPath}: ${err.message}`))
-        }
-        return prev
+    else {
+      const error = result.error;
+      const message = Parser.validatorErrorAsString(error)
+      let dataPath = error.keys.map(x => JSON.parse(x)).join('/')
+      errorCallback(new Error(`/data/${idx}: ${message}`))
+      if (dataPath) {
+        applyOperation(cur, { op: 'replace', path: dataPath, value: undefined })
```
elvece, line 139: get clarification on what lines 133-139 are doing in the first place to ensure proper transformation

@MattDHill @DrewAns can you explain the purpose of lines 131-137 in the original file?

Blu-J: I'm not sure if the errorCallback is right at all, or the applyOperation

PR #1070, opened 2022-01-13
fix: Add conversion for old type

### LINKED ISSUE #992: Properties migration failing in some cases
After migration, filebrowser properties will not render. These errors were thrown in the console, omitting massive stack traces.
There seems to be an incompatibility between the FE types expected for these values. When filebrowser is installed regularly (not from a migrated version), the missing values are populated correctly. See screenshots for comparison. 

```
Error: unknown properties version 1, attempting to parse as v2
Error: /data/0: should NOT have additional properties
Error: /data/0: should have required property 'type'
Error: /data/0: should have required property 'masked'
Error: /data/0: should have required property 'value'
Error: /data/0: should NOT have additional properties
Error: /data/0/value: should be object
Error: /data/0: should match some schema in anyOf
```

Migrated version properties response:
Screen Shot 2021-12-21 at 11 15 39 PM

Fresh install properties response:
Screen Shot 2021-12-21 at 11 15 29 PM

We could make the FE types more lenient, but these values are probably expected for UI renderings, so the solution is probably to write a migration for Filebrowser `stats.yaml`. Please confirm @dr-bonez.

Blu-J:
fixes #992

### DIFF
--- a/ui/src/app/util/properties.util.ts
+++ b/ui/src/app/util/properties.util.ts
@@ -1,8 +1,34 @@
 import * as Ajv from 'ajv'
 import { applyOperation } from 'fast-json-patch'
 
+
+export type ValidVersion = 1 | 2
+
+function has<Obj extends {}, K extends string>(obj: Obj, key: K): obj is (Obj & { [P in K]: unknown }) {
+  return key in obj
+}
+
+
 const ajv = new Ajv({ jsonPointers: true, allErrors: true, nullable: true })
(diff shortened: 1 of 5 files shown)


ui/src/app/util/properties.util.ts
```diff
@@ -1,8 +1,34 @@
 import * as Ajv from 'ajv'
 import { applyOperation } from 'fast-json-patch'
 
+
+export type ValidVersion = 1 | 2
+
+function has<Obj extends {}, K extends string>(obj: Obj, key: K): obj is (Obj & { [P in K]: unknown }) {
```
elvece, line 7: 👍

elvece (approved):

PR #1074, opened 2022-01-14
migrate root ca from 0.2.x

ProofOfKeags:


appmgr/src/setup.rs
```diff
@@ -471,6 +471,37 @@ async fn recover_v2(ctx: SetupContext, recovery_source: TmpMountGuard) -> Result
...
+    let root_ca_cert_path = recovery_source
+        .as_ref()
+        .join("root")
+        .join("agent")
+        .join("ca")
+        .join("certs")
+        .join("embassy-root-ca.cert.pem");
+    let (root_ca_key_bytes, root_ca_cert_bytes) = tokio::try_join!(
+        tokio::fs::read(root_ca_key_path),
+        tokio::fs::read(root_ca_cert_path)
```
dr-bonez, line 490: this probably doesn't afford any speedup.

dr-bonez (changes requested):
the root ca needs to be included in the return type instead of manually importing it. the recovery flow already accounts for importing CAs and tor addresses

ProofOfKeags:
I don't understand what you mean by this. The ticket is cut on the assumption that it didn't account for this. So which is it? Is the ticket invalid? or do we need to include this change?

dr-bonez:
for 0.3.0 backup recovery, this *is* accounted for. it just isn't for 0.2.x. the 0.2.x recovery just needs to splice its tor address and cert into that existing code

dr-bonez (approved):

ProofOfKeags:
closed in favor of #1105

PR #1075, opened 2022-01-14
Fix/show errors

Blu-J:
So, the errors that happen don't show in the front end, and it should. Don't know why it was running in a spawn, but it shouldn't and now we can see the errors.

This was a quick life quality that @kn0wmad was running into and I believe that this should be showing that something didn't work.

### DIFF
--- a/appmgr/src/net/wifi.rs
+++ b/appmgr/src/net/wifi.rs
@@ -69,22 +69,21 @@ pub async fn add(
         drop(wpa_supplicant);
         Ok(())
     }
-    tokio::spawn(async move {
-        match add_procedure(
-            &mut ctx.db.handle(),
-            ctx.wifi_manager.clone(),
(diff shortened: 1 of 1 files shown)


Blu-J:
image

ProofOfKeags:
The problem is that now it is synchronous, but the issue here is that you may lose the connection while it tries to connect. The reason it was in a spawn was so that we could let the FE know we received the request before fucking with the network stack. We should go over what this attempts to solve on Monday because I'm not up to speed with the rest of your WiFi changes and maybe this is no longer needed.

Blu-J:
* [ ] Need to async this message to frontend
* [ ]

Blu-J:
image

Blu-J:
Removed the rest of the wpa to network renaming

ProofOfKeags (approved):

PR #1077, opened 2022-01-15
fixes lan address divergence across setup scenarios

ProofOfKeags:
This code is cursed. I'm sorry.

### DIFF
--- a/appmgr/src/setup.rs
+++ b/appmgr/src/setup.rs
@@ -39,7 +39,7 @@ use crate::sound::BEETHOVEN;
 use crate::util::io::{dir_size, from_yaml_async_reader};
 use crate::util::Version;
 use crate::volume::{data_dir, VolumeId};
-use crate::{ensure_code, Error, ResultExt};
+use crate::{ensure_code, Error, ErrorKind, ResultExt};
 
 #[instrument(skip(secrets))]
 pub async fn password_hash<Ex>(secrets: &mut Ex) -> Result<String, Error>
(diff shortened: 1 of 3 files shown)


appmgr/src/setup.rs
```diff
@@ -222,12 +228,27 @@ pub async fn complete(#[context] ctx: SetupContext) -> Result<(), Error> {
...
-            .write_all(ctx.product_key().await?.as_bytes())
-            .await?;
-        pkey_file.sync_all().await?;
+        crate::hostname::set_product_key(&*ctx.product_key().await?).await?;
+    } else {
+        let key_on_disk = crate::hostname::get_product_key().await?;
+        let key_in_cache = ctx.product_key().await?;
+        if *key_in_cache != key_on_disk {
+            crate::hostname::set_product_key(&*ctx.product_key().await?).await?;
+        }
```
dr-bonez, line 237: why would these ever differ?

ProofOfKeags: in a case where someone built from source without NO_KEY and then they tried to migrate from 0.2.x

dr-bonez: is this a case we are trying to support?

ProofOfKeags: It is imperative that we don't just blackhole people's product keys that they have purchased from us. We need to make sure that the "correct" product key is in place when migrating. If they also built 02x from source and generated their own key then, this process does nothing to harm that case, but it does fix the cases where a user migrates from an old system and they built 0.3 from source. So yes. I think it is something we are trying to support.

ProofOfKeags:
closed in favor of #1105

PR #1102, opened 2022-01-19
return null for tor-key pointer if doesn't exist

dr-bonez:
relevant: https://github.com/Start9Labs/cups-messenger/issues/13


dr-bonez:
accidentally based this off of #1100. will wait to merge till after

ProofOfKeags (approved):

chrisguida:
Are we sure we want this? Doesn't this always mean the service is misconfigured?

PR #1108, opened 2022-01-21
Bugfix/mdns thrash

ProofOfKeags:
The presumed issue here is that we over aggressively reload the entire set of mdns services. Every time an add or a remove is called we reset the entire set of mdns addresses. This can happen in situations where a container is started or stopped, so in crash loop scenarios it can happen quite often. This PR changes the behavior to not reset the entry group when services are added, but rather we add the service. When we remove a service, we don't actually destroy the CNAME record until we've accumulated a thousand dead ones, or embassyd itself is restarted.

### DIFF
--- a/backend/src/net/mdns.rs
+++ b/backend/src/net/mdns.rs
@@ -107,38 +107,43 @@ impl MdnsControllerInner {
                 std::ffi::CStr::from_ptr(self.hostname_raw)
             );
             for key in self.services.values() {
-                let lan_address = key
-                    .public()
-                    .get_onion_address()
-                    .get_address_without_dot_onion()
-                    + ".local";
-                tracing::debug!("Adding mdns CNAME entry for {}", &lan_address);
-                let lan_address_ptr = std::ffi::CString::new(lan_address)
-                    .expect("Could not cast lan address to c string");
-                res = avahi_sys::avahi_entry_group_add_record(
-                    self.entry_group,
-                    avahi_sys::AVAHI_IF_UNSPEC,
-                    avahi_sys::AVAHI_PROTO_UNSPEC,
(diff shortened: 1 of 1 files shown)


appmgr/src/net/mdns.rs
```diff
@@ -107,38 +108,43 @@ impl MdnsControllerInner {
...
-                    avahi_sys::AVAHI_DNS_TYPE_CNAME as u16,
-                    avahi_sys::AVAHI_DEFAULT_TTL,
-                    self.hostname.as_ptr().cast(),
-                    self.hostname.len(),
+                self.load_one(key);
+            }
+        }
+    }
+    fn load_one(&self, key: &TorSecretKeyV3) {
+        unsafe {
```
Blu-J, line 116: Doesn't unsafe require the function be annotated too (Safety)?
https://doc.rust-lang.org/std/keyword.unsafe.html

appmgr/src/net/mdns.rs
```diff
@@ -210,18 +217,30 @@ impl MdnsControllerInner {
         pkg_id: &PackageId,
         interfaces: I,
     ) {
-        self.services.extend(
-            interfaces
-                .into_iter()
-                .map(|(interface_id, key)| ((pkg_id.clone(), interface_id), key)),
-        );
-        self.sync();
+        let mut diff = false;
```
Blu-J, line 220: Would this be better as an iter().any() combinator?

appmgr/src/net/mdns.rs
ProofOfKeags: that would make every function in rust unsafe

appmgr/src/net/mdns.rs
ProofOfKeags: I'd love to tbh. I haven't figured out the way to make that clear. I'll take a stab at it again.

PR #1132, opened 2022-01-25
feat: move all frontend projects under the same Angular workspace

waterplea:

PR #1141, opened 2022-01-26
feat: move all frontend projects under the same Angular workspace

dr-bonez:
rebased from #1132


elvece (approved):

elvece:
This has been tested by running `make frontend` on macOS - which successfully compiles, and also building a new image using `make`, quick flashing an already setup EOS card, and ensuring the UI still operates as expected.

PR #1144, opened 2022-01-27
Properly configure changed optional dependencies 

elvece:
add logic to accept optional dependency list and determine dependencies based on new service configurations

### DIFF
--- a/system-images/compat/src/config/mod.rs
+++ b/system-images/compat/src/config/mod.rs
@@ -45,6 +45,7 @@ pub fn validate_configuration(
                 .map(|(pkg_id, data)| depends_on.insert(pkg_id, data.health_checks));
         };
     }
+    dbg!(&depends_on);
 
     // check that all configuration rules
     let rule_check = rules
--- a/system-images/compat/src/config/mod.rs
+++ b/system-images/compat/src/config/mod.rs
@@ -1,10 +1,12 @@
 use std::borrow::Cow;
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, BTreeSet, HashMap};
 use std::path::Path;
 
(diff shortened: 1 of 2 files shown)


system-images/compat/src/config/mod.rs
```diff
@@ -1,26 +1,52 @@
...
     config_path: &Path,
+    maybe_deps_path: Option<&str>,
 ) -> Result<SetResult, anyhow::Error> {
     let rules: Vec<ConfigRuleEntry> = serde_yaml::from_reader(std::fs::File::open(rules_path)?)?;
     let mut cfgs = LinearMap::new();
     cfgs.insert(name, Cow::Borrowed(&config));
+
+    let mut depends_on = BTreeMap::new();
+    if maybe_deps_path.is_some() {
+        let deps_path = Path::new(maybe_deps_path.unwrap());
```
dr-bonez, line 38:
```suggestion
    if let Some(deps_path) = maybe_deps_path.map(Path::new) {
```

elvece: I can't grok this pattern for the life of me 😣 will update

ProofOfKeags: matches also work if you forget the `if let` construct.

ProofOfKeags: also:

```rust
if let Pat(...) = expr {
    do_x();
} else {
    do_y();
}
```
is *exactly* identical to
```rust
match expr {
    Pat(...) => { do_x(); }
    _ => { do_y(); }
}
```

dr-bonez (approved):

ProofOfKeags:
squash this one.

Issue #1036, opened 2022-01-06
[bug]: .local access sometimes breaks on all services.


chrisguida:
Not sure how to reproduce. I was testing spark and the .local address worked briefly, then stopped. I went and tried to visit other services but they don't work either. I tried on both mac and linux and that also doesn't work. All I get is this, on all services:

<img width="970" alt="Screen Shot 2022-01-06 at 11 39 36 AM" src="https://user-images.githubusercontent.com/7445670/148426794-8ae32dd6-9b35-460b-8f60-271acdb5b9cd.png">

Nothing that suspicious in the logs, though maybe I missed something:

```
Jan 06 17:30:51 embassy-30297ba3 embassyd[4493]: Jan 06 17:30:51.087 DEBUG run_main:remove{pkg_id=PackageId(Id("spark-wallet"))}: embassy::net::mdns: Loading services for mDNS
Jan 06 17:30:51 embassy-30297ba3 embassyd[4493]: Jan 06 17:30:51.091  INFO run_main:remove{pkg_id=PackageId(Id("spark-wallet"))}: embassy::net::mdns: Published "embassy-30297ba3.local"
Jan 06 17:30:51 embassy-30297ba3 embassyd[4493]: Jan 06 17:30:51.092 DEBUG run_main:remove{pkg_id=PackageId(Id("spark-wallet"))}: embassy::net::mdns: Adding mdns CNAME entry for pjzz3s232yrelgbrcjt2im74hnfe3t7crjlw55smrqmxojp5tbiehoad.local
Jan 06 17:30:51 embassy-30297ba3 embassyd[4493]: Jan 06 17:30:51.095  INFO run_main:remove{pkg_id=PackageId(Id("spark-wallet"))}: embassy::net::mdns: Published "pjzz3s232yrelgbrcjt2im74hnfe3t7crjlw55smrqmxojp5tbiehoad.local"
Jan 06 17:30:51 embassy-30297ba3 embassyd[4493]: Jan 06 17:30:51.096 DEBUG run_main:remove{pkg_id=PackageId(Id("spark-wallet"))}: embassy::net::mdns: Adding mdns CNAME entry for r7l7qgzlp3o7iciz53upx4t2hu7l7hj4fpuqiqksxhvirosodhlcg5id.local
[...]

PR #1145, opened 2022-01-27
adds some investigative logging and improves error handling for mdns

ProofOfKeags:
This is in service of #1036. This doesn't fix it but should improve insight as to what is happening. Upon digging deeper into the avahi docs it seems that it is possible to get into collision states from which we must repair and it is also possible for avahi-daemon's misbehavior to have adverse impact to embassyd's ability to support .local. These changes should tell us what kind of error states we are hitting. We should merge this so that Chris can continue his testing that produces this bug, from there we can scrape the logs and learn what is going on and address it.

### DIFF
--- a/backend/src/net/mdns.rs
+++ b/backend/src/net/mdns.rs
@@ -63,7 +63,6 @@ pub struct MdnsControllerInner {
     hostname_raw: *const libc::c_char,
     entry_group: *mut AvahiEntryGroup,
     services: BTreeMap<(PackageId, InterfaceId), TorSecretKeyV3>,
-    client_error: std::pin::Pin<Box<i32>>,
 }
 unsafe impl Send for MdnsControllerInner {}
 unsafe impl Sync for MdnsControllerInner {}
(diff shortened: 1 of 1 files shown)


chrisguida (approved):
Concept ACK

dr-bonez (approved):

backend/src/net/mdns.rs
```diff
@@ -191,18 +193,45 @@ impl MdnsControllerInner {
                 hostname_raw,
                 entry_group: group,
                 services: BTreeMap::new(),
-                client_error: box_err,
```
dr-bonez, line 194: don't do this

backend/src/net/mdns.rs
```diff
@@ -63,7 +63,6 @@ pub struct MdnsControllerInner {
     hostname_raw: *const libc::c_char,
     entry_group: *mut AvahiEntryGroup,
     services: BTreeMap<(PackageId, InterfaceId), TorSecretKeyV3>,
-    client_error: std::pin::Pin<Box<i32>>,
```
dr-bonez, line 66: don't do this

dr-bonez (changes requested):

dr-bonez (approved):

PR #1158, opened 2022-01-31
Changes to ubuntu 21.10 at the base image

ProofOfKeags:


ProofOfKeags:
Currently running my prod setup off of this and see if it makes anything better. I still have a bitcoin sync so it should be passively loaded.

dr-bonez (approved):

dr-bonez:
Note: requires full reflash! No quick flashing.

PR #1159, opened 2022-01-31
fix: Add in the masked just in case

Blu-J:


dr-bonez:
ok, I understand what's going on here, but what's motivating it? masked was added in properties v2, so there's nothing using v1 properties that should be using it, right?

Blu-J:
> ok, I understand what's going on here, but what's motivating it? masked was added in properties v2, so there's nothing using v1 properties that should be using it, right?

This is what I got, but this was something that Lucy was wondering why. It just looks like things might be remembered wrong?

MattDHill:
@elvece can you chime in here. I'm not sure why this would be needed since `masked` is only an attr of V2 properties.

Blu-J:
Looks like this didn't fix the potential issue that Lucy was seeing, closing

PR #1160, opened 2022-02-01
alt marketplace feature

DrewAns:


frontend/projects/ui/src/app/pages/server-routes/marketplaces/marketplaces.page.ts
```diff
@@ -96,7 +127,10 @@ export class MarketplacesPage {
...
       JSON.stringify(this.patch.data.ui.marketplace),
     )
-    const newMarketplace = marketplace['known-hosts'][id]
+
+    const url = id
+      ? marketplace['known-hosts'][id].url
+      : this.config.marketplace.url
 
     const loader = await this.loadingCtrl.create({
       spinner: 'lines',
```
MattDHill, line 136: Should say connecting..., not validating marketplace

ProofOfKeags:
Just so y'all know: Unfortunately since this PR has conflated a few unrelated things, some of which are necessary and some of which are not, it has created some new work on other systems that could have been able to be avoided. As such, if you could please be careful about separating concerns so that we don't entangle necessary and lower priority changes it will help our ability to ship a bit faster.

MattDHill (approved):

PR #1165, opened 2022-02-02
fix lock scope

dr-bonez:
This should fix a bug where the *second* app being backed up would sometimes deadlock.

### DIFF
--- a/backend/src/backup/backup_bulk.rs
+++ b/backend/src/backup/backup_bulk.rs
@@ -306,7 +306,9 @@ async fn perform_backup<Db: DbHandle>(
             .synchronize()
             .await;
 
-        installed_model.lock(&mut db, LockType::Write).await?;
+        let mut tx = db.begin().await?;
+
+        installed_model.lock(&mut tx, LockType::Write).await?;
 
         let guard = backup_guard.mount_package_backup(&package_id).await?;
         let res = manifest
@@ -328,7 +330,6 @@ async fn perform_backup<Db: DbHandle>(
             },
         );
 
-        let mut tx = db.begin().await?;
         if let Ok(pkg_meta) = res {
             installed_model
                 .last_backup()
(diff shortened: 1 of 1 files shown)


backend/src/backup/backup_bulk.rs
```diff
@@ -330,15 +332,14 @@ async fn perform_backup<Db: DbHandle>(
                 &manifest.volumes,
             )
             .await;
-        drop(guard);
+        guard.unmount().await?;
```
ProofOfKeags, line 335: I assume this is because `async` drop traits are not a thing?

ProofOfKeags (approved):

backend/src/backup/backup_bulk.rs
dr-bonez: Yes

Blu-J: https://github.com/bluss/scopeguard
Would the above help?

Blu-J:
Second try and it locks/ stops
image
image

ProofOfKeags:
@Blu-J are you saying this doesn't work?

Blu-J:
@ProofOfKeags Yes it doesn't seem to fix it
Sometimes it gets through;
Sometimes it gets stuck on btc pay server;
Sometimes gets stuck on Mastodon

ProofOfKeags:
how are you assessing stuck?

Part of the backup process is to save off the .s9pk and Mastodon is 1GB.

Blu-J:
Will do a few more, the mastodon was more premptive. The btc -pay one was stuck there an hour, but I have done 5 more backups without issue. So will test throughout the night

ProofOfKeags:
we have deadlock detection so maybe we need to kick up the logging around the lock acquisition.

Blu-J (approved):
Checked 5 times, was consistent with time, seems fixed

Issue #1168, opened 2022-02-04
[bug]: port mapping for lan is inverted from tor


dr-bonez:
currently, for tor config, the external port is on the left, and the internal port is on the right.
for lan, the internal port is on the left, and the external port is under "mapping"

this is highly confusing and inconsistent.

@elvece @chrisguida fixing this will require updating all services

marking this as p1 cause we can't change it after shipping.

ProofOfKeags:
Service checklist:
- [ ] bitcoind
- [ ] btc-rpc-proxy
- [ ] btcpayserver
- [ ] burn-after-reading
- [ ] c-lightning
- [ ] cups
- [ ] embassy-pages
- [ ] filebrowser
- [ ] lnd
- [ ] mastodon
- [ ] photoview
- [ ] ride-the-lightning
- [ ] spark-wallet
- [ ] sphinx-relay
- [ ] synapse
- [ ] thunderhub
- [ ] vaultwarden

PR #1181, opened 2022-02-08
bugfix: inverts mapping for lan config

dr-bonez:
addresses #1168


elvece (dismissed):

chrisguida (dismissed):
Leaving `mapping` for now since this would cause a database deserialization error and prevent embassy from booting

Will convert all services to use the new lan config format

Issue #1164, opened 2022-02-02
[bug]: /embassyd-data/main filling up with system logs


chrisguida:
Journald is supposed to limit logs to 1G

https://github.com/Start9Labs/embassy-os/pull/1156

```
[77995.436666] systemd-journald[1673]: Failed to open system journal: No space left on device
[77995.437406] systemd-journald[1673]: Failed to open system journal: No space left on device
[77995.438541] systemd-journald[1673]: Failed to open system journal: No space left on device
[77995.439436] systemd-journald[1673]: Failed to open system journal: No space left on device
[77995.518614] systemd-journald[1673]: Failed to open system journal: No space left on device
[77995.519349] systemd-journald[1673]: Failed to open system journal: No space left on device
[77996.623863] systemd-journald[1673]: Failed to open system journal: No space left on device
[77996.624564] systemd-journald[1673]: Failed to open system journal: No space left on device
[77999.440488] systemd-journald[1673]: Failed to open system journal: No space left on device
[77999.441123] systemd-journald[1673]: Failed to open system journal: No space left on device
[77999.588838] systemd-journald[1673]: Failed to open system journal: No space left on device
[77999.589465] systemd-journald[1673]: Failed to open system journal: No space left on device
[77999.601555] systemd-journald[1673]: Failed to open system journal: No space left on device
[77999.602032] systemd-journald[1673]: Failed to open system journal: No space left on device
[77999.604694] systemd-journald[1673]: Failed to open system journal: No space left on device
[77999.605225] systemd-journald[1673]: Failed to open system journal: No space left on device
[77999.654548] systemd-journald[1673]: Failed to open system journal: No space left on device
[77999.655184] systemd-journald[1673]: Failed to open system journal: No space left on device
[77999.655969] systemd-journald[1673]: Failed to open system journal: No space left on device
[77999.656570] systemd-journald[1673]: Failed to open system journal: No space left on device
[78000.142116] systemd-journald[1673]: Failed to open system journal: No space left on device
[78000.142765] systemd-journald[1673]: Failed to open system journal: No space left on device
[78000.143656] systemd-journald[1673]: Failed to open system journal: No space left on device
[78000.144120] systemd-journald[1673]: Failed to open system journal: No space left on device
[78000.437783] systemd-journald[1673]: Failed to open system journal: No space left on device
[78000.438368] systemd-journald[1673]: Failed to open system journal: No space left on device
[...]

PR #1188, opened 2022-02-09
Make machine-ids and ssh host keys unique

ProofOfKeags:
This PR is in response to #1164 but unintuitively.

The actual problem originally arose from having different machine ids every boot. This was due to running cloud init every boot back in the couple of weeks where we were on 21.10.

RaspiOS implicitly fixed that by doing away with cloud-init. However, the reason we had cloud-init in the first place is to make ssh host keys different per machine.

This PR ultimately ensures that every Embassy has a unique machine-id and set of ssh keys, and that those materials are preserved across OTA updates.

### DIFF
--- a/build/initialization.sh
+++ b/build/initialization.sh
@@ -58,6 +58,8 @@ ControlPort 9051
 CookieAuthentication 1
 EOF
 
+cat /embassy-os/product_key.txt | tr -d '\n' | sha256sum | head -c 32 | sed 's/$/\n/' > /etc/machine-id
+
 raspi-config nonint enable_overlayfs
 systemctl disable initialization.service
 sudo systemctl restart NetworkManager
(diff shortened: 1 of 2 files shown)


Blu-J (approved):

build/initialization.sh
```diff
@@ -58,6 +58,10 @@ ControlPort 9051
 CookieAuthentication 1
 EOF
 
+cat /embassy-os/product_key.txt | tr -d '\n' | sha256sum | head -c 32 | sed 's/$/\n/' > /etc/machine-id
+rm /etc/ssh/ssh_host_*
+dpkg-reconfigure openssh-server
```
dr-bonez, line 63: you don't need to do this in initialization

dr-bonez (changes requested):

build/initialization.sh
ProofOfKeags: If I don't then there won't be any ssh_host_* keys.

dr-bonez (approved):

dr-bonez (approved):

PR #1195, opened 2022-02-11
prevents packages from producing conflicting ssl configs

### LINKED ISSUE #1194: [bug]: Packages with conflicting SSL LAN configs bring down NGINX
### Prerequisites

- [X] I have searched for [existing issues](https://github.com/start9labs/embassy-os/issues) that already report this problem, without success.

### EmbassyOS Version

1d8512e1dd29f3620c58c7b2b44842b4b968b453

### Device

Laptop/Desktop

### Device OS

Linux

### Device OS Version

Ubuntu 21.10

### Browser

Brave

### Browser Version

1.19.92

### Current Behavior

Scenario:

- Package A requests a LAN config bound to external port 80 with `ssl: false`
- Package B requests a LAN config bound to external port 80 with `ssl: true`

This results in nginx crashing because it cannot listen on port 80 with ssl both enabled *and* disabled.

### Expected Behavior

Nginx should never crash as a result of package requests. Full Stop. Any instance of this is a security issue where malicious packages could render an Embassy unusable.

However, given that the OS has been given contradictory instructions, we cannot expect the system to install both packages successfully granting all of their requests. Best case scenario here is to fail the install of the packages that contradicts the current system state.

The way we intend to address this is to check if any of the LAN interfaces requested by a package's manifest contradict the set of LAN interfaces currently out on lease. If they do we reject the install. The tricky part will be communicating this problem to the user effectively. Unfortunately there is no way to communicate this to a user in a way that they will understand but we can say something to the effect of:

`Package B has requested a resource from EmbassyOS that contradicts (EmbassyOS|package: <package-id>)`

### Steps to Reproduce

see above

### Anything else?

_No response_

ProofOfKeags:
fixes #1194

### DIFF
--- a/backend/src/install/mod.rs
+++ b/backend/src/install/mod.rs
@@ -9,8 +9,8 @@ use std::time::{Duration, Instant};
 
 use color_eyre::eyre::eyre;
 use emver::VersionRange;
-use futures::future::BoxFuture;
-use futures::{FutureExt, StreamExt, TryStreamExt};
+use futures::future::{self, BoxFuture};
+use futures::{stream, FutureExt, StreamExt, TryStreamExt};
 use http::header::CONTENT_LENGTH;
 use http::{Request, Response, StatusCode};
 use hyper::Body;
@@ -43,7 +43,7 @@ use crate::s9pk::manifest::{Manifest, PackageId};
 use crate::s9pk::reader::S9pkReader;
 use crate::status::{MainStatus, Status};
 use crate::util::io::{copy_and_shutdown, response_to_reader};
-use crate::util::serde::{display_serializable, IoFormat};
+use crate::util::serde::{display_serializable, IoFormat, Port};
 use crate::util::{display_none, AsyncFileExt, Version};
 use crate::version::{Current, VersionT};
 use crate::volume::asset_dir;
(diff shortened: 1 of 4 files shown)


backend/src/install/mod.rs
```diff
@@ -124,6 +124,52 @@ pub async fn install(
         .json()
         .await
         .with_kind(crate::ErrorKind::Registry)?;
+    let mut db_handle = ctx.db.handle();
+    // TODO: Ask Aiden why this didn't work
+    // let manifests = stream::iter(crate::db::package::get_packages(&mut db_handle).await?)
+    //     .then(move |pkg| crate::db::package::get_manifest(&mut db_handle, &pkg))
+    //     .try_filter_map(|a| async { Ok(a) })
+    //     .try_collect::<Vec<Manifest>>();
```
dr-bonez, line 132:
```suggestion
    // let manifests = stream::iter(crate::db::package::get_packages(&mut db_handle).await?)
    //     .then(|pkg| async move { crate::db::package::get_manifest(&mut db_handle, &pkg).await })
    //     .try_filter_map(|a| async { Ok(a) })
    //     .try_collect::<Vec<Manifest>>();
```
🙃

dr-bonez (approved):

backend/src/install/mod.rs
ProofOfKeags: dude....what?!?! I tried `move |x| async {...}` and that didn't work. Wthat the actual fuck is the difference?

dr-bonez (approved):

PR #1196, opened 2022-02-11
implement delete-recovered

### LINKED ISSUE #1134: [bug]: Cannot delete recovered services
Steps to reproduce:

1. Migrate 02x box with package(s) you don't need
2. Once services page loads, click trash can to delete recovered package you don't want to keep
3. Error message is displayed, "method not found".

Screen Shot 2022-01-25 at 11 20 03 AM

On git hash 3bf6e3fb2bf83874a584d1358907e5651decefa9

dr-bonez:
closes #1134 

untested

### DIFF
--- a/backend/sqlx-data.json
+++ b/backend/sqlx-data.json
@@ -162,6 +162,16 @@
       "nullable": []
     }
   },
+  "56b986f2a2b7091d9c3acdd78f75d9842242de1f4da8f3672f2793d9fb256928": {
+    "query": "DELETE FROM tor WHERE package = ?",
+    "describe": {
+      "columns": [],
+      "parameters": {
+        "Right": 1
+      },
+      "nullable": []
+    }
+  },
   "5b114c450073f77f466c980a2541293f30087b57301c379630326e5e5c2fb792": {
     "query": "REPLACE INTO tor (package, interface, key) VALUES (?, ?, ?)",
     "describe": {
--- a/backend/src/install/cleanup.rs
+++ b/backend/src/install/cleanup.rs
@@ -3,6 +3,7 @@ use std::collections::{BTreeMap, HashMap};
 use bollard::image::ListImagesOptions;
 use color_eyre::eyre::eyre;
 use patch_db::{DbHandle, LockType, PatchDbHandle};
+use sqlx::{Executor, Sqlite};
 use tracing::instrument;
 
 use super::{PKG_ARCHIVE_DIR, PKG_DOCKER_DIR};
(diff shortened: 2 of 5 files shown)


backend/sqlx-data.json
```diff
@@ -162,6 +162,16 @@
       "nullable": []
     }
   },
+  "56b986f2a2b7091d9c3acdd78f75d9842242de1f4da8f3672f2793d9fb256928": {
```
Blu-J, line 165: Confirmation: This is the generation of sqlx right.

backend/src/install/cleanup.rs
```diff
@@ -233,12 +234,16 @@ pub async fn remove_from_current_dependents_lists<
...
+#[instrument(skip(ctx, secrets, db))]
+pub async fn uninstall<Ex>(
     ctx: &RpcContext,
     db: &mut PatchDbHandle,
+    secrets: &mut Ex,
     id: &PackageId,
-) -> Result<(), Error> {
+) -> Result<(), Error>
+where
+    for<'a> &'a mut Ex: Executor<'a, Database = Sqlite>,
```
Blu-J, line 245: Comment: First time seeing a for<'a>
Most of the time I would have had in line 241 something like `secrets: & mut impl Executor<'_, Database = Sqlite>`
Looks like the for is a way of tying the lifetimes together.

backend/src/install/cleanup.rs
```diff
@@ -288,5 +293,18 @@ pub async fn uninstall(
...
+    Ok(())
+}
+
+#[instrument(skip(secrets))]
+pub async fn remove_tor_keys<Ex>(secrets: &mut Ex, id: &PackageId) -> Result<(), Error>
+where
+    for<'a> &'a mut Ex: Executor<'a, Database = Sqlite>,
+{
+    let id_str = id.as_str();
+    sqlx::query!("DELETE FROM tor WHERE package = ?", id_str)
```
Blu-J, line 306: Comment: The query looks good. I would have thought you could pass the id as a reference instead of converting to as_str. Then make sure package_id can be used in sqlx queries.

backend/src/install/mod.rs
```diff
@@ -564,7 +564,17 @@ pub async fn uninstall_impl(ctx: RpcContext, id: PackageId) -> Result<WithRevisi
     drop(handle);
 
     tokio::spawn(async move {
-        if let Err(e) = cleanup::uninstall(&ctx, &mut ctx.db.handle(), &id).await {
+        if let Err(e) = async {
```
Blu-J, line 567: Comment/ Question: Is the reason for the double async wrap just to get the correct error out?

backend/src/install/mod.rs
```diff
@@ -593,6 +603,42 @@ pub async fn uninstall_impl(ctx: RpcContext, id: PackageId) -> Result<WithRevisi
     })
 }
 
+#[command(rename = "delete-recovered", display(display_none))]
+pub async fn delete_recovered(
```
Blu-J, line 607: Highest risk of having an error, I don't know if there are resources that need to be removed that shouldn't or aren't here. Wondering if there is a way of ensuring this?

Blu-J (approved):
Nothing looks wrong code wise as far as I can find, tried to be active reader.

backend/sqlx-data.json
dr-bonez: yes

PR #1201, opened 2022-02-13
backend: fix for loop iteration in ssl_port_status function

RhynoTech:
Not sure if this is the best/correct way to fix the issue I was experiencing. Essentially the nested for loops had trouble iterating over the manifest variable. The below code changes did fix my build.


ProofOfKeags:
What version of rust are you running?

RhynoTech:
I was just using the build process here: https://github.com/Start9Labs/embassy-os/tree/master/build#readme

I don't have rust installed on my Ubuntu 20.04 Desktop.

ProofOfKeags:
I ended up going with a different approach in #1202 due to the need to do an interactive rebase. But thank you for your contribution regardless!

PR #1207, opened 2022-02-14
Wip: Fix most of the wifi and 80% of initialization.sh

dr-bonez:


Blu-J (approved):
Seems like this fixes it.
image

PR #1210, opened 2022-02-15
add vim and jq

dr-bonez:
NO TICKET

these tools are often very useful when helping users debug

### DIFF
--- a/build/initialization.sh
+++ b/build/initialization.sh
@@ -27,7 +27,8 @@ apt-get install -y \
 	ntp \
 	network-manager \
 	vim \ 
-	jq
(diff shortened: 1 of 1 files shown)


Blu-J (dismissed):

Blu-J:
> Add ncdu as well
I had to look this up. Yes, this could be useful for a debug, but should it be part of the default install. 
Vim maybe because, well vim is super light and used by all the devs.
Jq, for sure on this tool, this can be

ProofOfKeags (dismissed):

ProofOfKeags (approved):

PR #1219, opened 2022-02-16
fix (de)serialization of min and max

dr-bonez:


kn0wmad (approved):
Patch works as expected

PR #1244, opened 2022-02-21
chore: Converting a bash script into a javascript file

Blu-J:


frontend/update-git-hash.sh
```diff
@@ -1,7 +0,0 @@
-#!/bin/bash
-
-cd "$(dirname "$0")"
```
Blu-J, line 3: Not sure that we moved anywhere, so this is the only thing that is missing in a 1-1 conversion

ProofOfKeags (dismissed):

dr-bonez (approved):

Issue #363, opened 2021-07-12
Ensure all external links in EmbassyOS point to appropriate documentation.


MattDHill:
app component 145-165

EDIT by @ProofOfKeags:

There are several places in the Embassy UI that presently render "broken links" to documentation. This is because the appropriate documentation hasn't been written yet (Start9Labs/documentation#39). However, upon that work being completed we need to make sure that the link targets provide sensible explanations for the situations that create these links to begin with.

kn0wmad:
Blocked by docs versioning - https://github.com/Start9Labs/documentation/issues/39

PR #1249, opened 2022-02-22
Fix links in UI to point to new website

kn0wmad:
It seems like there ought to be more links than what I found, but perhaps there is just room to add more links going forward.  I will test with this in mind


DrewAns (approved):

kn0wmad:
Will close #363

kn0wmad (commented):
Up to date

DrewAns (approved):

MattDHill (approved):

PR #1253, opened 2022-02-22
speed up shutdown

chrisguida:
Speed up shutdown sound from 120 to 200 qpm to make it sound more crisp


ProofOfKeags (approved):

ProofOfKeags:
I had been meaning to do this for a while. Thank you.

PR #1260, opened 2022-02-24
Bugfix/fix syntax change user

ProofOfKeags:
turns out that `users` shows logged in users, not the users on the system. So it requires you to ssh in in order to actually get it unjammed. This fixes that.


dr-bonez (approved):
lmao

Issue #1306, opened 2022-03-04
[feat]: SSH keys should go to start9 user (not root)


k0gen:
### Prerequisites

- [X] I have searched for [existing issues](https://github.com/start9labs/embassy-os/issues) that already suggest this feature, without success.

### Describe the Feature Request

Right now if you want to add SSH keys to your Embassy those keys are for root account. I suggest changing this to non-root `start9` user account.

### Describe the Use Case

Using root account is not a good practice security wise. The less we depend on `root` account the more secure the system will become.
Also longing as `start9@embassy-q1w2e3.local` looks cool.

### Describe Preferred Solution

Just add keys to /home/start9/.ssh/authorized_keys and disable root login via ssh.

### Describe Alternatives

none

### Anything else?

Elevate `start9` user permissions to use tools like embassy-cli, docker etc.

ProofOfKeags:
[Relevant Code](https://github.com/Start9Labs/embassy-os/blob/master/backend/src/ssh.rs) for anyone who wants to try and tackle this.

k0gen:
Thanks. I'll take it from here.

k0gen:
> [Relevant Code](https://github.com/Start9Labs/embassy-os/blob/master/backend/src/ssh.rs) for anyone who wants to try and tackle this.

[Work in progress ...](https://github.com/k0gen/embassy-os/commits/patch-1)
Should `backend/copy.sh` also be addressed regarding the ssh keys for start9 user?

dr-bonez:
That's a utility script that @elvece uses, I think. It's not part of the mainline development flow for embassyos

elvece:
Apologies, I am unfamiliar with that script. It's either a relic from a prior version of a build script, or an old utility for development purposes. I do not think it is used or needed any longer.

ProofOfKeags:
@k0gen feel free to delete that script as part of this PR.

k0gen:
Done.

PR #1321, opened 2022-03-10
Switching SSH keys to start9 user

k0gen:
PR for  #1306


dr-bonez:
looks good, but needs to be tested. Not worth the risk to add it to the patch release, but will likely include it in 0.3.1.

k0gen:
Right now I have build and tested it against `ENVIRONMENT=dev make` on a fresh  install

https://user-images.githubusercontent.com/3606313/157978179-4038638f-dbc8-4994-a8cd-bdd387eb49af.mp4

As you can see `.ssh` directory ownership is left for root for now so I don't expect any major problems with that.

dr-bonez (approved):

PR #1332, opened 2022-03-15
use consistent config file type

dr-bonez:
we were parsing a .yaml file as toml in some circumstances


ProofOfKeags (approved):

ProofOfKeags:
This is non-backwards compatible right?

dr-bonez:
I mean it is for anyone who put toml in a file called `config.yaml`. which I don't think anyone did

PR #1333, opened 2022-03-15
whitelist samsung T7

dr-bonez:
testing has shown no issues with UASP on the samsung T7

### DIFF
--- a/backend/src/disk/quirks.rs
+++ b/backend/src/disk/quirks.rs
@@ -11,11 +11,12 @@ use crate::Error;
 
 pub const QUIRK_PATH: &'static str = "/sys/module/usb_storage/parameters/quirks";
 
-pub const WHITELIST: [(VendorId, ProductId); 4] = [
+pub const WHITELIST: [(VendorId, ProductId); 5] = [
     (VendorId(0x1d6b), ProductId(0x0002)), // root hub usb2
     (VendorId(0x1d6b), ProductId(0x0003)), // root hub usb3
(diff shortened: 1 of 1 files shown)


dr-bonez:
note - if the T7 has already been blacklisted this will not update until disk.list is called. Ideally should have this fixed on startup. Also could be dangerous if it gets updated on a live data drive.

backend/src/disk/quirks.rs
```diff
@@ -11,11 +11,12 @@ use crate::Error;
 
 pub const QUIRK_PATH: &'static str = "/sys/module/usb_storage/parameters/quirks";
 
-pub const WHITELIST: [(VendorId, ProductId); 4] = [
+pub const WHITELIST: [(VendorId, ProductId); 5] = [
     (VendorId(0x1d6b), ProductId(0x0002)), // root hub usb2
     (VendorId(0x1d6b), ProductId(0x0003)), // root hub usb3
     (VendorId(0x2109), ProductId(0x3431)),
     (VendorId(0x1058), ProductId(0x262f)), // western digital black HDD
+    (VendorId(0x04e8), ProductId(0x4001)),
```
ProofOfKeags, line 19: put drive model in the comment here.

dr-bonez:
```suggestion
    (VendorId(0x04e8), ProductId(0x4001)), // Samsung T7
```

Blu-J (approved):

Issue #1335, opened 2022-03-15
[bug]: lan nginx proxy doesn't handle websockets


dr-bonez:
// TODO

PR #1340, opened 2022-03-16
allow websockets to containers

dr-bonez:
addresses #1335

### DIFF
--- a/backend/src/nginx/main-ui.conf.template
+++ b/backend/src/nginx/main-ui.conf.template
@@ -1,4 +1,8 @@
 
+map $http_upgrade $connection_upgrade {
+        default upgrade;
+        '' $http_connection;
+}
 
 server {{
         listen 443 ssl default_server;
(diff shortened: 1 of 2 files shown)


cryptodread:
Tested, works as expected over LAN and Tor. All other services still behaving as expected over LAN.

backend/src/nginx/main-ui.conf.template
```diff
@@ -1,4 +1,8 @@
 
+map $http_upgrade $connection_upgrade {{
+        default upgrade;
+        '' $http_connection;
+}}
```
elvece, line 5: just curious what this is doing?

elvece (approved):

backend/src/nginx/main-ui.conf.template
dr-bonez: this only sets the `Connection` header to `upgrade` if the `Upgrade` header is set.

Issue #1346, opened 2022-03-18
[feat]: Add options for performing filesystem repair on startup


dr-bonez:
## purpose
Goal is to improve ways to recover from disk i/o issues that lead to data corruption.

### backend
during disk mount, in embassy-init:
- if `/embassy-os/repair-disk` exists
  - if `/embassy-os/<guid>_<fs>.e2undo` exists
    - `e2undo -z /embassy-os/<guid>_<fs>.e2undo /dev/mapper/<guid>_<fs>`
  - else
    - `e2fsck -y -z /embassy-os/<guid>_<fs>.e2undo /dev/mapper/<guid>_<fs>`
- else
  - `e2fsck -p /dev/mapper/<guid>_<fs>`
Add endpoints `diagnostic.disk.repair`, and `disk.repair` that create `/embassy-os/repair-disk` and reboot.

### frontend
add a button to diagnostic ui if error code 2 (Filesystem I/O) or 48 (Disk Management) to repair disk.
add same button to main ui, next to "System Rebuild". 

Relevant warnings to be conveyed to the user:
- Likely to make things better, but could make things worse. 
- Recommend backing up beforehand if possible
- if anything happens to the device while repair is happening (between bep and chime) *will* make things worse, in an unrecoverable way.
  - such as unplugging the drive, unplugging the embassy, power surges, i/o issues due to hardware faults

@MattDHill please advise on exact messaging and UX

PR #1350, opened 2022-03-22
run fsck on startup

dr-bonez:
addresses #1346

### DIFF
--- a/backend/src/diagnostic.rs
+++ b/backend/src/diagnostic.rs
@@ -5,6 +5,7 @@ use rpc_toolkit::command;
 use rpc_toolkit::yajrc::RpcError;
 
 use crate::context::DiagnosticContext;
+use crate::disk::repair;
 use crate::logs::{display_logs, fetch_logs, LogResponse, LogSource};
 use crate::shutdown::Shutdown;
 use crate::util::display_none;
@@ -12,7 +13,7 @@ use crate::Error;
 
 pub const SYSTEMD_UNIT: &'static str = "embassy-init";
 
-#[command(subcommands(error, logs, exit, restart, forget_disk))]
+#[command(subcommands(error, logs, exit, restart, forget_disk, disk))]
 pub fn diagnostic() -> Result<(), Error> {
     Ok(())
 }
@@ -56,7 +57,12 @@ pub fn restart(#[context] ctx: DiagnosticContext) -> Result<(), Error> {
     Ok(())
 }
 
-#[command(rename = "forget-disk", display(display_none))]
+#[command(subcommands(forget_disk, repair))]
+pub fn disk() -> Result<(), Error> {
+    Ok(())
+}
+
+#[command(rename = "forget", display(display_none))]
 pub async fn forget_disk() -> Result<(), Error> {
     let disk_guid = Path::new("/embassy-os/disk.guid");
     if tokio::fs::metadata(disk_guid).await.is_ok() {
--- a/backend/src/disk/fsck.rs
+++ b/backend/src/disk/fsck.rs
@@ -15,27 +15,55 @@ impl std::ops::BitOrAssign for RequiresReboot {
     }
 }
 
-#[instrument]
-pub async fn e2fsck(
-    logicalname: impl AsRef<Path> + std::fmt::Debug,
-    repair: bool,
-) -> Result<RequiresReboot, Error> {
-    let mut e2fsck_cmd = Command::new("e2fsck");
-    if repair {
-        let undo_path = Path::new("/embassy-os")
-            .join(
-                logicalname
-                    .as_ref()
-                    .file_name()
-                    .unwrap_or(OsStr::new("unknown")),
-            )
-            .with_extension("e2undo");
-        e2fsck_cmd.arg("-y").arg("-z").arg(&undo_path);
-    } else {
-        e2fsck_cmd.arg("-p");
+#[derive(Debug, Clone, Copy)]
+pub enum RepairStrategy {
+    Preen,
+    Aggressive,
+}
+impl RepairStrategy {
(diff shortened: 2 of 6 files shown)


backend/src/diagnostic.rs
```diff
@@ -56,7 +57,12 @@ pub fn restart(#[context] ctx: DiagnosticContext) -> Result<(), Error> {
     Ok(())
 }
 
-#[command(rename = "forget-disk", display(display_none))]
+#[command(subcommands(forget_disk, repair))]
+pub fn disk() -> Result<(), Error> {
+    Ok(())
```
elvece, line 62: why do we not need to `tokio::fs::write(REPAIR_DISK_PATH, b"").await?;` here?

backend/src/disk/main.rs
```diff
@@ -267,27 +274,26 @@ pub async fn mount_fs<P: AsRef<Path>>(
...
     datadir: P,
+    repair: bool,
     password: &str,
-) -> Result<(), Error> {
-    mount_fs(guid, &datadir, "main", password).await?;
-    mount_fs(guid, &datadir, "package-data", password).await?;
-    Ok(())
+) -> Result<RequiresReboot, Error> {
+    let mut reboot = RequiresReboot::False;
+    reboot |= mount_fs(guid, &datadir, "main", repair, password).await?;
```
elvece, line 296: jw what `|=` does?

elvece (approved):

backend/src/disk/main.rs
dr-bonez: or-equals. Or the lhs with the rhs, then store the result in the lhs.

backend/src/diagnostic.rs
dr-bonez: that does happen in the repair function which is a child of this

backend/src/disk/fsck.rs
```diff
@@ -0,0 +1,78 @@
...
+            (RequiresReboot::False, RequiresReboot::False) => RequiresReboot::False,
+            _ => RequiresReboot::True,
+        }
+    }
+}
+
+#[instrument]
+pub async fn e2fsck(
+    logicalname: impl AsRef<Path> + std::fmt::Debug,
+    repair: bool,
```
Blu-J, line 27: Side: I question boolean params.
We could pass in an enum that tranforms to the correct e2fsck_cmd
Pass the command
Use Option, and the default is -p and not the repair side

backend/src/disk/fsck.rs
```diff
@@ -0,0 +1,78 @@
...
+use std::path::Path;
+
+use color_eyre::eyre::eyre;
+use tokio::process::Command;
+use tracing::instrument;
+
+use crate::{Error, ResultExt};
+
+#[derive(Debug, Clone, Copy)]
+pub enum RequiresReboot {
```
Blu-J, line 11: Comment: Just weird that we are doing this just for the or operator
Side: Could rethink the bool as 


```rs
pub enum RebootAction;
/// Then return Option<RebootAction>?

/// Or
pub enum Action {
  Nothing,
  Reboot,
}
```
Doesn't solve the boolean issues, except the blindness.
Since this is not far from a boolean, and we just get a smarter wrapper
```rs
pub struct RequiresReboot(bool);
````

File: backend/src/disk/main.rs [hunk 5]
```diff
@@ -267,27 +274,26 @@ pub async fn mount_fs<P: AsRef<Path>>(
...
     guid: &str,
     datadir: P,
+    repair: bool,
     password: &str,
-) -> Result<(), Error> {
-    mount_fs(guid, &datadir, "main", password).await?;
-    mount_fs(guid, &datadir, "package-data", password).await?;
-    Ok(())
+) -> Result<RequiresReboot, Error> {
+    let mut reboot = RequiresReboot::False;
```
Blu-J, line 295: Alternative: Could also do it like 
```rs
let reboot = 
   mount_fs(guid, &datadir, "main", repair, password).await?
   | mount_fs(guid, &datadir, "package-data", repair, password).await?;
```
```

**Review by Blu-J — APPROVED** [review]

File: backend/src/disk/fsck.rs [hunk 6]
```diff
@@ -0,0 +1,78 @@
...
+    fn bitor_assign(&mut self, rhs: Self) {
+        *self = match (*self, rhs) {
+            (RequiresReboot::False, RequiresReboot::False) => RequiresReboot::False,
+            _ => RequiresReboot::True,
+        }
+    }
+}
+
+#[instrument]
+pub async fn e2fsck(
```
Blu-J, line 25: Alternative: We could make this into two seperate functions, one for repair and one not. 
Then use helper functions to figure out the output to error.

backend/src/disk/main.rs
Blu-J: Typescript I think has this now
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#short-circuiting-assignment-operators

https://github.com/tc39/proposal-logical-assignment

Blu-J (approved):

backend/src/disk/fsck.rs
```diff
@@ -0,0 +1,100 @@
...
+#[derive(Debug, Clone, Copy)]
+pub enum RepairStrategy {
+    Preen,
+    Aggressive,
+}
+impl RepairStrategy {
+    pub async fn e2fsck(
+        &self,
+        logicalname: impl AsRef<Path> + std::fmt::Debug,
+    ) -> Result<RequiresReboot, Error> {
```
Blu-J, line 27: Didn't know we could gleam this, that's cool

Blu-J (approved):

PR #1359, opened 2022-03-26
refactor: decompose app component

waterplea:


elvece (approved):
Reviewed and tested on both desktop and mobile. Only bug discovered was the improper font displaying in the menu item titles, making that new class `!important` fixed. 

Looks good to merge, but after 0.3.0.2 release.

PR #1375, opened 2022-04-01
allow for multiple disk repairs

dr-bonez:


backend/src/disk/fsck.rs
```diff
@@ -39,21 +41,38 @@ pub async fn e2fsck_preen(
     e2fsck_runner(Command::new("e2fsck").arg("-p"), logicalname).await
 }
 
+fn backup_existing_undo_file<'a>(path: &'a Path) -> BoxFuture<'a, Result<(), Error>> {
```
Blu-J, line 44: Like the name is self explanatory, nice

Blu-J, line 44: :question:  why didn't you make this fn async?

backend/src/disk/fsck.rs
```diff
@@ -39,21 +41,38 @@ pub async fn e2fsck_preen(
...
+    async move {
+        if tokio::fs::metadata(path).await.is_ok() {
+            let bak = path.with_extension(format!(
+                "{}.bak",
+                path.extension()
+                    .and_then(|s| s.to_str())
+                    .unwrap_or_default()
+            ));
+            backup_existing_undo_file(&bak).await?;
```
Blu-J, line 53: :speech_balloon: Nice that you have fixed the issue with already existing backup with a recursive method.

Blu-J (approved):

backend/src/disk/fsck.rs
dr-bonez: it is async, but recursive, so it needs to be boxed

Blu-J: Glad I asked, now I learned something

PR #1401, opened 2022-04-15
⬐ smarter wget

k0gen:
If for some reason `make` will be interrupted, wget will try to download `.zip` file and cause the name is taken it will save it to `.zip.1` which wont be recognized. This fixes it + you save some internet bandwidth by continuing the download.


dr-bonez (approved):

dr-bonez:
neat! Thanks.

PR #1412, opened 2022-04-20
refactor: isolate network toast and login redirect to separate services

waterplea:


frontend/projects/shared/src/util/misc.util.ts
```diff
@@ -14,7 +14,7 @@ export function capitalizeFirstLetter(string: string): string {
   return string.charAt(0).toUpperCase() + string.slice(1)
 }
 
-export const exists = (t: any) => {
+export function exists<T>(t: T | undefined): t is T {
```
Blu-J, line 17: :speech_balloon:  I used to be a fan of the const lambda pattern, but after debugging the function guarantee's that the function is named during debug.

frontend/projects/ui/src/app/app.component.ts
```diff
@@ -133,19 +107,15 @@ export class AppComponent {
             ])
           })
         // UNVERIFIED
-      } else if (auth === AuthState.UNVERIFIED) {
```
Blu-J, line 136: :question:  I'm looking for changes that prevent default. So the previous pattern, doesn't it make sure that all possible types are checked, if not there should be a way last time I used ts. Though this may only work with disjoint types rather than the hacked TS Enum
```ts
function assert_never(x: never) {}
let auth: AuthState = auth;
if (auth === AuthState.VERFIED) {
  // ..
} else {
  assertNever(auth)
}
```

frontend/projects/ui/src/app/app.component.ts
```diff
@@ -133,19 +107,15 @@ export class AppComponent {
...
+      } else {
```
Blu-J, line 110: :speech_balloon: This is fine, since the type is bool. I tend not to like bools, but not seeing the downside here.

frontend/projects/ui/src/app/guards/auth.guard.ts
```diff
@@ -1,38 +1,29 @@
...
-  ) {
-    this.authService.watch$()
-    .pipe(
-      tap(auth => this.authState = auth),
-    ).subscribe()
-  }
+  ) {}
 
-  canActivate (): boolean {
+  canActivate(): Observable<boolean | UrlTree> {
```
Blu-J, line 16: :eye: That's a weird type

waterplea: Angular guards can return new `UrlTree` to be redirected to or `true` to let router know the navigation is allowed by the guard.

frontend/projects/shared/src/util/misc.util.ts
waterplea: What's the benefit of using lambda const?

frontend/projects/ui/src/app/app.component.ts
waterplea: There's really just two options there, that's why I switched to boolean. I believe if there's a third one introduced — we would need to refactor anyway.

Blu-J: Thanks

frontend/projects/shared/src/util/misc.util.ts
Blu-J: None that, except that it force you to create a function before you use it. And it is consistent with other things, only sometimes. It is better with function, so :+1:

frontend/projects/ui/src/app/guards/auth.guard.ts
Blu-J: This is why I ask questions, thank you.

frontend/projects/ui/src/app/guards/auth.guard.ts
```diff
@@ -1,38 +1,29 @@
...
-  private runAuthCheck (): boolean {
-    if (this.authState === AuthState.VERIFIED) {
-      return true
-    } else {
-      this.router.navigate(['/login'], { replaceUrl: true })
-      return false
-    }
+  private runAuthCheck(): Observable<boolean | UrlTree> {
+    return this.authService.isVerified$.pipe(
+      map(verified => verified || this.router.parseUrl('/login')),
```
elvece, line 26: 👍

frontend/projects/ui/src/app/pages/login/login.page.ts
```diff
@@ -13,30 +13,24 @@ export class LoginPage {
...
     private readonly api: ApiService,
-  ) { }
+  ) {}
 
-  ngOnDestroy () {
-    if (this.loader) {
-      this.loader.dismiss()
-      this.loader = undefined
-    }
-    if (this.patchConnectionSub) {
```
elvece, line 30: why is this no longer needed here?

elvece (approved):
Some questions, new organization looks great

frontend/projects/setup-wizard/src/app/app.component.ts
```diff
@@ -29,7 +29,7 @@ export class AppComponent {
         this.stateService.isMigrating = false
         await this.navCtrl.navigateForward(`/recover`)
       }
-    } catch (e) {
+    } catch (e: any) {
```
elvece, line 32: Why are these `any` and not the error type (eg. HttpError, RpcError, EncryptionError)

elvece: Ok, I think unknown is more appropriate in this case.

elvece:
There's a bug I'm investigating here revealed in regression testing (not reproducible on mocks). When you logout, then log back in, the app list page is displaying the "welcome" message despite packages being installed. This is just on the first view of the "Services" tab - if I navigate to another tab and then back to "Services", they appear as normal.
Screen Shot 2022-05-10 at 10 06 25 PM

elvece (approved):

PR #1419, opened 2022-04-28
Update index.html

mirkoRainer:
Increase proposed time to wait.

### DIFF
--- a/index.html
+++ b/index.html
@@ -10,6 +10,6 @@
   </head>
   <body>
     <h2 style="color: #e0e0e0; text-align: center; padding-top: 20px;">EmbassyOS is initializing...</h2>
-    <p style="color: #e0e0e0; text-align: center;">Please wait several minutes and refresh the page.</p>
+    <p style="color: #e0e0e0; text-align: center;">This process can take up to several minutes to complete. Please wait to refresh the page.</p>
   </body>
 </html>
--- a/index.html
+++ b/index.html
@@ -10,6 +10,6 @@
   </head>
   <body>
     <h2 style="color: #e0e0e0; text-align: center; padding-top: 20px;">EmbassyOS is initializing...</h2>
-    <p style="color: #e0e0e0; text-align: center;">Please wait a few seconds and refresh the page.</p>
+    <p style="color: #e0e0e0; text-align: center;">Please wait several minutes and refresh the page.</p>
   </body>
 </html>


mirkoRainer:
According the the chat, this could take seconds if pre-built or bought, but minutes if built from source.
I wasn’t sure keeping it open would be appropriate if the timing wasn’t consistent.

I’ll open up again for consideration.

-Mirko
________________________________

chrisguida:
How about just "this might take several minutes if you built from source"?

chrisguida:
Assuming the person made it to this screen, they probably built from source

chrisguida:
@dr-bonez @elvece @MattDHill @ProofOfKeags surely this message could be made more accurate

index.html
```diff
@@ -10,6 +10,6 @@
   </head>
   <body>
     <h2 style="color: #e0e0e0; text-align: center; padding-top: 20px;">EmbassyOS is initializing...</h2>
-    <p style="color: #e0e0e0; text-align: center;">Please wait a few seconds and refresh the page.</p>
+    <p style="color: #e0e0e0; text-align: center;">Please wait several minutes and refresh the page.</p>
```
elvece, line 13:
```suggestion
    <p style="color: #e0e0e0; text-align: center;">This process can take up to several minutes to complete. Please wait to refresh the page.</p>
```

elvece:
Made a suggestion. @mirkoRainer @chrisguida  If this change is satisfactory to you, commit the suggestion and we can get this merged.

chrisguida:
That works. Are there any situations in which it might take a few seconds? Should we handle that?

mirkoRainer:
> That works. Are there any situations in which it might take a few seconds? Should we handle that?

From @MattDHill in the chat
"If you get an Embassy or purchase a pre-built image, it should be a few seconds"

So yes, it could be a few seconds, but I'm not sure if there's a way to programmatically detect that. Maybe a boolean flag of some sort after initial boot? 
(Assuming that the delay is caused by initialization, and pre-built/bought images boot fast because they've been booted once to validate functionality.)

Thnx @elvece for the suggestion.

chrisguida (approved):
LGTM

elvece:
> > That works. Are there any situations in which it might take a few seconds? Should we handle that?
> 
> From @MattDHill in the chat "If you get an Embassy or purchase a pre-built image, it should be a few seconds"
> 
> So yes, it could be a few seconds, but I'm not sure if there's a way to programmatically detect that. Maybe a boolean flag of some sort after initial boot? (Assuming that the delay is caused by initialization, and pre-built/bought images boot fast because they've been booted once to validate functionality.)
> 
> Thnx @elvece for the suggestion.

This very simple static html page does not have any data context, it just serves in place of the default nginx page to avoid confusion (ie. knowing that eOS is doing something and is not broken). Sure, it would be nice to have more specific information at this point, but would be a bit of development to do so. I think adjusting this message in the meantime is sufficient.

@mirkoRainer The assumption you made is correct. Perhaps create a new issue that outlines what you commented so we can track that progress separately. If you would like some direction to implement, I can also offer some suggestions :)

elvece (approved):

chrisguida:
"eOS"

stealing...

PR #1420, opened 2022-04-29
attempt to heal when health check passes

### LINKED ISSUE #1383: [bug]: health check in dependencies.yaml not fulfulling requirement even when passing
given this `dependencies.yaml` in electrs:

```
bitcoind:
  condition: '''bitcoind.type = "internal-proxy"'
  health_checks: ["synced"]
```

and this health check in bitcoind:
```
health-checks:
  synced:
    name: Synced
    success-message: Bitcoin Core is synced with the network
    type: docker
    image: main
    system: false
    entrypoint: [check-synced.sh](http://check-synced.sh/)
    args: []
    mounts: {}
    io-format: yaml
    inject: true
```

I expect electrs' bitcoind dependency to be fulfilled by a passing `synced` health check. But this is not the case:
Screen Shot 2022-04-05 at 10 28 06 AM
Screen Shot 2022-04-05 at 10 28 24 AM

dr-bonez:
fixes #1383

### DIFF
--- a/backend/src/manager/health.rs
+++ b/backend/src/manager/health.rs
@@ -5,7 +5,7 @@ use patch_db::{DbHandle, LockType};
 use tracing::instrument;
 
 use crate::context::RpcContext;
-use crate::dependencies::{break_transitive, DependencyError};
+use crate::dependencies::{break_transitive, heal_transitive, DependencyError};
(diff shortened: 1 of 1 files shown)


chrisguida (approved):
tACK

PR #1421, opened 2022-04-30
Add nginx config for proxy redirect

yzernik:
The problem

My app (https://github.com/squeaknode/squeaknode-wrapper) uses Flask for the UI interface. The Flask server uses redirects for the login/logout/index routes. When a redirect happens in the Squeaknode app, it tries to redirect to "http://..." because Flask does not know that it is behind a reverse proxy.

The solution

Add a config to Nginx so that all redirects go to the correct "https://..." url.

I tested this by manually editing the `/etc/nginx/sites-enabled/squeaknode_main_443.conf` file to include:

```
proxy_redirect http://$host/ https://$host/;
```

Then I did `systemctl reload nginx`.

After that, the Squeaknode login/logout works fine.

### DIFF
--- a/backend/src/net/nginx.conf.template
+++ b/backend/src/net/nginx.conf.template
@@ -14,6 +14,6 @@ server {{
         proxy_http_version 1.1;
         proxy_set_header Upgrade $http_upgrade;
         proxy_set_header Connection $connection_upgrade;
-        proxy_redirect http://$host/ https://$host/;
+        proxy_redirect http://$host/ $scheme://$host/;
     }}
 }}
--- a/backend/src/net/nginx.conf.template
+++ b/backend/src/net/nginx.conf.template
@@ -14,5 +14,6 @@ server {{
         proxy_http_version 1.1;
         proxy_set_header Upgrade $http_upgrade;
         proxy_set_header Connection $connection_upgrade;
+        proxy_redirect http://$host/ https://$host/;
     }}
 }}
--- a/backend/src/net/nginx.conf.template
+++ b/backend/src/net/nginx.conf.template
@@ -14,6 +14,6 @@ server {{
         proxy_http_version 1.1;
         proxy_set_header Upgrade $http_upgrade;
         proxy_set_header Connection $connection_upgrade;
-        proxy_redirect http://$host/ $scheme://$host/;
+        {proxy_redirect_directive}
     }}
 }}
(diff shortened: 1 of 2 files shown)


ProofOfKeags:
Does this occur even when you use relative URLs? I ask because I'm not sure that even if we admit this change, if your app would work properly over the tor interface. If it does I think we would be fine with accepting this change. If it doesn't, we may need to discuss a more appropriate way to fix your issue that is consistent with the behavior of the other aspects of the system.

yzernik:
Thanks @ProofOfKeags 
My app currently works fine over the Tor interface.

I tested this change by manually editing the `/etc/nginx/sites-enabled/squeaknode_main_443.conf` file on the EmbassyOS and verifying that it fixes the issue.

I double-checked that all redirects are using relative urls here:
https://github.com/squeaknode/squeaknode/blob/28acf6f4fc82289107286accf1b0d16b08e0f7ec/squeaknode/admin/webapp/app.py#L74-L110

yzernik:
The way that Nginx is currently configured, it will break any app that uses HTTP redirects.

ProofOfKeags:
Actually it occurs to me that we cannot do this as is. We need this to be interpolated by some of the OS parameters. The reason for this is that we cannot guarantee that the app we are proxying works over http and not https. This is managed by the "LAN Config" entries within the application manifest. As a result we will need to put this directive in as a template variable and interpolate it in the OS code. If you are willing to dip your toes in, I'd recommend adding the following:

`{proxy_redirect_directive}` in place of the line you put here. From there if you open up the [embassyd](/src/backend) code, your typical rust infrastructure will point you to where changes need to be made. If you get stuck, feel free to ask questions here and I'll try to shed some light on what is going on.

yzernik:
The way I understand `proxy_redirect` is that is intercepts the `Location` header in the 3xx response from the server, and replaces the location using the given patterns.

So if the Location header already has a "https://" location, this config won't change the header.

ProofOfKeags:
> Is there any app that uses https for the web UI that I can use to test this?

I believe all of them, but the simplest ones to test will be filebrowser, and burn after reading.

dr-bonez:
Yeah, looks like we only want to include this if `ssl: true`

yzernik:
I updated the code, so now it only includes the proxy redirect directive when `ssl` is true.

I haven't tested it yet. I'm not sure if I can just re-flash my micoSD card and put it back in the Raspberry Pi?

chrisguida:
@yzernik yep, that's all you do

yzernik:
I just tested this on my Raspberry Pi:
- [x] Squeaknode app works
- [x] File Browser app works

dr-bonez (approved):

PR #1449, opened 2022-05-16
attempted fix for auth lockup

dr-bonez:
@elvece try this


elvece:
@dr-bonez I remember testing this at the time, but I forget what the original issue was, do you recall?

PR #1456, opened 2022-05-20
Fix/making js work

Blu-J:


backend/src/procedure/js_scripts.rs
```diff
@@ -346,26 +350,29 @@ mod js_runtime {
                 .build();
 
             let loader = std::rc::Rc::new(self.module_loader.clone());
-            let mut runtime = JsRuntime::new(RuntimeOptions {
+            let runtime_options = RuntimeOptions {
                 module_loader: Some(loader),
                 extensions: vec![ext],
+                startup_snapshot: Some(Snapshot::Static(SNAPSHOT_BYTES)),
```
Blu-J, line 356: This is the one line actual fix btw

Makefile
```diff
@@ -1,6 +1,7 @@
 EMBASSY_BINS := backend/target/aarch64-unknown-linux-gnu/release/embassyd backend/target/aarch64-unknown-linux-gnu/release/embassy-init backend/target/aarch64-unknown-linux-gnu/release/embassy-cli backend/target/aarch64-unknown-linux-gnu/release/embassy-sdk
 EMBASSY_UIS := frontend/dist/ui frontend/dist/setup-wizard frontend/dist/diagnostic-ui
 EMBASSY_SRC := raspios.img product_key.txt $(EMBASSY_BINS) backend/embassyd.service backend/embassy-init.service $(EMBASSY_UIS) $(shell find build)
+EMBASSY_V8_SNAPSHOTS := backend/src/procedure/js_scripts/AMR_JS_SNAPSHOT.bin
```
dr-bonez, line 4:
```suggestion
EMBASSY_V8_SNAPSHOTS := backend/src/procedure/js_scripts/ARM_JS_SNAPSHOT.bin
```

backend/src/procedure/js_scripts.rs
```diff
@@ -120,11 +121,16 @@ mod js_runtime {
 
     use crate::s9pk::manifest::PackageId;
     use crate::util::Version;
-    use crate::volume::{script_dir, Volumes};
+    use crate::volume::script_dir;
+    use crate::volume::Volumes;
 
     use super::super::ProcedureName;
     use super::{JsCode, JsError};
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
```
dr-bonez, line 129: These probably aren't compatible with eachother. So the snapshot is likely x86_64 only.

backend/src/procedure/js_scripts.rs
```diff
@@ -120,11 +121,16 @@ mod js_runtime {
...
+    const SNAPSHOT_BYTES: &[u8] = include_bytes!("./js_scripts/JS_SNAPSHOT.bin");
 
+    #[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
```
dr-bonez, line 132: Same here. Probably just aarch64

dr-bonez (changes requested):

backend/src/procedure/js_scripts.rs
Blu-J: K

backend/src/procedure/js_scripts.rs
```diff
@@ -346,26 +350,29 @@ mod js_runtime {
...
-                evaluated.await??;
                 res?;
+                evaluated.await??;
                 Ok::<_, AnyError>(())
             };
 
             tokio::runtime::Handle::current()
                 .block_on(future)
                 .map_err(|e| {
+                    dbg!(&e);
```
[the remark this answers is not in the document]
Blu-J: Good eye

backend/src/procedure/js_scripts.rs
```diff
@@ -605,8 +615,49 @@ mod js_runtime {
             let ctx = state.borrow::<JsContext>();
             Ok(ctx.sandboxed)
         }
+
+        /// We need to make sure that during the file accessing, we don't reach beyond our scope of control
+        fn is_subset(parent: impl AsRef<Path>, child: impl AsRef<Path>) -> bool {
```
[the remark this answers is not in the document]
Blu-J: If you have a better solution, that would be lovely. It is a marked improvement from the previous.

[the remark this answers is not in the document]
Blu-J: But will bring back the canoncialize since there is a possibility that one could create a link, and following that link would allow modification of another volume, maybe. But I don't want to take that chance so using the canonicalize now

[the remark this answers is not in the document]
dr-bonez: This looks *more* permissive than before. You're checking prefixes of path segments, doesn't that mean /foo/baz would match against /foobar/baz? That doesn't make sense to me.

backend/src/procedure/js_scripts.rs
```diff
@@ -605,12 +649,57 @@ mod js_runtime {
...
+            let child = normalize_path(child);
+            let parent = normalize_path(parent);
+            child
+                .zip(parent)
+                .map(|(child, parent)| child.starts_with(parent))
+                .unwrap_or(false)
+        }
+
+        /// Used during the subset, we are collapsing the state of the file paths. This will not work with soft link,
+        /// and the correct method, canonicalizePaths seems to freeze randomly, and is much slower.
```
dr-bonez, line 664: This is concerning, we use canonicalize in several places. Evidence?

dr-bonez (approved):

backend/src/procedure/js_scripts.rs
Blu-J: Issue was that the freezing was the order of checking the error from javascript was wrong. Fixed, and made sure the is parent is safer, slower but safer.

dr-bonez (approved):

PR #1464, opened 2022-05-27
Strict null checks

waterplea:


elvece:
Few bugs discovered:

1. Notifications do not load in associated tab despite notifications existing (reproducible in mocks).
2. In mocks, marketplace tab will not load.
3. When a service is uninstalled, it is quickly removed from the service list page, but if I navigate to that service again in the marketplace, it says it is in "Removing" state. Same if I click the "View Service" button from the marketplace page. This is fixed by a page refresh. Reproducible live, and might be reproducible in mocks, but blocked by #2.

frontend/projects/ui/src/app/services/marketplace.service.ts
```diff
@@ -27,33 +30,49 @@ import {
...
     switchMap(({ url }) =>
-      from(this.getMarketplaceData({ 'server-id': this.serverInfo.id }, url)),
+      this.serverInfo$.pipe(
+        switchMap(({ id }) =>
+          from(this.getMarketplaceData({ 'server-id': id }, url)),
+        ),
+      ),
     ),
-    shareReplay(),
+    map(({ categories }) => categories),
```
elvece, line 53: why do we need to spread this instead?

frontend/projects/shared/src/util/unused.ts
```diff
@@ -63,31 +63,6 @@ export function toObject<T>(t: T[], map: (t0: T) => string): Record<string, T> {
   }, {} as Record<string, T>)
 }
 
-export function deepCloneUnknown<T>(value: T): T {
```
elvece, line 66: all these were deleted because unused?

frontend/projects/ui/src/app/services/marketplace.service.ts
```diff
@@ -27,33 +30,49 @@ import {
...
     .watch$('ui', 'marketplace')
-    .pipe(
-      map(marketplace =>
-        marketplace?.['selected-id']
-          ? marketplace['known-hosts'][marketplace['selected-id']]
-          : this.config.marketplace,
-      ),
-      shareReplay(),
-    )
+    .pipe(take(1), shareReplay({ bufferSize: 1, refCount: true }))
```
elvece, line 35: with `bufferSize` and `refCount`, this essentially enables one subscription?

elvece:
Remaining bug: When switch marketplaces, the header and marketplace data do not get updated unless I refresh the page

frontend/projects/ui/src/app/services/marketplace.service.ts
waterplea: This enables auto unsubscribe if all subscribers have unsubscribed.

elvece (approved):
🎉

PR #1475, opened 2022-05-31
Chore/version 0 3 1 0

Blu-J:
Chore: Make the build faster by breaking out the v8 engine building to not require of all embassy os
Chore: make the feature for the v8 engine optional
Chore: Update the version to 3.1


backend/src/version/v0_3_1.rs
```diff
@@ -0,0 +1,36 @@
+use emver::VersionRange;
+
+use super::*;
+
+const V0_3_1: emver::Version = emver::Version::new(0, 3, 1, 0);
+lazy_static::lazy_static! {
+    pub static ref V0_3_1_COMPAT: VersionRange = VersionRange::Conj(
+        Box::new(VersionRange::Anchor(
+            emver::GTE,
+            emver::Version::new(0, 3, 1, 0),
```
dr-bonez, line 10: We're still compatible with 0.3.0 packages

dr-bonez (changes requested):

dr-bonez (approved):

PR #1481, opened 2022-06-03
Fix/patch db unwrap remove

Blu-J:
## About

Thanks for @chrisguida for finding the bug. We tracked it down to the wifi not having the correct fallback. So the patch-db needs to have the unwrap removed, then I propagated the type and dealt with the cases of None. I then did a flatten for the wifi, since the state of wifi could be none, and there didn't need to be an error.

## Proof Of Fix

@chrisguida got his machine working when he built from this branch.

## Related PR

https://github.com/Start9Labs/patch-db/pull/35


chrisguida:
This fix got my box working. Thanks @Blu-J !

Still has this frontend bug, but that's hopefully being fixed in another issue? @elvece ?

Screen Shot 2022-06-02 at 9 48 50 PM

^this just keeps popping up no matter how many times i restart. finally got it to go away by clicking outside of the modal

dr-bonez (approved):

Issue #62, opened 2020-09-15
Change password


MattDHill:
Complicated because it has implications on creating and restoring backups. Also disk encryption.

PR #1490, opened 2022-06-06
Reset password through setup wizard

MattDHill:
When user chooses to attach existing Embassy data drive, they are prompted to enter a password that will become their new master password.


dr-bonez:
closes #62

elvece (approved):

PR #1496, opened 2022-06-06
fix: potential fix for the docker leaking the errors and such

Blu-J:
## About

We are trying to solve that we where seeing a memory leak where certain packages where filling the memory over time.

## Solution Limitation

Right now because we are reading the lines in the read buffer there could be a possibility that the lines are gigantic, and would eat up memory


dr-bonez (approved):

dr-bonez:
might eventually want to move RingVec to utils

PR #1501, opened 2022-06-07
allow embassy-cli not as root

### LINKED ISSUE #523: Embassy SDK fails to pack/verify and gives inadequate error information
Filebrowser not packing. 

tried on m1 mac and linux x86, both say

`Filesystem I/O Error: No such file or directory (os error 2)`

Reported by @chrisguida

### LINKED ISSUE #1373: [bug]: [cli]: Cannot sideload using embassy-cli remotely
### EmbassyOS Version

cb5bb34ed82e7991d24f7cb5ad0ae37146b40c10

### Device

Laptop/Desktop

### Device OS

MacOS

### Device OS Version

11.6.4

### Current Behavior

CLI is hard/impossible to use remotely on Mac. Haven't tested on Linux.

1. Install `embassy-cli` at the above hash
2. create `/etc/embassy/config.yaml` and put the following in it: `host: "https://embassy-30297ba3.local"` (doesn't work without quotes)
3. `sudo chown -R $(whoami) /etc/embassy`
4. Trust root cert in system keychain
5. `embassy-cli auth login` and paste password
6. Bug number 1: embassy-cli tries to connect to localhost despite my embassy's hostname being defined in the config file

```
% embassy-cli auth login                                  
Password: 
General error: HTTP Error: error sending request for url (http://127.0.0.1/rpc/v1): error trying to connect: tcp connect error: Connection refused (os error 61)

```
7. `embassy-cli -h https://embassy-30297ba3.local/ auth login`
8. That seems to work
9. Bug number 2:
```
% embassy-cli -h https://embassy-30297ba3.local package install electrs.s9pk 
S9PK Parsing Error: Deserializing Manifest (CBOR): Semantic(None, "missing field `description`")
```

I can install this exact s9pk just fine by `scp`ing it to my box and sideloading.

### Expected Behavior

1. `embassy-cli` should honor the hostname set in the config so I don't have to use the `-h` flag
2. `embassy-cli install` should install an s9pk that otherwise works.

dr-bonez:
fixes #523 
fixes #1373

### DIFF
--- a/backend/src/context/sdk.rs
+++ b/backend/src/context/sdk.rs
@@ -1,5 +1,3 @@
-use std::fs::File;
-use std::io::Read;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
@@ -9,6 +7,7 @@ use rpc_toolkit::Context;
 use serde::Deserialize;
 use tracing::instrument;
 
+use crate::util::config::{load_config_from_paths, local_config_path};
 use crate::{Error, ResultExt};
 
 #[derive(Debug, Default, Deserialize)]
@@ -28,22 +27,24 @@ impl SdkContext {
     /// BLOCKING
     #[instrument(skip(matches))]
     pub fn init(matches: &ArgMatches) -> Result<Self, crate::Error> {
-        let cfg_path = Path::new(matches.value_of("config").unwrap_or(crate::CONFIG_PATH));
-        let base = if cfg_path.exists() {
-            serde_yaml::from_reader(
-                File::open(cfg_path)
-                    .with_ctx(|_| (crate::ErrorKind::Filesystem, cfg_path.display().to_string()))?,
-            )
-            .with_kind(crate::ErrorKind::Deserialization)?
-        } else {
-            SdkContextConfig::default()
-        };
+        let local_config_path = local_config_path();
+        let base: SdkContextConfig = load_config_from_paths(
+            matches
+                .values_of("config")
+                .into_iter()
+                .flatten()
+                .map(|p| Path::new(p))
+                .chain(local_config_path.as_deref().into_iter())
+                .chain(std::iter::once(Path::new(crate::util::config::CONFIG_PATH))),
+        )?;
         Ok(SdkContext(Arc::new(SdkContextSeed {
             developer_key_path: base.developer_key_path.unwrap_or_else(|| {
-                cfg_path
+                local_config_path
+                    .as_deref()
+                    .unwrap_or_else(|| Path::new(crate::util::config::CONFIG_PATH))
                     .parent()
                     .unwrap_or(Path::new("/"))
-                    .join(".developer_key")
+                    .join("developer.key.pem")
             }),
         })))
     }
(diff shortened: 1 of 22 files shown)


backend/src/context/sdk.rs
```diff
@@ -28,22 +27,24 @@ impl SdkContext {
...
-            serde_yaml::from_reader(
-                File::open(cfg_path)
-                    .with_ctx(|_| (crate::ErrorKind::Filesystem, cfg_path.display().to_string()))?,
-            )
-            .with_kind(crate::ErrorKind::Deserialization)?
-        } else {
-            SdkContextConfig::default()
-        };
+        let local_config_path = local_config_path();
+        let base: SdkContextConfig = load_config_from_paths(
```
Blu-J, line 31: :speech_balloon: :question: So just to verify, this is a pattern of optional. to iterator, then adding on the fallbacks, so a iterator of possible paths, interesting...

backend/src/developer/mod.rs
```diff
@@ -22,8 +23,17 @@ pub fn init(#[context] ctx: SdkContext) -> Result<(), Error> {
         tracing::info!("Generating new developer key...");
         let keypair = Keypair::generate(&mut rand::thread_rng());
         tracing::info!("Writing key to {}", ctx.developer_key_path.display());
+        let keypair_bytes = ed25519::KeypairBytes {
```
Blu-J, line 26: :notebook:  Part of new auth, ok

backend/src/error.rs
```diff
@@ -31,7 +31,7 @@ pub enum ErrorKind {
     InvalidOnionAddress = 22,
     Pack = 23,
     ValidateS9pk = 24,
-    DiskCorrupted = 25,
+    DiskCorrupted = 25, // Remove
```
Blu-J, line 34: :question: Why the remove now? Is the `deprecated` better than remove?

backend/src/util/config.rs
```diff
@@ -0,0 +1,58 @@
...
+
+pub fn local_config_path() -> Option<PathBuf> {
+    if let Ok(home) = std::env::var("HOME") {
+        Some(Path::new(&home).join(CONFIG_PATH_LOCAL))
+    } else {
+        None
+    }
+}
+
+/// BLOCKING
```
Blu-J, line 21: Thanks for the note

backend/src/util/config.rs
```diff
@@ -0,0 +1,58 @@
...
+pub fn load_config_from_paths<'a, T: for<'de> Deserialize<'de>>(
+    paths: impl IntoIterator<Item = impl AsRef<Path>>,
+) -> Result<T, Error> {
+    let mut config = Default::default();
+    for path in paths {
```
Blu-J, line 26: :question: So, if we are given multiple configs, will take the last?

backend/src/util/config.rs
```diff
@@ -0,0 +1,58 @@
...
+    }
+    serde_json::from_value(Value::Object(config)).with_kind(crate::ErrorKind::Deserialization)
+}
+
+pub fn merge_configs(mut first: Config, second: Config) -> Config {
+    for (k, v) in second.into_iter() {
+        if !first.contains_key(&k) {
+            first.insert(k, v);
+        } else {
+            let old = first.remove(&k).unwrap(); // because in else of !contains_key
```
Blu-J, line 47: :question: Wouldn't this better to be a match, then there would be no unwrap. I know this is safe, for now, but things change.
```rs
match first.remove(&k) {
  None => first.insert(k, v),
  Some(old) => {
    let new = match (old, v) {
                (Value::Object(first), Value::Object(second)) => {
                    Value::Object(merge_configs(first, second))
                }
                (first, _) => first,
            };
            first.insert(k, new);
  }
}

backend/src/context/sdk.rs
dr-bonez: yes

backend/src/error.rs
dr-bonez: it has never been used. and I was just copying the comment from below

Blu-J (approved):

PR #1505, opened 2022-06-07
Fix/memory leak docker

Blu-J:
## About

Found issues during the memory leak fix

## Proof 

This was having an issue, now returns
image
image

And as for the only error is also persistent in pre - memory leak fix
image

### DIFF
--- a/backend/src/procedure/docker.rs
+++ b/backend/src/procedure/docker.rs
@@ -17,7 +17,7 @@ use nix::unistd::Pid;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use std::collections::VecDeque;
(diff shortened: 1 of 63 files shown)


backend/src/procedure/docker.rs
```diff
@@ -479,27 +475,28 @@ async fn buf_reader_to_lines(
     Ok(output)
 }
 
-async fn max_buffer(
+async fn max_by_lines(
```
Blu-J, line 478: So make it so I don't have to go to vec to early and I avoided the issue that the buffer was wrong for the serialize, so glad that got fixed. And this makes more sense to keep it in string

dr-bonez (approved):

PR #1506, opened 2022-06-07
fixes a bug where nginx will crash if eos goes into diagnostic mode a…

dr-bonez:
…fter service init has completed


backend/src/nginx/diagnostic-ui.conf
```diff
@@ -1,3 +1,9 @@
+
+map $http_upgrade $connection_upgrade {
```
Blu-J, line 2: :disappointed:  I don't know enough about this conf to say if it is good or not. Would like a proof that it did work.

PR #1510, opened 2022-06-08
Adds typescript bundling to embassy-sdk pack flow

ProofOfKeags:
This PR bundles typescript procedures if `scripts/embassy.ts` exists and `scripts/embassy.js` does not.

This is a complete feature as is, but does less than we would have anticipated.

NOTE: this does NOT do any typechecking. This is a limitation of deno_emit and the typechecking story is going to be something that requires discussion.

### DIFF
--- a/backend/src/s9pk/mod.rs
+++ b/backend/src/s9pk/mod.rs
@@ -4,7 +4,6 @@ use color_eyre::eyre::eyre;
 use imbl::OrdMap;
 use rpc_toolkit::command;
 use serde_json::Value;
-use tokio::io::{AsyncRead, AsyncReadExt};
 use tracing::instrument;
 
 use crate::context::SdkContext;
@@ -61,6 +60,7 @@ pub async fn pack(#[context] ctx: SdkContext, #[arg] path: Option<PathBuf>) -> R
 
     let outfile_path = path.join(format!("{}.s9pk", manifest.id));
     let mut outfile = File::create(outfile_path).await?;
+    let key: ed25519_dalek::Keypair = ctx.developer_key()?;
     S9pkPacker::builder()
         .manifest(&manifest)
         .writer(&mut outfile)
@@ -170,7 +170,7 @@ pub async fn pack(#[context] ctx: SdkContext, #[arg] path: Option<PathBuf>) -> R
             }
         })
         .build()
-        .pack(&ctx.developer_key()?)
+        .pack(&key)
         .await?;
     outfile.sync_all().await?;
 
--- a/backend/src/s9pk/mod.rs
+++ b/backend/src/s9pk/mod.rs
@@ -4,6 +4,7 @@ use color_eyre::eyre::eyre;
 use imbl::OrdMap;
 use rpc_toolkit::command;
 use serde_json::Value;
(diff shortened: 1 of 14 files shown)


backend/src/s9pk/mod.rs
```diff
@@ -122,23 +123,52 @@ pub async fn pack(#[context] ctx: SdkContext, #[arg] path: Option<PathBuf>) -> R
...
-                (false, false) => None
+                (false, false, true) => {
+                    tracing::warn!("Manifest does not declare any actions that use scripts, but a script exists at ./scripts/embassy.ts");
+                    None
+                }
+                (true, false, false) => {
+                    return Err(Error::new(eyre!("Script is declared in manifest, but no such script exists at ./scripts/embassy.js"), ErrorKind::Pack))
+                }
+                (true, true, true) => {
+                    tracing::warn!("Both ./scripts/embassy.js and ./scripts/embassy.ts exist. Using embassy.js, skipping Typescript bundling...");
```
Blu-J, line 144: I'm 100% against the warning while the ts bundling in not working/ validating.

libs/ts-bundler/src/loader/bootstrap.ts
```diff
@@ -0,0 +1,15 @@
+import * as E from "./embassy.ts";
+import * as T from "./types.d.ts";
+type Maybe<T> = T | undefined;
```
Blu-J, line 3:
```suggestion
export * from './embassy.ts'
type Maybe<T> = T | undefined;
```

backend/src/s9pk/mod.rs
Blu-J, line 144:
```suggestion
```
I'm 100% against the warning while the ts bundling in not working/ validating.

Blu-J (changes requested):
These are my requests. Change or explain why not. I don't agree with this feature as is, but
I won't stop the pr

libs/ts-bundler/src/loader/bootstrap.ts
ProofOfKeags: Do the later exports override the ones with the same names from ./embassy.ts?

backend/src/s9pk/mod.rs
ProofOfKeags: Are you saying you want to disable the embassy.ts? I think giving a warning that tells the user how the thing is choosing to disambiguate a request is a very important idea. I would hazard a guess that your beef isn't with the warning but about the actual behavior. Can you describe what you'd like to be different about ti?

libs/ts-bundler/src/loader/bootstrap.ts
Blu-J: Yes, the later exports override.

backend/src/s9pk/mod.rs
Blu-J: Right now, with no checking, I feel that it would be better to bundle via `deno bundle` or `webpack pack` or `rollup`
It would be better because then they are getting the bundler's ability to lint and typecheck.

Blu-J:
I'm closing this PR until we can get typechecking.
I believe that the bundling by other tools like deno is sufficient hold over for now.

PR #1514, opened 2022-06-08
Feat/js known errors

Blu-J:
fixes: https://github.com/Start9Labs/embassy-os/issues/1512

## About

We where hoping that the return type could also be or'ed with a known error shape as a return. Indicating to rust that we know that there is an expected error. Other errors are things like the infamous using method of undefined. Or something else we didn't catch or mind that we will say uncaught.

## Proof of working

Now the error looks like 
image

When the code looked like 
```ts
type KnownError = { error: string };

export async function setConfig(
  effects: Effects,
  input: Config,
): Promise<SetResult | KnownError> {
  return { error: "This is a test" };
}
```


Blu-J:
Updated to use the agreed upon types in our research internal

Blu-J:
image
image

ProofOfKeags (approved):

PR #1527, opened 2022-06-13
fix: restart/ uninstall sometimes didn't work

Blu-J:
## About

When trying to recreate locally, noticing that there are locking issues with updating/ uninstalling/ installing some of the packages. 

https://github.com/Start9Labs/embassy-os/issues/1526

## What

This solution is migrating more to the bulk locking, since it was conflicting with locking reasons.

### DIFF
--- a/backend/src/install/update.rs
+++ b/backend/src/install/update.rs
@@ -1,20 +1,66 @@
 use std::collections::BTreeMap;
 
-use patch_db::{DbHandle, LockType};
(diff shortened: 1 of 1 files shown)


Blu-J:
### Tested

Got my local box to no longer have the issues, and only 2 of my services had issues, bitcoin and the lnd

Also tentatively looks like it might be solved when double checking with @kn0wmad 
We still haven't double checked with the box that did have the issues.

Issue #368, opened 2021-07-15
Add restart button to service details page


kn0wmad:
Add a restart button so that it is no longer necessary to first Stop and then Start a service again in order to restart it

elvece:
Add an RPC command on the backend for `restart` which will:

- preform a stop without changing sync status
- waits for stop to successfully complete
- then run start

Call this new rpc command from the front end.

MattDHill:
Adding Aiden for BE portion

PR #1536, opened 2022-06-14
add restart rpc and status

dr-bonez:
backend portion of #368 

this adds a `restarting` status

### DIFF
--- a/backend/src/manager/sync.rs
+++ b/backend/src/manager/sync.rs
@@ -31,6 +31,9 @@ async fn synchronize_once(shared: &ManagerSharedState) -> Result<Status, Error> 
             MainStatus::Stopping => {
                 *status = MainStatus::Stopped;
             }
+            MainStatus::Restarting => {
+                *status = MainStatus::Starting;
+            }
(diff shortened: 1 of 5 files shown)


backend/src/manager/sync.rs
```diff
@@ -41,7 +44,7 @@ async fn synchronize_once(shared: &ManagerSharedState) -> Result<Status, Error>
             MainStatus::BackingUp { .. } => (),
         },
         Status::Starting => match *status {
-            MainStatus::Stopped | MainStatus::Stopping => {
+            MainStatus::Stopped | MainStatus::Stopping | MainStatus::Restarting => {
```
elvece, line 47: are these all potential cases of from -> to states?

dr-bonez: kind of. The `Status` states are the state of the Manager, the `MainStatus` is the desired state of the package.

Blu-J (approved):
I'm trusting that rust has our back, looks like a rust will check. Thank you enums

elvece (approved):

PR #1539, opened 2022-06-14
chore: Update types to match embassyd

Blu-J:
While working on the @kn0wmad converting the https://github.com/Start9Labs/bitcoind-wrappe to use the js, had to update the types to match better.

### DIFF
--- a/libs/artifacts/types.d.ts
+++ b/libs/artifacts/types.d.ts
@@ -156,7 +156,7 @@ export type ValueSpecAny =
           WithDefault<
               {
                   values: string[];
-                  "value-names"?: {
+                  "value-names": {
                       [key: string]: string;
                   };
               },
@@ -211,7 +211,7 @@ export type ValueSpecUnion = {
       id: string;
       name: string;
       description?: string;
-      "variant-names"?: {
+      "variant-names": {
           [key: string]: string;
       };
   };
@@ -246,7 +246,7 @@ export type ValueSpecList =
           WithDefault<
               {
                   values: string[];
-                  "value-names"?: {
+                  "value-names": {
                       [key: string]: string;
                   };
               },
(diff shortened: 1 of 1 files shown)


libs/artifacts/types.d.ts
```diff
@@ -229,70 +229,75 @@ export type ValueSpecObject = {
...
+                  "value-names"?: {
+                      [key: string]: string;
+                  };
+              },
+              string[]
+          >
       >
-    >
+  >
+  | Subtype<
```
Blu-J, line 257: :speech_balloon: Missed this, so this was added.

libs/artifacts/types.d.ts
```diff
@@ -143,81 +143,81 @@ export type ValueSpecBoolean = {};
...
-            [key: string]: string;
-          };
-        },
-        string
+      "enum",
+      WithDescription<
+          WithDefault<
+              {
+                  values: string[];
+                  "value-names"?: {
```
Blu-J, line 159: :speech_balloon: Made the "value-names" conditional, because of how default works in serde and previous types of our config.yaml match this model better.

dr-bonez: it is only optional on the backend for backwards compatibility. It is preferred to be explicit

libs/artifacts/types.d.ts
```diff
@@ -143,81 +143,81 @@ export type ValueSpecBoolean = {};
...
-    id: string;
-    name: string;
-    description?: string;
-    "variant-names": {
-      [key: string]: string;
-    };
+      id: string;
+      name: string;
+      description?: string;
+      "variant-names"?: {
```
dr-bonez, line 214: same here

Blu-J (commented):
chore: revert the optional

Blu-J (commented):
chore: revert the optional

libs/artifacts/types.d.ts
Blu-J: reverted

dr-bonez (approved):

PR #1541, opened 2022-06-15
Remove app wiz and dry calls

MattDHill:


elvece (approved):
Reviewed more in depth in person

PR #1542, opened 2022-06-15
chore: Update the lite types to include the union and enum

Blu-J:
### About

Add the types for the union and enum

Rust looked like 
image

And the d.ts looks like
image


Blu-J:
Tested with bitcoind wrapper and the bitcoin proxy

dr-bonez (approved):

PR #1544, opened 2022-06-16
Re-mask generic input fields

elvece:
This PR addresses masking for input fields that use the `generic-input-component`.


elvece:
Closing this without merging as separating the two implementations out into two separate tickets

PR #1552, opened 2022-06-17
chore: Add tracing for debuging the js procedure slowness

Blu-J:
While debuging why configSet was slow, was using this trace. @chrisguida  thought it would be nice if others could trace this too.

### DIFF
--- a/backend/src/procedure/mod.rs
+++ b/backend/src/procedure/mod.rs
@@ -64,15 +64,7 @@ impl PackageProcedure {
         allow_inject: bool,
         timeout: Option<Duration>,
     ) -> Result<Result<O, (i32, String)>, Error> {
-        tracing::trace!(
-            "Procedure execute {} {} - {:?}",
-            match self {
-                PackageProcedure::Docker(_) => "docker",
-                PackageProcedure::Script(_) => "JS",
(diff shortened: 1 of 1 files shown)


chrisguida:
This will be extremely useful during our switch from docker to JS

backend/src/procedure/mod.rs
```diff
@@ -64,6 +64,15 @@ impl PackageProcedure {
...
     ) -> Result<Result<O, (i32, String)>, Error> {
+        tracing::trace!(
+            "Procedure execute {} {} - {:?}",
+            match self {
+                PackageProcedure::Docker(_) => "docker",
+                PackageProcedure::Script(_) => "JS",
+            },
+            pkg_id,
+            name
+        );
```
ProofOfKeags, line 75: yeesh this is a lot of visual space for a trace log. Is there any way to make it more innocuous?
I may encourage display instances for `Package Procedure` and then you can eliminate the case statement and probably smash this to a single line.

Blu-J: Thanks for the idea. It did reduce the vertical garbage

ProofOfKeags (approved):

PR #1570, opened 2022-06-23
Fix/mask generic inputs

elvece:
Tested on live box (not through proxy). Works perfectly as expected.


frontend/projects/ui/src/app/modals/generic-input/generic-input.component.ts
```diff
@@ -52,6 +77,7 @@ export class GenericInputComponent {
     try {
       await this.options.submitFn(value)
       this.modalCtrl.dismiss(undefined, 'success')
+      this.clearInput()
```
MattDHill, line 80: Is this necessary? Are these values persisting after modal destroy?

frontend/projects/ui/src/app/modals/generic-input/generic-input.component.ts
```diff
@@ -1,20 +1,26 @@
 import { Component, Input, ViewChild } from '@angular/core'
 import { ModalController, IonicSafeString, IonInput } from '@ionic/angular'
 import { getErrorMessage } from '@start9labs/shared'
+import { MaskPipe } from 'src/app/pipes/mask/mask.pipe'
 
 @Component({
   selector: 'generic-input',
   templateUrl: './generic-input.component.html',
   styleUrls: ['./generic-input.component.scss'],
+  providers: [MaskPipe],
```
MattDHill, line 10: You shouldn't need this

frontend/projects/ui/src/app/modals/generic-input/generic-input.component.ts
elvece: Never tested without it - would have to confirm.

frontend/projects/ui/src/app/modals/generic-input/generic-input.component.ts
elvece: Why?

frontend/projects/ui/src/app/modals/generic-input/generic-input.component.module.ts
```diff
@@ -5,6 +5,7 @@ import { IonicModule } from '@ionic/angular'
 import { RouterModule } from '@angular/router'
 import { SharedPipesModule } from '@start9labs/shared'
 import { FormsModule } from '@angular/forms'
+import { MaskPipeModule } from 'src/app/pipes/mask/mask.module'
```
MattDHill, line 8: I'm not sure you need to import this at all, since you are not using the pipe in html. Could be wrong. Try it.

frontend/projects/ui/src/app/modals/generic-input/generic-input.component.html
```diff
@@ -22,14 +22,25 @@ <h1>{{ options.title }}</h1>
...
             (ionChange)="error = ''"
           ></ion-input>
-          <!-- removing mask button until proper solution implemented -->
-          <!-- <ion-button
+          <ion-input
+            #mainInput
+            *ngIf="masked"
+            type="text"
+            name="value"
+            [ngModel]="maskedValue"
```
waterplea, line 38: You can just do `[ngModel]="masked ? maskedValue : value"`

frontend/projects/ui/src/app/modals/generic-input/generic-input.component.ts
```diff
@@ -37,13 +42,27 @@ export class GenericInputComponent {
...
 
   cancel() {
     this.modalCtrl.dismiss()
   }
 
+  transformInput(newValue: string) {
+    let i = 0
+    this.value = newValue
+      .split('')
+      .map(x => {
```
waterplea, line 56: `x => x === '●' ? this.value[i++] : x` for shorts :) Why does it return next char from old value for each '●' in new value though?

frontend/projects/ui/src/app/modals/generic-input/generic-input.component.ts
MattDHill: MaskPipe was needed here in order to use it in the component rather than the template

MattDHill (approved):

frontend/projects/ui/src/app/modals/generic-input/generic-input.component.module.ts
```diff
@@ -14,6 +15,7 @@ import { FormsModule } from '@angular/forms'
     FormsModule,
     RouterModule.forChild([]),
     SharedPipesModule,
+    MaskPipeModule,
```
waterplea, line 18: Pretty sure you don't need that when you provide and inject pipe like a service.

frontend/projects/ui/src/app/modals/generic-input/generic-input.component.ts
elvece: This is so we can accurately persist the actual text value when it is displaying as the masked value as the user is typing. Ie. `newValue` could be `maskedValue` (just dots)  or `value` (plaintext) depending if input is toggled to masked or not.

PR #1571, opened 2022-06-23
adjust service marketplace button for installation source relevance

elvece:
Tested working via FE proxy to backend.

### DIFF
--- a/frontend/projects/shared/styles/shared.scss
+++ b/frontend/projects/shared/styles/shared.scss
@@ -45,13 +45,13 @@ ion-modal::part(content) {
     position: absolute;
     height: 80% !important;
     top: 10%;
-    width: $wide-modal !important;
+    width: $wide-modal  !important;
     left: calc((100vw - $wide-modal) / 2) !important;
   }
 
   .alertlike-modal {
     &::part(content) {
-      width: $wide-modal !important;
+      width: $wide-modal  !important;
       left: calc((100vw - $wide-modal) / 2) !important;
     }
   }
@@ -110,4 +110,12 @@ ion-modal::part(content) {
 
 .montserrat {
   font-family: 'Montserrat', sans-serif !important;
+}
+
+.color-success-shade {
+  color: var(--ion-color-success-shade)
+}
+
+.color-primary-shade {
+  color: var(--ion-color-primary-shade)
 }
--- a/frontend/projects/shared/styles/shared.scss
+++ b/frontend/projects/shared/styles/shared.scss
@@ -45,13 +45,13 @@ ion-modal::part(content) {
     position: absolute;
     height: 80% !important;
     top: 10%;
-    width: $wide-modal  !important;
+    width: $wide-modal !important;
     left: calc((100vw - $wide-modal) / 2) !important;
   }
 
   .alertlike-modal {
     &::part(content) {
-      width: $wide-modal  !important;
+      width: $wide-modal !important;
       left: calc((100vw - $wide-modal) / 2) !important;
     }
   }
--- a/frontend/projects/ui/src/app/pages/apps-routes/app-show/app-show.page.ts
+++ b/frontend/projects/ui/src/app/pages/apps-routes/app-show/app-show.page.ts
@@ -14,15 +14,7 @@ import {
   ConnectionFailure,
   ConnectionService,
 } from 'src/app/services/connection.service'
-import {
-  map,
-  startWith,
-  filter,
-  first,
-  distinctUntilChanged,
-  switchMap,
-  skipWhile,
-} from 'rxjs/operators'
+import { map, startWith, filter, first } from 'rxjs/operators'
 import { ActivatedRoute } from '@angular/router'
 import { getPkgId } from '@start9labs/shared'
 import { MarketplaceService } from 'src/app/services/marketplace.service'
@@ -67,19 +59,8 @@ export class AppShowPage {
     ),
   )
 
-  skipNull =
-    () =>
-    <T>(source: Observable<T>): Observable<T> =>
-      source.pipe(skipWhile((value: any) => value === null))
-
   readonly currentMarketplace$: Observable<Marketplace> =
-    this.marketplaceService.getMarketplace().pipe(
-      // switchMap(m => m ? m : EMPTY),
-      // distinctUntilChanged(),
-      // filter<Marketplace>(Boolean),
-      this.skipNull(),
-      first(),
-    )
+    this.marketplaceService.getMarketplace().pipe(first())
 
   readonly connectionFailure$ = this.connectionService
     .watchFailure$()
--- a/frontend/projects/ui/src/app/pages/apps-routes/app-show/app-show.page.ts
+++ b/frontend/projects/ui/src/app/pages/apps-routes/app-show/app-show.page.ts
@@ -5,6 +5,7 @@ import {
   PackageDataEntry,
   PackageMainStatus,
   PackageState,
+  UIMarketplaceData,
 } from 'src/app/services/patch-db/data-model'
 import {
   PackageStatus,
@@ -60,7 +61,10 @@ export class AppShowPage {
   )
 
   readonly currentMarketplace$: Observable<Marketplace> =
-    this.marketplaceService.getMarketplace().pipe(first())
+    this.marketplaceService.getMarketplace()
+
+  readonly altMarketplaceData$: Observable<UIMarketplaceData | undefined> =
+    this.marketplaceService.getAltMarketplace()
 
   readonly connectionFailure$ = this.connectionService
     .watchFailure$()
--- a/frontend/projects/ui/src/app/pages/apps-routes/app-show/app-show.page.ts
+++ b/frontend/projects/ui/src/app/pages/apps-routes/app-show/app-show.page.ts
@@ -15,7 +15,7 @@ import {
   ConnectionFailure,
   ConnectionService,
 } from 'src/app/services/connection.service'
-import { map, startWith, filter, first } from 'rxjs/operators'
+import { map, startWith, filter } from 'rxjs/operators'
 import { ActivatedRoute } from '@angular/router'
 import { getPkgId } from '@start9labs/shared'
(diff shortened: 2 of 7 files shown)


frontend/projects/ui/src/app/pages/apps-routes/app-show/app-show.page.ts
```diff
@@ -53,6 +59,9 @@ export class AppShowPage {
     ),
   )
 
+  readonly currentMarketplace$: Observable<Marketplace> =
+    this.marketplaceService.getMarketplace().pipe(first())
```
MattDHill, line 63: Why only first? What if this changes while I'm on the page?

frontend/projects/ui/src/app/pages/apps-routes/app-show/app-show.page.ts
```diff
@@ -62,7 +71,9 @@ export class AppShowPage {
     private readonly navCtrl: NavController,
     private readonly patch: PatchDbService,
     private readonly connectionService: ConnectionService,
-  ) { }
+    @Inject(AbstractMarketplaceService)
```
MattDHill, line 74: Is @Inject() necessary? Maybe just `private readonly marketplaceService: AbstractMarketplaceService`

frontend/projects/ui/src/app/pages/apps-routes/app-show/pipes/to-buttons.pipe.ts
```diff
@@ -112,23 +115,67 @@ export class ToButtonsPipe implements PipeTransform {
...
+    currentMarketplace: Marketplace | null,
+  ): Button {
+    const pkgMarketplace = pkg.installed?.['marketplace-url']
+    const pkgM = pkgMarketplace
+      ? removeTrailingSlash(pkgMarketplace)
+      : pkgMarketplace
+    const currentM = currentMarketplace
+      ? removeTrailingSlash(currentMarketplace.url)
+      : currentMarketplace
+    if (!pkgMarketplace) {
```
MattDHill, line 129: Should probably start with this conditional, then you don't have to use the ternaries above.

frontend/projects/ui/src/app/pages/apps-routes/app-show/pipes/to-buttons.pipe.ts
```diff
@@ -112,23 +115,67 @@ export class ToButtonsPipe implements PipeTransform {
...
+      return {
+        disabled: true,
+        action: () => {},
+        title: 'Marketplace',
+        description:
+          'This package has been sideloaded and is not available in the Start9 Marketplace',
```
MattDHill, line 135: Misleading, it might be available in the Start9 Marketplace. "This package was not installed from a marketplace."

frontend/projects/ui/src/app/pages/apps-routes/app-show/pipes/to-buttons.pipe.ts
```diff
@@ -112,23 +115,67 @@ export class ToButtonsPipe implements PipeTransform {
...
+      ? removeTrailingSlash(pkgMarketplace)
+      : pkgMarketplace
+    const currentM = currentMarketplace
+      ? removeTrailingSlash(currentMarketplace.url)
+      : currentMarketplace
+    if (!pkgMarketplace) {
+      return {
+        disabled: true,
+        action: () => {},
+        title: 'Marketplace',
```
MattDHill, line 133: Title and icon are always the same. Should be abstracted out.

frontend/projects/ui/src/app/pages/apps-routes/app-show/pipes/to-buttons.pipe.ts
```diff
@@ -112,23 +115,67 @@ export class ToButtonsPipe implements PipeTransform {
...
+        description:
+          'This package has been sideloaded and is not available in the Start9 Marketplace',
+        icon: 'storefront-outline',
+      }
+    } else if (pkgM && currentM && pkgM !== currentM) {
+      return {
+        action: async () => {
+          const alert = await this.alertCtrl.create({
+            header: 'Marketplace Conflict',
```
MattDHill, line 142: Abstract function for readability

frontend/projects/ui/src/app/pages/apps-routes/app-show/pipes/to-buttons.pipe.ts
```diff
@@ -112,23 +115,67 @@ export class ToButtonsPipe implements PipeTransform {
...
+            message: `This service was installed from:
+            <br><br>
+            <span class="courier-new color-success-shade">${pkgM}</span>
```
MattDHill, line 145: Maybe better to display the Marketplace names rather than URLs?

frontend/projects/ui/src/app/pages/apps-routes/app-show/pipes/to-buttons.pipe.ts
```diff
@@ -112,23 +115,67 @@ export class ToButtonsPipe implements PipeTransform {
...
+            <br><br>but you are currently connected to:<br><br>
+            <span class="courier-new color-primary-shade">${currentM}</span>
+            <br><br>
+            Please change marketplaces to see the relevant details.`,
```
MattDHill, line 149: To view the marketplace listing for this service, visit your Marketplace Settings and change marketplaces.

frontend/projects/ui/src/app/pages/apps-routes/app-show/pipes/to-buttons.pipe.ts
```diff
@@ -112,23 +115,67 @@ export class ToButtonsPipe implements PipeTransform {
...
+            buttons: [
+              {
+                text: 'Cancel',
+                role: 'cancel',
+              },
+              {
+                text: 'Change Marketplace',
```
MattDHill, line 156: Go to Settings

frontend/projects/shared/styles/shared.scss
```diff
@@ -110,4 +110,12 @@ ion-modal::part(content) {
 
 .montserrat {
   font-family: 'Montserrat', sans-serif !important;
+}
+
+.color-success-shade {
```
MattDHill, line 115: I _think_ you can also do this inline with style="color: var(--ion-color-success-shade)" pretty sure.

MattDHill (changes requested):

frontend/projects/ui/src/app/pages/apps-routes/app-show/app-show.page.ts
elvece: I thought it made sense to close the subscription here, as navigating away from and back to the component (to, say, change marketplaces) would cause this data to be "reloaded".

The only case I can think of where this would be problematic is if the user has two views of the UI open simultaneously, one where they are changing the marketplace and one where they are watching the service show page. In that case, using `first()` would disable the live update in the service show page. Seems like a rare case, but perhaps a better UX to remove this pipe since it's not strictly necessary.

frontend/projects/ui/src/app/pages/apps-routes/app-show/app-show.page.ts
elvece: I thought it was necessary for getting the type of `getMarketplace()`, but it appears not. Will remove.

frontend/projects/ui/src/app/pages/apps-routes/app-show/pipes/to-buttons.pipe.ts
elvece: We don't always have that information from the installed package details, it's just the marketplace url. We could attempt to use that value to look up the name from the saved marketplaces, but it's not guaranteed to exist there.

frontend/projects/ui/src/app/pages/apps-routes/app-show/app-show.page.ts
elvece: Just kidding, it is necessary, otherwise you get this error: 

```
NullInjectorError: R3InjectorError(AppShowPageModule)[MarketplaceService -> MarketplaceService -> MarketplaceService -> MarketplaceService -> MarketplaceService]: 
  NullInjectorError: No provider for MarketplaceService!
NullInjectorError: R3InjectorError(AppShowPageModule)[MarketplaceService -> MarketplaceService -> MarketplaceService -> MarketplaceService -> MarketplaceService]: 
  NullInjectorError: No provider for MarketplaceService!
  ```

frontend/projects/shared/styles/shared.scss
elvece: You can. I abstracted it as I prefer to not use inline styles, for future usability, and for readability.

frontend/projects/ui/src/app/pages/apps-routes/app-show/app-show.page.ts
MattDHill: That's a circular reference error

elvece: Ok - this `@Inject` is used elsewhere when `MarketplaceService` is in the component constructor - is there a better pattern? Perhaps it's necessary because `MarketplaceService` extends `AbstractMarketplaceService` and this is shared?

MattDHill (approved):

frontend/projects/ui/src/app/pages/apps-routes/app-show/app-show.page.ts
waterplea: You can drop `@Inject` but use `AbstractMarketplaceService` as type and it will work. That is because DI works with `AbstractMarketplaceService` and we provide `MarketplaceService` as concrete implementation. Also the less concrete you define it here the better. Makes it more independent of implementation.

waterplea: Just noticed you used `getAltMarketplace` here which only exists in `MarketplaceService` implementation. Then you just leave it as is, current code is correct.

PR #1572, opened 2022-06-24
add dns server to embassy-os

dr-bonez:
this is not tested yet

this will allow containers and the host to resolve .embassys and .locals using systemd-resolved

### DIFF
--- a/build/initialization.sh
+++ b/build/initialization.sh
@@ -80,6 +80,9 @@ ControlPort 9051
 CookieAuthentication 1
 EOF
 
+# enable embassyd dns server
+systemd-resolve --set-dns 172.18.0.1
+
 if [ -f /embassy-os/product_key.txt ]
 then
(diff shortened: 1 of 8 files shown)


ProofOfKeags (dismissed):

build/initialization.sh
```diff
@@ -80,6 +80,9 @@ ControlPort 9051
 CookieAuthentication 1
 EOF
 
+# enable embassyd dns server
+systemd-resolve --set-dns 172.18.0.1
```
Blu-J, line 84: Are we installing the systemd-resolve on the pi along with the embassyd?

Blu-J (dismissed):

build/initialization.sh
ProofOfKeags: I'm pretty sure it was already there.

Blu-J (approved):

PR #1574, opened 2022-06-25
refactor: remove sync data access from PatchDbService

waterplea:


waterplea:
Gotta make sure to test this "sideloading" case in `app-show` on real data:
image

frontend/projects/ui/src/app/pages/server-routes/marketplaces/marketplaces.page.ts
```diff
@@ -188,11 +189,11 @@ export class MarketplacesPage {
...
 
-  private async delete(id: string): Promise<void> {
+  private async delete(id?: string): Promise<void> {
     if (!id) return
-    const marketplace: UIMarketplaceData = JSON.parse(
-      JSON.stringify(this.patch.getData().ui.marketplace),
-    )
+
+    const data = await getMarketplace(this.patch)
+    const marketplace: UIMarketplaceData = JSON.parse(JSON.stringify(data))
```
elvece, line 196: since we are using this JSON.parse/stringify a few times, can we make it a function with descriptive name? I'm not sure _why_ we need to do this, do you know @MattDHill ?

frontend/projects/ui/src/app/pipes/mask/mask.pipe.ts
```diff
@@ -5,8 +5,6 @@ import { Pipe, PipeTransform } from '@angular/core'
 })
 export class MaskPipe implements PipeTransform {
   transform(val: string, max = 16): string {
-    if (!val) return val
-    const times = val.length <= max ? val.length : max
-    return '●'.repeat(times)
+    return val && '●'.repeat(Math.min(val.length, max))
```
elvece, line 8: 👍

frontend/projects/ui/src/app/util/get-marketplace.ts
```diff
@@ -0,0 +1,9 @@
+import { first } from 'rxjs/operators'
+import { PatchDbService } from 'src/app/services/patch-db/patch-db.service'
+import { UIMarketplaceData } from 'src/app/services/patch-db/data-model'
+
+export function getMarketplace(
+  patch: PatchDbService,
+): Promise<UIMarketplaceData> {
+  return patch.watch$('ui', 'marketplace').pipe(first()).toPromise()
```
elvece, line 8: @MattDHill and I discussed changing the approach here - rather than using promises to fit in with the current schema, what do you think about having each component subscribing to this data, and using `async` in the html or using the `DestroyService` in the `<component>.ts` if we need to apply logic to the data. Matt can elaborate more on this idea.

elvece (changes requested):
Few comments

PR #1577, opened 2022-06-27
Fix/sideload icon type

elvece:
Main task here was to add the content type for icons.
Ended up refactoring how I was handling reading blobs to array buffers or data urls, so as to avoid using `as` typecasting. I wanted to accurately handle the error cases, so added proper logic to do so.

Tested via proxy to live backend.


frontend/projects/ui/src/app/pages/server-routes/sideload/sideload.page.ts
```diff
@@ -135,11 +130,22 @@ export class SideloadPage {
...
+    const contentType = `image/${this.toUpload.manifest?.assets.icon
+      .split('.')
+      .pop()}`
     const data = file.slice(
       Number(positions['icon'][0]),
       Number(positions['icon'][0]) + Number(positions['icon'][1]),
+      contentType,
     )
-    this.toUpload.icon = await readBlobAsDataURL(data)
+    try {
```
Blu-J, line 141: Don't need to try if we are just throwing it again right?

frontend/projects/ui/src/app/pages/server-routes/sideload/sideload.page.ts
```diff
@@ -175,23 +181,75 @@ async function getPositions(
...
+      return new Error('readBlobAsDataURL response should not be null')
+    if (typeof res === 'string') return res
+    return new Error('no possible blob to data url resolution found')
+  } catch (e: any) {
+    return new Error(e)
+  }
+}
+
+async function handleBlobToData(data: Blob | File): Promise<string> {
+  const res = await tryReadBlobAsDataUrl(data)
```
Blu-J, line 216: alternative, remove the try + catch in the tryReadBlobAsDataUrl
```ts
const asError = x => new Error(x)
const rest = await tryReadBlobAsDataUrl(data).catch(asError)

Blu-J: Or not return error, that works too

Blu-J (approved):

PR #1581, opened 2022-06-28
fix: Properties had a null description

Blu-J:
### About

During the current js version of the proxy, getting an error in the properties because it is failing to check the shape

Looks like the issue was that the value was coming back but as null, and I wasn't dealing with the null case, just the undefined case.

### Proof Of Fix 

Picture of the properties on the proxy working now

image

### DIFF
--- a/frontend/projects/ui/src/app/util/properties.util.ts
+++ b/frontend/projects/ui/src/app/util/properties.util.ts
@@ -149,7 +149,7 @@ function parsePropertiesV1Permissive(
       } else {
         const error = result.error
         const message = Parser.validatorErrorAsString(error)
-        const dataPath = '/' + error.keys.map(removeQuotes).join('/')
+        const dataPath = error.keys.map(removeQuotes).join('/')
         errorCallback(new Error(`/data/${idx}: ${message}`))
         if (dataPath) {
           applyOperation(cur, {
--- a/frontend/projects/ui/src/app/util/properties.util.ts
+++ b/frontend/projects/ui/src/app/util/properties.util.ts
@@ -24,9 +24,9 @@ const matchPropertiesV1 = shape(
   {
     name: string,
     value: string,
-    description: string,
-    copyable: boolean,
-    qr: boolean,
+    description: string.optional(),
+    copyable: boolean.optional(),
+    qr: boolean.optional(),
   },
   ['description', 'copyable', 'qr'],
   { copyable: false, qr: false } as const,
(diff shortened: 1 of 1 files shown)


frontend/projects/ui/src/app/util/properties.util.ts
```diff
@@ -149,12 +149,12 @@ function parsePropertiesV1Permissive(
       } else {
         const error = result.error
         const message = Parser.validatorErrorAsString(error)
-        let dataPath = error.keys.map(x => JSON.parse(x)).join('/')
+        const dataPath = '/' + error.keys.map(removeQuotes).join('/')
```
Blu-J, line 152:
```suggestion
        const dataPath = error.keys.map(removeQuotes).join('/')
```

frontend/projects/ui/src/app/util/properties.util.ts
```diff
@@ -41,11 +41,11 @@ const [matchPackagePropertiesV2, setPPV2] = deferred<PackagePropertiesV2>()
...
     type: literal('string'),
-    description: string,
+    description: string.optional(),
     value: string,
-    copyable: boolean,
-    qr: boolean,
-    masked: boolean,
+    copyable: boolean.optional(),
+    qr: boolean.optional(),
+    masked: boolean.optional(),
```
Blu-J, line 48: I'm going to probably make a fix for the ts-matches about this...

frontend/projects/ui/src/app/util/properties.util.ts
```diff
@@ -149,12 +149,12 @@ function parsePropertiesV1Permissive(
...
         const error = result.error
         const message = Parser.validatorErrorAsString(error)
-        let dataPath = error.keys.map(x => JSON.parse(x)).join('/')
+        const dataPath = error.keys.map(removeQuotes).join('/')
         errorCallback(new Error(`/data/${idx}: ${message}`))
         if (dataPath) {
           applyOperation(cur, {
             op: 'replace',
-            path: dataPath,
+            path: `/${dataPath}`,
```
Blu-J, line 157: I probably shouldn't make this change. Paths should start with /
Just noticed it was wrong when looking at this

frontend/projects/ui/src/app/util/properties.util.ts
```diff
@@ -196,6 +196,14 @@ function parsePropertiesV2Permissive(
   )
 }
 
+const removeRegex = /('|")/
+function removeQuotes(x: string) {
```
Blu-J, line 200: image

frontend/projects/ui/src/app/util/properties.util.ts
Blu-J: https://github.com/Blu-J/ts-matches/issues/505

frontend/projects/ui/src/app/util/properties.util.ts
elvece: No this is fine as it it complied with the specification

elvece (approved):

PR #1599, opened 2022-07-01
don't crash service if io-format is set for main

dr-bonez:
this is necessary because *many* old services use io-format yaml for main


Blu-J (commented):
This was talked about. Idea was that we don't want to drop the output till the end. The idea is that we will stop putting the information in a buffer, but not drop the buffer. It should mean once it ends we will see an error that it overflowed, but it won't kill the program

PR #1603, opened 2022-07-01
Fix/html logs

elvece:


elvece:
argh closing since i reused branch and its wrong

PR #1604, opened 2022-07-01
strip html from colors from logs

elvece:


elvece:
Tested on mocks, works as expected

elvece:
Also tested on proxy, works great

MattDHill (approved):

PR #1608, opened 2022-07-02
fix: Closed file

Blu-J:
### About

Potentially I am trying to fix 
image

And my theory is that since I am no longer consuming the buffer it is going into a state of EOF.

### Proof Of Fix

- [x] Test deployment to a pi, restart scanning of blocks

### DIFF
--- a/backend/src/procedure/docker.rs
+++ b/backend/src/procedure/docker.rs
@@ -187,20 +187,25 @@ impl DockerProcedure {
         let output = NonDetachingJoinHandle::from(tokio::spawn(async move {
             match async {
                 if let Some(format) = io_format {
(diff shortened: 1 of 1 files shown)


Blu-J:
image

Blu-J:
We get the git info of 
image
And the git log
image

Proving we are on the embassyd version that we are expecting

Blu-J:
For also testing that there is no memory leak after x hours


image

Blu-J:
Preliminary, looks like it works

Blu-J:
image
Seems to have fixed, not sure there is or not a memory leak though

elvece (approved):

PR #1610, opened 2022-07-03
Default to https:// urls for repositories, remove apt-transport-https

gStart9:
As of apt 1.5, the package apt-transport-https is no longer required because https:// is supported out of the box.
Reference: https://packages.debian.org/bullseye/apt-transport-https "This is a dummy transitional package - https support has been moved into the apt package in 1.5. It can be safely removed."  Apt is currently at 2.2.4.

Use a sed one-liner to convert all repos in /etc/apt/sources.list and /etc/apt/sources.list.d/*.list that are http:// to https:// (https:// is available for all http:// URLs currently referenced in EmbassyOS's apt repos).

### DIFF
--- a/build/initialization.sh
+++ b/build/initialization.sh
@@ -22,9 +22,11 @@ mount -o remount,rw /boot
 apt-mark hold raspberrypi-bootloader
 apt-mark hold raspberrypi-kernel
 
+# Convert all repos to use https:// before apt update
+sed -i "s/http:/https:/g" /etc/apt/sources.list /etc/apt/sources.list.d/*.list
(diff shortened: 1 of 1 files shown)


elvece:
@gStart9 Has this been tested by running a fresh image after building from source?

gStart9:
@elvece no it has not.  I will try to do that tonight.

elvece:
> @elvece no it has not. I will try to do that tonight.

Great - post results back here and we will get merged on success.

dr-bonez (approved):

PR #1614, opened 2022-07-05
release lock on update progress

dr-bonez:


Blu-J (approved):
#### Proof of fix

Can't get a video, but we got the recreation of the bug where during an install of something big, it would lock something that shouldn't be locked, like the js get config. 

After the fix, during the health check it no longer prevents the get config from getting the information that was needed

PR #1616, opened 2022-07-05
Fix/receipts health

Blu-J:


backend/src/install/progress.rs
```diff
@@ -44,10 +44,14 @@ impl InstallProgress {
         mut db: Db,
     ) -> Result<(), Error> {
         while !self.download_complete.load(Ordering::SeqCst) {
-            model.put(&mut db, &self).await?;
+            let mut tx = db.begin().await?;
```
Blu-J, line 47: These shouldn't be here in this diff?

dr-bonez (approved):

PR #1621, opened 2022-07-06
handle current-dependents properly during update

dr-bonez:
This one is a bit confusing, so I'll explain.

The issue here is that the &current_dependencies that we pass to add_dependent_to_current_dependents_lists is out of date, since it changes during configure. If we wanted to keep this code mostly the same, we would need to refetch current_dependencies after configure. However, since configure already does update this, we can just put it in the other side of an else.

### DIFF
--- a/backend/src/install/mod.rs
+++ b/backend/src/install/mod.rs
@@ -1360,21 +1360,22 @@ pub async fn install_s9pk<R: AsyncRead + AsyncSeek + Unpin>(
                 .await?;
             *main_status = prev.status.main;
             main_status.save(&mut tx).await?;
+        } else {
+            remove_from_current_dependents_lists(
+                &mut tx,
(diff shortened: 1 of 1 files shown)


Blu-J (approved):

Blu-J:
Solution feels like a band aid. Not preventing what was the source. But small improvements over nothing

elvece (approved):

dr-bonez:
Repro steps:
- Install electrs
- Install mempool@2.4.0.1
- run `embassy-cli db dump | jq '.value["package-data"].electrs.installed["current-dependents"]'`
- see mempool in output
- downgrade mempool to 2.4.0
- run `embassy-cli db dump | jq '.value["package-data"].electrs.installed["current-dependents"]'`
- see empty output

PR #1622, opened 2022-07-06
Fix/misc UI

elvece:
Fixes several FE issues:

1. font size in config header
2. updated package.json/lock for ts-matches to remove console.log
3. new option display for nested config elements (tested with updating to CLN w/ new option CLBOSS):
Screen Shot 2022-07-06 at 4 01 04 PM
4. navigation options for marketplace conflict action:
Screen Shot 2022-07-06 at 4 02 26 PM


frontend/projects/ui/src/app/modals/action-marketplace/action-marketplace.component.html
```diff
@@ -0,0 +1,39 @@
+<ion-header>
+  <ion-toolbar>
+    <ion-buttons slot="start">
+      <ion-button (click)="dismiss()" class="enter-click">
```
MattDHill, line 4: No `enter-click` on close buttons. This would be unique int he app and therefore be unexpected.

frontend/projects/ui/src/app/modals/action-marketplace/action-marketplace.component.html
```diff
@@ -0,0 +1,39 @@
...
+        <ion-icon name="close"></ion-icon>
+      </ion-button>
+    </ion-buttons>
+    <ion-title>{{ title }}</ion-title>
+  </ion-toolbar>
+</ion-header>
+<ion-content class="center1">
```
MattDHill, line 11: Where is this class?

elvece: nowhere anymore - removing

MattDHill (approved):

PR #1653, opened 2022-07-15
sdk: don't allow mounts in inject actions

chrisguida:
Fixes: https://github.com/Start9Labs/embassy-os/issues/557

Proof:
Screen Shot 2022-07-14 at 7 02 13 PM
Screen Shot 2022-07-14 at 7 02 46 PM

### DIFF
--- a/backend/src/procedure/docker.rs
+++ b/backend/src/procedure/docker.rs
@@ -85,6 +85,9 @@ impl DockerProcedure {
         if expected_io && self.io_format.is_none() {
             color_eyre::eyre::bail!("expected io-format");
         }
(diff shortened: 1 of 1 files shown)


dr-bonez:
since `verify()` is invoked on package install as well, we may need to only run this check if `manifest.eos-version >= 0.3.2`

chrisguida:
I'd prefer to know on the pack step, not on the install step

chrisguida:
Or perhaps I'm misunderstanding

dr-bonez:
It does both

dr-bonez (approved):

PR #1656, opened 2022-07-15
fix bug when config has more keys than the incoming spec

elvece:
Caught during downgrades where config has more keys than the incoming config spec

Test case: install LND v0.14.2.2, configure with defaults, downgrade to v0.13.3.1, attempt to configure, see errors below, extra options are watchtower
Screen Shot 2022-07-14 at 9 11 52 PM
Screen Shot 2022-07-14 at 9 12 03 PM

### DIFF
--- a/frontend/projects/ui/src/app/components/form-object/form-object.component.ts
+++ b/frontend/projects/ui/src/app/components/form-object/form-object.component.ts
@@ -35,7 +35,7 @@ interface Config {
   styleUrls: ['./form-object.component.scss'],
 })
 export class FormObjectComponent {
-  @Input() objectSpec: ConfigSpec | undefined
+  @Input() objectSpec?: ConfigSpec
   @Input() formGroup: FormGroup
   @Input() unionSpec?: ValueSpecUnion
   @Input() current?: Config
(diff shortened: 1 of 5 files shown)


frontend/projects/ui/src/app/components/form-object/form-object.component.ts
```diff
@@ -35,7 +35,7 @@ interface Config {
   styleUrls: ['./form-object.component.scss'],
 })
 export class FormObjectComponent {
-  @Input() objectSpec: ConfigSpec
+  @Input() objectSpec: ConfigSpec | undefined
```
MattDHill, line 38: By convention, this would be `objectSpec?: ConfigSpec`

frontend/projects/ui/src/app/components/form-object/form-object.component.ts
```diff
@@ -62,31 +62,36 @@ export class FormObjectComponent {
...
-      const spec = this.objectSpec[key]
-
-      if (spec.type === 'list' && ['object', 'union'].includes(spec.subtype)) {
-        this.objectListDisplay[key] = []
-        this.formGroup.get(key)?.value.forEach((obj: any, index: number) => {
-          const displayAs = (spec.spec as ListValueSpecOf<'object'>)[
-            'display-as'
-          ]
-          this.objectListDisplay[key][index] = {
+    if (this.objectSpec) {
```
MattDHill, line 65: Cleaner to do `Object.keys(this.objectSpec || {})` and remove higher conditional. Or another option would be `if (!this.objectSpec) return`

frontend/projects/ui/src/app/services/form.service.ts
```diff
@@ -508,51 +508,53 @@ function isUnion(spec: any): spec is ListValueSpecUnion {
 }
 
 export function convertValuesRecursive(
-  configSpec: ConfigSpec,
+  configSpec: ConfigSpec | undefined,
```
MattDHill, line 511: See above

frontend/projects/ui/src/app/services/form.service.ts
```diff
@@ -508,51 +508,53 @@ function isUnion(spec: any): spec is ListValueSpecUnion {
...
-      } else if (valueSpec.subtype === 'union') {
-        controls.forEach(formGroup => {
-          const unionSpec = valueSpec.spec as ListValueSpecUnion
-          const spec =
-            unionSpec.variants[
-              (formGroup as FormGroup).controls[unionSpec.tag.id].value
-            ]
-          convertValuesRecursive(spec, formGroup as FormGroup)
-        })
+  if (configSpec) {
```
MattDHill, line 514: See above

elvece:
Tested adding list items, also see [here](https://github.com/Start9Labs/lnd-wrapper/pull/66#issuecomment-1185136864).

MattDHill:
Closing in favor of  #1658

PR #1663, opened 2022-07-17
Fix `to` migration version input

elvece:
Presently, when running migrations using JS procedure scripts, the incoming `version` variable is the version you are currently on when upgrading (which seems correct), but is what you are downgrading to on downgrade (which seems incorrect). Confirmed by logging an error in either case to see what in the incoming version actually was.

I was under the impression that this variable was _always_ the version the package is current installed at ie. the version you are on / coming from.

So, this PR fixes the downgrade case, to allow the version that you are currently on to be passed in as the input.

This fix is needed to be able to accurately write downgrades in JS script files, because we need to know to run the downgrade based on the version you are currently on. 

Please elaborate and close this issue if my logic is wrong and this was intentional by design.

EDIT: maybe this is just a limitation of the scripts implementation, that we don't know if it's a from or to migration to set equivalent ranges for either case.

### DIFF
--- a/backend/src/install/mod.rs
+++ b/backend/src/install/mod.rs
@@ -1340,9 +1340,9 @@ pub async fn install_s9pk<R: AsyncRead + AsyncSeek + Unpin>(
             .migrations
             .to(
                 ctx,
-                version,
-                pkg_id,
                 &prev.manifest.version,
+                pkg_id,
+                version,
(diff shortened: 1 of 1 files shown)


dr-bonez:
the version supplied is always the version you are going *to* in a `to` block, and the version you are coming *from* in a `from` block.

The behavior specified is correct.

Issue #1434, opened 2022-05-10
Fix UI build to not rebuild every time `make` is invoked if nothing has changed


ProofOfKeags:
Right now the UI rebuilds every single time. This shouldn't be the case.

elvece:
Do you always run `make clean` first? Because it will always rebuild the frontend in that case.

Otherwise, curious, afaict the way the Makefile is currently setup, it should only rebuild on changes. 🤔

ProofOfKeags:
> Do you always run make clean first?

No.

PR #1664, opened 2022-07-17
run build checks only when relevant FE changes

elvece:
While attempting to add individual project build checks in the previous configuration (using the `lint-staged` key in `package.json`), I encountered this error:
```
error TS5042: Option 'project' cannot be mixed with source files on a command line.
```

Since we use projects and there is a current [limitation](https://github.com/microsoft/TypeScript/issues/27379), linting and build checks thus needed to be configured how addressed in this PR.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -84,11 +84,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "lint-staged && npm run check"
+      "pre-commit": "lint-staged --concurrent false"
     }
-  },
-  "lint-staged": {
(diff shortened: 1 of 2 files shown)


frontend/package.json
```diff
@@ -84,11 +84,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "lint-staged && npm run check"
+      "pre-commit": "lint-staged --concurrent false"
```
elvece, line 87: since we have multiple checks for `.ts` files, added so operations run in order

MattDHill (approved):

dr-bonez:
this doesn't actually close #1434

PR #1673, opened 2022-07-20
feat: atomic writing

Blu-J:
### About

Renames are safer and are more atomic than just writing to a file. 

Why not just using a `tmp` file:
- What if the service is writing it's own `.tmp` files and the file names conflict? we could clobber them and cause weird bugs in the service
- What if the service has behaviors that depend on stat-ing a directory? The new file could cause it to try to do something invalid

### Proof it works
Nothing needs to change on the api side. Thus the old tests still working is proof that it works
image

### DIFF
--- a/libs/js_engine/src/lib.rs
+++ b/libs/js_engine/src/lib.rs
@@ -514,7 +514,7 @@ mod fns {
             bail!("Volume {} is readonly", volume_id);
         }
 
-        let new_file = volume_path.join(path_in);
+        let new_file = volume_path.join(&path_in);
         let parent_new_file = new_file
             .parent()
             .ok_or_else(|| anyhow!("Expecting that file is not root"))?;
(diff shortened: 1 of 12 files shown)


libs/js_engine/src/lib.rs
```diff
@@ -525,7 +525,33 @@ mod fns {
                 volume_path.to_string_lossy(),
             );
         }
-        tokio::fs::write(new_file, write).await?;
+        let new_volume_tmp = {
+            let last = volume_path
+                .iter()
+                .last()
+                .ok_or_else(|| anyhow!("Volume is not able to have temp dir"))?
```
Blu-J, line 532:
```suggestion
                .ok_or_else(|| anyhow!("Unreachable: Volume path would have to be empty"))?
```

libs/js_engine/src/lib.rs
```diff
@@ -525,7 +525,33 @@ mod fns {
...
+                .to_str()
+                .ok_or_else(|| anyhow!("Volume is not able to have temp"))?;
```
Blu-J, line 534:
```suggestion
                .ok_or_else(|| anyhow!("Unreachable: Path would have to have things that aren't string -able"))?;
```

libs/js_engine/src/lib.rs
```diff
@@ -525,7 +525,33 @@ mod fns {
...
+            let mut tmp = volume_path
+                .parent()
+                .ok_or_else(|| anyhow!("Volume is not able to have temp"))?
```
Blu-J, line 537:
```suggestion
                .ok_or_else(|| anyhow!("Unreachable: Volumes should never be highest level nor empty"))?
```

dr-bonez (approved):

MattDHill (approved):

Issue #1032, opened 2022-01-05
[feat]: CI Build Checking/Testing


ProofOfKeags:
### Prerequisites

- [X] I have searched for [existing issues](https://github.com/start9labs/embassy-os/issues) that already suggest this feature, without success.

### Describe the Feature Request

CI Build/Testing

### Describe the Use Case

I'd like to get automated build verification onto the branch checks of PR's into the mainline. This would save us from accidentally merging broken builds to master or other protected branches

### Describe Preferred Solution

Travis CI

### Describe Alternatives

Circle CI
Jenkins 😬

moerketh:
Would you consider a GitHub Actions workflow?

A basic workflow using the existing scripts would cover your use case with little effort. This can be iterated on to make it faster and more complete. For example:

```yaml
name: Build Pipeline

on:
  push:
    branches:
    - master
  pull_request:
    branches:
    - master

jobs:
  build:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v3
      with:
        submodules: recursive
        
    - name: Set up QEMU
      uses: docker/setup-qemu-action@v1
      
    - name: Set up Docker Buildx        
      uses: docker/setup-buildx-action@v1
    
    - name: Build
      run: make
```

This runs the build using the default software installed on the GitHub ubuntu-latest environment, see [here](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md).

chrisguida:
Absolutely! Been thinking we could use Github Actions. Never tried it before though, how would it work? Firstly we actually need to run it on raspios rather than ubuntu. Would it be possible to "flash" an embassy-os image to a github actions server, then boot into it just like you would on a physical pi? If not we'll need to figure out how to adapt embassy-os to this environment.

ProofOfKeags:
Agreed. I think additionally we should build components in individual steps so we have better info on failures if/when they crop up. Luckily we already have make targets for a bunch of stuff.

kn0wmad:
> Great. Yes, in fact that is exactly what the above pipeline accomplishes. The build agent it uses for compilation is x64 ubuntu. However the scripts in this project already do the hard part -> aarch64 cross compilation via docker. The result image can be flashed to a physical pi.
> 
[...]

Issue #981, opened 2021-12-20
[feat]: Support x86_64 Architecture for Embassy Pro


kn0wmad:
### Prerequisites

- [X] I have searched for [existing issues](https://github.com/start9labs/embassy-os/issues) that already suggest this feature, without success.

### Describe the Feature Request

Self-explanitory

### Describe the Use Case

Expand hardware options

### Describe Preferred Solution

Soon

ProofOfKeags:
So in theory we already support x86. What inspired this ticket?

kn0wmad:
In theory?  This is prerequisite for a test device for potential next hardware

ProofOfKeags:
My point is that we haven't tested on an actual x86 device, but theres no reason I can think of that it won't work. However, testing on new hardware carries more requirements than just making it run on a new architecture, it also requires that things like the sound API works similarly/the same on the new hardware or a relaxing of the requirement that that particular feature works. If this is for testing a device on x86_64 we are already at a place where it can be tested.

chrisguida:
Perhaps we should turn this into 2 tickets: "test building embassy-os on x86", "make a build guide for x86"?

kn0wmad:
This is purely needed internally at the moment, and when I spoke with @dr-bonez about it, he insinuated it was trivial to 'add support for.'  I'm not sure what is entailed here, but the device is now specifically known

k0gen:
How about services? Will they include multi platform binaries or there will be a separate marketplace for x86_64 architecture?

ProofOfKeags:
> How about services? Will they include multi platform binaries or there will be a separate marketplace for x86_64 architecture?

We already have support in the marketplace protocol for architecture specification. Packages will need to provide s9pks for the architectures they want to support.

chrisguida:
I'm guessing embassy-os will know which architecture to request based on where it's running

PR #1682, opened 2022-07-24
add x86 build and run unittests to backend pipeline

moerketh:
Here is some more pipeline work for your consideration. This adds a x86_64 build for the backend, in addition to the existing aarch64 build. Also added running the backend unittests on both platforms.

Relates to #1032 and #981

## Added backend x86_64 build

The changes made are:
* Build the backend for x86_64
* Use the target names x86_64 and aarch64 (instead of amd64 and arm64) consistently in the pipeline
* Use the same rust version everywhere
* Remove some unnecessary build steps, or add an `if` when platform specific

The current Makefile only targets aarch64, so for the build and test commands I've had to ignore the Makefile.

## Added executing unittests in pipeline

To run the unittests I found that some tests expect eOS environment features, such as existing files or directories. Perhaps they should be classified as integration tests and run in the eOS environment. I skipped those tests for now.

Running the tests on aarch64 was a little tricky. I found a couple of ways that didn't work. The obvious one was running `cargo test` in a aarch64 container. However `cargo` detects a change in filetime or signature and rebuilds everything, taking multiple hours on QEMU. I don't know how to fix that. The solution I found was to use nextest's archive option. Unfortunatly nextest doesn't release aarch64 binaries, so I added a stage to create them.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -12,7 +12,7 @@ A matrix-strategy has been used to build for both x86_64 and aarch64 platforms i
 Unittests are run using [cargo-nextest]( https://nexte.st/). First the sources are (cross-)compiled and archived. The archive is then run on the correct platform.
 Because nextest currently does not publish a pre-built aarch64 binary, there is a stage for building this.
 
-## frontend-pr
+## frontend
 Runs: manually (on: workflow_dispatch) or called by product-pipeline (on: workflow_call)
 
 This workflow builds the frontends.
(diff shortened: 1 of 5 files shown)


kn0wmad:
There may be some considerations with the upcoming x86 builds, @dr-bonez will know more.  Thank you again for your contributions :)

sunshowers:
FYI nextest now publishes aarch64 Linux binaries.

dr-bonez (approved):

moerketh:
@dr-bonez , just a note after merge., I notice a few more "util::http_reader::*" unittests are now failing because the container image the unittests are running in does not have the ca-certificates apt package installed. This is intentional, because a unittest that goes onto the internet is a too large unit in my view. We could add them to the skip list for now, or fix them if possible.

dr-bonez:
I'd accept a pr that switches to using a local fileserver to test the code, but for now I'd rather fix the container to have ca certs, and keep the tests.

PR #1686, opened 2022-07-25
only validate mounts for inject if eos `>=0.3.1.1`

dr-bonez:


MattDHill (approved):

chrisguida:
What's the purpose of this? I thought we decided not to worry about backwards compatibility since inject docker actions literally won't run if they have mounts defined

dr-bonez:
@chrisguida the inject flag is ignored for certain types of actions

chrisguida:
Which actions? Did those actions exist prior to 0.3.1.1?

dr-bonez:
backup restores is where we ran into this

chrisguida:
Ok, so the effect of this PR is to allow inject actions to have mounts on packages older than 0.3.1.1? Do we want users to be able to install packages whose backup restores are invalid?

PR #1694, opened 2022-07-26
sync blockdev after update

dr-bonez:


MattDHill (approved):

Blu-J (approved):

dr-bonez:
It just makes sure all data is synced to the physical media before moving on. Otherwise there's a tiny window where if you yank power you could boot into a corrupt filesystem

PR #1695, opened 2022-07-26
fix: Add modification to the max_user_watches

Blu-J:
### About

Fixes https://docs.syncthing.net/users/faq.html#inotify-limits for
https://github.com/Start9Labs/syncthing-wrapper/issues/17

Using the notify of 2^20, 

### Proof That Works

image


dr-bonez (changes requested):
let's set this in initialization.sh instead

dr-bonez (approved):

PR #1704, opened 2022-07-27
typo

k0gen:


kn0wmad:
This is considered the 'correct' spelling in American English, and it seems, from the original Greek/Latin.  The French are responsible for this it seems -> https://grammarist.com/spelling/recognize-vs-recognise/

PR #1705, opened 2022-07-27
Update README.md

kn0wmad:
Updated release to 0.3.1


kn0wmad:
I was planning to manually update this on 'minor' version increases (v0.3.x.0), but if there is a way to automate to latest release tag, that may be better

dr-bonez (approved):

PR #1714, opened 2022-07-29
[Feat] follow logs

chrisguida:
Adds an `-f` (short) or `follow` (long) flag to the logs endpoint, which returns a guid which can then be called via /ws/rpc/<guid> to access a stream of realtime logs. Also includes a CLI implementation.

Since `before_flag` was changed to `before`, the UI will need to be updated to send this api key instead before this can be merged.

Proof:

https://user-images.githubusercontent.com/7445670/181671636-92f72e61-870f-4564-bb6d-1ee2014f5b86.mp4


frontend/projects/diagnostic-ui/src/app/services/api/api.service.ts
```diff
@@ -13,8 +13,8 @@ export interface GetErrorRes {
 }
 
 export type GetLogsReq = {
+  before: boolean
   cursor?: string
-  before_flag?: boolean
```
chrisguida, line 17: looks like you deleted the question mark, is that important?

backend/src/logs.rs
```diff
@@ -219,28 +409,53 @@ pub async fn fetch_logs(
...
     }
-    Ok(LogResponse {
+    Ok(LogResponse::NoFollow {
         entries,
         start_cursor,
         end_cursor,
     })
 }
 
+// #[tokio::test]
```
MattDHill, line 419: What's with the comment out?

chrisguida: This is an old test that doesn't compile anymore because `fetch_logs` now takes a `CliContext`, which is difficult to construct in a testing context. I'm leaving it because it's still useful documentation for how this module works, and it may be easier to fix this test in the future when we implement some kind of testing framework.

I can remove if you think it's not worth it to leave in.

frontend/projects/diagnostic-ui/src/app/services/api/api.service.ts
chrisguida: Ah yes, thanks 👍

backend/src/logs.rs
MattDHill: Yeah if it's useful, leave it, but you should include a comment about why it's being left.

MattDHill:
Last commit will only work once BE code to support the `tail` rpc methods

Blu-J (dismissed):
Saw demo with code built on my system

chrisguida:
What was the reason for splitting this out into its own rpc command?

MattDHill:
> What was the reason for splitting this out into its own rpc command?

The behavior of the endpoint when `follow: true` was passed was sufficiently different that it seemed cleaner to create a new method than clutter with conditionals.

chrisguida:
This looks good, I think we can merge?

MattDHill:
Let's wait for Lucy to review tomorrow. I'm investigating the browser tab crashing, even though I don't htink it's related to this PR

MattDHill (approved):

PR #1718, opened 2022-07-30
don't use rust-arm-cross docker buildx if arch is already arm64

chrisguida:
This allows fast(er) builds in a Linux VM on MacOS M1. It probably also works on a Pi, though that is untested.

Without this change, the user needs to build `rust-arm-cross` and then build the backend inside the resulting docker container using `buildx`. This process is much slower than simply building natively using `cargo build` (60 minutes -> 16 minutes on my machine).

This change also builds `cargo-deps` without `rust-arm-cross`, so `rust-arm-cross` is removed completely as a system build dependency with this feature, when run on arm64.

Unfortunately the `compat` image still requires the `rust-musl-cross` cross-build, as I couldn't figure out a straightforward way of removing that from the build process.

### DIFF
--- a/build-cargo-dep.sh
+++ b/build-cargo-dep.sh
@@ -13,9 +13,17 @@ if tty -s; then
 	USE_TTY="-it"
 fi
 
-mkdir -p cargo-deps
-alias 'rust-arm64-builder'='docker run $USE_TTY --rm -v "$HOME/.cargo/registry":/root/.cargo/registry -v "$(pwd)"/cargo-deps:/home/rust/src -P start9/rust-arm-cross:aarch64'
+if [[ "$(uname -m)" =~ ^(arm64|aarch64)$ ]]; then
+	echo "Detected arm64 system, skipping cross-build"
+	cargo install "$1"
+	mkdir -p cargo-deps/aarch64-unknown-linux-gnu/release/
+	cp $(which "$1") cargo-deps/aarch64-unknown-linux-gnu/release/
+else
+	mkdir -p cargo-deps
+	echo "Detected non-arm64 system, cross-building"
+	alias 'rust-arm64-builder'='docker run $USE_TTY --rm -v "$HOME/.cargo/registry":/root/.cargo/registry -v "$(pwd)"/cargo-deps:/home/rust/src -P start9/rust-arm-cross:aarch64'
+	rust-arm64-builder cargo install "$1" --target-dir /home/rust/src
+fi
 
-rust-arm64-builder cargo install "$1" --target-dir /home/rust/src
 sudo chown -R $USER cargo-deps
 sudo chown -R $USER ~/.cargo
--- a/build-cargo-dep.sh
+++ b/build-cargo-dep.sh
@@ -13,6 +13,7 @@ if tty -s; then
 	USE_TTY="-it"
 fi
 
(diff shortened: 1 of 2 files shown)


build-cargo-dep.sh
```diff
@@ -13,9 +13,17 @@ if tty -s; then
...
 fi
 
-mkdir -p cargo-deps
 alias 'rust-arm64-builder'='docker run $USE_TTY --rm -v "$HOME/.cargo/registry":/root/.cargo/registry -v "$(pwd)"/cargo-deps:/home/rust/src -P start9/rust-arm-cross:aarch64'
+if [[ "$(uname -m)" =~ ^(arm64|aarch64)$ ]]; then
+	echo "Detected arm64 system, skipping cross-build"
+	cargo install "$1"
+	mkdir -p cargo-deps/aarch64-unknown-linux-gnu/release/
+	cp $(which "$1") cargo-deps/aarch64-unknown-linux-gnu/release/
+else
```
dr-bonez, line 22: check the *whole triple* please. Don't want this to break in a super non-obvious way if built on BSD or musl

dr-bonez (changes requested):

build-cargo-dep.sh
chrisguida: by "whole triple" do you mean:

```
   -m, --machine
          print the machine hardware name

   -p, --processor
          print the processor type or "unknown"

   -i, --hardware-platform
          print the hardware platform or "unknown"
```

If so, my ubuntu vm in UTM on macos gives:
```
cguida@cg-m1-ubuntu:~/embassy-os$ uname -m
aarch64
cguida@cg-m1-ubuntu:~/embassy-os$ uname -p
aarch64
cguida@cg-m1-ubuntu:~/embassy-os$ uname -i
aarch64
```

while my embassy gives:
```
root@embassy-30297ba3:~# uname -m
aarch64
root@embassy-30297ba3:~# uname -p
unknown
root@embassy-30297ba3:~# uname -i
unknown
```

and my mac gives:
```
cguida@cg-mac 22-08-01T14:18 embassy-os % uname -m
arm64
cguida@cg-mac 22-08-01T14:18 embassy-os % uname -p
arm
cguida@cg-mac 22-08-01T14:18 embassy-os % uname -i
uname: illegal option -- i
usage: uname [-amnprsv]
```

Do we currently support building on BSD and musl? What do you expect to break in those builds?

dr-bonez: I have no idea if it would even build, but the binary would be incorrect if so. 

The whole triple would be something like aarch64-linux-gnu. Arch, kernel, libc implementation.

chrisguida: Do you have any suggestions as to how to check those things in bash?

chrisguida: Looks like just `uname` gives `Linux` on the embassy and the ubuntu VM, whereas my mac gives `Darwin`. So that works.

Now to determine libc implementation...

chrisguida: looks like `grep -q __GLIBC__ <<< "$(gcc -dM -E - <<< "#include <features.h>")"` is the shortest way to detect glibc

chrisguida:
Converting to draft, need to fix embassy* and nc-broadcast binaries, there seems to be a version mismatch between glibc versions on my build box and the embassy

PR #1720, opened 2022-08-04
fix cli install

chrisguida:
for some reason the base url now has a trailing slash, so calls to it that add a leading slash will end up calling `//rest/rpc`, with two slashes. i have no idea why this has changed, but this will fix cli installs.


k0gen (approved):
this fixes it

dr-bonez (approved):

PR #1733, opened 2022-08-15
Feat: HttpReader

redragonx:


backend/src/util/http_reader.rs
```diff
@@ -0,0 +1,221 @@
...
+
+        let mut fut = if let Some(fut) = this.read_in_progress.take() {
+            fut
+        } else {
+            HttpReader::get_range(
+                *this.range_unit,
+                this.http_client.clone(),
+                this.http_url.clone(),
+                *this.cursor_pos,
+                *this.total_bytes,
```
dr-bonez, line 164:
```suggestion
                std::cmp::min(*this.cursor_pos + buf.remaining(), *this.total_bytes),
```

dr-bonez (changes requested):

backend/src/util/http_reader.rs
```diff
@@ -0,0 +1,248 @@
...
+    }
+}
+
+impl AsyncSeek for HttpReader {
+    fn start_seek(self: Pin<&mut Self>, position: std::io::SeekFrom) -> std::io::Result<()> {
+        let this = self.project();
+
+        match position {
+            std::io::SeekFrom::Start(pos) => {
+                *this.cursor_pos = pos as usize;
```
dr-bonez, line 213: throw error if pos > total_bytes

backend/src/util/http_reader.rs
```diff
@@ -0,0 +1,248 @@
...
+                Ok(())
+            }
+            std::io::SeekFrom::End(pos) => {
```
dr-bonez, line 216:
```suggestion
            std::io::SeekFrom::Current(pos) => {
```

backend/src/util/http_reader.rs
```diff
@@ -0,0 +1,248 @@
...
+                // if the result of this < 0, it will error out
+                *this.cursor_pos -= pos as usize;
+                Ok(())
+            }
+            std::io::SeekFrom::Current(pos) =>
```
dr-bonez, line 221:
```suggestion
            std::io::SeekFrom::End(pos) => 
```

backend/src/util/http_reader.rs
```diff
@@ -0,0 +1,248 @@
...
+    fn start_seek(self: Pin<&mut Self>, position: std::io::SeekFrom) -> std::io::Result<()> {
+        let this = self.project();
+
+        match position {
+            std::io::SeekFrom::Start(pos) => {
+                *this.cursor_pos = pos as usize;
+                Ok(())
+            }
+            std::io::SeekFrom::End(pos) => {
+                // if the result of this < 0, it will error out
```
dr-bonez, line 217: it will panic. It would be preferred to throw an actual error

backend/src/util/http_reader.rs
```diff
@@ -0,0 +1,248 @@
...
+                *this.cursor_pos -= pos as usize;
```
dr-bonez, line 218:
```suggestion
                *this.cursor_pos = (*this.cursor_pos as i64 + pos) as usize;
```

dr-bonez (changes requested):
Some things here are not quite right. Also I would like you to throw errors for invalid seeks

backend/src/util/http_reader.rs
```diff
@@ -0,0 +1,298 @@
...
+                let new_pos = i64::try_from(*this.cursor_pos)
+                    .map_err(|err| StdIOError::new(std::io::ErrorKind::InvalidInput, err))?
+                    + offset;
+
+                if new_pos < 0 {
+                    return Err(StdIOError::new(
+                        std::io::ErrorKind::InvalidInput,
+                        "Can't read before byte 0",
+                    ));
+                }
```
dr-bonez, line 251: this is redundant with the usize conversion below

backend/src/util/http_reader.rs
```diff
@@ -0,0 +1,298 @@
...
+                    }
+                    Err(err) => return Err(StdIOError::new(std::io::ErrorKind::InvalidInput, err)),
+                }
+                Ok(())
+            }
+            std::io::SeekFrom::Current(offset) => {
+                // We explicitly check if we read before byte 0.
+                let new_pos = i64::try_from(*this.cursor_pos)
+                    .map_err(|err| StdIOError::new(std::io::ErrorKind::InvalidInput, err))?
+                    + offset;
```
dr-bonez, line 244: this will only fail if this.cursor_pos >= 2^63
I'm not saying this isn't worth handling, but I'm also not saying it is

backend/src/util/http_reader.rs
```diff
@@ -0,0 +1,298 @@
...
+
+                if new_pos < 0 {
+                    return Err(StdIOError::new(
+                        std::io::ErrorKind::InvalidInput,
+                        "Can't read before byte 0",
+                    ));
+                }
+                *this.cursor_pos = usize::try_from(new_pos)
```
dr-bonez, line 271: same redundancy here

backend/src/util/http_reader.rs
```diff
@@ -0,0 +1,298 @@
...
+
+    async fn get_range(
+        range_unit: Option<RangeUnit>,
+        http_client: Client,
+        http_url: Url,
+        start: usize,
+        len: usize,
+        total_bytes: usize,
+    ) -> Result<Vec<u8>, Error> {
+        if start > len {
```
dr-bonez, line 120: I think you want `start > end` and put it after your end calculation

backend/src/util/http_reader.rs
```diff
@@ -0,0 +1,298 @@
...
+            return Ok(Vec::new());
+        }
+
+        let mut data = Vec::new();
```
dr-bonez, line 124: You can save yourself a heap allocation if you use `Vec::with_capacity(len)`

backend/src/util/http_reader.rs
```diff
@@ -0,0 +1,298 @@
...
+        let end = min((start + len) - 1, total_bytes);
```
dr-bonez, line 125:
```suggestion
        let end = min(start + len, total_bytes) - 1;
```

dr-bonez (changes requested):

backend/src/util/http_reader.rs
redragonx: thought it would be nice if it was a case on it's own, is that being too verbose?

backend/src/util/http_reader.rs
redragonx: Yeah, but since this is an OS basically, my logic was that this code could be ported to other archs or to 32 bits where 2^63 might not be the case?

What do you think?

backend/src/util/http_reader.rs
dr-bonez: that's fine, but I'd just use `as` below then, instead of using `usize::try_from`

backend/src/util/http_reader.rs
dr-bonez: in that case it would be actually impossible to error, because the target is `i64`, not the source.

backend/src/util/http_reader.rs
```diff
@@ -0,0 +1,294 @@
...
+        start: usize,
+        len: usize,
+        total_bytes: usize,
+    ) -> Result<Vec<u8>, Error> {
+
+        let mut data = Vec::with_capacity(len);
+   
+        let end = min(start + len, total_bytes) - 1;
+   
+        if start > len {
```
dr-bonez, line 126:
```suggestion
        if start > end {
```

dr-bonez (changes requested):

dr-bonez (approved):

PR #1736, opened 2022-08-15
Mino Feat: Change the Manifest To have a new type

Blu-J:
* [x] Fix the build (dependencies:634 map for option)
* [x] Test
  * [x] Start a docker sleep infinity
  * [x] Stop/start LR
  * [x] Stop/start Reg
  * [x] More


Blu-J:
# Proofs/ Checks
- [x]  When stopped, docker still running in background?
image
Image shows the image is stopped, docker is running in sleep infinity and the curl shows no running ui, therefore not running service in background
- [x] Started, we can reach the service inside like normal

image
Image shows the image is running, docker is running in sleep infinity and the curl shows running ui, therefore running service in background

- [x] Start/ Stop/ Restart

https://user-images.githubusercontent.com/2364004/185174042-bea137c8-dea3-4c2b-bd4e-e455509e6178.mp4


- [x] Ensure we can Backup and restore
image

# TODO
- [ ] Uninstall

PR #1738, opened 2022-08-15
[Fix] websocket connecting and patchDB connection monitoring

MattDHill:


frontend/projects/ui/src/app/app/global/services/patch-monitor.service.ts
```diff
@@ -14,7 +13,7 @@ export class PatchMonitorService extends Observable<boolean> {
   private readonly stream$ = this.authService.isVerified$.pipe(
     switchMap(verified => {
       if (verified) {
-        return from(this.patch.start()).pipe(mapTo(true))
+        return from(this.patch.start()).pipe(map(() => true))
```
waterplea, line 16: There doesn't seem to be anything async left in `patch.start`. Looks like we can remove `async` from it and just use `map` instead of `switchMap` here.

frontend/projects/ui/src/app/app/global/services/offline.service.ts
```diff
@@ -1,39 +1,54 @@
...
+    .pipe(startWith(null), distinctUntilChanged())
+
+  private failureInfo$ = combineLatest([
+    this.connectionService.watchFailure$,
+    this.updateProgress$,
+  ]).pipe(
+    map(([connectionFailure, progress]) => {
+      if (connectionFailure === ConnectionFailure.None) {
+        return null
+      } else if (!!progress && progress.downloaded === progress.size) {
```
waterplea, line 37: No need for `else` since there are `return` statements.

frontend/projects/ui/src/app/app/global/services/offline.service.ts
```diff
@@ -1,39 +1,54 @@
...
+import { PatchDbService } from 'src/app/services/patch-db/patch-db.service'
 
 // Watch for connection status
 @Injectable()
 export class OfflineService extends Observable<unknown> {
   private current?: HTMLIonToastElement
 
-  private readonly connection$ = this.connectionService
-    .watchFailure$()
-    .pipe(distinctUntilChanged(), debounceTime(500))
```
waterplea, line 27: Do we no longer need `debounceTime`?

frontend/projects/ui/src/app/pages/marketplace-routes/marketplace-list/marketplace-list.page.ts
```diff
@@ -31,7 +32,7 @@ export class MarketplaceListPage {
     .pipe(
       filter(data => exists(data) && !isEmptyObject(data)),
       first(),
-      switchMapTo(this.marketplaceService.getPackages()),
+      switchMap(() => this.marketplaceService.getPackages()),
```
waterplea, line 35: This could be the reason why now it requests packages upon reconnection — as opposed to `switchMapTo` now `getPackages` is getting called everytime stream gets here.

frontend/projects/ui/src/app/services/connection.service.ts
```diff
@@ -5,66 +5,61 @@ import {
...
+  private readonly patchConnection$ = new BehaviorSubject<PatchConnection>(
+    PatchConnection.Initializing,
+  )
   private readonly networkState$ = merge(
-    fromEvent(window, 'online').pipe(mapTo(true)),
-    fromEvent(window, 'offline').pipe(mapTo(false)),
+    fromEvent(window, 'online'),
+    fromEvent(window, 'offline'),
   ).pipe(
-    startWith(null),
```
waterplea, line 27: `startWith(null)` was here so that upon initial subscription it would run through following `map` and emit current network state.

frontend/projects/ui/src/app/services/patch-db/patch-db.factory.ts
```diff
@@ -1,50 +1,57 @@
...
 
-export function realSourceFactory(
-  embassyApi: ApiService,
-  config: ConfigService,
-  { defaultView }: Document,
-): Source<DataModel>[] {
-  const { patchDb } = config
-  const host = defaultView?.location.host
-  const protocol = defaultView?.location.protocol === 'http:' ? 'ws' : 'wss'
+      return of('').pipe(
```
waterplea, line 38: What's the purpose of `of('')` and `switchMap`? Why not `from` at the beginning?

frontend/projects/ui/src/app/services/patch-db/patch-db.service.ts
```diff
@@ -1,165 +1,49 @@
...
-        )
-        .subscribe({
-          complete: () => {
-            console.warn('patchDB: SYNC COMPLETE')
-          },
+    this.sub = this.patchDb.cache$
+      .pipe(
+        debounceTime(420),
+        tap(cache => {
+          this.bootstrapper.update(cache)
```
waterplea, line 35: The subscription is now here only to update the bootstrapper?

MattDHill: Correct

waterplea: I wonder if it makes sense to pass it to `patch-db` to be added here:
```ts
  public cache$ = this.source$.pipe(
    tap(res => this.store.update(res)),
    map(_ => this.store.cache),
    tap(cache => this.bootstrapper.update(cache),
    shareReplay(1),
  )
```
Because we're not really *starting* anything here...

waterplea (approved):
Looks good to me, added a few cosmetic comments.

frontend/projects/ui/src/app/pages/notifications/notifications.page.ts
```diff
@@ -1,4 +1,4 @@
-import { Component } from '@angular/core'
+import { ChangeDetectionStrategy, Component } from '@angular/core'
```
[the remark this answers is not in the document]
MattDHill: No, I tested it and doesn't work on this page. Will remove the import

frontend/projects/ui/src/app/services/patch-db/patch-db.service.ts
MattDHill: I'm game, will require some thought as to keeping PatchDB generic with regard to initial cache and bootstrapper updating.

PR #1741, opened 2022-08-17
Bugfix/dns

### LINKED ISSUE #1727: [bug]: DNS Issue in Services
### Prerequisites

- [X] I have searched for [existing issues](https://github.com/start9labs/embassy-os/issues) that already report this problem.

### EmbassyOS Version

0.3.1.1

### Device

Laptop/Desktop

### Device OS

Linux

### Device OS Version

5.18

### Browser

Firefox

### Browser Version

102

### Current Behavior

` WARN embassy::net::dns: Non A-Record requested for vaultwarden.embassy. `

Device is bep/chiming every 10-15min, becoming unavailable shortly before bep/chime.  The above error message is prevalent

### Expected Behavior

Device runs without constantly restarting

### Steps to Reproduce

1. Latest image
2. Recover / Use Drive
3. See issue

### Anything else?

This is a symptom of a larger, yet unknown issue.  @gStart9 has run into several production errors that indicate DNS problems, i.e. involving SMB share

dr-bonez:
should close #1727

### DIFF
--- a/backend/src/net/wifi.rs
+++ b/backend/src/net/wifi.rs
@@ -801,7 +801,6 @@ pub async fn synchronize_wpa_supplicant_conf<P: AsRef<Path>>(
         .arg("up")
         .invoke(ErrorKind::Wifi)
         .await?;
-    Command::new("dhclient").invoke(ErrorKind::Wifi).await?;
     if let Some(last_country_code) = last_country_code {
         tracing::info!("Setting the region");
         let _ = Command::new("iw")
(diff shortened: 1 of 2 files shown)


backend/src/net/wifi.rs
```diff
@@ -801,7 +801,6 @@ pub async fn synchronize_wpa_supplicant_conf<P: AsRef<Path>>(
         .arg("up")
         .invoke(ErrorKind::Wifi)
         .await?;
-    Command::new("dhclient").invoke(ErrorKind::Wifi).await?;
```
Blu-J, line 804: If I remember, this is redundant because the initialization.sh

Blu-J (approved):

backend/src/net/wifi.rs
dr-bonez: We aren't using dhclient anymore. Relying on systemd-resolved

PR #1746, opened 2022-08-18
[feat]: resumable downloads

dr-bonez:
this modifies HttpReader to automatically resume downloads if the connection closes before the message completes. It also reduces the memory footprint of HttpReader by writing chunks directly to the buffer, instead of allocating one.

### DIFF
--- a/backend/src/util/http_reader.rs
+++ b/backend/src/util/http_reader.rs
@@ -285,16 +285,32 @@ impl AsyncSeek for HttpReader {
 
 #[tokio::test]
 async fn main_test() {
-    use tokio::io::AsyncReadExt;
     let http_url = Url::parse("https://start9.com/latest/_static/css/main.css").unwrap();
 
     println!("Getting this resource: {}", http_url);
     let mut test_reader = HttpReader::new(http_url).await.unwrap();
 
-    let mut buf = vec![0; test_reader.total_bytes];
-    let bytes_read = test_reader.read(&mut buf).await.unwrap();
+    let mut buf = Vec::new();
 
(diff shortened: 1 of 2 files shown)


backend/src/util/http_reader.rs
```diff
@@ -167,51 +169,88 @@ impl AsyncRead for HttpReader {
...
+
+            let res_poll = in_progress.as_mut().poll(cx);
+
+            match res_poll {
+                Poll::Ready(result) => match result {
+                    Ok(body) => {
+                        *this.read_in_progress = ReadInProgress::Complete(body);
                     }
+                    Err(err) => {
+                        break Poll::Ready(Err(StdIOError::new(
```
redragonx, line 243: I really like this

backend/src/util/http_reader.rs
```diff
@@ -167,51 +169,88 @@ impl AsyncRead for HttpReader {
...
+                                "Error reading bytes from {}: {}, attempting to resume download",
+                                this.http_url,
+                                e
+                            );
+                            tracing::debug!("{:?}", e);
+                        }
+                        if *this.cursor_pos == *this.total_bytes {
+                            return Poll::Ready(Ok(()));
+                        }
+                        continue;
```
redragonx, line 220: i think you can remove this in a enum variant, but i like it explicit anyway.

redragonx (dismissed):
:+1:

backend/src/util/http_reader.rs
dr-bonez: What do you mean?

redragonx (approved):

PR #1750, opened 2022-08-19
remove product key from setup flow

MattDHill:


backend/src/logs.rs
```diff
@@ -120,7 +120,7 @@ pub struct LogResponse {
     end_cursor: Option<String>,
 }
 #[derive(serde::Serialize, serde::Deserialize, Debug, Clone)]
-#[serde(rename_all = "kebab-case", tag = "type")]
```
Blu-J, line 123: @dr-bonez i This should still have the tag right ?

PR #1752, opened 2022-08-19
feat: add declarative `Toast` and `Alert` components

waterplea:


frontend/projects/ui/src/app/components/toast-container/offline-toast/offline-toast.service.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  constructor(
+    private readonly authService: AuthService,
+    private readonly connectionService: ConnectionService,
+  ) {
+    super(subscriber => this.stream$.subscribe(subscriber))
+  }
+}
+
+function getMessage(failure: ConnectionFailure): OfflineMessage | null {
```
waterplea, line 43: This is probably the only place where this code differs from your source refactor, @MattDHill. Hence `TODO` on this file.

PR #1753, opened 2022-08-20
Set pipeline job timeouts and add ca-certificates to test container

moerketh:
As a follow-up on #1682 this adds a more sensible timeout on the pipelines jobs (the default apparently was 6 hours) and adds the ca-certificates apt package to the test container.


dr-bonez:
@moerketh `test_logs` should be skipped

really appreciate the work, btw

dr-bonez (approved):

PR #1775, opened 2022-08-31
pinning cargo dep versions for CLI

redragonx:


moerketh:
Perhaps the pipeline should also be updated? See [here](https://github.com/Start9Labs/embassy-os/blob/705653465a6bdfcaa2b5efbe63beb54919ace922/.github/workflows/backend.yaml#L118) for x86 and [here ](https://github.com/Start9Labs/embassy-os/blob/705653465a6bdfcaa2b5efbe63beb54919ace922/.github/workflows/backend.yaml#L132)for aarch64

redragonx:
> Perhaps the pipeline should also be updated? See [here](https://github.com/Start9Labs/embassy-os/blob/705653465a6bdfcaa2b5efbe63beb54919ace922/.github/workflows/backend.yaml#L118) for x86 and [here ](https://github.com/Start9Labs/embassy-os/blob/705653465a6bdfcaa2b5efbe63beb54919ace922/.github/workflows/backend.yaml#L132)for aarch64

Thanks! I'll update those files.

dr-bonez (approved):

PR #1789, opened 2022-09-09
simplify backend build

dr-bonez:


moerketh:
Looks like the updated start9/rust-arm-cross:aarch64 (which is an amd64 image with a aarch64 tag) currently has the aarch64 cargo-nextest binary installed. That's why the build fails.  Running `curl -LsSf https://get.nexte.st/latest/linux | tar zxf - -C ${CARGO_HOME:-~/.cargo}/bin` on that image would fix it.

After this change different rustc versions would be used (1.62.1 for x86_64 and 1.63 for aarch64). I would consider it a feature that the build currently overrides and unifies this.

dr-bonez:
The goal here is just to reduce build times. Currently, the final image build step recompiles the backend because of the mismatched rust versions, which I'd prefer to avoid. A configurable rust version isn't worth the added time of reinstalling the toolchains every time, or rerunning the build. If we want to change the rust version, we can update the rust-arm-builder image.

moerketh:
Oh I see, hadn't noticed it started doing that. I expect it's rebuilding because an embassy bin "backend/target/aarch64-unknown-linux-gnu/release/avahi-alias" has been added since pipeline 340 that is not caught by the embassy* filter and thus not copied to the image job, so make compiles everything again. Having this logic outside of the Makefile is painful. 
See: https://github.com/Start9Labs/embassy-os/blob/8b794c229974e2a64f2d1a7bbd83ff47e3550a00/.github/workflows/backend.yaml#L136

dr-bonez:
closing in favor of #1886

PR #1806, opened 2022-09-15
add support for gzip compression

dr-bonez:


Blu-J (approved):

dr-bonez:
We cannot merge this. The registry does not support content length with content-encoding gzip

PR #1833, opened 2022-09-26
give name to logs file

### LINKED ISSUE #1830: [feat]: Make logfile download filename correspond to the log's service name
### Prerequisites

- [X] I have searched for [existing issues](https://github.com/start9labs/embassy-os/issues) that already suggest this feature.

Currently pressing the Download button on a Log in the web interface presents you with a download where the default filename offered is logs.html

I suggest we change this to reflect at least the service name of where the log came from, eg. bitcoin-core_logs.html or oslogs_logs.html

It may be even better for support purposes, if people have multiple embassies on their network, to also include the Embassy's name, eg. embassy-lavish-warts_bitcoin-core_logs.html

### Describe the Use Case

When a user goes to look at a log file it might be good to describe where it came from.

My vote is for embassyname_servicename_logs.html

### Describe Alternatives

_No response_

### Anything else?

_No response_

MattDHill:
closes #1830


elvece (approved):
siiick

elvece (approved):

PR #1841, opened 2022-09-28
Feature/setup migrate

elvece:
Just includes logic updates, no styling changes.


frontend/projects/setup-wizard/src/app/pages/transfer/transfer.page.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  async select(target: DiskInfo) {
+    const { logicalname, guid } = target
+
+    if (!logicalname) return
+
+    const alert = await this.alertCtrl.create({
+      header: 'Warning',
+      message:
+        'Data from this drive will not be deleted, but you will not be able to use this drive to run embassyOS after the data is transferred.',
```
elvece, line 52: @MattDHill review copy ^

frontend/projects/setup-wizard/src/app/pages/embassy/embassy.page.ts
```diff
@@ -31,10 +33,14 @@ export class EmbassyPage {
     private readonly stateService: StateService,
     private readonly loadingCtrl: LoadingController,
     private readonly errorToastService: ErrorToastService,
+    private route: ActivatedRoute,
   ) {}
 
   async ngOnInit() {
     await this.getDrives()
+    this.route.queryParams.subscribe(params => {
```
MattDHill, line 41: This creates a subscription that never gets unsubscribed. Instead do

`this.incomingAction = this.route.snapshot.paramMap.get('action')`

Or better yet, just do the snapshot when you need it below. ngOnInit unnecessary.

frontend/projects/setup-wizard/src/app/pages/embassy/embassy.page.ts
```diff
@@ -129,7 +139,9 @@ export class EmbassyPage {
     try {
       await this.stateService.setupEmbassy(logicalname, password)
       if (!!this.stateService.recoverySource) {
-        await this.navCtrl.navigateForward(`/loading`)
+        await this.navCtrl.navigateForward(`/loading`, {
+          queryParams: { action: this.incomingAction },
```
MattDHill, line 143: `{ action: this.route.snapshot.paramMap.get('action') }`

Also, what happens if the query param is not there. Is that possible?

frontend/projects/setup-wizard/src/app/pages/loading/loading.page.ts
```diff
@@ -8,12 +9,18 @@ import { StateService } from 'src/app/services/state.service'
...
+  incomingAction = undefined
+
   constructor(
     public stateService: StateService,
     private navCtrl: NavController,
+    private route: ActivatedRoute,
   ) {}
 
   ngOnInit() {
+    this.route.queryParams.subscribe(params => {
```
MattDHill, line 21: Same thing here as above. Don't subscribe.

frontend/projects/setup-wizard/src/app/pages/transfer/transfer.page.html
```diff
@@ -0,0 +1,63 @@
...
+            <ion-spinner
+              *ngIf="loading"
+              class="center-spinner"
+              name="lines"
+            ></ion-spinner>
+
+            <!-- loaded -->
+            <ion-item-group *ngIf="!loading" class="ion-text-center">
+              <!-- drives -->
+              <h2 class="target-label">Physical Drives</h2>
```
MattDHill, line 28: This heading seems unnecessary since the whole page is physical drives only.

frontend/projects/setup-wizard/src/app/pages/transfer/transfer.page.scss
```diff
@@ -0,0 +1,4 @@
+.target-label {
```
MattDHill, line 1: Delete if deleting the heading from the html.

frontend/projects/setup-wizard/src/app/pages/transfer/transfer.page.ts
MattDHill: Might want to explain the consequences of trying to use the old drive after transfer. Let's talk about that.

frontend/projects/setup-wizard/src/app/pages/transfer/transfer.page.ts
```diff
@@ -0,0 +1,74 @@
+import { Component } from '@angular/core'
```
MattDHill, line 1: This component seems very similar to others. Maybe there is an abstraction here.

MattDHill (changes requested):

frontend/projects/setup-wizard/src/app/pages/transfer/transfer.page.ts
elvece: We can. Aiden and I discussed a way we could mark the drive data as unusable. It would still technically be there if needed for a manual recovery, but it would be rendered useless by our normal setup flows. I forget the exact implementation he described, but it was possible.

frontend/projects/setup-wizard/src/app/pages/embassy/embassy.page.ts
elvece: Not in reality. Can assert that it will always be there. Otherwise, yes would have to handle the case of non-existence.

frontend/projects/setup-wizard/src/app/pages/transfer/transfer.page.ts
elvece: There probably is a base component that could be abstracted, which could be extended with different `getDrive`, and `select`/`chooseDrive` implementations, between the embassy, recover, and transfer pages. They are distinct enough that I decided to keep them separate for now, mainly for clarity.

I think the entire approach here is roundabout and repetitive, and think an overall refactor would be ideal, but that is out of scope for this task.

frontend/projects/setup-wizard/src/app/pages/transfer/transfer.page.ts
elvece: We should discuss, but see updated copy for now.

elvece:
waiting for completion of BE components

Blu-J (approved):

PR #1844, opened 2022-09-29
handle multiple image tags having the same hash and increase timeout

dr-bonez:


backend/src/install/cleanup.rs
```diff
@@ -119,10 +119,20 @@ pub async fn cleanup(ctx: &RpcContext, id: &PackageId, version: &Version) -> Res
...
+            images
+                .into_iter()
+                .flatten()
+                .flat_map(|image| image.repo_tags)
+                .filter(|tag| {
+                    tag.starts_with(&format!("start9/{}/", id))
+                        && tag.ends_with(&format!(":{}", version))
+                })
+                .map(|tag| async {
+                    let tag = tag; // move into future
```
Blu-J, line 132: Isn't that the point of 
```rs
move |tag| async{
}
//instead of 
|tag| async move {
}

Blu-J: Great point, thanks.

Blu-J (approved):

PR #1850, opened 2022-10-01
💫 Rebranding to embassyOS

k0gen:


MattDHill (changes requested):
Please undo changes made to non user facing code.

MattDHill (changes requested):
Definitely still some types that were changed from E -> e. In some places, you actually changed the code from e -> E.

This PR should never change e -> E and only change E -> e for text the user will see.

PR #1866, opened 2022-10-10
Feature/multi platform

dr-bonez:


backend/Cargo.toml
```diff
@@ -42,7 +42,7 @@ path = "src/bin/avahi-alias.rs"
 
 [features]
 avahi = ["avahi-sys"]
-default = ["avahi", "sound", "metal", "js_engine"]
+default = ["avahi", "metal", "js_engine"]
```
Blu-J, line 45:
```suggestion
default = ["avahi", "metal", "sound", "js_engine"]
```

backend/src/disk/util.rs
```diff
@@ -245,42 +246,55 @@ pub async fn list() -> Result<Vec<DiskInfo>, Error> {
...
-            }
-            if !disks.contains_key(&disk) {
-                disks.insert(disk.clone(), IndexSet::new());
-            }
-            if let Some(part_path) = part_path {
-                let part_path = Path::new(DISK_PATH).join(part_path);
-                let part = tokio::fs::canonicalize(&part_path).await.with_ctx(|_| {
+    .try_fold(
+        (BTreeMap::new(), Vec::new()),
+        |(mut disks, mut internal), dir_entry| async move {
```
Blu-J, line 251: I was going to complain this tuple and the names, but the fact that these are two seperate lines makes it readable, kinda nice :)

backend/src/s9pk/docker.rs
```diff
@@ -0,0 +1,94 @@
...
+                        })
+                    }
+                    .boxed()
+                })
+                .try_next()
+                .await?
+            {
+                Ok(Self::MultiArch(image))
+            } else {
+                Err(Error::new(
```
Blu-J, line 73: :question: Is this running the packing or the running of the system. If it is running, how would it be possible that this error would have been hit, or is this an error that should never have been called.

backend/src/s9pk/mod.rs
```diff
@@ -94,33 +99,73 @@ pub async fn pack(#[context] ctx: SdkContext, #[arg] path: Option<PathBuf>) -> R
...
-                .await
-                .with_ctx(|_| {
-                    (
-                        crate::ErrorKind::Filesystem,
-                        manifest.assets.docker_images_path().display().to_string(),
-                    )
-                })?,
-        )
+        .docker_images({
+            let docker_images_path = path.join(manifest.assets.docker_images_path());
```
Blu-J, line 103: :thought_balloon: Wondering if this this be part of the struct of something, the logic being here feels like an abstraction fail.

libs/js_engine/src/lib.rs
```diff
@@ -808,6 +808,15 @@ mod fns {
         let parent = tokio::fs::canonicalize(parent).await?;
         Ok(child.starts_with(parent))
     }
+
```
Blu-J, line 811:
```suggestion

```

Blu-J, line 811: Delete

system-images/binfmt/manifest.json
```diff
@@ -0,0 +1 @@
+[{"Config":"3f7fd8db05afd7a5fa804ebfb0e40038068af52b45d8f9a04c1c9fa8d316e511.json","RepoTags":["multitest/binfmt:latest"],"Layers":["753827173463aa2b7471a5be8ac2323a1cf78ccedfdfcc7bb21831d9e0141732/layer.tar","e0747212aee318099c729893bb4cca2ff164f6bd154841d5777636f82b7f18d4/layer.tar"]},{"Config":"143329dfbcc6abb472070a2f0bae5cf9865f178e5391a15592e75b349c803c69.json","RepoTags":["multitest/binfmt:latest"],"Layers":["de6e72d83667be34f782cf1f0a376e0507d189872142e1934f14d3be4813715a/layer.tar","12670876b666af35f388c3d2160436257b390a163a4c7178863badd05394577a/layer.tar"]}]
```
Blu-J, line 1: This doesn't seem right

system-images/compat/Makefile
```diff
@@ -2,10 +2,17 @@
 
 .DELETE_ON_ERROR:
 
-all: compat.tar
+all: docker-images/aarch64.tar
 
-compat.tar: Dockerfile target/aarch64-unknown-linux-musl/release/compat
-	DOCKER_CLI_EXPERIMENTAL=enabled docker buildx build --tag start9/x_system/compat --platform=linux/arm64 -o type=docker,dest=compat.tar .
+clean:
```
Blu-J, line 7: :speech_balloon: Not sure if this build things will work with both, but I would rather verify than just reason on the code. I'm assuming that these work, if they don't it will be a fail fast, so little harm

Blu-J (approved):
We need the building working in the github, but I think that this works. Most of the logic seems to be that we have multi arch and different notes.

backend/src/s9pk/docker.rs
dr-bonez: this is the running of the system. and it should never be hit

system-images/binfmt/manifest.json
dr-bonez: yeah that wasn't supposed to be committed

Makefile
```diff
@@ -18,45 +18,81 @@ $(shell sudo true)
...
 	rm -f product_key.txt
 	rm -f system-images/**/*.tar
-	sudo rm -f $(EMBASSY_BINS)
+	rm -rf system-images/compat/target
+	rm -rf backend/target
+	rm -rf frontend/.angular
 	rm -f frontend/config.json
 	rm -rf frontend/node_modules
 	rm -rf frontend/dist
+	rm -rf libs/target
```
Blu-J, line 44: :+1:

backend/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 authors = ["Aiden McClelland <me@drbonez.dev>"]
 description = "The core of the Start9 Embassy Operating System"
 documentation = "https://docs.rs/embassy-os"
-edition = "2018"
+edition = "2021"
```
Blu-J, line 5: :+1: Woot

Blu-J (approved):

Issue #1863, opened 2022-10-08
[chore] UI data migration


MattDHill:
Drop `pkg-order`

Change `marketplace` to
```
  'selected-url': string
  'known-hosts': {
    'https://registry.start9.com/': string
    'https://community-registry.start9.com/': string
    [url: string]: string
  }
```

PR #1890, opened 2022-10-26
Feat/patch migration

Blu-J:
fixes: #1863


backend/src/install/mod.rs
```diff
@@ -1436,38 +1390,6 @@ pub async fn install_s9pk<R: AsyncRead + AsyncSeek + Unpin + Send + Sync>(
             &receipts.config.update_dependency_receipts,
         )
         .await?;
-    } else if let Some(recovered) = {
```
Blu-J, line 1439: This is why I don't think that just being dumb and removing the recoveredPackages is a good idea

backend/src/install/mod.rs
```diff
@@ -1436,38 +1390,6 @@ pub async fn install_s9pk<R: AsyncRead + AsyncSeek + Unpin + Send + Sync>(
...
-        receipts
-            .recovered_packages
-            .get(&mut tx)
-            .await?
-            .remove(pkg_id)
-    } {
-        handle_recovered_package(
```
Blu-J, line 1446: This is the only thing that we are removing now?

dr-bonez (approved):

PR #1907, opened 2022-11-04
feat: Exposing the rsync that we have to the js

Blu-J:
### Proof it works

Via the test that is added to the js_scripts.rs


dr-bonez (approved):

Blu-J:
closes: https://github.com/Start9Labs/embassy-os/issues/1713

PR #1917, opened 2022-11-08
Fix/remove merge with seed during init

Blu-J:
During talking in testing, realizing that we may not want to merge with the seed during all the init for the ui.

### DIFF
--- a/backend/src/init.rs
+++ b/backend/src/init.rs
@@ -203,24 +203,6 @@ pub async fn init(cfg: &RpcContextConfig) -> Result<InitResult, Error> {
         .lock(&mut handle, LockType::Write)
         .await?;
 
-    let defaults: serde_json::Value =
-        serde_json::from_str(include_str!("../../frontend/patchdb-ui-seed.json")).map_err(|x| {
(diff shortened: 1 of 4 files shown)


backend/src/init.rs
```diff
@@ -203,24 +203,6 @@ pub async fn init(cfg: &RpcContextConfig) -> Result<InitResult, Error> {
         .lock(&mut handle, LockType::Write)
         .await?;
 
-    let defaults: serde_json::Value =
```
Blu-J, line 206: We don't want this to run on all the nationalizations now.

backend/src/version/v0_3_2.rs
```diff
@@ -25,6 +49,14 @@ impl VersionT for Version {
         crate::hostname::ensure_hostname_is_set(db, &receipts).await?;
         receipts.id.set(db, generate_id()).await?;
 
+        let mut ui = crate::db::DatabaseModel::new()
```
Blu-J, line 52: And we move this here since this was the intended behavior, or what was the case during the 0.3.2

PR #1931, opened 2022-11-09
disable efi

dr-bonez:


chrisguida (approved):
Looks good, awaiting iso for proper test

PR #1950, opened 2022-11-16
changing ip addr type

redragonx:


backend/src/net/proxy_controller.rs
```diff
@@ -104,34 +104,16 @@ impl ProxyController {
             // connection be upgraded, so we can't return a response inside
             // `on_upgrade` future.
             match host_addr_fqdn(&req) {
```
dr-bonez, line 106: don't even call host_addr_fqdn

dr-bonez (changes requested):

backend/src/net/proxy_controller.rs
redragonx: good point now we just use addr.

dr-bonez (approved):

PR #1952, opened 2022-11-18
📺 Full screen mode by default

k0gen:


build/lib/scripts/enable-kiosk
```diff
@@ -58,6 +58,7 @@ while ! curl "http://localhost" > /dev/null; do
     sleep 1
 done
 matchbox-window-manager -use_titlebar no &
+xdotool search --sync --onlyvisible --class "Firefox" windowactivate key F11 &
```
redragonx, line 61: I would use a variable for the class name.

k0gen: ok look again 👇

dr-bonez:
deciding not to do this. users can press f11 if they want fullscreen.

PR #1958, opened 2022-11-21
feature: 0.3.2 -> 0.3.3 upgrade

dr-bonez:


build/raspberry-pi/033-upgrade.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+
+set -e
+
+if grep 'cb15ae4d-03' /media/boot-rw/cmdline.txt; then
+    BLOCK_COUNT=$(tune2fs -l /dev/mmcblk0p3 | grep "^Block count:" | awk '{print $3}')
+    BLOCK_SIZE=$(tune2fs -l /dev/mmcblk0p3 | grep "^Block size:" | awk '{print $3}')
```
Blu-J, line 7: Interesting...

build/raspberry-pi/033-upgrade.sh
```diff
@@ -0,0 +1,68 @@
...
+rsync -acvAXUH --info=progress2 /media/root-rw/current/boot/ /media/boot-rw/
+
+sync
+
+umount /media/root-rw
+
+e2label /dev/mmcblk0p3 rootfs
+
+(
+    echo d
```
Blu-J, line 44: This is readable :+1:

Blu-J (approved):
Let me know when you update, but seems reasonable, I thought would be more complicated
Also, what happens when the main drive is on p4 instead of p3 or vice versa?

dr-bonez:
> Let me know when you update, but seems reasonable, I thought would be more complicated Also, what happens when the main drive is on p4 instead of p3 or vice versa?

if we are on p3, we switch over to p4, to give us room, since the new partition overwrites p3

PR #1959, opened 2022-11-21
Embassy-init typo

redragonx:


dr-bonez:
not using `tokio::main` was intentional. we want to wait for the runtime to shutdown before calling `shutdown.execute`

redragonx:
afaik, it still worked that way, but I will undo it since it's more explicit that way.

dr-bonez (approved):

PR #1963, opened 2022-11-22
Disk fix part 1 (handle all kinds of disk partition signatures) 

redragonx:


k0gen (dismissed):
`wipefs -a -f` is the way to go here

MattDHill:
Closing as doesn't seem to be an issue

PR #1979, opened 2022-11-26
fix: Add the locking to the package during the backup.

Blu-J:


backend/src/install/mod.rs
```diff
@@ -601,6 +601,11 @@ pub async fn uninstall_dry(
 pub async fn uninstall_impl(ctx: RpcContext, id: PackageId) -> Result<(), Error> {
     let mut handle = ctx.db.handle();
     let mut tx = handle.begin().await?;
+    crate::db::DatabaseModel::new()
```
Blu-J, line 604: Makes sense that there should be a lock during the uninstall. It used to get this lock by default by getting the model on the next line. Now we have to make it more explicit.

backend/src/backup/backup_bulk.rs
```diff
@@ -350,6 +350,11 @@ async fn perform_backup<Db: DbHandle>(
             .await;
 
         let mut tx = db.begin().await?;
+        let lock_package = crate::db::DatabaseModel::new()
```
Blu-J, line 353: This should be the locking that we need for the system since we are needed to lock during the backup. Again this used to be covered by the system, and we don't want to remove the package while we are backing it up.

backend/src/install/cleanup.rs
```diff
@@ -342,6 +342,11 @@ where
     for<'a> &'a mut Ex: Executor<'a, Database = Postgres>,
 {
     let mut tx = db.begin().await?;
+    crate::db::DatabaseModel::new()
+        .package_data()
+        .idx_model(&id)
+        .lock(&mut tx, LockType::Write)
```
Blu-J, line 348: Lower down we get the package as a mut. Again I believe that this system used to lock the package. Now we are making this more explicit.

dr-bonez (approved):

PR #1983, opened 2022-11-28
Update/marketplace info

elvece:
This PR updates the marketplace info endpoint to include eos version with the request and ensures that the system architecture is always accurate and present in config so that the registry can use this information to log metrics.


frontend/projects/ui/src/app/services/api/embassy-live-api.service.ts
```diff
@@ -177,22 +178,32 @@ export class LiveApiService extends ApiService {
 
   // marketplace URLs
 
-  async marketplaceProxy<T>(path: string, qp: {}, baseUrl: string): Promise<T> {
```
Blu-J, line 180: Isn't qp a type of `Record<String, String>`

PR #2016, opened 2022-12-03
fix: Service isn't dying

Blu-J:


backend/src/manager/mod.rs
```diff
@@ -438,6 +440,9 @@ async fn manager_thread_loop(mut recv: Receiver<OnStop>, thread_shared: &Arc<Man
                     }
                 }
                 tracing::error!("service crashed: {}: {}", e.0, e.1);
+                if &e.1 == KILLED && e.0 == 137 {
+                    break;
```
Blu-J, line 444: If there was our killed by the kill signal, expecting that we are either killing or exiting so we shouldn't continue on this loop

Blu-J:
Not the solution at this point

Blu-J:
Repro: 


1.    Install Specter from prod
1.   Configure it
1.    Start it.
1.    Wait a minute
1.    Update Specter from Alpha or Beta
1.   wait a minute
1.    open specter show page in emabssyOS

Blu-J:
Something about the state machine is that it is killing after we created. Seems wrong
image

Blu-J:
Best that I could figure out,
is that during the install there is a period where the previous manager for this service is running and the new one. 
This makes the the new manager get into a state that it is running main, but it isn't working right !?
The previous add was only used during the installation. And the removal only removed the exact package id and version. 
I don't know why that was the expected. I figure the add for the service managers only wants one. That way we are not fighting the db of shared db or resources. And as far as I can tell at this point the service would be valid, so the only errors that would escape are db rare ones if ever.

backend/src/manager/mod.rs
```diff
@@ -87,13 +87,17 @@ impl ManagerMap {
         manifest: Manifest,
         tor_keys: BTreeMap<InterfaceId, TorSecretKeyV3>,
     ) -> Result<(), Error> {
+        let keys = self.0.read().await.keys().cloned().collect::<Vec<_>>();
         let mut lock = self.0.write().await;
-        let id = (manifest.id.clone(), manifest.version.clone());
-        if let Some(man) = lock.remove(&id) {
-            if !man.thread.is_empty().await {
-                man.exit().await?;
+        for id in keys {
```
Blu-J, line 92: Not sure that we want to leave the previous ones running and start another running service. Also, wondering why the keys are packageId and the Versions.

PR #2028, opened 2022-12-05
make procfs an optional dependency so sdk can build on macos

elvece:
It seems like `procfs` was added in this [commit]( https://github.com/Start9Labs/embassy-os/commit/a3d1b2d6714649e1a554a1b2e74c76fa81cf25a1), but on master, the only reference is appearing in Cargo.lock. Perhaps the lockfile was just not updated post removing this dependency? @Blu-J or @dr-bonez please confirm.

If it is needed, this PR reintroduces the dependency so that the SDK can be built on macos machines. Tested and this does indeed fix running `backend/install-sdk.sh` on a mac machine.

### DIFF
--- a/backend/Cargo.lock
+++ b/backend/Cargo.lock
@@ -1234,7 +1234,6 @@ dependencies = [
  "pin-project",
  "pkcs8",
  "prettytable-rs",
- "procfs",
  "proptest",
  "proptest-derive",
  "rand 0.7.3",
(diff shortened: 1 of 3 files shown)


dr-bonez:
pretty sure this won't work. it's `embassy_container_init` that needs to be made an optional dependency

Blu-J (approved):

chrisguida (approved):
Tested, works

:shipit:

PR #2037, opened 2022-12-06
fix: Make the restores limited # restore at a time

Blu-J:
Trying to fix the error seen in here with timeouts
image


dr-bonez (approved):

gStart9:
This fixed the connection pool timeouts!  That error has disappeared upon the newest full-restore-from-cifs-backup-at-setup.

PR #2076, opened 2022-12-18
Add logs window to setup wizard loading screen

MattDHill:


frontend/projects/setup-wizard/src/app/services/api/live-api.service.ts
```diff
@@ -87,6 +90,14 @@ export class LiveApiService extends ApiService {
     })
   }
 
+  async followLogs(): Promise<string> {
+    return this.rpcRequest({ method: 'setup.logs.follow', params: {} })
+  }
+
+  openLogsWebsocket$(guid: string): Observable<Log> {
+    return webSocket(`http://embassy.local/ws/${guid}`)
```
elvece, line 98:
```suggestion
    return webSocket(`http://start.local/ws/${guid}`)
```

elvece (approved):

PR #2098, opened 2023-01-05
fix config bug for number 0 and also maybe conenction icon

### LINKED ISSUE #2097: [bug]: embassyOS Setup's Password confirmation field is not length restricted
### Prerequisites

- [X] I have searched for [existing issues](https://github.com/start9labs/embassy-os/issues) that already report this problem.

### embassyOS Version

v0.3.3

### Device

Laptop/Desktop

### Device OS

Linux

### Device OS Version

5

### Browser

Firefox

### Browser Version

108

### Current Behavior

A user wrote us:

> when making a password during setup the first text box limits the maximum # of characters and the second does not. Wasn’t sure why my pasted password wasn’t matching at first until I revealed and inspected. 
>

### Expected Behavior

Both password and password confirmation text boxes should limit the # of chars.  We should probably also have a tooltip about the password length (and any complexity) constraints.

### Steps to Reproduce

Setup embassy OS
Proceed to Set Password screen
Paste a very long password into both fields

### Anything else?

_No response_

MattDHill:
Also closes #2097

### DIFF
--- a/frontend/projects/setup-wizard/src/app/modals/password/password.page.html
+++ b/frontend/projects/setup-wizard/src/app/modals/password/password.page.html
@@ -48,8 +48,9 @@
             [(ngModel)]="passwordVer"
             [ngModelOptions]="{'standalone': true}"
             [type]="!unmasked2 ? 'password' : 'text'"
-            (ionChange)="checkVer()"
             placeholder="Retype Password"
+            (ionChange)="checkVer()"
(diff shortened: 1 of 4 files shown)


elvece (dismissed):

elvece (approved):

dr-bonez (approved):
rebased

PR #2102, opened 2023-01-06
only do standby mode for pi

### LINKED ISSUE #2044: [fix] Shutdown for Pro should not reboot into standby mode

dr-bonez:
closes #2044


Blu-J (dismissed):
Seems like the obvious solution, good job.

PR #2105, opened 2023-01-09
Feat/marketplace show links

### LINKED ISSUE #2084: Add "replace legacy service" filters to marketplace
> Either way, I like it.
> 
> I believe I was originally trying to solve a Telegram Group problem more than a marketplace problem, if you will.
> There was a lot of "Where's Matrix?" "How is Synapse different from Matrix?" etc in the community and I think that streamlining the way these esoteric concepts are discussed will help the layperseon. But yeah, hard problem.
> 
> Since you're going with tags, what if you added some curated layperson buttons at top of the marketplace like "replace whatsapp" and "replace icloud", and those would filter the marketplace list on anything tagged #whatsapp etc. And tie that to the url, a la `/marketplace#whatsapp` so the link can be shared.
> 

_Originally posted by @spencerflagg in https://github.com/Start9Labs/embassy-os/issues/1561#issuecomment-1358150284_

MattDHill:
- closes #2084
- rearranges marketplace show and app show
- adds donation link for Start9

### DIFF
--- a/backend/src/s9pk/manifest.rs
+++ b/backend/src/s9pk/manifest.rs
@@ -76,6 +76,9 @@ pub struct Manifest {
     pub dependencies: Dependencies,
     #[model]
     pub containers: Option<DockerContainers>,
+
+    #[serde(default)]
+    pub replaces: Option<Vec<String>>,
 }
 
 impl Manifest {
--- a/backend/src/s9pk/manifest.rs
+++ b/backend/src/s9pk/manifest.rs
@@ -78,7 +78,7 @@ pub struct Manifest {
     pub containers: Option<DockerContainers>,
 
     #[serde(default)]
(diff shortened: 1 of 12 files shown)


elvece (dismissed):

elvece (approved):

backend/src/s9pk/manifest.rs
```diff
@@ -76,6 +76,9 @@ pub struct Manifest {
     pub dependencies: Dependencies,
     #[model]
     pub containers: Option<DockerContainers>,
+
+    #[serde(default)]
+    pub replaces: Option<Vec<String>>,
```
dr-bonez, line 81: Why `Option<Vec>` and not just `Vec`? Do we really want it to be `null` instead of `[]` by default?

PR #2137, opened 2023-01-26
Night theme

waterplea:


kn0wmad:
> _ No description provided. _

It only gets so much darker...

kn0wmad:
Red light filter?

waterplea:
> Red light filter?

We will be able to make all sorts of themes in the future, right now it's only about basic light/dark mode. Especially when we ditch iconic and come up with brand palette it would be easy.

MattDHill:
Light mode is too light haha. Need to soften it up a little with some light grays, etc

elvece:
Experiencing some odd behavior here (on mocks) - login screen is always white, but UI is always dark once logged in. The icon to toggle no longer seems to be present? Running the latest code on this branch. Made sure to clear cache. 

This palette might help inform a better scheme of gray than the stark white:

[Start9 Colors.pdf](https://github.com/Start9Labs/embassy-os/files/10799061/Start9.Colors.pdf)

backend/src/version/v0_3_3_1.rs
```diff
@@ -61,6 +61,8 @@ impl VersionT for Version {
                 .put(db, &parsed_url)
                 .await?;
         }
+        ui["theme"] = json!("Dark".to_string());
+        ui["widgets"] = json!([]);
```
Blu-J, line 65: I believe that I had a fix to save the migration, hopefully I don't forget this migration for the big refactor we are in.

backend/src/version/v0_3_3_1.rs
```diff
@@ -87,6 +89,11 @@ impl VersionT for Version {
                 .await?;
         }
 
+        if let Value::Object(ref mut obj) = *ui {
+            obj.remove("theme");
+            obj.remove("widgets");
+        }
+
         ui["marketplace"]["known-hosts"][COMMUNITY_URL].take();
         ui.save(db).await?;
```
Blu-J, line 98: This is what was forgotten on the up.

backend/src/version/v0_3_3_1.rs
Blu-J: :notebook: These are notes for me, but the code looks great. I'm going to have to merge this into the big refactor

MattDHill (approved):

PR #2146, opened 2023-02-01
feat: make widgets sidebar width togglable

waterplea:


elvece:
A few things:

1. On first load with live data, the service list page arrangement is off (if widgets are opened or closed):
Screen Shot 2023-02-02 at 7 00 02 PM
Screen Shot 2023-02-02 at 7 03 06 PM


2. For the desktop view of the widgets sidebar, rather than expanding the tile to responsively fill the whole screen, we want the tile to stay at its predetermined fixed side. For example with the health widget, it's configured as 2x2 for both desktop and mobile, so it should look the same in both views. Here is how it looks now, compared to how I Matt and I envisioned, expressed in the mockups below:
Screen Shot 2023-02-02 at 7 07 09 PM

Mockups:
Dashboard-1
Dashboard

elvece:
@dr-bonez Do I need to propagate the change you made to limit the build name somewhere else? It seems to be failing here for the same issue, but I cannot find where you committed the fix.

elvece (approved):

PR #2151, opened 2023-02-05
Better Updates Tab and updates count

MattDHill:
**Goals**
1. Registries containing the string 'alpha' or 'beta' should only show in the Updates Tab and count towards updates badge count if dev tools enabled.
2. In the Updates Tab, service updates should show in _every_ registry where there is an update available (not just the registry from which it was originally installed)
3. Updates badge count should be de-duplicated. For example, if there is an update available for Bitcoin in 3 different registries, the badge should only count that as 1.

### DIFF
--- a/frontend/projects/ui/src/app/app/menu/menu.component.ts
+++ b/frontend/projects/ui/src/app/app/menu/menu.component.ts
@@ -1,14 +1,12 @@
 import { ChangeDetectionStrategy, Component, Inject } from '@angular/core'
 import { EOSService } from '../../services/eos.service'
 import { PatchDB } from 'patch-db-client'
-import { combineLatest, map, Observable, of, startWith } from 'rxjs'
+import { combineLatest, first, map, Observable } from 'rxjs'
 import { AbstractMarketplaceService } from '@start9labs/marketplace'
 import { MarketplaceService } from 'src/app/services/marketplace.service'
 import { DataModel } from 'src/app/services/patch-db/data-model'
 import { SplitPaneTracker } from 'src/app/services/split-pane.service'
-import { Emver, sameDomain } from '@start9labs/shared'
-import { versionLower } from '../../pages/updates/updates.page'
-import { ClientStorageService } from 'src/app/services/client-storage.service'
+import { Emver } from '@start9labs/shared'
 
 @Component({
   selector: 'app-menu',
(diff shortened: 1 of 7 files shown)


frontend/projects/ui/src/app/app/menu/menu.component.ts
```diff
@@ -54,22 +53,25 @@ export class MenuComponent {
   readonly showEOSUpdate$ = this.eosService.showUpdate$
 
   readonly updateCount$: Observable<number> = combineLatest([
-    this.marketplaceService.getMarketplace$(),
-    this.patch.watch$('package-data'),
+    this.marketplaceService.getMarketplace$(true),
+    this.patch.watch$('package-data').pipe(first()),
```
MattDHill, line 57: Curious why this wasn't `first()` before. Without `first()` I would expect this count to go up forever, since it would fire every time package-data is updated. @waterplea do you know?

waterplea: The count will not go up, it will just get refreshed when an update appears since it recalculates the number on each emit. This way updates count should actually go down when we install one of the updates available. At least that's what I think it was supposed to do.

frontend/projects/ui/src/app/app/menu/menu.component.ts
```diff
@@ -54,22 +53,25 @@ export class MenuComponent {
...
-          (store?.packages.filter(
-            ({ manifest }) =>
-              marketplaceSame(manifest, local, url) &&
-              versionLower(manifest, local, this.emver),
-          ).length || 0),
-        0,
-      ),
+      Object.entries(marketplace).reduce((list, [_, store]) => {
+        store?.packages.forEach(({ manifest: { id, version } }) => {
+          if (!local[id]) return
```
waterplea, line 62: Looks like this `if` can be joined with the next one:
```ts
if (
  local[id] && 
  this.emver.compare(
    version,
    local[id].installed?.manifest.version || '',
  ) === 1
)
```

waterplea: Or even just `local[id]?.installed?.manifest.version || ''`

waterplea: Also can we reuse `versionLower`?

frontend/projects/ui/src/app/services/marketplace.service.ts
```diff
@@ -55,6 +56,20 @@ export class MarketplaceService implements AbstractMarketplaceService {
       }),
     )
 
+  private readonly filteredKnownHosts: Observable<StoreIdentity[]> =
```
waterplea, line 59: If we follow the convention, let's add `$` at the end.

frontend/projects/ui/src/app/services/marketplace.service.ts
```diff
@@ -55,6 +56,20 @@ export class MarketplaceService implements AbstractMarketplaceService {
...
+    combineLatest([
+      this.clientStorageService.showDevTools$,
+      this.knownHosts$,
+    ]).pipe(
+      map(([devMode, knownHosts]) => {
```
waterplea, line 64: I would use ternary operator and drop brackets and `return` keyword here and in `filteredMarketplace$`, but that's just my preferential code style:
```ts
map(([devMode, knownHosts]) => devMode
  ? knownHosts
  : knownHosts.filter(...)
),
```

frontend/projects/ui/src/app/services/marketplace.service.ts
```diff
@@ -93,6 +108,23 @@ export class MarketplaceService implements AbstractMarketplaceService {
     shareReplay(1),
   )
 
+  private readonly filteredMarketplace$ = combineLatest([
```
waterplea, line 111: Maybe a comment or something would be helpful, so later it is clear what we filter here at a glance?

frontend/projects/ui/src/app/services/marketplace.service.ts
```diff
@@ -110,18 +142,19 @@ export class MarketplaceService implements AbstractMarketplaceService {
...
+    return filtered ? this.filteredKnownHosts : this.knownHosts$
   }
 
   getSelectedHost$(): Observable<StoreIdentity> {
     return this.selectedHost$
   }
 
-  getMarketplace$(): Observable<Marketplace> {
-    return this.marketplace$
+  getMarketplace$(filtered = false): Observable<Marketplace> {
```
waterplea, line 156: Maybe a comment or something would be helpful, so later it is clear what we filter here at a glance?

waterplea (approved):
Looks fine to me, just some minor mostly aesthetic comments.

frontend/projects/ui/src/app/pages/updates/updates.page.ts
```diff
@@ -191,33 +175,21 @@ export class FilterUpdatesPipe implements PipeTransform {
...
-    return pkgs.filter(
-      ({ manifest }) =>
-        marketplaceSame(manifest, local, url) &&
-        versionLower(manifest, local, this.emver),
-    )
+    return pkgs.filter(({ manifest }) => {
+      const { id, version } = manifest
+      const localPkg = local[id]
+
+      if (!localPkg) return false
```
waterplea, line 184: Same comment as before, probably can just use `?.` accessor in the `return` statement below.

elvece (approved):

PR #2159, opened 2023-02-19
fix(shared): fix responsive column size

waterplea:


elvece:
Fixed up the responsiveness on a few more pages. The `logs` component needs investigating though - `with-widgets` class does not seem to be working on the `ion-content` there, nor on the outer element for os/kernel logs. There ahould be a download button that shows on the lower right hand side.

elvece:
Possibly unimportant because we will probably not release with the feature flag off, but if `"enableWidgets": false`, the skeleton loading window display is off, and all of the content stays adjusted as if the widgets sidebar was present:
Screen Shot 2023-02-21 at 5 14 45 PM
Screen Shot 2023-02-21 at 5 56 41 PM

waterplea:
@elvece fixed your comments

MattDHill (approved):

PR #2224, opened 2023-03-16
update display obj on union change

MattDHill:


frontend/projects/ui/src/app/components/form-object/form-object.component.ts
```diff
@@ -63,6 +64,37 @@ export class FormObjectComponent {
...
+    // setTimeout hack to avoid ExpressionChangedAfterItHasBeenCheckedError
+    setTimeout(() => {
+      if (this.original) {
+        Object.keys(this.current || {}).forEach(key => {
+          if ((this.original as Config)[key] === undefined) {
+            this.hasNewOptions.emit()
+          }
+        })
+      }
+    }, 10)
```
waterplea, line 78: You don't need any time, just default 0 is enough to move it to the next frame.

frontend/projects/ui/src/app/components/form-object/form-object.component.ts
```diff
@@ -63,6 +64,37 @@ export class FormObjectComponent {
...
 
   ngOnInit() {
+    this.setDisplays()
+
+    // setTimeout hack to avoid ExpressionChangedAfterItHasBeenCheckedError
+    setTimeout(() => {
+      if (this.original) {
+        Object.keys(this.current || {}).forEach(key => {
+          if ((this.original as Config)[key] === undefined) {
+            this.hasNewOptions.emit()
```
waterplea, line 74: Do we need it for every new option?

frontend/projects/ui/src/app/components/form-object/form-object.component.ts
MattDHill: fixed

frontend/projects/ui/src/app/components/form-object/form-object.component.ts
MattDHill: No, if any options are new, we can emit and abort

elvece (approved):

PR #2226, opened 2023-03-16
fix loading authcookie into cookie store on ssh

dr-bonez:


backend/src/context/cli.rs
```diff
@@ -54,7 +54,8 @@ impl Drop for CliContextSeed {
             true,
         )
         .unwrap();
-        let store = self.cookie_store.lock().unwrap();
+        let mut store = self.cookie_store.lock().unwrap();
+        store.remove("localhost", "", "local");
```
Blu-J, line 58: Do we need to save the cookie store?

Blu-J (approved):

PR #2240, opened 2023-03-22
fix: Add type of archive

Blu-J:
### About 

So, we have noticed the following error. 
We back up a service that uses rsync, like nostr.
We backup to exfat, fat32, and (probably ntfs)
Then the backup fails.

### Research

There are so many options that are part of the archive, and windows won't support almost most of them.
This still doesn't solve the length and name too long. But if we do just a simple recursive it will work on the exfat drives.

### Downsides

Now the default is not going to preserve dates, user information, permissions, or links (soft/hard)

### Proof of working 

https://user-images.githubusercontent.com/2364004/227014330-0f55c09e-a85d-4a0c-b6f4-fa94191dfad2.mp4

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -183,7 +183,7 @@ frontend/config.json: $(GIT_HASH_FILE) frontend/config-sample.json
 	npm --prefix frontend run-script build-config
 
 frontend/patchdb-ui-seed.json: frontend/package.json
-	jq '."ack-welcome" = "$(shell yq '.version' frontend/package.json)"' frontend/patchdb-ui-seed.json > ui-seed.tmp
+	jq '."ack-welcome" = $(shell yq '.version' frontend/package.json)' frontend/patchdb-ui-seed.json > ui-seed.tmp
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -183,7 +183,7 @@ frontend/config.json: $(GIT_HASH_FILE) frontend/config-sample.json
 	npm --prefix frontend run-script build-config
 
 frontend/patchdb-ui-seed.json: frontend/package.json
-	jq '."ack-welcome" = "$(shell yq '.version' frontend/package.json)"' frontend/patchdb-ui-seed.json > ui-seed.tmp
+	jq '."ack-welcome" = $(shell yq '.version' frontend/package.json)' frontend/patchdb-ui-seed.json > ui-seed.tmp
```
MattDHill, line 186: What were the consequences of this?

Blu-J: Thought I commented, the issue might be my yq

PR #2246, opened 2023-03-29
Feature/pwa

elvece:


frontend/package-lock.json
elvece, line 1: curious - this might be because I'm using node v18.15.0 now?

MattDHill: I'm also using that and don't have this issue

MattDHill: Oh, maybe because it's already been done on `integration/refactors`

frontend/projects/ui/src/environments/environment.ts
```diff
@@ -4,6 +4,7 @@
 
 export const environment = {
   production: false,
+  useServiceWorker: true,
```
MattDHill, line 7: Is this dev?

frontend/package-lock.json
elvece: oh yeah, that makes sense. Hm, I think its fine to make this change now (after I fix the build).

frontend/projects/shared/assets/img/logo_solid.png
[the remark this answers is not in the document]
elvece: It is not, it just needs to be a min of 144x144. I can shrink it a bit though since it will never render this large.

frontend/projects/ui/src/environments/environment.ts
elvece: yeah - well, any environment that is not prod.

MattDHill (approved):

elvece:
@waterplea Since we have the `RefreshAlertService`, the user is prompted to refresh the page on any version mismatch between the FE and BE. From what I have researched and in testing, this refresh is sufficient for the service worker to update the cache. Any browser refresh or PWA reload works to get the new file versions. 

Questions:

1. Matt and I had discussed programmatically refreshing the cache within this service, but I do not think it is as safe or necessary. Let me know if you think otherwise. 
2. Do you recommend adding a service worker configuration for [data groups](https://angular.io/guide/service-worker-config#datagroups), possibly for marketplace data?

PR #2263, opened 2023-05-05
Add sans-serif font fallback

BBlackwo:
While the fonts are loading, headings default to a serif font, so it looks a bit funny and it's more jarring when the sans-serif fonts load.

Some places already had a fallback (e.g. the [side navigation](https://github.com/Start9Labs/embassy-os/blob/master/frontend/projects/shared/styles/shared.scss#L114)), this PR just adds it everywhere else.

### Before

Screenshot from 2023-05-05 13-45-24

### After

Screenshot from 2023-05-05 13-45-47

## Resources

- https://css-tricks.com/css-basics-fallback-font-stacks-robust-web-typography/

### DIFF
--- a/frontend/projects/diagnostic-ui/src/styles.scss
+++ b/frontend/projects/diagnostic-ui/src/styles.scss
@@ -7,7 +7,7 @@
 
 /** Ionic CSS Variables overrides **/
 :root {
-  --ion-font-family: 'Montserrat';
+  --ion-font-family: 'Montserrat', sans-serif;
 
   --ion-color-primary: #0075e1;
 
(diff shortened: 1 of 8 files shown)


waterplea (approved):

frontend/projects/diagnostic-ui/src/styles.scss
```diff
@@ -7,7 +7,7 @@
 
 /** Ionic CSS Variables overrides **/
 :root {
-  --ion-font-family: 'Montserrat';
+  --ion-font-family: 'Montserrat', sans-serif;
```
BBlackwo, line 10: A future improvement would be to put this into a `variables.scss` file that can be imported and used throughout the codebase rather than hard coding everywhere.

MattDHill (approved):

MattDHill:
Thank you @BBlackwo this will go into 040

PR #2271, opened 2023-05-12
display icons based on mime type

elvece:
This is an MVP solution for displaying icons based on their mime type (file extension). It does not handle dependencies because a registry feature is needed to accommodate this. Dependencies are hard coded to use `.png` files,  and this is fine because they all currently this type.

Impetus: Needed for 0342 release so that new community services can use svg icons.

### DIFF
--- a/frontend/projects/marketplace/src/pipes/mime-type.pipe.ts
+++ b/frontend/projects/marketplace/src/pipes/mime-type.pipe.ts
@@ -6,16 +6,14 @@ import { MarketplacePkg } from '../types'
 })
 export class MimeTypePipe implements PipeTransform {
   transform(pkg: MarketplacePkg): string {
-    switch (pkg.manifest.assets.icon.split('.')[1]) {
+    switch (pkg.manifest.assets.icon.split('.').pop()) {
       case 'png':
         return `data:image/png;base64,${pkg.icon}`
       case 'jpeg':
       case 'jpg':
         return `data:image/jpeg;base64,${pkg.icon}`
-        break
       case 'gif':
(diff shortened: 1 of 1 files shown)


MattDHill (dismissed):

frontend/projects/marketplace/src/pipes/mime-type.pipe.ts
```diff
@@ -0,0 +1,32 @@
+import { NgModule, Pipe, PipeTransform } from '@angular/core'
+import { MarketplacePkg } from '../types'
+
+@Pipe({
+  name: 'mimeType',
+})
+export class MimeTypePipe implements PipeTransform {
+  transform(pkg: MarketplacePkg): string {
+    switch (pkg.manifest.assets.icon.split('.')[1]) {
```
dr-bonez, line 9: you technically want the last here, not `[1]`

frontend/projects/marketplace/src/pipes/mime-type.pipe.ts
```diff
@@ -0,0 +1,32 @@
...
+      case 'png':
+        return `data:image/png;base64,${pkg.icon}`
+      case 'jpeg':
+        return `data:image/jpeg;base64,${pkg.icon}`
+      case 'jpg':
+        return `data:image/jpeg;base64,${pkg.icon}`
```
dr-bonez, line 15:
```suggestion
      case 'jpeg':
      case 'jpg':
        return `data:image/jpeg;base64,${pkg.icon}`
```

dr-bonez (approved):

PR #2278, opened 2023-05-17
wait for tor

dr-bonez:


gStart9 (approved):
Obviously I don't truly understand what's going on in init but I can "Review" the PR based on the new logic you put in... which I think I follow:
you made a new function check_tor_is_ready
you wait up to a minute for tor to start, in which case you continue doing whatever it was that tor not being started previously prevented.  If tor's control port doesn't become available, it times out.

gStart9:
The fix worked!

PR #2309, opened 2023-06-18
Add qemu-guest-agent for advanced VM shutdown options

gStart9:
This may help anybody running StartOS in a VM to:
* properly shutdown the guest, instead of relying on ACPI commands
* freeze the guest file system when making a backup/snapshot, entering suspend, and thawing filesystems
* In the phase when the guest (VM) is resumed after pause (for example after shapshot) it immediately synchronizes its time with the hypervisor using qemu-guest-agent (as first step).

More info:
https://wiki.libvirt.org/Qemu_guest_agent.html
https://pve.proxmox.com/wiki/Qemu-guest-agent

### DIFF
--- a/build/lib/depends
+++ b/build/lib/depends
@@ -34,6 +34,7 @@ nyx
 openssh-server
 postgresql
 psmisc
+qemu-guest-agent
 rsync
 samba-common-bin
 smartmontools


gStart9:
As an initial test I added qemu-guest-agent via /usr/lib/embassy/scripts/chroot-and-upgrade.sh

Installing qemu-guest-agent also installs 1 dependency: `liburing1`

When the VM came back up, `ps auxfw` shows the new process running:
`root         458  0.0  0.0  80076  3316 ?        Ssl  19:31   0:00 /usr/sbin/qemu-ga`

Which is controlled through a new systemd service:
`root@prepaid-pleats:~# systemctl status qemu-guest-agent
● qemu-guest-agent.service - QEMU Guest Agent
     Loaded: loaded (/lib/systemd/system/qemu-guest-agent.service; static)
     Active: active (running) since Sun 2023-06-18 19:31:10 UTC; 11min ago
   Main PID: 458 (qemu-ga)
      Tasks: 2 (limit: 23925)
     Memory: 1.4M
        CPU: 379ms
     CGroup: /system.slice/qemu-guest-agent.service
             └─458 /usr/sbin/qemu-ga

Notice: journal has been rotated since unit was started, output may be incomplete.
```

I wonder if there's a way we can detect if the installation is running as a VM and optionally turn qemu-guest-agent on or off, since it doesn't need to be running on a bare metal server.  Will research.

**Comment by k0gen** [comment]
> I wonder if there's a way we can detect if the installation is running as a VM and optionally turn qemu-guest-agent on or off, since it doesn't need to be running on a bare metal server. Will research.

Sure we can, by using `sudo dmidecode -s system-product-name`
Though on RPi this will not output anything (there is no SMBIOS nor DMI entry point on RPi) but on anything else it will work. Some examples:

Start9 Pure
```
sudo dmidecode -s system-product-name
librem_mini_v2
```
StartOS on M1 Mac in UTM
```
sudo dmidecode -s system-product-name
QEMU Virtual Machine
```

dr-bonez:
I have tested that the agent does not activate on non-virtualized hardware

PR #2337, opened 2023-07-06
Update README.md

kn0wmad:


MattDHill (commented):
I think many will want to copy/paste the email instead of launching the default email app on their device. So maybe drop the link and just display the address?

kn0wmad:
> I think many will want to copy/paste the email instead of launching the default email app on their device. So maybe drop the link and just display the address?

My reasoning here was to obscure the email from a basic text search of the page.  You can right-click and copy email addresses.  However, I changed to a regular link as requested.

MattDHill:
That's a good point and would probably be fine as such. But I'm fine either way, approving.

MattDHill (approved):

PR #2348, opened 2023-07-12
disable encryption for new raspi setups

dr-bonez:


Blu-J (approved):
It's okay, could refactor out options and booleans, but not sure worth it.

PR #2360, opened 2023-07-17
Fix bugs

MattDHill:
fix reset tor, delete http redirect, show message for tor http, update release notes


frontend/projects/ui/src/app/app.component.ts
```diff
@@ -39,14 +39,6 @@ export class AppComponent implements OnDestroy {
   ) {}
 
   async ngOnInit() {
-    if (location.hostname !== 'localhost' && location.protocol === 'http:') {
```
elvece, line 42: are we moving the redirect to the BE? or abandoning it alltogether?

frontend/projects/ui/src/app/modals/os-welcome/os-welcome.page.html
```diff
@@ -26,12 +26,12 @@ <h4>0.3.4.4</h4>
...
-    <li>Enabled use of IP address for local network backups</li>
-    <li>Added change password flow to the UI</li>
-    <li>Disabled LUKS for new Raspberry Pi setups</li>
-    <li>Fixed miscellaneous bugs</li>
+    <li>Https over Tor for faster UI loading times</li>
+    <li>Change password through UI</li>
+    <li>Use IP address for Network Folder backups</li>
+    <li>
+      Multiple bug fixes, performance enhancements, and other small features
+    </li>
```
elvece, line 34: grammar + tense should be the same throughout

```suggestion
    <li>Enable https over Tor for faster UI loading times</li>
    <li>Change password through UI</li>
    <li>Use IP address for Network Folder backups</li>
    <li>
     Fix multiple bugs, enhance performance, and add other small features
    </li>
```

frontend/projects/ui/src/app/pages/apps-routes/app-show/app-show.page.html
```diff
@@ -52,21 +52,17 @@
...
+                <ion-text color="warning">Http detected</ion-text>
               </h2>
               <p class="ion-padding-bottom">
-                Click the button below to switch to https. Your browser may warn
-                you that the page is insecure. You can safely bypass this
-                warning. It will go away after you
+                Your connection is insecure.
                 <a
-                  [routerLink]="['/system', 'lan']"
+                  href="https://docs.start9.com/latest/user-manual/connecting/connecting-lan/"
```
elvece, line 60: We ideally should update these links with the proper page in the new docs refactor before release, or at least need to make sure that page will no longer exist so a redirect can be implemented.

frontend/projects/ui/src/app/pages/apps-routes/app-show/app-show.page.html
```diff
@@ -52,21 +52,17 @@
           <ion-row class="ion-align-items-center">
             <ion-col class="ion-text-center">
               <h2>
-                <ion-text color="warning">
-                  You are using an unencrypted http connection
-                </ion-text>
+                <ion-text color="warning">Http detected</ion-text>
```
elvece, line 55: Haha I think this language is better....but 😄 
```suggestion
                <ion-text color="warning">Unencrypted connection</ion-text>
```

frontend/projects/ui/src/app/pages/apps-routes/app-show/app-show.page.html
elvece: also we dont additionally want to link to where to download the cert here?

frontend/projects/ui/src/app/pages/server-routes/server-show/server-show.page.html
```diff
@@ -15,15 +15,20 @@
 
   <!-- loaded -->
   <ion-item-group *ngIf="server$ | async as server; else loading">
-    <ion-item *ngIf="!secure" color="warning">
+    <ion-item *ngIf="!secure || isTorHttp" color="warning">
       <ion-icon slot="start" name="warning-outline"></ion-icon>
       <ion-label>
-        <h2 style="font-weight: bold">You are using unencrypted http</h2>
+        <h2 style="font-weight: bold">Http detected</h2>
```
elvece, line 21: same comments as above for this section - language & links

frontend/projects/ui/src/app/app.component.ts
MattDHill: Abandoning. It's unnecessary and the code was problematic

frontend/projects/ui/src/app/pages/apps-routes/app-show/app-show.page.html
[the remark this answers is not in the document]
dr-bonez: http over tor is still E2EE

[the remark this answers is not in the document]
elvece: ah ok

frontend/projects/ui/src/app/app.component.ts
elvece: ok, could you relay why? just curious what changed @dr-bonez? for my own understanding

frontend/projects/ui/src/app/pages/apps-routes/app-show/app-show.page.html
MattDHill: I believe the docs are changing quite a bit, so redirects will be necessary everywhere. If you know the new link and are confident it will be done in time for this release, you can change it here.

I'll take another look, but givent he choice I would rather link to the instructions than to the cert page. But that reminds me, we should rename the LAN page

frontend/projects/ui/src/app/app.component.ts
MattDHill: 1. It just didn't work in my testing
2. If it did work, the UX would be that after a period of time and with no indication why, the page would suddenly redirect/refresh
3. There was no catch block, so if the fetch failed, it would be an unhandled promise rejection.

We don't need to *force* people away from http. We just need to tell them why they should be using https

frontend/projects/ui/src/app/pages/apps-routes/app-show/app-show.page.html
MattDHill: Reverted to local links

frontend/projects/ui/src/app/app.component.ts
elvece: Cool thanks for explaining. Those concerns could for sure be addressed in code, I definitely made a mistake not adding the catch block. In my testing, it only redirected once on initial load and worked great. Perhaps the lag over Tor would be abrupt. I think we had wanted to do this for a few reasons, rather than making the user manually change, but understood its not ideal to force the behavior.

elvece (approved):

frontend/projects/ui/src/app/pages/server-routes/server-show/server-show.page.html
```diff
@@ -15,15 +15,20 @@
...
         <p style="font-weight: 600">
-          Click the button on the right to switch to https. Your browser may
-          warn you that the page is insecure. You can safely bypass this
-          warning. It will go away after you download and trust your server's
-          certificate
+          {{ isTorHttp ? 'Tor is faster over https.' : 'Your connection is
```
elsirion, line 23: What's the reasoning behind the claim that "Tor is faster over HTTPs"? All it does is adding another round-trip to establish a secure connection over an already secure connection. 

I'd get it if it was about accessing the clearnet version via HTTPs over Tor, that might reduce the latency (no Tor between rendezvous point and hidden service), but as it is it's just accessing the hidden service with an extra layer of TLS.

dr-bonez: It's faster because browsers only support http2 over https. http2 allows better batching and tcp connection reuse than 1.1 which results in faster page load times

elsirion: TIL, thx!

PR #2372, opened 2023-07-21
add cli & rpc to edit db with jq syntax

dr-bonez:


Blu-J (approved):
Just waiting for actual test, but if that works gtg

PR #2379, opened 2023-07-31
add qr code to insights->about->tor

Reckless-Satoshi:
This PR adds a QR scan button to Insights -> About -> Tor onion property.

Scanning the QR codes is very handy for the apps. However, I did not find where to scan the main startos entrypoint QR.

This PR is **not tested**, I have no experience with Angular/Ionic and just edited code without setting up a dev environment . Hopefully it is not far from working.

### DIFF
--- a/frontend/projects/ui/src/app/pages/server-routes/server-specs/server-specs.page.ts
+++ b/frontend/projects/ui/src/app/pages/server-routes/server-specs/server-specs.page.ts
@@ -2,6 +2,7 @@ import { ChangeDetectionStrategy, Component } from '@angular/core'
 import { ToastController } from '@ionic/angular'
 import { PatchDB } from 'patch-db-client'
 import { ConfigService } from 'src/app/services/config.service'
+import { QRComponent } from 'src/app/components/qr/qr.component'
 import { copyToClipboard } from '@start9labs/shared'
 import { DataModel } from 'src/app/services/patch-db/data-model'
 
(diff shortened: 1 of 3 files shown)


frontend/projects/ui/src/app/pages/server-routes/server-specs/server-specs.page.ts
```diff
@@ -2,6 +2,7 @@ import { ChangeDetectionStrategy, Component } from '@angular/core'
 import { ToastController } from '@ionic/angular'
```
elvece, line 2:
```suggestion
import { ModalController, ToastController } from '@ionic/angular'
```

elvece: I cannot comment on the direct line, but you also need to add the following to the class constructor:
```
private readonly modalCtrl: ModalController,
```

so its finalized form should look like:
```
  constructor(
    private readonly toastCtrl: ToastController,
    private readonly modalCtrl: ModalController,
    private readonly patch: PatchDB<DataModel>,
    private readonly config: ConfigService,
  ) {}
```

elvece (changes requested):
Need a few changes to get it working but works great once fixed!

frontend/projects/ui/src/app/pages/server-routes/server-specs/server-specs.page.ts
Reckless-Satoshi: Hey @elvece thank you for testing it so quickly! :rocket: Committed the fix.

elvece (approved):

PR #2401, opened 2023-08-18
Refactor/service manager

Blu-J:
### About

The idea about this is that the service manager should be a state machine, so we shall make the code actually a state machine. The code ends up being simpler, the thing then acts like a actor, and the state can be observed.

### Proof

Here we show that we can stop and start services.
image

image

And the restart works

and backup/ restore
image

Then deleted then restored 2/3 
image

Proof that the config is working

https://github.com/Start9Labs/start-os/assets/2364004/f94609c9-d271-4e1a-81f1-0c6ffd29d430


backend/src/manager/mod.rs
```diff
@@ -1,143 +1,682 @@
...
+    fn _transition_backup(
+        &self,
+        backup_guard: BackupGuard,
+    ) -> (TransitionState, BoxFuture<BackupReturn>) {
+        let (send, done) = oneshot::channel();
+        (
+            TransitionState::BackingUp(
+                tokio::spawn(
+                    self.perform_backup(backup_guard)
+                        .then(finnish_up_backup_task(self.transition.clone(), send)),
```
dr-bonez, line 314: 🇫🇮

Blu-J: Fixed mispelling

dr-bonez (approved):

PR #2412, opened 2023-09-07
Add tor-check to CLI

k0gen:
Screenshot 2023-09-07 at 14 45 57

### DIFF
--- a/backend/src/bins/start_init.rs
+++ b/backend/src/bins/start_init.rs
@@ -38,6 +38,12 @@ async fn setup_or_init(cfg_path: Option<PathBuf>) -> Result<(), Error> {
         .arg("/usr/local/bin/aptitude")
         .invoke(crate::ErrorKind::OpenSsh)
         .await?;
+    Command::new("ln")
+        .arg("-sf")
+        .arg("/usr/lib/embassy/scripts/tor-check.sh")
+        .arg("/usr/local/bin/tor-check")
+        .invoke(crate::ErrorKind::OpenSsh)
+        .await?;
 
     Command::new("make-ssl-cert")
         .arg("generate-default-snakeoil")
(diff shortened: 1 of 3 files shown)


backend/src/bins/start_init.rs
```diff
@@ -38,6 +38,12 @@ async fn setup_or_init(cfg_path: Option<PathBuf>) -> Result<(), Error> {
         .arg("/usr/local/bin/aptitude")
         .invoke(crate::ErrorKind::OpenSsh)
         .await?;
+    Command::new("ln")
```
dr-bonez, line 41: Put this in postinst instead. With `/usr/bin` instead of `/usr/local/bin`

build/lib/scripts/tor-check.sh
```diff
@@ -0,0 +1,40 @@
...
+echo "Testing connection to Onion Pages ..."
+
+for data in "${onion_list[@]}"; do
+    name="${data%%|*}"
+    url="${data#*|}"
+
+    curl --socks5-hostname localhost:9050 "$url" > /dev/null 2>&1
+    HC=$?
+
+    if [ $HC -ne 0 ]; then
```
dr-bonez, line 32:
```suggestion
    if curl --socks5-hostname localhost:9050 "$url" > /dev/null 2>&1; then
```

dr-bonez (changes requested):

build/lib/scripts/tor-check.sh
k0gen: Done

backend/src/bins/start_init.rs
k0gen: done

dr-bonez (approved):

backend/src/bins/start_init.rs
k0gen: done

Blu-J (approved):

PR #2424, opened 2023-09-26
hack it in

MattDHill:


frontend/projects/ui/src/app/services/dep-error.service.ts
```diff
@@ -0,0 +1,210 @@
...
+    }, {} as DependencyErrors)
+  }
+
+  private getDepError(
+    pkgs: DataModel['package-data'],
+    pkgInstalled: InstalledPackageDataEntry,
+    depId: string,
+    outerErrors: PackageDependencyErrors,
+  ): DependencyError | null {
+    const depInstalled = pkgs[depId]?.installed
```
Blu-J, line 67: 🤔 We could have broken this up into 4 fns, is it better?
I thought of it because you where commenting sections of code, and the built in way of naming sections of code is fns

```ts
return this.getDepErrorInstalled(pkgs, pkgInstalled, depId, outerErrors) |
 this.getDepErrorIncorrectVersion(pkgs, pkgInstalled, depId, outerErrors) |
  this.getDepErrorInvalidConfig(pkgs, pkgInstalled, depId, outerErrors) |
    this.getDepErrorNotRunning(pkgs, pkgInstalled, depId, outerErrors) |
    null
}
private getDepErrorInstalled(
    pkgs: DataModel['package-data'],
    pkgInstalled: InstalledPackageDataEntry,
    depId: string,
    outerErrors: PackageDependencyErrors,
  ): DependencyError | null {
      const depInstalled = pkgs[depId]?.installed
     if (depInstalled) return null;
     return {
        type: DependencyErrorType.NotInstalled,
      }
  }
  private getDepErrorIncorrectVersion(
    pkgs: DataModel['package-data'],
    pkgInstalled: InstalledPackageDataEntry,
    depId: string,
    outerErrors: PackageDependencyErrors,
  ): DependencyError | null {
      const depInstalled = pkgs[depId]?.installed
     if (!depInstalled) return null;
     
    const pkgManifest = pkgInstalled.manifest
    const depManifest = depInstalled.manifest
    if (
      this.emver.satisfies(
        depManifest.version,
        pkgManifest.dependencies[depId].version,
      )
    )  return null
    
      return {
        type: DependencyErrorType.IncorrectVersion,
        expected: pkgManifest.dependencies[depId].version,
        received: depManifest.version,
      }
    
  }

File: frontend/projects/ui/src/app/services/dep-error.service.ts [hunk 2]
```diff
@@ -0,0 +1,210 @@
...
+  PackageMainStatus,
+} from './patch-db/data-model'
+
+export type PackageDependencyErrors = Record<string, DependencyErrors>
+export type DependencyErrors = Record<string, DependencyError | null>
+
+@Injectable({
+  providedIn: 'root',
+})
+export class DepErrorService {
```
Blu-J, line 19: 🤔 So, other possible solutions that I can think of. 
For the finding the depErrors, you could have done the truly recursive, and that is repulsive. Sounds like you already explored it. 
Your [dynamic solution ](https://www.geeksforgeeks.org/dynamic-programming/) seems to work.

The downside is that we are doing a recompute every change. We could do a memoization on if the pkgs are the same object before, but that requires that we have a quick equal. Or the other is that the pointer equal would work, but then we need a guarentee that the pkgs are updated via a immutable style. I wouldn't trust this beyond using something like immutablejs 
Ok, the other is do the memoization deeper into the values that are easier run the equality checks. So we would could do a way of doing per pkg id, a combined observable based on each variable that we could do... Sounds like the the rxjs would make this a pain.
There are observable that might work, like [mobx](https://mobx.js.org/observable-state.html). It solves the diamond problem better, but the library would be new and unlike the rest of the thing. But it would do the leg work of the nested composted observable into something that feels natural. 

In summary, solution seems good. But the only downside is that we are not efficient at all (at least it looks like it)

frontend/projects/ui/src/app/services/dep-error.service.ts
```diff
@@ -0,0 +1,210 @@
...
+  private getDepErrors(
+    pkgs: DataModel['package-data'],
+    pkgId: string,
+    outerErrors: PackageDependencyErrors,
+  ): DependencyErrors {
+    const pkgInstalled = pkgs[pkgId].installed
+
+    if (!pkgInstalled) return {}
+
+    return currentDeps(pkgs, pkgId).reduce((innerErrors, depId) => {
```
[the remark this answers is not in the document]
MattDHill: Needed for type safety. TS seems unable to provide safety in your example. Try setting the value of [depId] to an arbitrary string. TS is fine with it, but it should be prohibited.

Blu-J (approved):

Blu-J (approved):

PR #2425, opened 2023-09-27
do not require auth for cert

dr-bonez:


backend/src/net/static_server.rs
```diff
@@ -274,69 +290,56 @@ async fn main_embassy_ui(req: Request<Body>, ctx: RpcContext) -> Result<Response
...
+                    .headers(
+                        request_parts
+                            .headers
+                            .iter()
+                            .filter(|(h, _)| {
+                                !PROXY_STRIP_HEADERS
+                                    .iter()
+                                    .any(|bad| h.as_str().eq_ignore_ascii_case(bad))
+                            })
+                            .map(|(h, v)| (h.clone(), v.clone()))
```
Blu-J, line 324: ❓  Isn't there an operator of cloned()
Might just be an Iterators crate?

dr-bonez: cloned is `impl Iterator<Item = &T> -> impl Iterator<Item = T>` not `impl Iterator<Item = (&T, &T)> -> impl Iterator<Item = (T, T)>`

Blu-J (approved):

PR #2427, opened 2023-09-28
feat: Change all the dependency errors at once

Blu-J:


backend/src/context/rpc.rs
```diff
@@ -279,8 +278,25 @@ impl RpcContext {
             })
             .await?;
         let peek = self.db.peek().await?;
+        let mut all_dependency_config_errs = BTreeMap::new();
```
Blu-J, line 281: So, the idea is that I want to accumulate all the package dependency errors. For the mutate below, a kind of global reset.

dr-bonez (dismissed):

Blu-J:
# Proof of working

So I start with my VM that is in a 3.4.4.4 state, and install lnd, configure, then install bitcoin. 
Then I pull image from build server, and updated to that. 
Now we should be on our latest cod, cb723a8 (the build is 1008b2, but that is a merge + master)  
image

Then in the LND we see the message that we are expecting, 
image

dr-bonez (approved):

PR #2438, opened 2023-10-04
set governor to "performance" if available

dr-bonez:


backend/src/init.rs
```diff
@@ -341,6 +344,23 @@ pub async fn init(cfg: &RpcContextConfig) -> Result<InitResult, Error> {
         .await?;
     tracing::info!("Enabled Docker QEMU Emulation");
 
+    if current_governor()
+        .await?
```
Blu-J, line 348: Wouldn't a is_performance method be better for what is returned on current_govenrnor()

Blu-J (approved):

Blu-J (approved):

PR #2443, opened 2023-10-05
enable trimming in luks

dr-bonez:


dr-bonez:
I have tested that this at least doesn't break anything. As for whether there is any measurable performance improvement? time will tell

Blu-J (approved):

PR #2455, opened 2023-10-12
add firmware updater

dr-bonez:


Blu-J (approved):
So, works on my pro with an update over usb

PR #2458, opened 2023-10-13
fix some causes of start wonkiness on update

dr-bonez:


Blu-J (approved):
Seen that the feature is fixed and several other testing with start/ stop;

PR #2461, opened 2023-10-16
add tokio-console if `unstable`

dr-bonez:


Blu-J (approved):
ran and used on my prod box

PR #2471, opened 2023-10-24
Bugfix/output timeout

dr-bonez:


Blu-J:
### Proof of fix for the buf reader to lines
image

Blu-J (approved):
Only one of the bug fixes doesn't have a proof of working.
But @dr-bonez Ran it on his machine and it should work

PR #2482, opened 2023-10-27
use old secret key derivation function

dr-bonez:


dr-bonez:
Proof of fix:
image
where `gitea` has no key in the `tor` table

Blu-J (approved):

PR #2489, opened 2023-11-01
use existing dependency icon if available

dr-bonez:


Blu-J (approved):

Blu-J:
## Proof that it works
image

PR #2504, opened 2023-11-09
feature: s9pk v2

dr-bonez:


backend/src/s9pk/merkle_archive/test.rs
```diff
@@ -0,0 +1,126 @@
...
+            let s2: Arc<[u8]> = s2.into();
+            ensure_code!(s1 == s2, ErrorKind::Pack, "s1 does not match s2");
+
+            Ok(())
+        })
+}
+
+proptest::proptest! {
+    #[test]
+    fn property_test(files: Vec<(PathBuf, String)>) {
```
ProofOfKeags, line 90: rename this function to describe what the property is actually testing. I think it's saying that you're asserting no error is ever thrown from this function? What's curious about that is that you should be able to assert that by dropping the Error from the type signature, no?

PR #2516, opened 2023-11-15
feat(portal): add notifications sidebar

waterplea:


web/projects/ui/src/app/apps/portal/components/header/header.component.ts
```diff
@@ -9,22 +20,93 @@ import {
...
   ],
 })
-export class HeaderComponent {}
+export class HeaderComponent {
+  @ViewChild(HeaderNotificationsComponent, { read: ElementRef })
+  private readonly panel?: ElementRef<HTMLElement>
+
+  private readonly navigation = inject(Router).events.subscribe(() => {
+    this.open$.next(false)
+  })
```
elvece, line 98: What is this needed for?

web/projects/ui/src/app/app/sidebar-host.component.ts
```diff
@@ -0,0 +1,35 @@
...
+  selector: 'sidebar-host',
+  template: '<ng-container #viewContainer></ng-container>',
+  styles: [':host { position: fixed; top: 0; }'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  standalone: true,
+  providers: [
+    { provide: AbstractTuiPortalService, useExisting: SidebarService },
+  ],
+})
+export class SidebarHostComponent extends AbstractTuiPortalHostComponent {}
```
elvece, line 35: Can you explain why this approach instead of just using the [tuiSidebar](https://taiga-ui.dev/directives/sidebar) as is?

waterplea: This way it appears under the dialogs and not over them, as dropdowns/portals do by default. Since we want to open dialogs from this sidebar and have it remain open — I've made a special container for it above all content and below dialogs.

web/projects/ui/src/app/apps/portal/components/header/header.component.ts
waterplea: So that if we navigate to service page or use native browser back button — the sidebar would close.

PR #2524, opened 2023-11-16
improve robustness of install

dr-bonez:
* perform old package cleanup after commit to patchdb
* unpack full s9pk on rebuild


Blu-J (approved):
Correct pattern, delete after things are done.

Blu-J (approved):
## Proof

Installed/ Removed/ Downgraded/ Upgraded package/ System Rebuilt done

PR #2528, opened 2023-11-20
fix ntp sync status check daemon

dr-bonez:
the previous implementation launched the ntp status check daemon during `start-init`. since `start-init` uses a different tokio runtime than `startd`, this daemon is shutdown after initialization completes. I have moved the ntp status check daemon to `RpcContext::init`.


MattDHill (approved):

Blu-J (commented):
👍

PR #2540, opened 2023-12-18
fix: Docker stopping will include a timeout

Blu-J:
So the timeout that was included in the original is not working therefore we move to a doubling with a timeout

image

### DIFF
--- a/core/startos/src/manager/manager_seed.rs
+++ b/core/startos/src/manager/manager_seed.rs
@@ -29,8 +29,15 @@ impl ManagerSeed {
         )
         .await
         {
-            Err(e) if e.kind == ErrorKind::NotFound => (), // Already stopped
-            Err(e) if e.kind == ErrorKind::Timeout => (),  // Already stopped In theory
+            Err(e) if e.kind == ErrorKind::NotFound => {
+                tracing::info!(
+                    "Command for package {command_id} should already be stopped",
+                    command_id = &self.manifest.id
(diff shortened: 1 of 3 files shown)


core/startos/src/util/docker.rs
```diff
@@ -115,7 +115,9 @@ pub async fn stop_container(
     let mut cmd = Command::new(CONTAINER_TOOL);
     cmd.arg("stop");
     if let Some(dur) = timeout {
-        cmd.arg("-t").arg(dur.as_secs().to_string());
+        cmd.arg("-t")
+            .arg(dur.as_secs().to_string())
+            .timeout(Some(dur));
```
dr-bonez, line 120: This won't work, timeout lifts it into a wrapper type. You'll need to reassign cmd

core/startos/src/manager/manager_seed.rs
```diff
@@ -30,6 +30,7 @@ impl ManagerSeed {
         .await
         {
             Err(e) if e.kind == ErrorKind::NotFound => (), // Already stopped
+            Err(e) if e.kind == ErrorKind::Timeout => (),  // Already stopped In theory
```
dr-bonez, line 33: I think we would actually like to log an error in the timeout case

dr-bonez (changes requested):

dr-bonez (approved):

PR #2554, opened 2024-01-24
Frontend changes for 036

MattDHill:
- Remove warning related to using Tor over http
- Accommodate new interfaces paradigm for LAN port forwarding upgrade


web/projects/ui/src/app/pages/apps-routes/app-interfaces/app-interfaces.page.ts
```diff
@@ -126,3 +100,65 @@ export class AppInterfacesItemComponent {
...
   }
 }
+
+function getAddresses(
+  serviceInterface: ServiceInterfaceWithHostInfo,
+): MappedAddress[] {
+  const host = serviceInterface.hostInfo
+  const addressInfo = serviceInterface.addressInfo
+  const username = addressInfo.username ? addressInfo.username + '@' : ''
+  const suffix = addressInfo.suffix ? '/' + addressInfo.suffix : ''
```
dr-bonez, line 110: I think the suffix might already include a /

web/projects/ui/src/app/pages/apps-routes/app-interfaces/app-interfaces.page.ts
```diff
@@ -126,3 +100,65 @@ export class AppInterfacesItemComponent {
...
+      } else {
+        name = h.hostname.kind
+        hostname =
+          h.hostname.kind === 'domain'
+            ? `${h.hostname.subdomain}.${h.hostname.domain}`
+            : h.hostname.value
+      }
+
+      if (h.hostname.sslPort) {
+        const port = h.hostname.sslPort === 443 ? '' : `:${h.hostname.sslPort}`
```
dr-bonez, line 138: Don't hardcode 443. Only erase the port if it matches the defaultPort for a KnownProtocol. So look up the scheme in knownProtocols, and check the port against KnownProtocol.defaultPort

web/projects/ui/src/app/services/config.service.ts
```diff
@@ -57,24 +56,60 @@ export class ConfigService {
...
+
+    if (host.kind === 'multi') {
+      const onionHostname = host.hostnames.find(
+        h => h.kind === 'onion',
+      ) as HostnameInfoOnion
+
+      if (this.isTor() && onionHostname) {
+        url.hostname = onionHostname.hostname.value
+      } else {
+        const ipHostname = host.hostnames.find(
```
dr-bonez, line 88: Just taking the first ip hostname you find? Not trying to match it to the address they're currently on?

dr-bonez (changes requested):

web/projects/ui/src/app/pages/apps-routes/app-interfaces/app-interfaces.page.ts
MattDHill: These are always UIs

web/projects/ui/src/app/pages/apps-routes/app-interfaces/app-interfaces.page.ts
MattDHill: Why is that. Are you expecting package devs to write it, or is the BE adding it universally to the string?

web/projects/ui/src/app/services/config.service.ts
MattDHill: Look at line 94. Just trying to figure out if they have an IP hostname. This won't work once we have domains.

web/projects/ui/src/app/pages/apps-routes/app-interfaces/app-interfaces.page.ts
dr-bonez: It gets built by the sdk from other components, and that function adds a leading / if appropriate

web/projects/ui/src/app/services/config.service.ts
dr-bonez: Hmm, ok, I guess that works

web/projects/ui/src/app/pages/apps-routes/app-interfaces/app-interfaces.page.ts
dr-bonez: Then I suppose it technically would work, though I'd document the reliance on all uis being http. Personally I'd abstract url construction to a different function and handle it there though cause you'll need it for displaying interface urls too

dr-bonez (approved):

PR #2587, opened 2024-04-01
Feat/implement rest of poly effects

Blu-J:


Blu-J:
image
image


image
image

container-runtime/src/Adapters/Systems/SystemForEmbassy/MainLoop.ts
```diff
@@ -74,6 +63,7 @@ export class MainLoop {
       daemon,
       wait: daemon.wait().finally(() => {
         this.clean()
+        console.error("BLUJ It ways it is done?")
```
dr-bonez, line 66: Probably should remove this

core/startos/src/service/config.rs
```diff
@@ -10,10 +10,27 @@ impl Service {
         ConfigureContext { timeout, config }: ConfigureContext,
     ) -> Result<(), Error> {
         let container = &self.seed.persistent_container;
+        let package_id = &self.seed.id;
+
         container
-            .execute::<Value>(ProcedureName::SetConfig, to_value(&config)?, timeout)
+            .execute::<SetResult>(ProcedureName::SetConfig, to_value(&config)?, timeout)
```
dr-bonez, line 16: Is this type still correct for the new api?

dr-bonez (changes requested):

core/startos/src/service/config.rs
Blu-J: Not sure, but it is at the moment

dr-bonez (approved):

core/startos/src/service/config.rs
Blu-J: So, going back to value to say that I don't care

PR #2602, opened 2024-04-15
Feat/logging

Blu-J:
So, we have taken the system that is in a state where the logging that is in the services doesn't work because of the lxc. After the changes to the debian, we got the systemd, so we just have to make sure that we mount the journal to be stored between sessions and to be able to be read during running via lxc-attach.

Proof that it works
image


core/startos/src/lxc/mod.rs
```diff
@@ -145,6 +198,18 @@ impl LxcContainer {
...
         .await?;
+        let log_mount_point = container_dir.join("var/log/journal");
+        let log_mount = MountGuard::mount(
+            &Bind::new(data_dir.join("package-data").join("logs").join(package_id)),
+            &log_mount_point,
+            MountType::ReadWrite,
+        )
+        .await?;
+        Command::new("chown")
+            .arg("100000:100999")
```
Blu-J, line 209:
```suggestion
            // This was needed as 100999 because the group id of journald
            .arg("100000:100999")
```

dr-bonez (approved):

PR #2604, opened 2024-04-18
feat: implement metrics on the dashboard

waterplea:


MattDHill (changes requested):
1. Currently, if I navigate to another component from the dashboard, then navigate back, the metrics component reloads. It would be preferable to maintain metrics and a global subscription and not re-initialize on dashboard view.
2. On mobile, the "Metrics" tab will be the detailed view, not the dashboard widget. Howver, the detailed component is not yet designed, so I think it's fine as-is for now.
3. As mentioned privately, the uptime metrics should be calculated using the server.time RPC endpoint.

MattDHill (approved):

PR #2609, opened 2024-04-22
Feature/sdk040dependencies

Blu-J:
Adding in the new sdk for the 040 and checking the dependencies that are needed hopefully.


core/startos/src/service/service_effect_handler.rs
```diff
@@ -1217,3 +1231,119 @@ async fn set_dependencies(
...
+        })
+        .try_collect()
+}
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize, Parser, TS)]
+#[serde(rename_all = "camelCase")]
+#[command(rename_all = "camelCase")]
+#[ts(export)]
+struct CheckDependenciesParam {
+    #[ts(type = "string[]")]
```
dr-bonez, line 1280: This shouldn't be necessary

core/startos/src/service/service_effect_handler.rs
```diff
@@ -1217,3 +1231,119 @@ async fn set_dependencies(
...
+    package_ids: Vec<PackageId>,
+}
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize, Parser, TS)]
+#[serde(rename_all = "camelCase")]
+#[command(rename_all = "camelCase")]
+#[ts(export)]
+struct CheckDependenciesResult {
+    #[ts(type = "string")]
```
dr-bonez, line 1288: Same here

core/startos/src/service/service_effect_handler.rs
```diff
@@ -1217,3 +1231,119 @@ async fn set_dependencies(
...
+    package_id: PackageId,
+    is_installed: bool,
+    is_running: bool,
+    health_checks: Vec<HealthCheckResult>,
+}
+impl ValueParserFactory for CheckDependenciesResult {
```
dr-bonez, line 1294: Why is this getting parsed as a command at all?

core/startos/src/service/service_effect_handler.rs
```diff
@@ -1217,3 +1231,119 @@ async fn set_dependencies(
...
+        let Some(package) = db.as_public().as_package_data().as_idx(&package_id) else {
+            results.push(CheckDependenciesResult {
+                package_id,
+                is_installed: false,
+                is_running: false,
+                health_checks: vec![],
+            });
+            continue;
+        };
+        if package.as_state_info().as_installing_info().is_none() {
```
dr-bonez, line 1318: This is only if the service is installing or updating. This stores install progress. Instead, grab the current manifest and check that the version matches. If the service is not installed at that version, the manifest won't exist

sdk/lib/dependencies/dependencies.ts
```diff
@@ -0,0 +1,115 @@
...
+  SetHealth,
+  CheckDependencyResult,
+} from "../types"
+
+export type CheckAllDependencies = {
+  notRunning: () => Promise<CheckDependencyResult[]>
+
+  notInstalled: () => Promise<CheckDependencyResult[]>
+
+  errorMessages: () => Promise<{ [id: string]: SetHealth[] }>
```
dr-bonez, line 14: Why SetHealth?

dr-bonez (changes requested):

sdk/lib/dependencies/dependencies.ts
Blu-J: Because that is the return type of the result, I'm just passing through the types
image
https://github.com/Start9Labs/start-os/pull/2609/files/9b94b8035039c436ebee4db1a0fdb71e552564ef#diff-448f8e9b112af283ba61fd29e758ef1a0c6f33206f3acd03fff0b9e7fb9f7a80R603

dr-bonez: It seems the field is improperly named then, if it is specific to health checks

PR #2610, opened 2024-04-23
More svc effect handlers

Dominion5254:


core/startos/src/net/host/address.rs
```diff
@@ -2,7 +2,7 @@ use serde::{Deserialize, Serialize};
 use torut::onion::OnionAddressV3;
 use ts_rs::TS;
 
-#[derive(Debug, Deserialize, Serialize, PartialEq, Eq, PartialOrd, Ord, TS)]
+#[derive(Clone, Copy, Debug, Deserialize, Serialize, PartialEq, Eq, PartialOrd, Ord, TS)]
```
dr-bonez, line 5:
```suggestion
#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq, PartialOrd, Ord, TS)]
```
This will likely not be able to be copy once we add clearnet hosts

core/startos/src/net/host/mod.rs
```diff
@@ -86,3 +86,17 @@ impl Model<HostInfo> {
...
+    pub fn get_host_primary(&self, host_id: &HostId) -> Option<HostAddress> {
+        match self.0.get(&host_id) {
+            Some(h) => {
+                match h.primary {
+                    Some(ha) => Some(ha),
+                    None => None,
+                }
+            }
+            None => None,
+        }
```
dr-bonez, line 100:
```suggestion
        self.0.get(&host_id).and_then(|h| h.primary.clone())
```

core/startos/src/service/service_effect_handler.rs
```diff
@@ -148,6 +151,7 @@ pub fn service_effect_handler() -> ParentHandler {
                 .no_display()
                 .with_remote_cli::<ContainerCliContext>(),
         )
+        .subcommand("setSystemSmtp", from_fn_async(set_system_smtp).no_cli())
```
dr-bonez, line 154: this should be an os rpc, not a service rpc

core/startos/src/service/service_effect_handler.rs
```diff
@@ -343,17 +403,52 @@ async fn export_service_interface(
 async fn get_primary_url(
     context: EffectContext,
     data: GetPrimaryUrlParams,
-) -> Result<Value, Error> {
-    todo!()
+) -> Result<HostAddress, Error> {
```
dr-bonez, line 406: I feel like this should return `Option<HostAddress>`. Does that not match the sdk?

core/startos/src/service/service_effect_handler.rs
```diff
@@ -310,11 +321,44 @@ struct MountParams {
...
+            model.ser(&mut Some(data.smtp))
+        })
+        .await
+}
 async fn get_system_smtp(
     context: EffectContext,
     data: GetSystemSmtpParams,
-) -> Result<Value, Error> {
-    todo!()
+) -> Result<String, Error> {
```
dr-bonez, line 343: is this really just a string?

dr-bonez (changes requested):

dr-bonez:
I'd also like you to include cli bindings for your rpcs. just remove `.no_cli()` and derive `Parser` for your params. you may also need to consider how to display the result. `.with_display_serializable()` is often the easiest for complex types

core/startos/src/service/service_effect_handler.rs
Dominion5254: That makes sense. This os rpc call should be in `core/startos/src/system.rs`, correct? 

Similarly, should `getSystemSmtp` also be an os rpc then?

core/startos/src/service/service_effect_handler.rs
Dominion5254: The base branch (next/minor) has [getPrimaryUrl](https://github.com/Start9Labs/start-os/blob/next/minor/sdk/lib/types.ts#L422-L426
) returning a `Promise<UrlString | null>`, where `UrlString` is an alias for `string` - This is consistent with the `address` field of [HostAddress](https://github.com/Start9Labs/start-os/blob/next/minor/core/startos/src/net/host/address.rs#L5-L14).

So I believe the return type should actually be updated to `<Option<String>`, correct? Or would this need to be wrapped in a `Result` to handle errors?

core/startos/src/service/service_effect_handler.rs
Dominion5254: Good catch. This should follow the TS type defined [here](https://github.com/Start9Labs/start-os/blob/next/minor/sdk/lib/types.ts#L153-L159).

I will update pending clarification whether this should be an os rpc call similar to `setSystemRpc`

core/startos/src/service/service_effect_handler.rs
dr-bonez: getSystemSmtp is a service rpc, so it is in the right place

dr-bonez:
let's merge this for now and wrap up those other issues in a separate PR

Blu-J (approved):

PR #2622, opened 2024-05-16
Feature/network

Blu-J:


Blu-J:
Show that we have the config set correctly
image

And that the ui can now run.
image

container-runtime/src/Adapters/Systems/SystemForEmbassy/index.ts
```diff
@@ -335,6 +335,113 @@ export class SystemForEmbassy implements System {
...
+              networkInterfaceId: `${id}-lan-${key}`,
+              public: true,
+              hostname: {
+                kind: "local",
+                value: `${id}-lan-${key}-value`,
+                port: value.internal,
+                sslPort: null,
+              },
+            }),
+          ),
```
dr-bonez, line 429: this shouldn't be necessary - the os can get from the hostId on addressInfo

container-runtime/src/Adapters/Systems/SystemForEmbassy/index.ts
```diff
@@ -335,6 +335,113 @@ export class SystemForEmbassy implements System {
       await this.migration(effects, previousVersion, timeoutMs)
     await effects.setMainStatus({ status: "stopped" })
     await this.exportActions(effects)
+    await this.exportNetwork(effects)
```
dr-bonez, line 338: should this not be part of setupInterfaces in MainLoop? or is that in the wrong place?

core/startos/src/net/net_controller.rs
```diff
@@ -209,6 +209,21 @@ impl NetService {
...
+        let package_id = &self.id;
+        self.net_controller()?
+            .db
+            .mutate(|db| {
+                db.as_public_mut()
+                    .as_package_data_mut()
+                    .as_idx_mut(package_id)
+                    .or_not_found(package_id)?
+                    .as_hosts_mut()
+                    .ser(&Default::default())
```
dr-bonez, line 222: this is probably too aggressive, we probably want to keep addresses. just wipe out the binds on each host

dr-bonez: also, `update` should be called after to do the actual network changes

core/startos/src/service/service_effect_handler.rs
```diff
@@ -639,39 +640,32 @@ async fn bind(
...
+    let ctx = ctx.deref()?;
+    let package_id = package_id.unwrap_or_else(|| ctx.id.clone());
+    let db = ctx.ctx.db.peek().await;
+
+    let interface = db
+        .as_public()
+        .as_package_data()
+        .as_idx(&package_id)
+        .or_not_found(&package_id)?
+        .as_service_interfaces()
```
dr-bonez, line 664: is this really `ServiceInterfaceWithHostInfo`? seems redundant to have HostInfo duplicated. I'm worried they will get out of sync

dr-bonez: the db should probably just have the ServiceInterface, and the host info should be queried and added here

sdk/lib/interfaces/Origin.ts
```diff
@@ -80,6 +81,8 @@ export class Origin<T extends Host> {
         addressInfo,
         type,
         masked,
+        hostKind: "multi",
+        hostnames: [],
```
dr-bonez, line 85: shouldn't be necessary - can be looked up from addressInfo.hostId

sdk/lib/util/getServiceInterface.ts
```diff
@@ -101,57 +94,58 @@ const addressHostToUrl = (
...
+      return addresses
         .filter((h) => h.kind === "onion")
-        .map((h) => stringifyHostname(h))
+        .flatMap(stringifyHostname)
     },
     get localHostnames() {
-      return hostnameInfo
-        .filter((h) => h.kind === "ip" && h.hostname.kind === "local")
-        .map((h) => stringifyHostname(h))
+      return []
```
dr-bonez, line 112: maybe drop a todo on these so we don't forget. what was wrong with them?

container-runtime/src/Adapters/Systems/SystemForEmbassy/index.ts
Blu-J: Needs to be ran before... config. So therefore can't be main. Good idea, but then the package known as Vaultwarden shoots you

core/startos/src/net/net_controller.rs
Blu-J: Logic is still not right here, but pretty sure at the moment we don't use this

container-runtime/src/Adapters/Systems/SystemForEmbassy/index.ts
Blu-J: Should be cleaned up now

sdk/lib/interfaces/Origin.ts
Blu-J: Cleaned up

core/startos/src/service/service_effect_handler.rs
Blu-J: The information is removed

core/startos/src/net/service_interface.rs
```diff
@@ -5,25 +5,13 @@ use serde::{Deserialize, Serialize};
...
-}
-
-#[derive(Clone, Debug, Deserialize, Serialize, TS)]
-#[ts(export)]
-#[serde(rename_all = "camelCase")]
-pub struct ExportedHostInfo {
-    pub id: HostId,
-    pub kind: HostKind,
-    pub hostnames: Vec<ExportedHostnameInfo>,
 }
```
dr-bonez, line 15: you should leave this type as is - just use `ServiceInterface` instead of it in patch-db

core/startos/src/net/net_controller.rs
```diff
@@ -209,6 +209,33 @@ impl NetService {
...
+                    .as_hosts_mut()
+                    .ser(&Default::default())?;
+                Ok(old_hosts)
+            })
+            .await?;
+        for (id, host) in old_hosts.0 {
+            self.update(id, host).await?;
+        }
+        Ok(())
+    }
```
dr-bonez, line 238: yeah, still not right. looking for foreach host, set bindings to empty. then call `update` on all the *resulting* hosts

core/startos/src/service/service_effect_handler.rs
```diff
@@ -639,39 +626,32 @@ async fn bind(
...
-                "preferredExternalPort": 80,
-                "addSsl":Value::Null,
-                "secure": false,
-                "ssl": false
-            }),
-            "suffix": "http"
-        }),
-        "type": "api"
-    }))
+) -> Result<ServiceInterfaceWrapper, Error> {
```
dr-bonez, line 640: this should still return `ServiceInterfaceWithHostInfo`. just perform the double get in the body of this function

core/startos/src/db/model/package.rs
```diff
@@ -11,7 +11,7 @@ use serde::{Deserialize, Serialize};
 use ts_rs::TS;
 
 use crate::net::host::HostInfo;
-use crate::net::service_interface::ServiceInterfaceWithHostInfo;
+use crate::net::service_interface::ServiceInterfaceWrapper;
```
dr-bonez, line 14:
```suggestion
use crate::net::service_interface::ServiceInterface;
```

core/startos/src/db/model/package.rs
```diff
@@ -333,7 +333,7 @@ pub struct PackageDataEntry {
     pub last_backup: Option<DateTime<Utc>>,
     pub current_dependencies: CurrentDependencies,
     pub actions: BTreeMap<ActionId, ActionMetadata>,
-    pub service_interfaces: BTreeMap<ServiceInterfaceId, ServiceInterfaceWithHostInfo>,
+    pub service_interfaces: BTreeMap<ServiceInterfaceId, ServiceInterfaceWrapper>,
```
dr-bonez, line 336:
```suggestion
    pub service_interfaces: BTreeMap<ServiceInterfaceId, ServiceInterface>,
```

core/startos/src/net/net_controller.rs
Blu-J: I'm going to just do this todo, I really don't need it

core/startos/src/service/service_effect_handler.rs
```diff
@@ -475,7 +461,7 @@ async fn get_primary_url(
 async fn list_service_interfaces(
     context: EffectContext,
     data: ListServiceInterfacesParams,
-) -> Result<BTreeMap<ServiceInterfaceId, ServiceInterfaceWithHostInfo>, Error> {
+) -> Result<BTreeMap<ServiceInterfaceId, ServiceInterface>, Error> {
```
dr-bonez, line 464:
```suggestion
) -> Result<BTreeMap<ServiceInterfaceId, ServiceInterfaceWithHostInfo>, Error> {
```

this still needs to return `ServiceInterfaceWithHostInfo`. readd that type, and interpolate the host info from the host id in the body of this function

core/startos/src/service/service_effect_handler.rs
```diff
@@ -639,39 +624,32 @@ async fn bind(
...
-                "preferredExternalPort": 80,
-                "addSsl":Value::Null,
-                "secure": false,
-                "ssl": false
-            }),
-            "suffix": "http"
-        }),
-        "type": "api"
-    }))
+) -> Result<ServiceInterface, Error> {
```
dr-bonez, line 638:
```suggestion
) -> Result<ServiceInterfaceWithHostInfo, Error> {
```

same here, it may be worth adding a reusable function for `ServiceInterface -> ServiceInterfaceWithHostInfo`

dr-bonez (changes requested):

dr-bonez:
ignore the last review

Blu-J:
Still works. 
Deleted config values for web
Re-configured
And the values still come back in the config set.

So, that leaves the system still working in the vaultwarden

Blu-J:
image

Blu-J:
image

dr-bonez (approved):

PR #2635, opened 2024-06-04
Feature/registry analytics

Dominion5254:
Add function for tracking MAU


dr-bonez (changes requested):
this still needs to be moved back to the registry, to the os.version.get rpc

dr-bonez (approved):

MattDHill (approved):

PR #2636, opened 2024-06-06
Fix sigterm-timeout type

chrisguida:
Fixes this:
sigterm-timeout-type-mismatch


container-runtime/src/Models/DockerProcedure.ts
```diff
@@ -31,12 +31,12 @@ export const matchDockerProcedure = object(
       "toml",
       "toml-pretty",
     ),
-    "sigterm-timeout": number,
```
Blu-J, line 34: Normal type is number, could be string, and more like Duration.

Blu-J:
I'm doing this in my own correctly with duration and parsers.

chrisguida:
Not sure what that means, but it sound like you have things handled :+1:

PR #2641, opened 2024-06-09
feat: implement state service, alert and routing

waterplea:


Zoiediez:
How long does it usually take for a transfer to take place? Thank you.

Zoiediez:
I'm sorry I really don't understand what that message you sent. If you need me to do anything, please let me know.

Zoiediez (approved):

Zoiediez (approved):

Zoiediez (approved):

Zoiediez (commented):
Fast/Combine-UIS Next-Minor

web/projects/ui/src/app/app-routing.module.ts
```diff
@@ -17,25 +18,27 @@ const routes: Routes = [
   },
   {
     path: 'diagnostic',
+    canActivate: [stateNot(['initializing'])],
```
MattDHill, line 21: Also "running"

web/projects/ui/src/app/pages/init/init.service.ts
```diff
@@ -0,0 +1,48 @@
...
+  takeWhile,
+} from 'rxjs'
+import { ApiService } from 'src/app/services/api/embassy-api.service'
+
+@Injectable({ providedIn: 'root' })
+export class InitService extends Observable<number> {
+  private readonly api = inject(ApiService)
+  private readonly errorService = inject(ErrorToastService)
+  private readonly progress$ = interval(500).pipe(
+    exhaustMap(() =>
```
MattDHill, line 22: Use guid to open websocket and get ongoing progress.

Should also return phases

web/projects/ui/src/app/pages/init/init.page.ts
```diff
@@ -1,11 +1,33 @@
...
+  readonly sub = this.progress$
+    .pipe(
+      filter(progress => progress === 1),
+      delay(500),
+    )
+    .subscribe(() => {
+      this.router.navigate([''])
+    })
+
+  getMessage(progress: number | null): string {
```
MattDHill, line 24: Messaging here is not correct and unnecessary. Messaging will come from phase names, which should be returned from progress sub

web/projects/ui/src/app/pages/init/init.page.html
```diff
@@ -0,0 +1,18 @@
...
+  </div>
+
+  <progress
+    tuiProgressBar
+    class="progress"
+    [style.max-width.rem]="40"
+    [style.margin]="'1rem auto'"
+    [attr.value]="progress && progress < 1 ? progress : null"
+  ></progress>
+  <p>{{ getMessage(progress) }}</p>
```
MattDHill, line 16: Should display phase name/percentage, see setup wizard for example.

web/projects/ui/src/app/pages/init/logs/logs.service.ts
```diff
@@ -0,0 +1,40 @@
...
+    )
+    .join('<br />')
+}
+
+@Injectable({ providedIn: 'root' })
+export class LogsService extends Observable<readonly string[]> {
+  private readonly api = inject(ApiService)
+  private readonly log$ = defer(() => this.api.followServerLogs({})).pipe(
+    switchMap(({ guid }) => this.api.openWebsocket$(guid, {})),
+    bufferTime(1000),
```
MattDHill, line 32: Cna lower the buffer to 250, I think. Will give a more live feel.

web/projects/ui/src/app/services/patch-db/patch-db.factory.ts
```diff
@@ -27,36 +27,34 @@ export function sourceFactory(
...
-                from(api.getState()).pipe(catchError(() => EMPTY)),
-              ),
-              take(1),
-              switchMap(() => watch$),
-            )
-          }),
-          tap(() => connectionService.websocketConnected$.next(true)),
-          startWith([dump]),
-        )
+    return defer(() => from(api.subscribeToPatchDB({}))).pipe(
```
MattDHill, line 35: `subscribeToPatchDB()` will fail if user is unauthenticated.

web/projects/ui/src/app/services/patch-db/patch-db.factory.ts
```diff
@@ -27,36 +27,34 @@ export function sourceFactory(
...
+      switchMap(({ dump, guid }) =>
+        auth.isVerified$.pipe(
```
MattDHill, line 37: Auth observable should be outside

web/projects/ui/src/app/services/patch-db/patch-db.factory.ts
```diff
@@ -27,36 +27,34 @@ export function sourceFactory(
...
+          switchMap(verified =>
+            verified
+              ? api.openWebsocket$<Revision>(guid, {}).pipe(
+                  bufferTime(250),
+                  filter(revisions => !!revisions.length),
+                  tap(() => connectionService.websocketConnected$.next(true)),
```
MattDHill, line 43: With global state monitoring, I don't think the websocketConnected$ observable is necessary anymore.

web/projects/ui/src/app/services/patch-db/patch-db.factory.ts
```diff
@@ -27,36 +27,34 @@ export function sourceFactory(
...
+                  startWith([dump]),
+                )
+              : EMPTY,
+          ),
+        ),
+      ),
+      catchError((_, original$) => {
+        connectionService.websocketConnected$.next(false)
```
MattDHill, line 51: See above

web/projects/ui/src/app/services/state.service.ts
```diff
@@ -0,0 +1,116 @@
...
+  paths: 'subset',
+  queryParams: 'exact',
+  fragment: 'ignored',
+  matrixParams: 'ignored',
+}
+
+@Injectable({
+  providedIn: 'root',
+})
+export class StateService extends Observable<RR.ServerState | null> {
```
MattDHill, line 39: Why `| null`?

web/projects/ui/src/app/services/state.service.ts
```diff
@@ -0,0 +1,116 @@
...
+  private readonly alerts = inject(TuiAlertService)
+  private readonly api = inject(ApiService)
+  private readonly router = inject(Router)
+  private readonly connection = inject(ConnectionService)
+  private readonly trigger$ = new BehaviorSubject<void>(undefined)
+  private readonly stream$ = this.trigger$.pipe(
+    switchMap(() =>
+      from(this.api.getState()).pipe(retry({ delay: 2000 }), startWith(null)),
```
MattDHill, line 47: Why `startWith(null)`?

web/projects/ui/src/app/services/state.service.ts
```diff
@@ -0,0 +1,116 @@
...
+    ),
+    tap(state => {
```
MattDHill, line 49: If null is necessary, shouldn't there be a case for it in the switch statement? At least a default case?

web/projects/ui/src/app/services/state.service.ts
waterplea: `null` means we do not know the state yet. If it was just initial query — we would just wait for the response, but we cache the last response with `shareReplay` and if we lose connection to the server we need to be able to reset state to something that means "we currently don't know".

web/projects/ui/src/app/services/state.service.ts
waterplea: So that our route guards receive `null` as in "the state is not yet known" and let you through to your routes, as we discussed.

web/projects/ui/src/app/services/state.service.ts
waterplea: No, it doesn't do anything, it means we don't know the state and have no place to redirect you yet, so it is just ignored.

MattDHill (approved):

PR #2644, opened 2024-06-13
Chore/refactoring effects

Blu-J:
### Proof it works

Configured
image

Start
image

Properties
image


container-runtime/src/Adapters/RpcListener.ts
```diff
@@ -247,7 +247,7 @@ export class RpcListener {
           })),
       )
       .when(exitType, async ({ id }) => {
-        if (this._system) await this._system.exit(this.effects)
+        if (this._system) await this._system.exit(this.effects("exitTypeId"))
```
dr-bonez, line 250: Use `null` and make it nullable. OS will accept null if it gets passed

Blu-J: Thanks

Blu-J:
```suggestion
        if (this._system) await this._system.exit(this.effects(null))
```

dr-bonez (approved):

PR #2647, opened 2024-06-18
Feature/fe new registry

elvece:
There are several todos dependent on additional information needed from BE types. Interim types were implemented to get most of the way there. It will be straightforward to remove these and use the original new types.

Todo summary:
- [x] `donationUrl`, `publishedAt`, and `dependencyMetadata` (title, icon, hidden, description), and `alerts` are missing in `PackageVersionInfo`.
- [x] `getStatic` endpoint needs to be updated with eventual BE implementation, possibly parsing info from s9pk on FE
- [x] A reworking of loading packages at particular versions is in order, particularly to handle toggling between alternate implementations smoother
- [x] Use emver helper (tbd) to determine if a version has a prefix and is an alternate implementation
- [x] Hitting endpoint for server metrics needs to be added
- [x] more robust fixtures for mocks
- [x] error: `emver.pipe.ts:29 emver comparison failed Parse Error: "knots-26.1.20240513" is not a valid Version 0.21.0 knots-26.1.20240513`


dr-bonez:
remaining TODOs:
- indicate there are no previous versions for `Past Release Notes` and `Other Versions`
- change `View Installed` to `View Installing` for service in `Installing` state
- fix progress bars not moving for service install (patch-db data is correct, issue is frontend)

dr-bonez (approved):

PR #2648, opened 2024-06-20
Boot param for logs subscription

MattDHill:


samsartor (approved):
matt assures me there is no malware

PR #2662, opened 2024-07-06
Feat/exver js parser

Dominion5254:
Adds module to parse `ExtendedVersion` and `VersionRange` from a string as well as expose a `satisfiedBy` function which returns a boolean indicating whether a given `ExtendedVersion` meets the criteria of a `VersionRange`.


MattDHill:
FE also uses a `compare()` fn to determine if a a version is higher/lower/same as another version. This is complicated by the addition of flavors, so we will need another fn to get the flavor of a version as well.

PR #2663, opened 2024-07-08
Adding in the transformation of the getConfig to the new types.

Blu-J:


Blu-J:
Ok, this code works, but something in the fix/config is broken. While debugging, it just worked once, so, this is a probablistic just to test it, but once you can get the config to work, it works...
image

PR #2669, opened 2024-07-12
init UI increase logs buffer and don't throw on websocket unsubscribe

MattDHill:


dr-bonez (approved):
error message fixed, and performance issue improved

PR #2670, opened 2024-07-12
use notification system for OS updates

MattDHill:
Awaiting BE portion.

- Introduces new notification code `2` to indicate OS welome notification, which includes amrkdown `data` of type `string`.


Blu-J:
image
image

elvece (approved):

PR #2679, opened 2024-07-22
refactor: Changing the types for the system to make sure the api's match

Blu-J:
between the system.

### DIFF
--- a/container-runtime/src/Adapters/Systems/SystemForEmbassy/index.ts
+++ b/container-runtime/src/Adapters/Systems/SystemForEmbassy/index.ts
@@ -48,6 +48,7 @@ import {
   transformConfigSpec,
   transformOldConfigToNew,
 } from "./transformConfigSpec"
+import { x25519 } from "@noble/curves/ed25519"
 
 type Optional<A> = A | undefined | null
 function todo(): never {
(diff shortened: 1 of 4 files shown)


container-runtime/src/Adapters/Systems/SystemForEmbassy/index.ts
```diff
@@ -179,7 +180,38 @@ const matchProperties = object({
...
+  "error-code": [number, string] | readonly [number, string]
+}>({
+  "error-code": tuple(number, string),
+})
 
+const assertNever = (
+  x: never,
+  message?: "Not expecting to get here",
+): never => {
+  throw new Error("")
```
dr-bonez, line 199: Maybe include an actual error message in for better debugging in case it's tripped

container-runtime/src/Adapters/Systems/SystemForEmbassy/index.ts
```diff
@@ -203,149 +235,122 @@ export class SystemForEmbassy implements System {
...
-          }
-        if (matchRpcResult.test(error)) return error
-        return {
-          error: {
-            code: 0,
-            message: String(error),
-          },
-        }
-      })
+  dependenciesMetadata(effects: T.Effects): Promise<T.ActionMetadata[]> {
```
dr-bonez, line 238: Do you mean actionsMetadata?

dr-bonez (commented):
For the most part looks good. Also, I realize this refactor makes the confusion around container runtime init vs service init even more confusing, so I'll try to come up with a different name for the former

PR #2682, opened 2024-07-24
default to all category and fix rounding for progress

MattDHill:


web/projects/ui/src/app/pipes/install-progress/install-progress.pipe.ts
```diff
@@ -19,9 +19,9 @@ export class InstallingProgressDisplayPipe implements PipeTransform {
...
 })
 export class InstallingProgressPipe implements PipeTransform {
-  transform(progress: T.Progress): number | null {
-    if (progress === true) return 1
-    if (progress === false || progress === null || !progress.total) return null
-    return Number((progress.done / progress.total).toFixed(2))
+  transform(progress: T.Progress): number {
+    if (progress === true) return 100
+    if (progress === false || progress === null || !progress.total) return 0
+    return Math.round((100 * progress.done) / progress.total)
```
dr-bonez, line 25: I might recommend Math.floor so it doesn't show 100% until it's actually complete

dr-bonez (approved):

PR #2707, opened 2024-08-13
Fix/overlay destroy

Blu-J:
Chore: Add more logging for error in js
Fix: The overlays that got destroyed during health checks


Blu-J (commented):
Should be good

dr-bonez (approved):

PR #2716, opened 2024-08-19
chore: reimplement refactor for the changes

Blu-J:
## Why

I want to make all the routes for each of the service adapters for the embassy and start9 services use fn's instead of routes

### DIFF
--- a/container-runtime/src/Adapters/RpcListener.ts
+++ b/container-runtime/src/Adapters/RpcListener.ts
@@ -19,7 +19,7 @@ import * as fs from "fs"
 
 import { CallbackHolder } from "../Models/CallbackHolder"
 import { AllGetDependencies } from "../Interfaces/AllGetDependencies"
-import { jsonPath } from "../Models/JsonPath"
+import { jsonPath, unNestPath } from "../Models/JsonPath"
 import { RunningMain, System } from "../Interfaces/System"
 import {
   MakeMainEffects,
@@ -64,7 +64,7 @@ const runType = object({
       input: any,
       timeout: number,
     },
-    ["timeout", "input"],
+    ["timeout"],
   ),
 })
 const sandboxRunType = object({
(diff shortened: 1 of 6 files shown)


container-runtime/src/Adapters/RpcListener.ts
```diff
@@ -342,3 +340,88 @@ export class RpcListener {
       })
   }
 }
+function getResult(
```
Blu-J, line 343: Should be the tranformations that existed for the other two adapters, so all the error adaption should be here

container-runtime/src/Adapters/RpcListener.ts
```diff
@@ -342,3 +340,88 @@ export class RpcListener {
...
+  procedure: typeof jsonPath._TYPE,
+  system: System,
+  effects: T.Effects,
+  timeout: number | undefined,
+  input: any,
+) {
+  return (async () => {
+    switch (procedure) {
+      case "/backup/create":
```
Blu-J, line 352: Here all the routes should be

container-runtime/src/Adapters/Systems/SystemForEmbassy/index.ts
```diff
@@ -61,6 +61,39 @@ const MANIFEST_LOCATION = "/usr/lib/startos/package/embassyManifest.json"
...
+  "error-code": tuple(number, string),
+})
+
+const assertNever = (
+  x: never,
+  message = "Not expecting to get here",
+): never => {
+  throw new Error(message)
+}
+const fromReturnType = <A>(a: U.ResultType<A>): A => {
```
Blu-J, line 82:
```suggestion
/**
  Should be changing the type for specific properties, and this is mostly a transformation for the old return types to the newer one.
*/
const fromReturnType = <A>(a: U.ResultType<A>): A => {
```

dr-bonez (approved):

PR #2732, opened 2024-08-30
add support for remote attaching to container

dr-bonez:
TODO:
* indicate options if multiple subcontainers
* allow specifying which subcontainer if multiple
* use default shell if no command provided


dr-bonez:
TODOs are handled. tty support is still not great, ideally should be fixed in a future pr

Blu-J (approved):

PR #2737, opened 2024-09-17
Fix/flavors

elvece:
Fixes to only display alternate implementations section of marketplace show page if there are flavors of that service.


web/projects/ui/src/app/services/marketplace.service.ts
```diff
@@ -42,7 +42,7 @@ export class MarketplaceService implements AbstractMarketplaceService {
   private readonly knownHosts$: Observable<StoreIdentity[]> = this.patch
     .watch$('ui', 'marketplace', 'knownHosts')
     .pipe(
-      map(hosts => {
+      map((hosts: { [url: string]: UIStore }) => {
```
MattDHill, line 45: This should not be necessary. Types show fine for me

MattDHill (approved):

PR #2759, opened 2024-10-18
Fix: Error Messages in HealthCheck

Blu-J:
Before
image
After
image


sdk/package/lib/util/SubContainer.ts
```diff
@@ -263,13 +263,8 @@ export class SubContainer implements ExecSpawnable {
     const appendData =
       (appendTo: { data: string | Buffer }) =>
```
dr-bonez, line 264:
```suggestion
      (appendTo: { data: string }) =>
```

dr-bonez (changes requested):

dr-bonez (dismissed):

MattDHill:
We were also seeing this in "loading" state, not sure if this fix addresses that as well.

MattDHill (approved):

PR #2762, opened 2024-10-22
sdk: fix piping stdio of Daemons, support onStdOut/onStderr

remcoros:
The default 'stdio' option of a node ChildProcess is 'pipe', causing no logs to be captured from service daemons.

Added an option 'stdio', defaulting to 'inherit' (this pipes stdio to the parent process, causing the logs to appear again).
If onStdout/onStderr hooks are used, then we switch to 'pipe' and it is the responsibility of whoever uses those to deal with logging.

### DIFF
--- a/sdk/package/lib/mainFn/CommandController.ts
+++ b/sdk/package/lib/mainFn/CommandController.ts
@@ -17,7 +17,7 @@ export class CommandController {
     readonly runningAnswer: Promise<unknown>,
     private state: { exited: boolean },
     private readonly subcontainer: SubContainer,
-    private process: cp.ChildProcessWithoutNullStreams,
+    private process: cp.ChildProcess,
     readonly sigtermTimeout: number = DEFAULT_SIGTERM_TIMEOUT,
   ) {}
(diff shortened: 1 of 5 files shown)


sdk/package/lib/mainFn/CommandController.ts
```diff
@@ -62,16 +62,25 @@ export class CommandController {
...
         })
       } else {
         childProcess = await subc.spawn(commands, {
           env: options.env,
+          stdio: options.onStdout || options.onStderr ? "pipe" : "inherit",
         })
       }
+
+      const childProcessWithStdio =
+        childProcess as cp.ChildProcessWithoutNullStreams
```
Blu-J, line 78: ❓ Is this safe as a cast to go from ChildProcess to ChildProcessWithoutNullStreams

remcoros: it worked, but realize now it's incorrect and not even needed, removed.

Blu-J (approved):

dr-bonez (approved):

PR #2763, opened 2024-10-22
sdk: checkPortListening: check tcp6/udp6 ports

remcoros:
ipv6 checks were missing.

This is from the Uptime Kuma container:

```
root@IMHWWUJCO33JPYQ5DBIF3N3B6PEVJBAE:~# netstat -l
Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State
tcp        0      0 localhost:domain        0.0.0.0:*               LISTEN
tcp        0      0 localhost:domain        0.0.0.0:*               LISTEN
tcp        0      0 0.0.0.0:5355            0.0.0.0:*               LISTEN
tcp6       0      0 [::]:3001               [::]:*                  LISTEN
tcp6       0      0 [::]:5355               [::]:*                  LISTEN
udp        0      0 0.0.0.0:5355            0.0.0.0:*
udp        0      0 localhost:domain        0.0.0.0:*
udp        0      0 localhost:domain        0.0.0.0:*
udp        0      0 IMHWWUJCO33JPYQ5:bootpc 0.0.0.0:*
udp6       0      0 [::]:5355               [::]:*
raw6       0      0 [::]:ipv6-icmp          [::]:*
```

notice that it only listens on a tcp6 port (3001).


Blu-J (dismissed):

dr-bonez:
my concern here is that the os does not support ipv6 networking to the containers rn. so a success via this metric would be misleading

remcoros:
not sure what's going on then. I packaged uptime kuma and the port healthcheck on 3001 was failing, but I can visit the ui fine and `netstat` shows only listening on tcp6 [::]:3001, so it does seem to work.

dr-bonez (approved):

PR #2785, opened 2024-11-12
sdk input spec improvements

dr-bonez:


sdk/base/lib/actions/input/builder/value.ts
```diff
@@ -122,19 +98,19 @@ export class Value<Type, Store> {
       boolean,
     )
   }
-  static text<Required extends RequiredDefault<DefaultString>>(a: {
+  static text<Required extends boolean>(a: {
```
Blu-J, line 101: So, good bye default

sdk/base/lib/actions/input/builder/value.ts
```diff
@@ -200,7 +175,8 @@ export class Value<Type, Store> {
         name: string
         description?: string | null
         warning?: string | null
-        required: RequiredDefault<DefaultString>
+        default: DefaultString | null
```
Blu-J, line 178: Kinda like that default will always be in the system, ok

MattDHill (approved):

PR #2796, opened 2024-11-28
More SDK comments

MattDHill:


sdk/package/lib/mainFn/Mounts.ts
```diff
@@ -30,9 +30,13 @@ export class Mounts<Manifest extends T.SDKManifest> {
   }
 
   addVolume(
+    /** The ID of the volume to mount. Much be one of the volume IDs defined in the manifest */
     id: Manifest["volumes"][number],
+    /** @TODO Bonez */
```
dr-bonez, line 35: if you want to only mount part of a volume - so some path within the volume

sdk/package/lib/mainFn/Mounts.ts
```diff
@@ -45,8 +49,11 @@ export class Mounts<Manifest extends T.SDKManifest> {
   }
 
   addAssets(
+    /** @TODO Bonez. I thought the assets array in the manifest was an array of path to asset directories. They have IDs? */
```
dr-bonez, line 52: yes, they have ids. the path to the asset directory uses the id as the folder name when we build the s9pk, so they are kind of the same

elvece (approved):

PR #2868, opened 2025-04-10
reduce task leaking

dr-bonez:


sdk/base/lib/util/getServiceInterfaces.ts
```diff
@@ -82,10 +82,15 @@ export class GetServiceInterfaces {
    */
   async *watch() {
     const { packageId } = this.opts
-    while (true) {
```
Blu-J, line 85: Probably the biggest win

Blu-J (approved):
I like the better implementation of the children as well

PR #2877, opened 2025-04-15
sdk improvements

dr-bonez:


core/startos/src/service/transition/restore.rs
```diff
@@ -73,9 +68,9 @@ impl Handler<Restore> for ServiceActor {
         if let Some(t) = old {
             t.abort().await;
         }
-        match transition.await {
-            None => Err(Error::new(eyre!("Restoring canceled"), ErrorKind::Unknown)),
-            Some(x) => Ok(x),
+        match recv_res.await {
```
elvece, line 71: can you explain the difference between using a transition vs oneshot here?

sdk/package/lib/util/fileHelper.ts
```diff
@@ -322,6 +326,41 @@ export class FileHelper<A> {
...
+  ) {
+    return new FileHelper<A>(
+      path,
+      (inData) => INI.stringify(inData, options),
+      (inString) => INI.parse(inString, options),
+      (data) => shape.unsafeCast(data),
+    )
+  }
+
+  static env<A extends Record<string, string>>(
```
elvece, line 343: 👏

elvece (approved):

core/startos/src/service/transition/restore.rs
dr-bonez: the transition was just a handle to the completion of the future but did not actually return the success or failure of it

PR #2884, opened 2025-04-21
use docker for build steps that require linux when not on linux

dr-bonez:


dr-bonez:
closing in favor of #2892

PR #2914, opened 2025-04-25
update marketplace url to reflect build version

elvece:
- updates makefile to set default marketplace url to the same registry as build version (ie. alpha/beta)
- updates FE code so messaging is still accurate for these testing marketplaces
- fixes bug with border-radius cutting off marketplace icons


MattDHill (commented):
I don't think we should include the community registry either

PR #2938, opened 2025-05-09
Fix/controls

elvece:


web/projects/ui/src/app/routes/portal/routes/marketplace/components/controls.component.ts
```diff
@@ -36,7 +36,7 @@ import { ApiService } from 'src/app/services/api/embassy-api.service'
   template: `
     @if (localPkg) {
       @if (
-        localPkg.stateInfo.state === 'installed' && (localPkg | toManifest);
+        localPkg.stateInfo.state && (localPkg | toManifest);
```
MattDHill, line 39: Youre just checking if state exists?

web/projects/ui/src/app/routes/portal/routes/marketplace/components/controls.component.ts
```diff
@@ -78,7 +78,11 @@ import { ApiService } from 'src/app/services/api/embassy-api.service'
         appearance="secondary-grayscale"
         (click)="showService()"
       >
-        {{ 'View Installed' | i18n }}
+        {{
+          'View ' +
+            localPkg.stateInfo.state.charAt(0).toUpperCase() +
```
MattDHill, line 83: Pretty sire there is a titlecase pipe for this. Also does it even need to be capitalized? I think we have been lower casing the second word

web/projects/ui/src/app/routes/portal/routes/notifications/item.component.ts
```diff
@@ -32,7 +32,7 @@ import { i18nPipe } from '@start9labs/shared'
     <td class="service">
       @if (manifest$ | async; as manifest) {
         <a tuiLink [routerLink]="getLink(manifest.id)">
-          {{ manifest.title }}
+          {{ notificationItem.packageId }}
```
MattDHill, line 35: No, we should show the title if we have it

MattDHill (changes requested):

MattDHill:
Also, we already have translations for view and the various statuses. You can just use the i18n pipe twice instead if creating new translations

web/projects/ui/src/app/routes/portal/routes/marketplace/components/controls.component.ts
elvece: yes since `state: "installing" | "restoring" | "installed" | "removing" | "updating"`, and its needed to complete the `localManifest` type

web/projects/ui/src/app/routes/portal/routes/marketplace/components/controls.component.ts
elvece: 👍 titlecase pipe
its quite varied and inconsistent across buttons actually, seems to depend on the phrase

web/projects/ui/src/app/routes/portal/routes/marketplace/components/controls.component.ts
MattDHill: but it's always there, you're just testing it for truthiness here, and it's always truthy...

elvece:
> Also, we already have translations for view and the various statuses. You can just use the i18n pipe twice instead if creating new translations

we could, but its technically not grammatically correct for all languages, eg. german

MattDHill (dismissed):

MattDHill (approved):

PR #2945, opened 2025-05-15
Add french translation

nbxl21:
I am a native French speaker, so I wanted to spend some time helping with the French translation.

I started with the output from an LLM and then reviewed it line by line to make all the necessary modifications.

I took notes while working so I could later verify the correctness of some text fragments in the context of the UI. Sometimes it was not easy to know if a message would be displayed as a button, a title, or a loading message, and the final result depends on this. I will try to make the required changes as soon as I can update to 4.0.0. Most of the translations were unambiguous, so the few remaining imprecisions should not impact the UX too much.

Aside the "fr.ts" file, I tried to make the other required changes by following the README but it's not clear to me if I addressed all the ramifications.  For instance, it seems like I need to add the translation for “French” in each dictionary, as is done for the other languages at keys 488 to 491. However, I wasn’t sure about the exact requirements or whether it was necessary, so I skipped this step.

Note: seems like the `  1. Export the dictionary in src/public-api.ts` line of the README can be removed, or at least I didn't do it.

Let me know anything.


MattDHill (changes requested):
Amazing, thank you!

You are correct, please go ahead and remove that line from the README, it is outdated.

MattDHill (approved):

MattDHill (changes requested):
Build is failing, looks like you need to add the word "french" to the English dictionary, as well as its translations to the other dictionaries.

I suppose this step should be added to the README as well, sorry about that, translations are brand new.

nbxl21:
Makes sense, I will do it now and update the instructions

MattDHill (approved):

PR #2969, opened 2025-06-24
postinit adjustments

gStart9:
`ID_LIKE` is intended for downstream software, automation tools, or scripts to infer behavior or compatibility when they don't recognize the main ID directly.


gStart9:
I wanna add something else too but gotta wait for the build I guess...

dr-bonez:
Do you have any examples or scripts that use this ID_LIKE parameter? Because treating startos as a Debian system in a script can get you in trouble

PR #3064, opened 2025-12-01
fix: pass --allow-discards to luksOpen for trim support

remcoros:
It was already passed when creating the fs here: https://github.com/remcoros/start-os/blob/9fb1773637e763b8094978b1b4fdc7ba42fd6742/core/startos/src/disk/main.rs#L129 
but not when mounting


remcoros:
aarch task failure is unrelated. disk is full: https://github.com/Start9Labs/start-os/actions/runs/19831188148/job/56921670173#step:11:3974

PR #3073, opened 2025-12-14
fix: shutdown order

remcoros:
Draft to fix shutdown sequence and implement correct shutdown order based on Daemon dependencies.

Leaving some in-line comments.

It's based on `next/major` so I could test this on current alpha. It needs some conflict resolving after `feature/sdk-idmap` is merged.

### DIFF
--- a/sdk/package/lib/mainFn/Daemon.ts
+++ b/sdk/package/lib/mainFn/Daemon.ts
@@ -37,7 +37,7 @@ export class Daemon<
     return this.oneshot
   }
   static of<Manifest extends T.SDKManifest>() {
-    return async <C extends SubContainer<Manifest> | null>(
+    return <C extends SubContainer<Manifest> | null>(
       effects: T.Effects,
       subcontainer: C,
       exec: DaemonCommandType<Manifest, C>,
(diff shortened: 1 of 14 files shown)


sdk/package/lib/mainFn/Daemon.ts
```diff
@@ -50,9 +50,6 @@ export class Daemon<
           exec,
         )
       const res = new Daemon(subc, startCommand)
-      effects.onLeaveContext(() => {
```
remcoros: removed here and in Daemons.build(). Use only 'term' and 'stop' to explicitly stop daemons.

the calls to 'stop' were already handled in RpcListener, SystemForStartOs and MainLoop

sdk/package/lib/mainFn/Daemons.ts
```diff
@@ -413,26 +413,50 @@ export class Daemons<Manifest extends T.SDKManifest, Ids extends string>
 
   async term() {
     try {
-      for (let result of await Promise.allSettled(
-        this.healthDaemons.map((x) => x.term()),
-      )) {
-        if (result.status === "rejected") {
-          console.error(result.reason)
+      const remaining = new Set(this.healthDaemons)
```
remcoros: this is basically a topological sort on the dependencies, but in reverse.

daemons without dependants are terminated concurrently.

sdk/package/lib/mainFn/Daemons.ts
```diff
@@ -413,26 +413,50 @@ export class Daemons<Manifest extends T.SDKManifest, Ids extends string>
...
+          }),
+        )
       }
     } finally {
       this.effects.setMainStatus({ status: "stopped" })
     }
   }
 
   async build() {
-    this.effects.onLeaveContext(() => {
```
remcoros: this causes duplicate calls to term.

PR #3110, opened 2026-02-04
fix: take into account service dependencies when loading the service …

remcoros:
…containers and load them in correct order

this was done with help of claude. i'm not familiar with how 'db' works exactly...

Testing so far, this fixes the issue where dependants of bitcoind that use createTask to require specific bitcoin.conf values, fail to create the service container


remcoros:
If you can take a look, I wouldn't know where to start tbh. This was the quickest I could come up with to confirm if this was the race condition we were seeing.

PR #3161, opened 2026-04-04
fix: use Tls::Required for STARTTLS in test_smtp

MattDHill:
## Summary
- Fixed `test_smtp` using `Tls::Wrapper` (immediate TLS) for both STARTTLS and TLS modes
- STARTTLS now correctly uses `Tls::Required`, which connects plaintext first then upgrades via the STARTTLS command
- `Tls::Wrapper` on port 587 caused the TLS handshake to fail (server sends a plaintext SMTP banner instead of TLS ServerHello), crashing the server and dropping the PatchDB WebSocket — the UI showed the reconnection/loading screen

## Test plan
- [x] Configure System SMTP with Protonmail (STARTTLS, port 587) and send a test email
- [x] Configure System SMTP with a TLS provider (port 465) and send a test email
- [x] Verify errors are shown as toasts instead of crashing the UI

### DIFF
--- a/core/src/system/mod.rs
+++ b/core/src/system/mod.rs
@@ -1233,12 +1233,15 @@ pub async fn test_smtp(
         .header(ContentType::TEXT_PLAIN)
         .body("This is a test email sent from your StartOS Server".to_owned())?;
 
+    let tls_parameters = TlsParameters::new(host.clone())?;
     let transport = match security {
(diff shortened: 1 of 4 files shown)


stupleb:
Works - Proton - STARTTLS

dr-bonez (approved):

PR #3190, opened 2026-04-28
feat(net): reverse-proxy auth gate (Bearer / Basic) with X-Forwarded-User

helix-nine:
## Summary

Adds an optional **`auth` gate** on `AddSslOptions` so a binding can require Bearer or Basic auth at the OS reverse proxy. Authenticated Basic users are forwarded upstream as `X-Forwarded-User`, in the same spirit as the existing `addXForwardedHeaders` plumbing.

```ts
await multi.bindPort(8080, {
  protocol: 'http',
  addSsl: {
    auth: {
      type: 'basic',
      credentials: [
        { username: 'alice', password: 'hunter2' },
        { username: 'bob',   password: 'swordfish' },
      ],
    },
  },
})
// upstream sees:  Authorization: Basic ...   AND   X-Forwarded-User: alice
```

Unauthenticated / unknown-credential requests get `401 Unauthorized` with a `WWW-Authenticate: Basic realm="StartOS"` (or `Bearer …`) challenge.

## Design

- **`ProxyAuth`** in `core/src/net/host/binding.rs`:
  - `Basic { credentials: Vec<BasicCredential> }` — list of accepted `(username, password)` pairs. Matched username is forwarded as `X-Forwarded-User`.
  - `Bearer { tokens: Vec<String> }` — list of accepted tokens. No user is forwarded (no user concept).
- **`AuthGate`** (in `core/src/net/http.rs`): compiled once per stream from `ProxyAuth`. Stores a `HashMap<HeaderValue, Option<HeaderValue>>` keyed by the *full* `Authorization` value (`"Basic dXNlcjpwYXNz"` / `"Bearer foo"`) and a pre-built `WWW-Authenticate` header value. Per-request cost is one `HeaderMap::get` + one `HashMap` probe — no per-request base64 / no per-credential loop.
- **`apply_request_policy`**: single function called from both the HTTP/1 and HTTP/2 service_fn. It (a) strips client-supplied `X-Forwarded-*` headers, (b) runs the gate, (c) sets `X-Forwarded-Proto`/`X-Forwarded-For` if `add_forwarded`, (d) sets `X-Forwarded-User` on a successful Basic match.
- **Response unification**: service_fn returns `Response<BoxBody<Bytes, …>>` so we can return either an upstream `hyper::body::Incoming` or a synthetic `Full<Bytes>` 401 body from the same closure.

### Security notes

- Client-supplied `X-Forwarded-User` is **always stripped**, even on bindings without a gate and even when `addXForwardedHeaders` is `false`. An upstream service that trusts that header can never be tricked by a malicious client.
- Lookup is plain `HashMap` (not constant-time). These are operator-controlled accept-lists, not user-supplied passwords being verified against a database, so timing leakage of the credential set size is not in the threat model. If we want constant-time later we can swap the inner check.
- Compile errors (e.g. a credential containing chars that don't fit in an `HeaderValue`) are logged and the gate is dropped, leaving the binding behaving as if `auth` were unset rather than 401-ing every request — i.e. configuration errors never make a binding *more* permissive than the operator intended.

## What changed

- `core/src/net/host/binding.rs`: `ProxyAuth` enum + `BasicCredential` struct, both `#[ts(export)]`.
- `core/src/net/http.rs`: `AuthGate`, `apply_request_policy`, `BoxBody`-based response type, refactored HTTP/1 + HTTP/2 service_fn, **5 unit tests**.
- `core/src/net/vhost.rs`: compiles `AuthGate` once per stream and routes through `run_http_proxy` whenever forwarded-headers OR auth-gate is set.
- `core/src/net/net_controller.rs`, `core/src/db/model/public.rs`: default `auth: None` on `AddSslOptions` literals.
- `sdk/base/lib/osBindings/`: regenerated bindings (new `ProxyAuth.ts`, new `BasicCredential.ts`, updated `AddSslOptions.ts`).
- `sdk/base/lib/interfaces/Host.ts`: re-exports `ProxyAuth` and `BasicCredential`. `bindPortForKnown` defaults `auth: null` so package devs override per-binding via the `addSsl` partial.
- `container-runtime/`, `web/projects/ui/src/app/services/api/mock-patch.ts`: default `auth: null` on existing literals.

## Verification

- `cargo check -p start-os --lib` ✅
- `cargo test --lib --features test net::` (incl. all `export_bindings_*` and the 5 new gate tests) ✅
- `cd sdk/base && npm run tsc && npm test` ✅
- `cd sdk/package && npm run tsc` ✅
- `cd container-runtime && npx tsc --noEmit` ✅

## Notes

- HTTP/1.1 → HTTP/2 upgrade path: the inner `run_http2_proxy` is called with `add_forwarded=false, gate=None` — the outer hop already validated the request, and the inner call only carries the upgraded byte stream.
- No SDK version bump in this PR.


core/src/net/host/binding.rs
```diff
@@ -236,6 +236,38 @@ pub struct AddSslOptions {
...
     pub add_x_forwarded_headers: bool, // TODO
     pub alpn: Option<AlpnInfo>,
+    /// Optional reverse-proxy auth gate. When set, the OS reverse proxy
+    /// will validate the `Authorization` header on incoming HTTP requests
+    /// against this configuration before forwarding them upstream.
+    /// Unauthenticated requests get `401 Unauthorized` with an appropriate
+    /// `WWW-Authenticate` challenge. For `Basic`, the authenticated
+    /// username is forwarded to the upstream service as `X-Forwarded-User`.
+    /// Setting this implies HTTP-aware proxying (same path as
+    /// `add_x_forwarded_headers`).
```
dr-bonez: make a similar doc comment on add_x_forwarded_headers, and drop the "same path as" parenthetical

dr-bonez (changes requested):

helix-nine:
Done — pushed as `2016d65ec`.

- Added a parallel doc comment to `add_x_forwarded_headers` describing what it actually does (`X-Forwarded-Proto: https` + `X-Forwarded-For: <client-ip>`, and that it implies HTTP-aware proxying). Also dropped the stale `// TODO`.
- Dropped the `(same path as add_x_forwarded_headers)` parenthetical from `auth`'s doc — both fields are now self-documenting so the cross-reference isn't needed.
- Regenerated `sdk/base/lib/osBindings/AddSslOptions.ts` so the same docs surface in the TS bindings for package developers.

dr-bonez (approved):

helix-nine:
Thanks for the review, @dr-bonez! 🙏

core/src/net/http.rs
```diff
@@ -1,17 +1,157 @@
...
+        use base64::Engine;
+        let mut valid: HashMap<HeaderValue, Option<HeaderValue>> = HashMap::new();
+        let challenge = match auth {
+            ProxyAuth::Bearer { tokens } => {
+                for token in tokens {
+                    let v = HeaderValue::from_str(&format!("Bearer {token}"))
+                        .with_kind(ErrorKind::InvalidRequest)?;
+                    valid.insert(v, None);
+                }
+                HeaderValue::from_static("Bearer realm=\"StartOS\"")
```
dr-bonez: Should the realm be defined by the package?

core/src/net/http.rs
```diff
@@ -1,17 +1,157 @@
...
+fn apply_request_policy<B>(
+    req: &mut Request<B>,
+    src_ip: Option<IpAddr>,
+    add_forwarded: bool,
+    gate: Option<&AuthGate>,
+) -> Result<(), Response<ProxyBody>> {
+    // Always strip — never trust client-supplied forwarded identity.
+    let h = req.headers_mut();
+    h.remove("X-Forwarded-User");
+    if add_forwarded || gate.is_some() {
```
dr-bonez: this check is redundant with the fact we don't do an http-level proxy unless this condition is true

core/src/net/vhost.rs
```diff
@@ -815,14 +822,37 @@ where
...
+        // gate, which means upstream is reachable; that is the same
+        // behaviour as if `auth` were never set, so it never makes a
+        // misconfigured binding *more* permissive than the operator's
+        // intent (a misconfigured gate would otherwise 401 every request,
+        // but configuring credentials that don't fit in a header value is
+        // an authoring bug, not a runtime condition).
+        let auth_gate = match self.auth.as_ref().map(crate::net::http::AuthGate::from_auth) {
+            Some(Ok(g)) => Some(g),
+            Some(Err(e)) => {
+                tracing::error!("Failed to compile proxy auth gate: {e}");
```
dr-bonez: should probably fail the connection in this scenario. way safer option

dr-bonez (changes requested):

helix-nine:
Pushed `efecc3c53` addressing all three pieces of feedback:

**1. Package-defined realm** (`http.rs:65`)
`ProxyAuth::{Bearer,Basic}` now each carry an optional `realm` field that flows into the `WWW-Authenticate` challenge — defaults to `"StartOS"` when unset. Realm strings are sanitized against `quoted-string` metacharacters (`"`, `\`, control bytes are dropped) so a hostile or sloppy realm can't break out of the challenge header. Bindings re-exported from `osBindings/ProxyAuth.ts`.

**2. Redundant strip guard** (`http.rs:128`)
You're right — `apply_request_policy` is only ever reached on the HTTP-aware path, which is itself gated on `add_forwarded || gate.is_some()`. Dropped the inner `if` and `X-Forwarded-{For,Proto,User}` are now unconditionally stripped at the top of the function.

**3. Fail-closed on gate compile failure** (`vhost.rs:839`)
Agreed — silently-no-gate is way too dangerous. On `AuthGate::from_auth` returning `Err` we now log and `return` from `handle_stream` before spawning, dropping `stream` and `prev` so the TLS-accepted connection terminates. Operators see the error in logs; clients see a connection close instead of an exposed upstream.

Test suite: existing 5 tests updated for the new `realm: …` field, plus a new `realm_is_sanitized_against_quoted_string_metacharacters` test (`hax"\…or` → `Basic realm="haxor"`) and the rejection test now asserts the package-supplied realm round-trips into the challenge verbatim. All 6 `net::http` tests green; SDK + container-runtime tsc clean.

core/src/net/http.rs
```diff
@@ -1,17 +1,175 @@
...
+                None => Err(self.unauthorized()),
+            },
+            None => Err(self.unauthorized()),
+        }
+    }
+
+    fn unauthorized(&self) -> Response<ProxyBody> {
+        Response::builder()
+            .status(http::StatusCode::UNAUTHORIZED)
+            .header(http::header::WWW_AUTHENTICATE, self.challenge.clone())
```
dr-bonez, line 120: should this be PROXY_AUTHENTICATE instead?

helix-nine:
I think `WWW-Authenticate` / `401` is correct here, but let me lay out the reasoning since it's a fair question.

Per RFC 7235:

- **`WWW-Authenticate` + `401 Unauthorized`** — auth required to access the *target resource*. Used by origin servers and reverse proxies that are gating access on behalf of an origin.
- **`Proxy-Authenticate` + `407 Proxy Authentication Required`** — auth required to use *the proxy itself*. Used by forward proxies (think corporate egress, an HTTP `CONNECT` tunnel, etc.) where the client has explicitly configured the proxy.

The StartOS reverse proxy is transparent to clients — from a client's POV they're talking to `https://<package-host>/…` and the proxy is acting as the origin. They didn't configure a proxy; they don't know one exists. So semantically we're a reverse proxy gating an origin, which is the `WWW-Authenticate` case. This also matches what nginx (`auth_basic`), Traefik (`BasicAuth` middleware), Caddy (`basicauth`), Apache (`mod_auth_basic`), and HAProxy all emit when they gate upstream access.

There's a practical consequence too: browsers treat the two very differently. `401`/`WWW-Authenticate` triggers the standard credential prompt scoped to the origin + realm, with credentials cached per-origin so the user only sees the prompt once. `407`/`Proxy-Authenticate` is gated through the OS/browser proxy stack and in many configurations never prompts at all — for a transparent reverse proxy it would just produce a broken UX.

So I'd like to leave this as-is. Happy to push back if I'm misreading the intent though — were you thinking about a forward-proxy use case I'm not seeing, or is there a specific client behaviour we're trying to match?

dr-bonez (approved):

helix-nine:
Thanks for the second look, @dr-bonez! 🙏

PR #3191, opened 2026-04-29
perf(net/vhost): skip cert resolution for TLS passthrough

helix-nine:
## Summary

Cuts vhost passthrough TLS handshake latency from ~2.7s down to ~1×RTT-to-backend by reordering the per-connection TLS pipeline so the routing decision (passthrough vs terminate) runs **before** cert resolution, instead of after.

## Diagnosis

For every incoming TLS connection on a vhost port, the listener was built as `TlsHandlerWrapper { inner: ChainedHandler(AcmeTlsHandler, RootCaTlsHandler), wrapper: VHostConnector }`. `TlsHandlerWrapper::get_config` ran the inner cert chain first, then asked `VHostConnector::wrap` whether the SNI mapped to a passthrough. For passthrough hits, the freshly produced `ServerConfig` was **discarded** — but the work to build it was not free:

- `RootCaTlsHandler::get_config` calls `db.mutate(|db| cert_for(...))`, which serializes through the single patch-db writer queue and pays the fsync cost on commit.
- Inside `Model<CertStore>::cert_for`, on a cold tuple `{SNI, local_addr.ip(), wan_ip}` (happens whenever the WAN IP wasn't populated last time, on dual-stack/IPv6 temp-addr connections, on first hits, etc.), it generates an ed25519 keypair + an nistp256 keypair, signs two leaf certs against the intermediate, and writes the new entry back into `leaves`.

With any concurrent DB write activity, this comfortably lands in the seconds range — observed as ~2.7s client-visible TLS handshake latency on a vhost passthrough between two StartOS servers.

## Fix

Promote `VHostConnector` from a `WrapTlsHandler` (post-processor) into a real `TlsHandler` (`VHostTlsHandler`) that owns the cert chain as its `inner`:

- On each `get_config`, look up the SNI in the mapping immediately. If the matched target is a passthrough, build a cheap stub `ServerConfig` (no keygen, no DB I/O — uses a `NoCertResolver`) just to satisfy the `VHostTarget::preprocess` signature, and skip the inner cert chain entirely.
- Only invoke `inner.get_config` for terminating-TLS targets where the cert is actually consumed.
- Keep ACME `acme-tls/1` ALPN challenges going straight to the inner chain so they're answered locally and never tunneled.

The remaining latency on a passthrough is dominated by the TCP connect to the backend plus the real TLS handshake against it — i.e. roughly 1×RTT-to-backend on top of a normal handshake.

Drops the now-unused `WrapTlsHandler` / `TlsHandlerWrapper` abstractions from `net::tls` (only consumer was `VHostConnector`).

## Tests

- `cargo check --lib -p start-os`: clean
- `cargo test --lib -p start-os net::`: 64/64 pass

### DIFF
--- a/core/src/net/vhost.rs
+++ b/core/src/net/vhost.rs
@@ -980,20 +980,7 @@ where
         hello: &'a ClientHello<'a>,
         metadata: &'a <A as Accept>::Metadata,
     ) -> Option<TlsHandlerAction> {
-        // ACME `acme-tls/1` challenge must be answered locally with the
-        // challenge cert produced by `AcmeTlsHandler` in the inner chain;
-        // skip the routing/passthrough decision entirely so we don't
-        // accidentally tunnel the challenge to a backend.
-        if hello
-            .alpn()
-            .into_iter()
-            .flatten()
-            .any(|a| a == ACME_TLS_ALPN_NAME)
-        {
-            return self.inner.get_config(hello, metadata).await;
-        }
-
-        let (target, rc) = self.mapping.peek(|m| {
+        let routed = self.mapping.peek(|m| {
             m.get(&hello.server_name().map(InternedString::from))
                 .or_else(|| m.get(&None))
                 .into_iter()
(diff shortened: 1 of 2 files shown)


core/src/net/vhost.rs
```diff
@@ -910,66 +910,159 @@ impl<A: Accept + 'static> GetAcmeProvider for GetVHostAcmeProvider<A> {
...
         }
+
+        // Terminating target. For an `acme-tls/1` challenge we serve the
+        // challenge cert directly from the inner chain and skip
+        // `preprocess` — there's no point opening a backend TCP
+        // connection just to satisfy a one-shot validation handshake the
+        // ACME client will tear down immediately.
+        if acme_alpn {
+            return self.inner.get_config(hello, metadata).await;
+        }
```
dr-bonez:
```suggestion
        // passthroughs should not intermediate acme challenges
        if routed.as_ref().map_or(false, |t| t.0.is_passthrough()) {
            let stub = passthrough_stub_config(&self.crypto_provider).log_err()?;
            let (_, store) = target.into_preprocessed(rc, stub, hello, metadata).await?;
            self.preprocessed = Some(store);
            return Some(TlsHandlerAction::Passthrough);
        }
        
        if acme_alpn {
            return self.inner.get_config(hello, metadata).await;
        }
        
        let Some((target, rc)) = routed else {
            return None;
        };
```

dr-bonez (changes requested):

helix-nine:
Pushed [`019d5a2`](https://github.com/Start9Labs/start-os/pull/3191/commits/019d5a25e). Adopted your structure — single passthrough check up front, single `acme_alpn` fallback to inner, then the terminating cert path. Net `-30 / +5` in the body of `get_config`. Tests still 3/3 in `net::vhost`.

core/src/net/vhost.rs
```diff
@@ -910,66 +910,129 @@ impl<A: Accept + 'static> GetAcmeProvider for GetVHostAcmeProvider<A> {
...
+            .alpn()
+            .into_iter()
+            .flatten()
+            .any(|a| a == ACME_TLS_ALPN_NAME);
 
-        self.1 = Some(store);
+        // Passthroughs should not intermediate ACME challenges — the
+        // backend is the ACME client and holds the challenge cert.
+        if routed.as_ref().is_some_and(|(t, _)| t.0.is_passthrough()) {
+            let (target, rc) = routed.unwrap();
```
dr-bonez:
```suggestion
        if let Some((target, rc)) = routed.as_ref().filter(|(t, _)| t.0.is_passthrough()) {
```

dr-bonez (changes requested):

helix-nine:
Pushed [`079fe5b`](https://github.com/Start9Labs/start-os/pull/3191/commits/079fe5b40). Folded the predicate into the `if let` with `.as_ref().filter(...)` per your suggestion; `target.clone().into_preprocessed(rc.clone(), ...)` because `routed` still needs to be available for the terminating path below (`DynVHostTarget` is `Arc`-backed and `Weak` is refcounted, so the clones are cheap). Tests still 3/3.

dr-bonez (approved):

helix-nine:
Thanks for the review!

PR #3192, opened 2026-04-29
feat(core): limited os update via diagnostic api

dr-bonez:
## Summary

- Adds `diagnostic.update` RPC, mirroring `server.update`'s shape (`UpdateSystemParams` in, `UpdateSystemRes` out with a WS continuation guid for progress). Lets users recover from broken state without reinstalling.
- Refactors `do_update` to take `reqwest::Client` instead of `RpcContext` — it only ever used the client. The diagnostic path reuses it directly.
- Fixes a latent registry-call bug in `server.update`: `os.version.get` was sent `{source, target}` but the registry deserializes `{sourceVersion, targetVersion}` (camelCase). The normal flow only worked because the registry falls back to the `DeviceInfo` header. Diagnostic mode has no such fallback, so the field names had to be correct anyway — fixing it in both spots.

## Constraints in diagnostic mode

- **No patch-db**: progress streams from `FullProgressTracker.stream()` straight into the WS instead of mirroring to `/public/serverInfo/statusInfo/updateProgress`.
- **No signing key**: `os.version.get` goes out unsigned via a one-off `reqwest::Client` and an `UnsignedRegistry` shim around `signature::call_remote`. The endpoint is public.
- **No DeviceInfo header**: source/target/platform passed explicitly in the JSON params.
- **Single-flight**: `SyncMutex<bool>` on `DiagnosticContext` prevents concurrent updates; collisions error with `update.already-updating`.
- **No auto-restart**: caller invokes `diagnostic.restart` afterward, matching the rest of the diagnostic surface.

## Test plan

- [ ] `cargo check -p start-os` (passes locally)
- [ ] Boot into diagnostic mode, run `start-cli diagnostic update --registry <url>` against a registry with a newer version available — verify download/verify/apply works and progress streams to the CLI
- [ ] Verify `diagnostic.restart` after success boots into the new version
- [ ] Concurrent `diagnostic.update` call returns `update.already-updating` error
- [ ] Normal-mode `server.update` still works end-to-end (regression check on the registry call field rename)

### DIFF
--- a/core/src/context/diagnostic.rs
+++ b/core/src/context/diagnostic.rs
@@ -17,10 +17,7 @@ pub struct DiagnosticContextSeed {
     pub error: Arc<RpcError>,
     pub disk_guid: Option<InternedString>,
     pub rpc_continuations: RpcContinuations,
-    /// Held by an in-flight `diagnostic.update`. The strong count is `1`
-    /// when no update is running; the update task holds a clone for its
-    /// entire lifetime, and dropping the clone (success, failure, or
-    /// cancellation) automatically frees the slot.
+    /// Single-flight token for `diagnostic.update`; strong count > 1 means an update is running.
     pub update_in_progress: SyncMutex<Arc<()>>,
 }
 
--- a/core/src/context/diagnostic.rs
+++ b/core/src/context/diagnostic.rs
@@ -10,12 +10,14 @@ use crate::context::config::ServerConfig;
 use crate::prelude::*;
 use crate::rpc_continuations::RpcContinuations;
 use crate::shutdown::Shutdown;
+use crate::util::sync::SyncMutex;
 
 pub struct DiagnosticContextSeed {
     pub shutdown: Sender<Shutdown>,
     pub error: Arc<RpcError>,
     pub disk_guid: Option<InternedString>,
     pub rpc_continuations: RpcContinuations,
+    pub update_in_progress: SyncMutex<bool>,
 }
 
 #[derive(Clone)]
@@ -40,6 +42,7 @@ impl DiagnosticContext {
             disk_guid,
             error: Arc::new(error.into()),
             rpc_continuations: RpcContinuations::new(),
+            update_in_progress: SyncMutex::new(false),
         })))
     }
 }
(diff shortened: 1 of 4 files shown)


core/src/context/diagnostic.rs
```diff
@@ -10,12 +10,14 @@ use crate::context::config::ServerConfig;
...
 use crate::rpc_continuations::RpcContinuations;
 use crate::shutdown::Shutdown;
+use crate::util::sync::SyncMutex;
 
 pub struct DiagnosticContextSeed {
     pub shutdown: Sender<Shutdown>,
     pub error: Arc<RpcError>,
     pub disk_guid: Option<InternedString>,
     pub rpc_continuations: RpcContinuations,
+    pub update_in_progress: SyncMutex<bool>,
```
dr-bonez: @helix-nine  use an Arc<()> and use the reference count to know if an update is in progress.

helix-nine:
Done. https://github.com/Start9Labs/start-os/pull/3192

@dr-bonez switched the single-flight gate to an `Arc<()>` refcount as suggested — pushed as `7d25662db`.

- `DiagnosticContextSeed::update_in_progress` is now `SyncMutex<Arc<()>>` (the mutex is just for the atomic check-and-clone window, not for the running update).
- Claim path: `mutate(|slot| if Arc::strong_count(slot) > 1 { None } else { Some(slot.clone()) })`. The clone is moved into the spawned update task as `_guard`.
- Release is automatic — the guard drops with the task, so success, `do_update` errors, panics, and all the early-return paths in `run_update` (no-op `target: None`, `validate` failure, etc.) all free the slot without an explicit clear-on-exit branch. That removed the `if result.target.is_none() { ... }` branch and the per-task `ctx_for_task` clone that existed solely to flip the bool back.

`cargo check -p start-os` (lib) is clean; the only build errors are the pre-existing `web/dist/static/...` proc-macro panics unrelated to this change.

helix-nine:
Trimmed the verbose comments in `4962d11ba`.

helix-nine:
Posted a brief follow-up on the PR pointing at the trim commit (`4962d11ba`).

https://github.com/Start9Labs/start-os/pull/3192

PR #3193, opened 2026-04-29
fix(setup): deduplicate install-os calls via shared future

dr-bonez:
## Summary

- Wraps `install_os` in a `Shared<BoxFuture>` stored on `SetupContext` so retries from the wizard (after a browser HTTP timeout aborts the previous request) attach to the running install instead of starting a parallel one against the same disk.
- Spawns the shared future so it keeps making progress even if every awaiter drops.
- Bumps the `setup.install-os` web client timeout to 5 minutes so the common case completes on a single request.
- Adds `Clone` to `SetupInfo` so the cached result can be returned to all awaiters.
- Drops a stale fixed-values table from the physical device test plan.

## Test plan

- [ ] Boot the setup wizard, start an OS install, simulate a browser HTTP timeout (e.g. throttle/disconnect network briefly mid-install) and retry — verify the second `setup.install-os` returns the same `SetupInfo` rather than racing a fresh install
- [ ] Verify a single happy-path install completes within the 5-minute timeout window
- [ ] Verify dropping all awaiters mid-install (closing the browser tab) doesn't abort the install — wizard reopened later sees the completed result

### DIFF
--- a/core/src/os_install/mbr.rs
+++ b/core/src/os_install/mbr.rs
@@ -8,15 +8,11 @@ use crate::os_install::partition_for;
 use crate::os_install::quiesce::{quiesce_disk, update_partition_table};
 use crate::prelude::*;
 
-/// Boot partition starts at sector 2048 and ends at this sector (exclusive).
-/// 2 GiB at 512-byte sectors.
+/// Boot end (exclusive). 2 GiB at 512-byte sectors.
 const BOOT_END_SECTOR: u32 = 4196352;
-/// Target end sector for the OS root partition (i.e. start of the data region in the
-/// happy path). 14 GiB of root + 2 GiB of boot.
+/// Happy-path root end. 14 GiB of root after boot.
 const ROOT_TARGET_END_SECTOR: u32 = 33556480;
-/// Minimum acceptable end sector for the OS root partition. We let the protected
-/// partition push root's end up to this floor, but no further. 12 GiB at 512-byte
-/// sectors = 25_165_824 sectors of root.
+/// Floor for elastic-root shrink. 12 GiB of root after boot.
 const ROOT_MIN_END_SECTOR: u32 = BOOT_END_SECTOR + 12 * 1024 * 1024 * 2;
 
 pub async fn partition(
@@ -37,9 +33,6 @@ pub async fn partition(
         }
     }
 
-    // Drop every kernel-side reference to partitions on this disk before rewriting the
-    // table. partx --update afterwards is what actually makes the kernel see the new
-    // layout, and it tolerates a still-busy protected partition.
     quiesce_disk(disk_path).await?;
 
     let disk_path = disk_path.to_owned();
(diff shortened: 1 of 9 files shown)


dr-bonez:
@helix-nine please test this using a VM

helix-nine:
Tested in a qemu/UEFI VM (4 CPU, 4 GiB, two 32 GiB virtio disks) with a dev ISO from `da817acb6`. All 4 scenarios pass:

1. Happy path — completes in 30 s.
2. Re-install over a previous install — completes in 24 s (the new "drop stale `install_rootfs` guards" prelude is what unblocks this).
3. Two concurrent calls 1 s apart — both return identical `SetupInfo`, total wallclock 25 s (one install).
4. First call aborted via `curl -m 5`, second re-issued — second attaches to in-flight future, total wallclock 26 s.

Partitions land on 1 MiB boundaries with no slack, confirming the alignment commit.

Not exercised: the protect-existing-data-partition shrink path. After a `wipe=true` install the data partition has only `LINUX_LVM` (no `STARTOS_` GUID until `setup.execute`), so the `protect` selector returns `None` on re-install. Can add a synthetic-disk test for that if you want.

core/src/os_install/mbr.rs
```diff
@@ -66,28 +81,35 @@ pub async fn partition(
...
+        {
+            if starting_lba < ROOT_MIN_END_SECTOR {
                 return Err(Error::new(
                     eyre!(
                         concat!(
-                            "Protected partition {} starts at sector {}",
-                            " which would be overwritten by OS partitions ending at sector {}"
+                            "Protected partition {} starts at sector {}, leaving less ",
+                            "than the minimum room for the OS root partition (must end ",
+                            "at or before sector {}, minimum is sector {})"
```
dr-bonez: @helix-nine  this error needs i18n

core/src/os_install/mod.rs
```diff
@@ -418,12 +421,50 @@ pub async fn install_os_to(
...
+                return existing.clone();
+            }
+        }
+        let ctx = ctx.clone();
+        let new_fut = async move { install_os_inner(ctx, params).await.map_err(Arc::new) }
+            .boxed()
+            .shared();
+        // Drive the future even if all RPC callers drop their await
+        // (e.g. browser HTTP timeout aborts the request).
+        tokio::spawn(new_fut.clone());
```
dr-bonez: @helix-nine  just have the shared future itself contain a non-detaching join handle. that way the types are more correct-by-construction. no installs can leak, and there's always a reference on the context to any running install, but the install still runs in the background.

helix-nine:
_Queued — will be delivered after the current turn finishes._

helix-nine:
Both review comments were addressed before merge:

- `mbr.rs` error i18n → b186707e2 (`fix(os-install): i18n elastic-root error messages`) — wraps the elastic-root errors with translation keys instead of bare strings.
- Shared future owns a non-detaching join handle → 4ae8f0e60 (`fix(setup): own install task via NonDetachingJoinHandle inside Shared`) — the `Shared<BoxFuture>` now wraps a `NonDetachingJoinHandle`, so the install task is always owned by the `SetupContext` reference; no detached spawns, no leaks, install still progresses if every awaiter drops.

helix-nine:
The PR was already merged. My two follow-up commits addressing the inline review comments did make it onto the PR before merge:

- `b186707e2 fix(os-install): i18n elastic-root error messages` — addresses the `mbr.rs` i18n comment
- `4ae8f0e60 fix(setup): own install task via NonDetachingJoinHandle inside Shared` — addresses the "shared future should own a non-detaching join handle" comment

Let me post a brief confirmation comment so the thread reflects that both points were handled.

Posted a confirmation comment on PR https://github.com/Start9Labs/start-os/pull/3193 noting that both inline review comments were addressed before merge:

- **`mbr.rs` i18n** → commit `b186707e2`, wraps the elastic-root errors with translation keys.
- **Non-detaching join handle inside the shared future** → commit `4ae8f0e60`, the `Shared<BoxFuture>` now owns a `NonDetachingJoinHandle` so the install task is always referenced by `SetupContext` (no detached spawns, can't leak), while still progressing if all awaiters drop.

PR was merged by @dr-bonez at 16:40 UTC today.

PR #3194, opened 2026-04-29
fix(net/acme): honor Retry-After after failed ACME orders

helix-nine:
## Problem

`AcmeTlsHandler::get_cert` deduplicates concurrent ACME orders via a shared future stored in `in_progress`, but a **failed** future is immediately replaced by a fresh order attempt on the very next call:

```rust
if let Some(fut) = map.get(san_info).cloned() {
    if fut.peek().map_or(true, |f| f.is_some()) { return fut; }
    // Some(None) — failed — falls through and starts a new order
}
```

Every TLS handshake landing on an ACME-managed SNI calls `get_cert`. Combined with the above, a single transient `429 Too Many Requests` from the ACME server turns into a self-sustaining retry storm: failed order → next inbound TLS handshake fires another order → 429 → … The retry rate is bounded only by inbound traffic, so the rate-limit window can never expire and the cert never issues. While that's happening, every TLS handshake also pays the multi-RTT cost of the failing ACME round-trip.

## Fix

Restructure the per-SAN state from a bare `Shared<Future>` into an `OrderEntry` that tracks both the in-flight order and a single cooldown timestamp. On failure, arm the cooldown from the server's `Retry-After` header when present (only `429 Too Many Requests` carries one), otherwise a fixed 60s fallback that's just long enough to break the per-connection retry loop.

While inside the cooldown window, `get_cert` returns `None` immediately so the `ChainedHandler` falls through to a self-signed cert from `RootCaTlsHandler` — inbound TLS handshakes stay fast while the ACME issue is investigated.

The order future writes back into its own `OrderEntry` on completion (clears `in_flight`, arms or resets the cooldown), so callers arriving after a future has resolved see fresh state without needing a separate cleanup task. Server-supplied `Retry-After` is capped at 24h so a misbehaving directory can't indefinitely silence cert issuance.

## Upstream change in `async-acme`

Surfacing `Retry-After` required a small change in [`Start9Labs/async-acme#1`](https://github.com/Start9Labs/async-acme/pull/1): a new non-breaking `AcmeError::RateLimited { retry_after: Option<Duration> }` variant for HTTP 429 responses. `jose_req` reads `Retry-After` from the response before discarding the body and parses the delta-seconds form (HTTP-date is reported as `None`). Other non-2xx responses continue to use `HttpStatus(u16)` unchanged.

This PR also **moves the dependency from `dr-bonez/async-acme` to `Start9Labs/async-acme`**, pinning to the commit that has the change. Bringing the fork into the org so iteration doesn't depend on an individual contributor's account.

Drive-by: lifted the cached-cert lookup out of a `loop` it was never actually iterating.

## Tests

- `cargo check --lib -p start-os`: clean.
- `cargo test --lib -p start-os net::acme`: 7/7 pass (5 new tests cover Retry-After extraction from `RateLimited`, the 24h cap, the no-Retry-After fallback, non-429 HTTP statuses, and non-HTTP errors).
- `cargo test --lib -p start-os net::`: 68/68 pass.

## Notes

- Background-driven renewal (so an idle host actually retries after the cooldown without inbound traffic) is intentionally out of scope — flagged for a separate change.
- Independent of #3191 (vhost passthrough handler refactor) but solves a related pain point. Together they're the difference between "ACME failure makes every TLS handshake on this server slow forever" and "ACME failure has no effect on TLS handshake latency until the operator fixes it".

### DIFF
--- a/core/src/net/acme.rs
+++ b/core/src/net/acme.rs
@@ -2,9 +2,10 @@ use std::collections::{BTreeMap, BTreeSet};
 use std::net::IpAddr;
 use std::str::FromStr;
 use std::sync::Arc;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
-use async_acme::acme::{ACME_TLS_ALPN_NAME, Identifier};
+use async_acme::acme::{ACME_TLS_ALPN_NAME, AcmeError, Identifier};
+use async_acme::rustls_helper::OrderError;
 use clap::Parser;
 use clap::builder::ValueParserFactory;
 use futures::future::{BoxFuture, Shared};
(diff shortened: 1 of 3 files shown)


core/src/net/acme.rs
```diff
@@ -39,13 +40,55 @@ use crate::util::sync::{SyncMutex, Watch};
...
+/// transient `429 Too Many Requests` into a self-sustaining retry storm,
+/// because a failed `Shared` future was immediately replaced by a new one
+/// the next time anyone called `get_cert`.
+#[derive(Default)]
+pub struct OrderEntry {
+    /// The currently in-flight order, if any. Cleared by the order future
+    /// itself when it resolves (success or failure).
+    in_flight: Option<Shared<BoxFuture<'static, Option<CertifiedKey>>>>,
+    /// If set, no new order may be started before this instant. Cleared
+    /// on success. On failure: set to the server-supplied `Retry-After`
```
dr-bonez: the whole entry is removed on success, no?

dr-bonez (changes requested):
comments are excessively verbose. trim them down.

helix-nine:
Pushed [`64b6fad`](https://github.com/Start9Labs/start-os/pull/3194/commits/64b6fad2b). Two changes:

- **Dropped the success-side update in the order future.** You're right — on success the cached-cert path on the next `get_cert` call removes the entry via `map.remove(san_info)`. The future now only touches the entry on failure (to arm the cooldown). Doc on `OrderEntry` notes this so a future reader doesn't reintroduce the redundant clear.
- **Trimmed the comments** throughout (`-87 / +40` lines net). Kept just enough to anchor the design (why the cooldown exists, why the cap, why `Action::Backoff` returns `None`); dropped the historical/storyline prose. Tests `net::acme` still 7/7 green.

dr-bonez (approved):

helix-nine:
Thanks for the review!

PR #3196, opened 2026-04-29
feat(core): drive setup mode from start-cli for unattended bring-up

helix-nine:
## Why

A freshly booted StartOS install ISO can't be brought to a clean baseline
state purely over the network today: the setup wizard's API endpoints are
all marked `.no_cli()`, so the only thing that can drive them is the
embedded web UI in setup mode. That blocks unattended bring-up — CI
runners, automated test fixtures, fleet provisioning, and the helix VM
template flow all need to drive setup with no human at the device.

This PR teaches `start-cli` to drive setup mode end-to-end against a
remote IP, so the unattended path is:

```bash
# 1. Boot ISO, peer host gets its IP.
# 2. Inspect.
start-cli --host <ip> setup status
start-cli --host <ip> setup disk list

# 3. Install OS to disk (uses /run/live/medium/live/filesystem.squashfs).
start-cli --host <ip> setup install-os /dev/vda --data-drive /dev/vda

# 4. Finalise (or `setup attach` if reusing an existing data drive).
start-cli --host <ip> setup execute --guid <GUID> --password <PW>

# 5. Wait for completion, then commit.
start-cli --host <ip> setup status      # poll until Complete
start-cli --host <ip> setup complete
start-cli --host <ip> setup exit        # reboots into installed OS
```

## What

Three commits, one capability per commit, behaviour over JSON-RPC
unchanged so the web UI keeps working:

1. **expose read-only and lifecycle setup commands to CLI** — `status`,
   `disk list`, `get-pubkey`, `complete`, `exit`, `restart`, `shutdown`,
   `set-language`, `set-keyboard`. Pure plumbing: drop `.no_cli()` and
   add `.with_call_remote::<CliContext>()`, with `.with_display_serializable()`
   for endpoints that return structured data.

2. **expose `setup install-os` to CLI** — `InstallOsParams` already
   derives `clap::Parser`, just needed the CLI binding. Also drops
   `#[group(skip)]` from the inner `DataDrive` struct, which clap
   rejects when the wrapper field is `Option<DataDrive>` and uses
   `#[command(flatten)]` (panics at parse time as soon as the command
   becomes reachable from CLI).

3. **expose `setup execute` and `setup attach` to CLI; accept plaintext
   passwords** — switches `Option<EncryptedWire>` to `Option<PasswordType>`
   on both. `PasswordType` is the existing untagged enum used by `auth`:
   it deserialises either a plain string or an `EncryptedWire` JWE blob,
   so the web UI's encrypted payloads keep working unchanged. The CLI
   passes plaintext over the LAN (setup mode is unauthenticated and runs
   plain HTTP regardless — the JWE was a half-measure, not a security
   boundary). Recovery flows stay JSON-only via `#[arg(skip)]` on
   `recovery_source`; typical fresh-install is a single CLI call.

## Validation

`cargo check` and `cargo build --bin start-cli` clean. End-to-end
validation on a libvirt VM driven from a peer host is in progress; will
report follow-ups as separate commits if more gaps surface.

## Notes for reviewers

- Setup mode has only CORS middleware, not auth, so no signature/cookie
  dance is needed for the CLI side.
- `PasswordType: FromStr` already exists, so clap parses `--password
  <plain>` correctly. JSON input from web UI still routes through serde.
- I deliberately did not touch `verify_cifs` (CIFS recovery params) —
  it's still `EncryptedWire`-only, since CIFS recovery isn't on the
  unattended bring-up critical path. Happy to follow up if desired.

### DIFF
--- a/core/src/setup.rs
+++ b/core/src/setup.rs
@@ -8,6 +8,8 @@ use josekit::jwk::Jwk;
 use patch_db::json_ptr::ROOT;
 use rpc_toolkit::yajrc::RpcError;
 use rpc_toolkit::{Context, Empty, HandlerExt, ParentHandler, from_fn_async};
+
+use crate::context::CliContext;
 use serde::{Deserialize, Serialize};
 use tokio::io::AsyncWriteExt;
 use tokio::process::Command;
@@ -42,7 +44,7 @@ use crate::system::{KeyboardOptions, SetLanguageParams, save_language, sync_kios
 use crate::util::Invoke;
 use crate::util::crypto::EncryptedWire;
 use crate::util::io::{Counter, create_file, dir_copy, dir_size, read_file_to_string};
-use crate::util::serde::{IoFormat, Pem};
+use crate::util::serde::{HandlerExtSerde, IoFormat, Pem};
 use crate::{DATA_DIR, Error, ErrorKind, MAIN_DATA, PACKAGE_DATA, PLATFORM, ResultExt};
 
 pub fn setup<C: Context>() -> ParentHandler<C> {
@@ -51,7 +53,9 @@ pub fn setup<C: Context>() -> ParentHandler<C> {
             "status",
             from_fn_async(status)
                 .with_metadata("authenticated", Value::Bool(false))
-                .no_cli(),
+                .with_display_serializable()
+                .with_about("about.setup-status")
+                .with_call_remote::<CliContext>(),
         )
         .subcommand("disk", disk::<C>())
         .subcommand("attach", from_fn_async(attach).no_cli())
@@ -61,14 +65,28 @@ pub fn setup<C: Context>() -> ParentHandler<C> {
         )
         .subcommand("execute", from_fn_async(execute).no_cli())
         .subcommand("cifs", cifs::<C>())
-        .subcommand("complete", from_fn_async(complete).no_cli())
+        .subcommand(
+            "complete",
+            from_fn_async(complete)
+                .with_display_serializable()
+                .with_about("about.setup-complete")
+                .with_call_remote::<CliContext>(),
+        )
         .subcommand(
             "get-pubkey",
             from_fn_async(get_pubkey)
                 .with_metadata("authenticated", Value::Bool(false))
-                .no_cli(),
+                .with_display_serializable()
+                .with_about("about.setup-get-pubkey")
+                .with_call_remote::<CliContext>(),
         )
-        .subcommand("exit", from_fn_async(exit).no_cli())
+        .subcommand(
+            "exit",
+            from_fn_async(exit)
+                .no_display()
+                .with_about("about.setup-exit")
+                .with_call_remote::<CliContext>(),
+        )
         .subcommand("logs", crate::system::logs::<SetupContext>())
         .subcommand(
             "logs",
@@ -76,10 +94,34 @@ pub fn setup<C: Context>() -> ParentHandler<C> {
                 .no_display()
                 .with_about("about.display-os-logs"),
         )
-        .subcommand("restart", from_fn_async(restart).no_cli())
-        .subcommand("shutdown", from_fn_async(shutdown).no_cli())
-        .subcommand("set-language", from_fn_async(set_language).no_cli())
-        .subcommand("set-keyboard", from_fn_async(set_keyboard).no_cli())
+        .subcommand(
+            "restart",
+            from_fn_async(restart)
+                .no_display()
+                .with_about("about.setup-restart")
+                .with_call_remote::<CliContext>(),
+        )
+        .subcommand(
+            "shutdown",
+            from_fn_async(shutdown)
+                .no_display()
+                .with_about("about.setup-shutdown")
+                .with_call_remote::<CliContext>(),
+        )
+        .subcommand(
+            "set-language",
+            from_fn_async(set_language)
+                .no_display()
+                .with_about("about.setup-set-language")
+                .with_call_remote::<CliContext>(),
+        )
+        .subcommand(
+            "set-keyboard",
+            from_fn_async(set_keyboard)
+                .no_display()
+                .with_about("about.setup-set-keyboard")
+                .with_call_remote::<CliContext>(),
+        )
 }
 
 pub fn disk<C: Context>() -> ParentHandler<C> {
@@ -87,7 +129,9 @@ pub fn disk<C: Context>() -> ParentHandler<C> {
         "list",
         from_fn_async(list_disks)
             .with_metadata("authenticated", Value::Bool(false))
-            .no_cli(),
+            .with_display_serializable()
+            .with_about("about.setup-disk-list")
+            .with_call_remote::<CliContext>(),
     )
 }
 
--- a/core/src/setup.rs
+++ b/core/src/setup.rs
@@ -527,12 +527,33 @@ pub struct SetupExecuteCliParams {
     hostname: Option<InternedString>,
 }
 
-fn read_password_env_or_prompt() -> Result<String, RpcError> {
+/// Read a password for an existing data drive: `PASSWORD` env first,
+/// otherwise prompt once on the TTY.
+fn read_password_env_or_prompt(prompt: &str) -> Result<String, RpcError> {
     if let Ok(p) = std::env::var("PASSWORD") {
         Ok(p)
     } else {
-        Ok(rpassword::prompt_password("Password: ")?)
+        Ok(rpassword::prompt_password(prompt)?)
     }
+}
+
+/// Read a *new* password (for fresh setup): `PASSWORD` env if set,
+/// otherwise prompt twice on the TTY and confirm. Mirrors
+/// `cli_reset_password` in `auth.rs`.
+fn read_new_password_env_or_prompt() -> Result<String, RpcError> {
+    if let Ok(p) = std::env::var("PASSWORD") {
+        return Ok(p);
+    }
+    let new_password = rpassword::prompt_password(&t!("auth.prompt-new-password"))?;
+    let confirm = rpassword::prompt_password(&t!("auth.prompt-confirm"))?;
+    if new_password != confirm {
+        return Err(crate::Error::new(
+            eyre!("{}", t!("auth.passwords-do-not-match")),
+            crate::ErrorKind::IncorrectPassword,
+        )
+        .into());
+    }
+    Ok(new_password)
 }
 
 fn print_remote_result(res: imbl_value::Value) -> Result<(), RpcError> {
@@ -565,7 +586,7 @@ async fn cli_attach(
     let password = if no_password && std::env::var_os("PASSWORD").is_none() {
         None
     } else {
-        Some(read_password_env_or_prompt()?)
+        Some(read_password_env_or_prompt("Data drive password: ")?)
     };
 
     let res = ctx
@@ -597,7 +618,7 @@ async fn cli_execute(
         ..
     }: HandlerArgs<CliContext, SetupExecuteCliParams>,
 ) -> Result<(), RpcError> {
-    let password = read_password_env_or_prompt()?;
+    let password = read_new_password_env_or_prompt()?;
 
     let res = ctx
         .call_remote::<SetupContext>(
--- a/core/src/setup.rs
+++ b/core/src/setup.rs
@@ -18,7 +18,7 @@ use tracing::instrument;
 use ts_rs::TS;
 
 use crate::account::AccountInfo;
-use crate::auth::write_shadow;
+use crate::auth::{PasswordType, write_shadow};
 use crate::backup::restore::recover_full_server;
 use crate::backup::target::BackupTargetFS;
 use crate::bins::set_locale;
@@ -43,6 +43,8 @@ use crate::shutdown::Shutdown;
 use crate::system::{KeyboardOptions, SetLanguageParams, save_language, sync_kiosk};
 use crate::util::Invoke;
 use crate::util::crypto::EncryptedWire;
+
+use clap::Parser;
 use crate::util::io::{Counter, create_file, dir_copy, dir_size, read_file_to_string};
 use crate::util::serde::{HandlerExtSerde, IoFormat, Pem};
 use crate::{DATA_DIR, Error, ErrorKind, MAIN_DATA, PACKAGE_DATA, PLATFORM, ResultExt};
@@ -58,7 +60,13 @@ pub fn setup<C: Context>() -> ParentHandler<C> {
                 .with_call_remote::<CliContext>(),
         )
         .subcommand("disk", disk::<C>())
-        .subcommand("attach", from_fn_async(attach).no_cli())
+        .subcommand(
+            "attach",
+            from_fn_async(attach)
+                .with_display_serializable()
+                .with_about("about.setup-attach")
+                .with_call_remote::<CliContext>(),
+        )
         .subcommand(
             "install-os",
             from_fn_async(crate::os_install::install_os)
@@ -66,7 +74,13 @@ pub fn setup<C: Context>() -> ParentHandler<C> {
                 .with_about("about.setup-install-os")
                 .with_call_remote::<CliContext>(),
         )
-        .subcommand("execute", from_fn_async(execute).no_cli())
+        .subcommand(
+            "execute",
+            from_fn_async(execute)
+                .with_display_serializable()
+                .with_about("about.setup-execute")
+                .with_call_remote::<CliContext>(),
+        )
         .subcommand("cifs", cifs::<C>())
         .subcommand(
             "complete",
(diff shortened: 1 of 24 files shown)


helix-nine:
Pushed a 4th commit covering one more gap I hit during validation:

**`feat(cli): add --insecure / --root-ca for talking to a freshly set-up StartOS`** (1c092c6)

Right after `setup complete` returns, the installed OS comes up on HTTPS with a self-signed CA whose root cert has just been handed back. `start-cli`'s reqwest Client trusts only system roots and had no flag to pin a custom CA or skip verification, so the very next call (`auth login` on the running server) failed with a TLS error. That breaks the unattended hand-off between setup mode (plain HTTP :80) and the installed OS (HTTPS :443).

Adds:
- `--root-ca <PEM_PATH>` — repeatable, calls `Client::add_root_certificate`. Right answer for production: pin the `rootCa` returned by `setup complete`.
- `--insecure` — `danger_accept_invalid_certs(true)`. For ephemeral test fixtures (CI, helix-vm template baselines).

Both new fields are `#[serde(default)]` so existing `config.yaml`s keep deserialising.

### End-to-end run that reaches "ready to install packages"

With this branch built, against a freshly-booted x86_64 ISO under libvirt:

```
$ start-cli --host http://192.168.122.214 setup status
{ "status": "needs-install" }

$ start-cli --host http://192.168.122.214 setup disk list
[{ "logicalname": "/dev/vda", "capacity": 34359738368, ... }]

$ start-cli --host http://192.168.122.214 setup install-os /dev/vda --data-drive /dev/vda --wipe
{ "guid": "STARTOS_5N4...", "attach": false, "mokEnrolled": false }

$ start-cli --host http://192.168.122.214 setup execute --guid STARTOS_5N4... \
    --password testpassword123 --hostname helix-startos-test
{ "progress": ..., "guid": "..." }

$ start-cli --host http://192.168.122.214 setup status
{ "status": "complete", "hostname": "helix-startos-test", "rootCa": "-----BEGIN CERT...", "needsRestart": true }

$ start-cli --host http://192.168.122.214 setup complete
{ "hostname": "helix-startos-test", "rootCa": "...", "needsRestart": true }

$ start-cli --host http://192.168.122.214 setup exit         # reboots into installed OS
$ # ... reboot, then on installed OS ...
$ PASSWORD=testpassword123 start-cli --host https://192.168.122.214 --insecure auth login
$ start-cli --host https://192.168.122.214 --insecure server time
+--------+-------------------------------------+
|  NOW   | 2026-04-29T22:31:29.559079940+00:00 |
+--------+-------------------------------------+
| UPTIME | 7 minutes, 31 seconds               |
```

That's an end-to-end unattended flow with no web UI and no human prompt, which is what was missing.

### Things I deliberately did **not** change

- `verify_cifs` still takes `EncryptedWire`-only — recovery from CIFS isn't on the unattended bring-up critical path.
- `setup install-os` panicked with `cannot #[flatten] an Option<Args> with #[group(skip)]` the first time it became reachable from CLI; fixed by dropping `group(skip)` on the inner `DataDrive`. Worth a quick eyeball to confirm the change is fine for the existing JSON-RPC consumers.

### Open design question (not in this PR)

`setup install-os` is currently `flash-os <SQUASHFS> <DISK>` for the local CLI version, and `setup install-os <OS_DRIVE> --data-drive <DISK>` for the network version (uses the live ISO's mounted squashfs). The naming/argument shape divergence is a bit awkward — happy to unify it in a follow-up if there's an opinion on whether the local version should also default to `/run/live/medium/live/filesystem.squashfs` when run on a live ISO, or whether they should stay distinct.

helix-nine:
End-to-end validation passed cold, including release + re-acquire from a finalized template:

```
$ helix-vm new-template startos-test --iso ...iso --memory 4096 --cpus 2 --disk-gb 32
$ start-cli --host http://192.168.122.159 setup install-os /dev/vda --data-drive /dev/vda --wipe
{ "guid": "STARTOS_A2EQNNOGUBTLVH6NGAQ464LETVTJMWCC", "attach": false, "mokEnrolled": false }
$ start-cli --host http://192.168.122.159 setup execute --guid STARTOS_A2EQ... \
    --password testpassword123 --hostname helix-baseline
$ start-cli --host http://192.168.122.159 setup status     # poll → "complete"
$ start-cli --host http://192.168.122.159 setup complete
$ start-cli --host http://192.168.122.159 setup exit
$ # virsh reset (see gotcha below)
$ PASSWORD=testpassword123 start-cli --host https://192.168.122.159 --insecure auth login
$ start-cli --host https://192.168.122.159 --insecure server time
+--------+-------------------------------------+
|  NOW   | 2026-04-30T00:33:43.775750713+00:00 |
| UPTIME | 47 seconds                          |
+--------+-------------------------------------+

$ helix-vm finalize-template startos-test
$ helix-vm release && helix-vm startos-test       # re-acquire from baseline
$ # NEW slot, NEW IP 192.168.122.143
$ PASSWORD=testpassword123 start-cli --host https://192.168.122.143 --insecure auth login
$ start-cli --host https://192.168.122.143 --insecure server time
+--------+-------------------------------------+
|  NOW   | 2026-04-30T00:36:06.917839143+00:00 |
| UPTIME | 1 minutes, 12 seconds               |
+--------+-------------------------------------+
```

Same password, same root CA, fresh slot, no human input — that's the unattended baseline this work was aiming at. Ready for review.

### Two design questions surfaced during validation (separate from this PR)

1. **`setup exit` doesn't reliably reboot.** The handler sends a `Shutdown` to a tokio channel that only `start_init.rs` listens on; if the live system has moved past that loop, the signal is dropped and the VM keeps running the live ISO. End user has to `virsh reset` to actually transition into the installed OS. Worth either making `setup exit` wait for the channel to drain (and fall back to `systemctl reboot`), or documenting the requirement explicitly.

2. **First-boot grub-install isn't `--removable`.** On UEFI, the install writes its boot entry into NVRAM via efibootmgr but doesn't put a fallback at `/EFI/BOOT/BOOTX64.EFI`. Any consumer that captures the disk image without the per-VM NVRAM (helix-vm template baselines, image exports, dd-style clones) ends up with a disk that drops to the UEFI shell. `grub-install --removable` would write the fallback path and avoid that whole class of issue, at the cost of slightly weirder secure-boot semantics. Happy to follow up either way.

Neither is part of this PR's scope; flagging them here so they don't get lost.

### helix-vm issues I reported separately (not start-os bugs)

- `qemu:///system` + disks under `/home` (mode 700) → libvirt-qemu can't traverse, requires ACLs.
- `helix-vm finalize-template` doesn't merge external snapshots → baseline empty if you snapshotted during build.
- UEFI templates lose boot entries on finalize (per-VM NVRAM not captured).
- `helix-vm new-template` `waitForAccess` blocks longer than the agent harness allows; recoverable via direct virsh, but worth surfacing.

These are tracked in `memories/start-os/MEMORY.md` and will get reported against the helix repo separately.

helix-nine:
Pushed a follow-up to drop `--password` and read the password from `PASSWORD` env (matching `auth login`'s convention), per review feedback. Passwords on argv leak via `ps`/shell history and shouldn't have been the default.

Commit `57bef1e9` on this branch:

- Server-side handlers and JSON-RPC params are unchanged — `AttachParams.password` and `SetupExecuteParams.password` keep `Option<PasswordType>` so the web UI's encrypted-wire form still works.
- `setup execute` / `setup attach` are split: `from_fn_async(execute).no_cli()` for the server path, plus sibling `cli_execute` / `cli_attach` handlers (same subcommand name) that take a slimmer params struct (no `password`), read `PASSWORD` env / fall back to `rpassword::prompt_password`, and forward via `ctx.call_remote::<SetupContext>`.
- `setup attach` gains `--no-password` for the `_UNENC` data drive case where there genuinely is no password.

Re-validated end-to-end against a freshly booted setup-mode VM:

```
$ PASSWORD=envtest1 start-cli --host http://192.168.122.18 setup execute \
    --guid STARTOS_IPGD3UABHDK7Z3BDTBKCN3PY7Z5EE6MJ --hostname env-test
{ "progress": {...}, "guid": "..." }
$ start-cli --host http://192.168.122.18 setup status
{ "status": "complete", ... }

$ start-cli setup execute --password foo --guid X --hostname Y
error: unexpected argument '--password' found

$ unset PASSWORD; </dev/null start-cli --host http://<ip> setup execute --guid X
General error: No such device or address (os error 6)   # rpassword: no TTY, exits clean
```

The earlier validated unattended pipeline now reads:

```bash
PASSWORD=$PW start-cli --host http://$IP setup execute --guid "$GUID" --hostname helix-baseline
PASSWORD=$PW start-cli --host https://$IP --insecure auth login
```

helix-nine:
Polished the interactive path per review — TTY fallback was already there via `rpassword::prompt_password`, but `setup execute` is setting a *new* password and should confirm by re-prompting (matching `cli_reset_password`).

Commit `49f88ff0` on this branch:

- `cli_execute` now prompts twice with confirmation when PASSWORD env isn't set, reusing the existing i18n keys `auth.prompt-new-password` / `auth.prompt-confirm` / `auth.passwords-do-not-match`. Mismatch returns a clean `Incorrect Password: Passwords do not match`.
- `cli_attach` keeps a single prompt ("Data drive password: ") since it's an existing password.
- PASSWORD env still short-circuits both, so unattended runs are unchanged.

Validated under a real PTY (expect(1)):

```
$ start-cli --host http://192.168.122.19 setup execute --guid STARTOS_… --hostname tty-test
New Password: ********
Confirm:      ********
{ "progress": {...}, "guid": "..." }

$ start-cli --host http://192.168.122.20 setup execute --guid STARTOS_… --hostname xyz
New Password: abc
Confirm:      different
Incorrect Password: Passwords do not match
```

So: **`PASSWORD=… start-cli setup execute …`** for unattended/CI, **`start-cli setup execute …`** on a real terminal prompts you twice and confirms, **piped stdin without env** errors cleanly via rpassword instead of hanging.

helix-nine:
Good catch — manpages are checked-in regenerated artifacts (`core/man/start-cli/`, generated by the `export_manpage_start_cli` test in `core/src/bins/start_cli.rs`). Regenerated and committed in `c29f900f`:

- 12 new files for the setup subcommands exposed in this PR (`setup-attach`, `setup-complete`, `setup-disk-list`, `setup-execute`, `setup-exit`, `setup-get-pubkey`, `setup-install-os`, `setup-restart`, `setup-set-keyboard`, `setup-set-language`, `setup-shutdown`, `setup-status`).
- `start-cli.1` updated for the new global `--root-ca` / `--insecure` flags.
- `start-cli-setup.1` lists the new subcommands.
- `start-cli-setup-disk.1` reflects the `list` subcommand.
- 3 unrelated drifts under `start-cli-registry-os-asset-get-*.1` where current clap_mangen now emits `[default: .]` for `--download`. Not from this PR but caught by the regen — happy to revert if you'd rather keep them out of scope, but leaving them in keeps `core/man/` consistent with the regen output.

Re #3196's `os_install::DataDrive` change (dropping `#[group(skip)]`): I confirmed the `start-cli-setup-install-os.1` page still renders cleanly with `--data-drive`, `--wipe`, and the positional `OS_DRIVE`. The reason it's still fine is that `DataDrive` is private to the module and only ever appears flattened into `InstallOsParams` (which itself keeps `#[group(skip)]`); clap_mangen renders the flattened fields against the parent group either way.

Regenerated via:
```
cargo test --manifest-path core/Cargo.toml --lib --features test 'export_manpage_start_cli'
```

helix-nine:
Pushed `2cbbc7b8` — actually fixed this rather than just confirming the manpage happens to look fine.

Restored `#[group(skip)]` on `os_install::DataDrive` and used the same split-CLI pattern that's already in place for `setup execute` / `setup attach` / `auth login`:

- Server-side `InstallOsParams` (with `#[group(skip)]` and `Option<DataDrive>`) is untouched, JSON-RPC contract unchanged.
- New `cli_install_os` + `SetupInstallOsCliParams` in `setup.rs` give the CLI a flat clap surface (`--data-drive: Option<PathBuf>`, no nested struct, no Option<flatten> trap), and the handler builds the JSON body explicitly: `{ osDrive, dataDrive: { logicalname, wipe } | omitted }`.
- `--data-drive` is now genuinely optional in CLI (the previous version required it because clap couldn't see through the Option<flatten>), which matches what JSON-RPC always supported.
- Regenerated `start-cli-setup-install-os.1`; every other manpage was already correct.

Validated:
```
$ start-cli --host http://192.168.122.21 setup install-os /dev/vda --data-drive /dev/vda --wipe
{ "guid": "STARTOS_3AFISG44MOBYAFTPH5YY3UMF63U3CD5E", "attach": false, "mokEnrolled": false }

$ cargo test --manifest-path core/Cargo.toml --lib --features test 'export_manpage_start_cli'
test bins::start_cli::export_manpage_start_cli ... ok

$ start-cli setup install-os --help
Usage: start-cli setup install-os [OPTIONS] <OS_DRIVE>
Arguments:
  <OS_DRIVE>  Path to the OS drive
Options:
      --data-drive <DATA_DRIVE>  Path to the data drive (omit for an OS-only install)…
      --wipe                     Wipe the data drive before use
```

Now every `Parser`-derived struct in this PR carries `#[group(skip)]`, matching bcdeabfe's invariant.

helix-nine:
Sorry — retargeted to `master`. The misleading signal was `CONTRIBUTING.md` line 42:

```
git clone --recursive https://github.com/Start9Labs/start-os.git --branch next/major
```

I read that as "next/major is the active dev branch, target it for PRs." Looking at recent PRs, every one of them targets `master` — mine was the outlier. My branch was actually based on `master` HEAD (`50df13936`), so no rebase was needed, just a base change (`PATCH /repos/.../pulls/3196` with `base=master`). GitHub now reports mergeable = true.

Worth fixing CONTRIBUTING.md to not start the dev setup with a `next/major` checkout, or at least add a "PRs target master" line — happy to send a one-liner if useful.

dr-bonez (changes requested):
you forgot to add i18n for newly added i18n keys

helix-nine:
Pushed.

@dr-bonez right, sorry — should have caught that when the help text was rendering `about.setup-*` literally. Fixed in `2dc38481`:

- Reused existing keys where the semantics already match: `setup get-pubkey` → `about.get-pubkey-from-server`, `setup set-language` → `about.set-language`, `setup set-keyboard` → `about.set-keyboard`.
- Added the remaining 9 with full `en_US` / `de_DE` / `es_ES` / `fr_FR` / `pl_PL` coverage, slotted alphabetically around `about.setup-acme-certificate`:
  - `about.setup-attach`, `about.setup-complete`, `about.setup-disk-list`, `about.setup-execute`, `about.setup-exit`, `about.setup-install-os`, `about.setup-restart`, `about.setup-shutdown`, `about.setup-status`.
- Regenerated `core/man/start-cli/`; the description lines now read e.g. `Display the current state of the setup wizard` instead of `about.setup-status`. `start-cli setup --help` confirms.

I went with conservative phrasings on de/es/fr/pl, modelled on neighbouring entries (`about.setup-acme-certificate`, `about.shutdown-server`, `about.set-language`). If a native speaker on the team wants to tighten any of them I'm happy to fold in suggestions, or you can adjust on merge.

core/src/setup.rs
```diff
@@ -414,13 +499,201 @@ pub async fn setup_data_drive(
 #[ts(export)]
 pub struct SetupExecuteParams {
     guid: InternedString,
-    password: Option<EncryptedWire>,
-    recovery_source: Option<RecoverySource<EncryptedWire>>,
+    password: Option<PasswordType>,
```
dr-bonez: I'd actually prefer we properly support encrypted wire from the cli

dr-bonez (changes requested):

helix-nine:
Pushed `242ea3243`.

@dr-bonez agreed — the CLI now does the same JWE round-trip the web UI does, and the server-side params/handlers are reverted to exactly what they were on `master` before this PR.

- Added `EncryptedWire::encrypt(plaintext, &Jwk)` in `core/src/util/crypto.rs` — josekit ECDH-ES + A128CBC-HS256, flattened-JSON, producing the same `{ encrypted: { protected, iv, ciphertext, tag } }` shape `EncryptedWire::decrypt` already accepts. Round-trip unit test (`test_encrypt_decrypt_roundtrip`) over empty / ASCII / multibyte UTF-8.
- Reverted `AttachParams.password` / `SetupExecuteParams.password` back to `Option<EncryptedWire>` (and `RecoverySource<EncryptedWire>`), and the corresponding `attach` / `execute` decrypt blocks back to the original Option-style. `PasswordType` import gone from `setup.rs`. JSON-RPC contract is now byte-identical to what the web UI already speaks.
- `cli_attach` / `cli_execute` now read the password (PASSWORD env or interactive prompt, with confirmation for execute), then call `setup.get-pubkey` to fetch the recipient JWK, encrypt locally via `EncryptedWire::encrypt`, and send the wire blob. `--no-password` on `setup attach` still skips both the prompt and the get-pubkey call for the `_UNENC` drive case.

Validated end-to-end against a freshly booted setup-mode VM:

```
$ start-cli --host http://192.168.122.22 setup get-pubkey
{ "kty": "EC", "crv": "P-256", "x": "otA9ur-…", "y": "3hMLdntM…" }

$ PASSWORD=jwetestpw1 start-cli --host http://192.168.122.22 \
      setup execute --guid STARTOS_3KG46QMT7L4UOIQ6VM4GHYUUG263T2F6 --hostname jwe-test
{ "progress": ..., "guid": "..." }

$ start-cli --host http://192.168.122.22 setup status
{ "status": "complete", ... }
```

Plaintext password now never leaves the local process — whatever's on the wire is the same JWE payload an unmodified setup-wizard browser would have sent.

dr-bonez (approved):

helix-nine:
Thanks @dr-bonez! Appreciate the careful reviews — both the env-var convention and the proper JWE round-trip are clear improvements over what I first proposed.

PR #3197, opened 2026-04-29
Feat/3155 preferences in services tab filtering

Blu-J:
Phase 1: ServicesPreferencesService with Persistence
 - [ ] localStorage.getItem('_startos/services/preferences') returns expected JSON after calling setSort() and setQuery() in the console

 Phase 2: Sort Persistence End-to-End
 - [ ] Click a column header to sort → navigate away from Services tab → navigate back → table is sorted by the previous column
 - [ ] Close browser, reopen → sort is restored from localStorage

 Phase 3: Filter UI End-to-End
 - [ ] Type in search box → table filters in real-time
 - [ ] Click lifecycle toggles → only matching services shown
 - [ ] Clear search + reset toggles → all services visible
 - [ ] Navigate away and back → filter state restored

### DIFF
--- a/web/projects/shared/src/i18n/dictionaries/de.ts
+++ b/web/projects/shared/src/i18n/dictionaries/de.ts
@@ -399,7 +399,6 @@ export default {
   425: 'Ausführen',
   426: 'Aktion kann nur ausgeführt werden, wenn der Dienst',
   427: 'Verboten',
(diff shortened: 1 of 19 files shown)


web/projects/shared/src/i18n/dictionaries/de.ts
Blu-J, line 402: Need to not have this change...

PR #3202, opened 2026-05-02
fix(cifs): use vers=3 to support older SMB servers

helix-nine:
## Summary

- Change the CIFS mount option from `vers=3.1.1` to `vers=3` in `core/src/disk/mount/filesystem/cifs.rs` so the kernel negotiates any SMB 3.x dialect (3.0 / 3.0.2 / 3.1.1) instead of pinning to 3.1.1.
- This lets StartOS mount shares from older SMB servers that don't speak 3.1.1, while still using SMB3 (no fallback to insecure SMB1/2).

## Test plan

- [ ] Verify backups/restore against an SMB 3.0 server (older NAS)
- [ ] Verify backups/restore still work against an SMB 3.1.1 server


MattDHill (approved):

helix-nine:
Thanks for the review, @MattDHill! 🎉

PR #3203, opened 2026-05-02
fix(net/acme): validate contact entries on init

helix-nine:
## Summary

- ACME `contact` entries must be RFC 8555 URIs, and Let's Encrypt only accepts the `mailto:` scheme. The web UI prepends `mailto:` for the user, but the `net acme init` API/CLI stored whatever the caller passed verbatim, so a bare email made it into the DB and later blew up the ACME `newAccount` request with `urn:ietf:params:acme:error:unsupportedContact`.
- Validate each contact at the `init` handler — must parse as a URL, scheme `mailto`, and the path must contain `@`. Bad input now returns `InvalidRequest` instead of being silently persisted.
- Per-the-ask, no migration/coercion of existing entries: only one user is hitting this and they can re-run `start-cli net acme init --provider <…> --contact mailto:<…>` (or omit `--contact` entirely) to overwrite.

New i18n key `acme.invalid-contact` covers all 5 locales.

## Test plan

- [x] `cargo check -p start-os --lib`
- [x] `cargo test -p start-os --lib net::acme::` — 5 new `validate_contact_*` cases pass alongside the existing retry-after suite

### DIFF
--- a/core/src/net/acme.rs
+++ b/core/src/net/acme.rs
@@ -580,10 +580,39 @@ pub struct InitAcmeParams {
     pub contact: Vec<String>,
 }
 
+// RFC 8555 §7.3 requires `contact` entries to be URLs, and Let's Encrypt
+// (the only provider start-os ships shorthand for) only supports `mailto:`.
+// Without this gate, a bare email here gets stored verbatim and eventually
+// fails the `newAccount` request with `unsupportedContact`.
+fn validate_contact(contact: &str) -> Result<(), Error> {
+    let url = contact.parse::<Url>().map_err(|_| {
+        Error::new(
+            eyre!("{}", t!("acme.invalid-contact", contact = contact)),
+            ErrorKind::InvalidRequest,
+        )
+    })?;
+    if url.scheme() != "mailto" {
+        return Err(Error::new(
+            eyre!("{}", t!("acme.invalid-contact", contact = contact)),
+            ErrorKind::InvalidRequest,
+        ));
+    }
+    if !url.path().contains('@') {
+        return Err(Error::new(
+            eyre!("{}", t!("acme.invalid-contact", contact = contact)),
+            ErrorKind::InvalidRequest,
+        ));
+    }
+    Ok(())
+}
+
 pub async fn init(
     ctx: RpcContext,
     InitAcmeParams { provider, contact }: InitAcmeParams,
 ) -> Result<(), Error> {
+    for c in &contact {
+        validate_contact(c)?;
+    }
     ctx.db
         .mutate(|db| {
             db.as_public_mut()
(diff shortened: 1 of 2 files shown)


core/src/net/acme.rs
```diff
@@ -580,10 +580,39 @@ pub struct InitAcmeParams {
...
+            ErrorKind::InvalidRequest,
+        )
+    })?;
+    if url.scheme() != "mailto" {
+        return Err(Error::new(
+            eyre!("{}", t!("acme.invalid-contact", contact = contact)),
+            ErrorKind::InvalidRequest,
+        ));
+    }
+    if !url.path().contains('@') {
```
dr-bonez: Use a regex for the email

core/src/net/acme.rs
```diff
@@ -580,10 +580,39 @@ pub struct InitAcmeParams {
...
+    }
+    Ok(())
+}
+
 pub async fn init(
     ctx: RpcContext,
     InitAcmeParams { provider, contact }: InitAcmeParams,
 ) -> Result<(), Error> {
+    for c in &contact {
+        validate_contact(c)?;
```
dr-bonez: Try prepending mailto on failure

dr-bonez (changes requested):

core/src/net/acme.rs
helix-nine: Done — `EMAIL_REGEX = ^[^@\s]+@[^@\s]+\.[^@\s]+$` (same `lazy_static` + `regex::Regex` shape as `id` and `wifi`). Now also rejects no-TLD addresses, whitespace, double-`@`. See 93a3d3f.

core/src/net/acme.rs
helix-nine: Done — added `normalize_contact`: validates as-given first, retries with `mailto:` prepended on failure, only errors if both fail. Bare emails now silently canonicalize to `mailto:<email>` at the `init` boundary. Tests cover passthrough, fallback, and unrecoverable garbage.

helix-nine:
Both review comments addressed in 93a3d3f (force-pushed after rebase onto latest `master`):

- **Regex email check** — replaced `path().contains('@')` with a `lazy_static` `EMAIL_REGEX = ^[^@\s]+@[^@\s]+\.[^@\s]+$`, matching the pattern `id`/`wifi` already use. Now also catches no-TLD, whitespace, and double-`@`.
- **`mailto:` auto-prefix** — new `normalize_contact` tries the contact as given first; on failure it retries with `mailto:` prepended and uses that if it validates. Only errors if both fail. Bare emails are silently canonicalized at the `init` boundary, so the API/CLI is now as forgiving as the web UI.

`cargo test -p start-os --lib net::acme::` — all 15 tests green (5 new `normalize_contact_*` and tightened `validate_contact_*`).

core/src/net/acme.rs
helix-nine: Done — switched to `lazy_static` + `regex::Regex` (`^[^@\s]+@[^@\s]+\.[^@\s]+$`), consistent with how `id` and `wifi` validate email-like fields elsewhere in the crate. It tightens the old `contains('@')` check to also reject no-TLD addresses, whitespace, and double-`@`.

dr-bonez (approved):

helix-nine:
Thanks for the review and quick approval, @dr-bonez! 🙏

PR #3201, opened 2026-05-02
chore: bump StartOS to 0.4.0-beta.8 and SDK to 1.3.4

helix-nine:
## Summary

Releases **StartOS 0.4.0-beta.8** and **SDK 1.3.4**.

### `bump version to beta.8`

Standard version bump per `docs/VERSION_BUMP.md`:

- `core/Cargo.toml` / `core/Cargo.lock` → `0.4.0-beta.8`
- New `core/src/version/v0_4_0_beta_8.rs` migration module (no-op `up`/`down`); `core/src/version/mod.rs` registry updated in all 5 spots, `// VERSION_BUMP` marker moved
- `web/package.json` / `web/package-lock.json` → `0.4.0-beta.8` (lockfile regenerated via `npm install`)

### `bump sdk version to 1.3.4`

- `sdk/package/package.json` → `1.3.4`; lockfile regenerated via `npm install`
- `container-runtime/package-lock.json` → embedded `../sdk/dist` version updated to `1.3.4`
- `sdk/package/lib/StartSdk.ts` → `OSVersion` raised from `0.4.0-beta.6` to `0.4.0-beta.8`
- `sdk/CHANGELOG.md` → new `1.3.4` entry

The SDK surface picked up the reverse-proxy auth gate that landed in master via #3190:

- `ProxyAuth` and `BasicCredential` re-exported from `interfaces/Host`
- New optional `auth: ProxyAuth | null` on `AddSslOptions` — set it on an HTTPS bind to put a `Bearer` or `Basic` gate in front of the upstream service. The OS validates `Authorization` before forwarding and (for `Basic`) injects the matched username as `X-Forwarded-User`. Setting `auth` implies HTTP-aware proxying. Optional `realm` controls the `WWW-Authenticate` challenge

The min-OS bump is required because services using `auth` need the OS reverse proxy on beta.8 to honor it.

## Test plan

- [ ] `make` builds clean (core, web, sdk, container-runtime)
- [ ] `make test-core` passes
- [ ] Existing services (no `auth`) still bind and proxy normally on a beta.8 host
- [ ] A package built with SDK 1.3.4 setting `auth: { type: 'basic', credentials: [...], realm: null }` returns 401 with a `WWW-Authenticate: Basic` challenge to anonymous requests, forwards to upstream with `X-Forwarded-User` once authenticated
- [ ] Bearer flow: `auth: { type: 'bearer', tokens: [...], realm: null }` accepts matching `Authorization: Bearer <token>` and 401s otherwise
[...]

PR #3214, opened 2026-05-06
fix(core/service, sdk): mirror ExtendedVersion uninit-target cascade for range data versions

MattDHill:
## Summary

Two paired fixes for the cross-flavor data-version dead-end:

### Host (`core/src/service/service_map.rs`)

Mirror the three-branch uninit-target dispatch from the `ExtendedVersion` arm of `ServiceMap::install` for range-typed on-disk data versions:

1. **`data_ver_range.intersects(next.canMigrateFrom)`** — hand the data through (analog of `data_ver_ev.satisfies(next.canMigrateFrom)`).
2. **`next_version.satisfies(prev.canMigrateTo)`** — target the concrete `next_version`.
3. **otherwise** — target `prev.canMigrateTo AND next.canMigrateFrom` (meet in the middle).

Diff is now structurally a 1:1 of the `ExtendedVersion` arm directly above, with `data_ver_ev.satisfies(...)` swapped for `data_ver_range.intersects(...)`.

Bumps `exver` from `0.2.0` → `0.2.1` (Start9Labs/exver-rs#2, now on crates.io), which adds `VersionRange::satisfiable()` and `intersects()` — the Rust crate's smart `and`/`or`/`not` constructors only fold `Any`/`None` and don't detect contradictions like `Flavor(#knots) AND Flavor(null)`, so the range arm's branch 1 (intersection test) needed the new primitive.

### SDK (`sdk/package/lib/version/VersionGraph.ts`)

Tighten `VersionGraph.uninit`'s guard on unsatisfiable range targets — throw with a "host contract violation" message instead of silently returning. The silent skip masked host bugs that produced contradictory targets, leaving stale data on disk for the destination's init to choke on. Throwing surfaces the violation at the source with a useful message. Bumps SDK 1.4.1 → 1.4.2.

## Background

Reproduced via Knots ↔ bip-110 round-trip on the Bitcoin packages. After one round-trip the on-disk data version was a flavored range (e.g. `^#knots:29:0`); a subsequent Knots → Core switch then failed with `cannot migrate from ^#knots:29:0 to 29.3:10`.

Walking the chain:

1. `service_map.rs` (pre-PR) computed `target = AND(data_ver_range, next.can_migrate_from)`. With `data_ver_range = ^#knots:29:0` (flavored) and `next.can_migrate_from` flavor-less (Core's), that AND has no satisfying version — flavor-less ranges only match `flavor === null` versions and flavored ranges only match their own flavor; `ExtendedVersion::compare` returns `null` ("incomparable") on flavor mismatch and every comparison helper treats `null` as `false`.
2. The SDK's `VersionGraph.uninit` short-circuited on `!target.satisfiable()`, returning before invoking the registered `migrations.other.down` (the Knots → Core migration the source package owns) and before writing the data version. The on-disk value stayed at the stuck flavored range — and there was no signal to the host that anything had gone wrong.
3. Core's `init` then read that range as `from`, found no graph vertex overlapping it, and `migrate()` threw at the destination — far from the source-side root cause.

The host fix prevents the bad target from being computed; the SDK fix ensures any future host-side bug producing a similar bad target gets surfaced loudly at the source instead of repeating the silent-skip → destination-throw pattern.

## Out of scope (dropped from earlier revisions)

- The first revision had `VersionGraph.init` always rewrite the on-disk data version to the concrete current version after init completed. @drbonez flagged that as semantically suspect — the data version is the data **shape**, not the running version, and a range left on disk by a `migrations.down` edge legitimately means "the data is in a shape any version in this range can read". Forcing it to concrete would mislead future migrations. Reverted.
- A second revision deleted the range arm entirely, falling through to the parse-failure path. That mirrored only the last two branches of the `ExtendedVersion` cascade, dropping the passthrough case. The current revision restores it as branch 1 via `intersects()`.

### DIFF
--- a/core/src/service/service_map.rs
+++ b/core/src/service/service_map.rs
@@ -324,22 +324,17 @@ impl ServiceMap {
                                 .can_migrate_to;
                             let next_version = &s9pk.as_manifest().version;
                             let next_can_migrate_from = &s9pk.as_manifest().can_migrate_from;
-                            if let Ok(data_ver_ev) = data_ver.parse::<exver::ExtendedVersion>() {
-                                if data_ver_ev.satisfies(next_can_migrate_from) {
-                                    ExitParams::target_str(data_ver)
-                                } else if next_version.satisfies(prev_can_migrate_to) {
-                                    ExitParams::target_version(&s9pk.as_manifest().version)
-                                } else {
-                                    ExitParams::target_range(&VersionRange::and(
-                                        prev_can_migrate_to.clone(),
-                                        next_can_migrate_from.clone(),
-                                    ))
-                                }
+                            let next_accepts_data_ver = if let Ok(data_ver_ev) =
+                                data_ver.parse::<exver::ExtendedVersion>()
+                            {
+                                data_ver_ev.satisfies(next_can_migrate_from)
                             } else if let Ok(data_ver_range) = data_ver.parse::<VersionRange>() {
-                                ExitParams::target_range(&VersionRange::and(
-                                    data_ver_range,
-                                    next_can_migrate_from.clone(),
-                                ))
+                                data_ver_range.intersects(next_can_migrate_from)
+                            } else {
+                                false
+                            };
+                            if next_accepts_data_ver {
+                                ExitParams::target_str(data_ver)
                             } else if next_version.satisfies(prev_can_migrate_to) {
                                 ExitParams::target_version(&s9pk.as_manifest().version)
                             } else {
--- a/container-runtime/package-lock.json
+++ b/container-runtime/package-lock.json
@@ -37,7 +37,7 @@
     },
     "../sdk/dist": {
       "name": "@start9labs/start-sdk",
-      "version": "1.3.4",
+      "version": "1.4.2",
       "license": "MIT",
       "dependencies": {
         "@iarna/toml": "^3.0.0",
@@ -50,7 +50,7 @@
         "isomorphic-fetch": "^3.0.0",
         "mime": "^4.1.0",
         "yaml": "^2.8.3",
-        "zod": "^4.3.6",
+        "zod": "4.3.6",
         "zod-deep-partial": "^1.2.0"
       },
       "devDependencies": {
@@ -58,7 +58,6 @@
         "jest": "^30.0.0",
         "peggy": "^3.0.2",
         "prettier": "^3.8.3",
-        "shx": "^0.4.0",
         "ts-jest": "^29.4.9",
         "ts-node": "^10.9.1",
         "ts-pegjs": "^4.2.1",
@@ -111,7 +110,6 @@
       "integrity": "sha512-l+lkXCHS6tQEc5oUpK28xBOZ6+HwaH7YwoYQbLFiYb4nS2/l1tKnZEtEWkD0GuiYdvArf9qBS0XlQGXzPMsNqQ==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@ampproject/remapping": "^2.2.0",
         "@babel/code-frame": "^7.26.2",
@@ -1202,7 +1200,6 @@
       "dev": true,
       "hasInstallScript": true,
       "license": "Apache-2.0",
-      "peer": true,
       "dependencies": {
         "@swc/counter": "^0.1.3",
         "@swc/types": "^0.1.17"
@@ -2146,7 +2143,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "caniuse-lite": "^1.0.30001688",
         "electron-to-chromium": "^1.5.73",
@@ -3994,7 +3990,6 @@
       "integrity": "sha512-NIy3oAFp9shda19hy4HK0HRTWKtPJmGdnvywu01nOqNC2vZg+Z+fvJDxpMQA88eb2I9EcafcdjYgsDthnYTvGw==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@jest/core": "^29.7.0",
         "@jest/types": "^29.6.3",
@@ -6555,7 +6550,6 @@
       "integrity": "sha512-84MVSjMEHP+FQRPy3pX9sTVV/INIex71s9TL2Gm5FG/WG1SqXeKyZ0k7/blY/4FdOzI12CBy1vGc4og/eus0fw==",
       "dev": true,
       "license": "Apache-2.0",
-      "peer": true,
       "bin": {
         "tsc": "bin/tsc",
         "tsserver": "bin/tsserver"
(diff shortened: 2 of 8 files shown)


helix-nine:
Restructured per the review.

- Dropped the OR-fallback. The range arm now collapses into the parse-failure path, leaving the same two branches as the bottom of the `ExtendedVersion` arm directly above (concrete-`next_version` when allowed by `prev.canMigrateTo`, else `prev.canMigrateTo AND next.canMigrateFrom`). Diff is now `-5 / +0` in `service_map.rs`.
- Reverted the `VersionGraph.init` change (and the SDK 1.4.2 / minimum-StartOS bump that was tied to it). Agreed the always-normalize behavior was overstepping: a range left on disk by `migrations.down` is a real description of the data shape, and rewriting it to concrete current would mislead future migrations.
- Branch 1 of the `ExtendedVersion` arm (passthrough when data is already in a shape `next` can read) isn't mirrored for ranges — doing so cleanly needs a `satisfiable()` / intersection method on `exver::VersionRange` that the Rust crate doesn't currently expose. Targeting concrete `next_version` is correct in the same scenarios, just slightly less efficient when the AND was already satisfiable. Happy to add `satisfiable()` to `exver-rs` as a follow-up if you want the passthrough arm too.

Updated PR description has the full diagnosis chain.

helix-nine:
Pushed the proper three-branch mirror.

The range arm now reads as a structural 1:1 of the `ExtendedVersion` arm directly above, swapping `data_ver_ev.satisfies(...)` for `data_ver_range.intersects(...)`:

```rust
} else if let Ok(data_ver_range) = data_ver.parse::<VersionRange>() {
    if data_ver_range.intersects(next_can_migrate_from) {
        ExitParams::target_str(data_ver)                              // (1) handoff
    } else if next_version.satisfies(prev_can_migrate_to) {
        ExitParams::target_version(&s9pk.as_manifest().version)       // (2) concrete next
    } else {
        ExitParams::target_range(&VersionRange::and(                  // (3) meet in middle
            prev_can_migrate_to.clone(),
            next_can_migrate_from.clone(),
        ))
    }
}
```

Cross-repo dep: opened **Start9Labs/exver-rs#2** which adds `VersionRange::satisfiable()` and `intersects()`. Ports the truth-table algorithm from the SDK exver — each flavor atom gets its own table over the version number line, AND/OR/NOT operate pointwise via merge-sort-style zip, satisfiable iff some segment in some table is true. Includes the actual reproducer `^#knots:29:0 && <=29.3:10` as a unit test, plus a `witness_implies_satisfiable` proptest for soundness over random formulas. All 22 exver-rs tests pass.

`core/Cargo.toml` is currently pinned to my exver-rs branch (`helix/d5QEC52Ho4Tu`, commit `16c842f`). Rebase onto `master` once exver-rs#2 merges.

Open question still on the table from earlier in the thread: the SDK's `VersionGraph.uninit` silently returns on `!target.satisfiable()`. With this fix the host won't produce unsatisfiable targets in the reproduced case, but branch 3 (`prev.canMigrateTo AND next.canMigrateFrom`) can still be empty for some cross-flavor scenarios where `next_version` isn't in `prev.canMigrateTo`. Want me to also tighten that guard (drop the silent return so `migrate()` throws naturally) in a separate PR?

helix-nine:
SDK guard tightened. `VersionGraph.uninit` now throws on unsatisfiable range target with a "host contract violation" message instead of silently returning:

```ts
if (isRange(target) && !target.satisfiable()) {
  throw new Error(
    `uninit target range \`${target.toString()}\` is unsatisfiable — no version can satisfy it (host contract violation)`,
  )
}
```

Together with the host-side three-branch fix:
- The reproduced cross-flavor case takes branch 2 (concrete `next_version`), satisfiable, uninit runs the registered down migration, all good.
- If a future host bug or third-party consumer ever computes an empty-AND target (e.g. branch 3 in some unusual cross-flavor topology), the source's uninit throws at the contract violation instead of silently leaving stale data for the destination to choke on.

SDK bumped 1.4.1 → 1.4.2 with a CHANGELOG entry. SDK type-check (`npm run check`) clean.

Final diff:
- `core/src/service/service_map.rs`: +10/-4 (range arm restructure to 3-branch)
- `core/Cargo.toml`: pin to exver-rs PR branch
- `core/Cargo.lock`: regenerate
- `sdk/package/lib/version/VersionGraph.ts`: +3/-1 (silent return → throw)
- `sdk/CHANGELOG.md`, `sdk/package/package.json`, `sdk/package/package-lock.json`: 1.4.1 → 1.4.2

helix-nine:
Repointed at `exver = "0.2.1"` from crates.io (dropped the git source). `cargo check --lib` clean.

helix-nine:
Ran the make targets that touch tracked lockfiles after the SDK bump:

- `make sdk/dist/package.json sdk/baseDist/package.json` — regenerated SDK dist (gitignored, just verifies the bump propagates).
- `make container-runtime/package-lock.json` — refreshed; this one was tracked.
- `make container-runtime/dist/package-lock.json` — gitignored.
- `make web/package-lock.json` — no change (web pins the SDK via `file:../sdk/baseDist`, which is a path-not-version reference).

Only `container-runtime/package-lock.json` actually changed in tree. It had been stale since SDK 1.3.4 (predates this PR), so the diff sweeps up the SDK bumps from #3201/#3205/#3206 alongside the 1.4.1 → 1.4.2 here. Force-pushed.

core/src/service/service_map.rs
dr-bonez: the latter 2 branches on both the Version and VersionRange cases are equivalent to the latter 2 fallback branches. this can probably be simplified

dr-bonez (changes requested):

core/src/service/service_map.rs
helix-nine: Good catch — collapsed.

The data-ver-shape-specific test (`satisfies` for `ExtendedVersion`, `intersects` for `VersionRange`) is the only thing that differs between the two parses, so I hoisted it into a single `next_can_read_data_ver: bool` and let the shared "concrete next or meet-in-middle" tail run once:

```rust
let next_can_read_data_ver = if let Ok(data_ver_ev) =
    data_ver.parse::<exver::ExtendedVersion>()
{
    data_ver_ev.satisfies(next_can_migrate_from)
} else if let Ok(data_ver_range) = data_ver.parse::<VersionRange>() {
    data_ver_range.intersects(next_can_migrate_from)
} else {
    false
};
if next_can_read_data_ver {
    ExitParams::target_str(data_ver)
} else if next_version.satisfies(prev_can_migrate_to) {
    ExitParams::target_version(&s9pk.as_manifest().version)
} else {
    ExitParams::target_range(&VersionRange::and(
        prev_can_migrate_to.clone(),
        next_can_migrate_from.clone(),
    ))
}
```

`cargo check --lib` clean. Force-pushed.

dr-bonez (approved):

helix-nine:
Thanks for the review — over to you for merge.

PR #3216, opened 2026-05-07
fix(net/vhost): bound idle proxy tasks via per-target cap, active cancel, and hyper header_read_timeout

helix-nine:
## Why this PR

Counter-proposal to #3211. That PR addresses a real bug — proxy connections leak when their peer is L4-responsive but L7-idle, e.g. an HTTP backend whose application abandoned the socket but whose TCP stack still ACKs `proxy_keepalive` probes. The reported symptom is 16 660 stuck `ESTABLISHED` sockets pinned in `startd`'s reverse proxy, climbing toward the soft fd limit.

The mechanism #3211 reaches for is a per-connection idle-progress timeout (wrapping each side of `copy_bidirectional` in `TimeoutStream(10 min)`). That makes the proxy the arbiter of "is this protocol still legitimate," and will close real long-lived idle connections — Bisq, Matrix federation, opportunistic pubsub, idle SSH-over-WS, etc. — that intentionally hold sockets open for hours without bytes flowing. We can't impose a per-connection clock on traffic the proxy doesn't understand.

The reframe: the user's actual problem is **fd exhaustion**, not "this connection is too old." Bound the resource, not the connection.

## What changes

Three complementary levers, none of which time-bound an individual connection:

### 1. Per-target connection cap with LRU eviction

`MAX_PROXY_CONNS_PER_TARGET = 4096`. A new `ConnRegistry` is attached to every vhost target. Each spawned proxy task registers on entry and deregisters on exit; when a registration would exceed the cap, the entry with the oldest `last_active` timestamp is cancelled. fd usage is bounded at `cap × distinct targets`, with no per-connection clock applied — a connection that exchanges *any* bytes is preserved over genuinely stuck peers. `ActivityStream` wraps the client-facing socket and ticks `last_active` on every successful read or write; both `copy_bidirectional` and the hyper-aware path ferry every upstream byte through this socket as a write, so wrapping one side captures activity in either direction.

### 2. Active cancellation propagation tied to target lifecycle

The previous `WeakFuture(Weak<()>)` design only re-checks the strong count on the *next poll* of the spawned future, so a parked, L7-idle `copy_bidirectional` never noticed when its target had been removed. A new `ProxyContext` carries a `CancellationToken` whose cancellation is fired when the target is removed from the vhost mapping (`gc()` and the dead-entry sweep in `add()` both call `cancel_dead`). Spawned tasks `tokio::select!` on `target_cancel.cancelled()` (lifecycle-tied) and `conn_cancel` (per-connection child token, used by LRU eviction). Both wake parked tasks immediately rather than waiting for I/O.

This doesn't fix the user's exact symptom (their package was still running), but it kills a whole adjacent class of leaks that today wait for the next byte to clean up after a stop/restart.

### 3. Hyper `header_read_timeout(60s)` on the HTTP-aware path

`hyper::server::conn::http1::Builder::header_read_timeout` is armed only between requests / while waiting for a complete request head, and is disarmed the moment a body or upgrade is in flight. So it can never close an active stream. It bounds the slowloris case and idle keep-alive sockets that would otherwise stay open indefinitely. The HTTP/2 path already has matching `keep_alive_interval` / `keep_alive_timeout` knobs.

For HTTP traffic that hits the hyper-managed path (no upgrade), this is the dominant lever. For SNI passthrough and post-upgrade tunnels, (1) + (2) carry the load.

## What didn't change

- Public API of `VHostController::add` is still `Result<Arc<()>, _>`. The cancel + registry side-table is internal.
- `VHostTarget::handle_stream`'s last parameter changed from `Weak<()>` to `ProxyContext`. Only one impl exists (`ProxyTarget`).
- TCP `proxy_keepalive` (60s / 10s × 6) from beta.7 stays as the L4 silent-peer escape hatch. The three levers here are layered on top, not replacing.
- No config surface, no RPC change, no SDK change.

## Tests

Four new unit tests (in `core/src/net/vhost.rs`):
- `lru_eviction_cancels_only_the_oldest` — fill registry to cap, verify (cap+1)th registration cancels the oldest entry and only that one.
- `activity_stream_keeps_busy_conn_out_of_lru_seat` — a busy connection registered second beats a stale connection registered first when the cap overflows; LRU policy honors byte progress.
- `drop_handle_deregisters` — `ConnRegHandle::Drop` removes the entry, so the registry doesn't over-report size after task exit.
- `target_cancel_wakes_idle_proxy_task` — a parked `copy_bidirectional` between two duplexes wakes within ms once `cancel.cancel()` fires; without the lever it would park indefinitely (companion to the existing `copy_bidirectional_hangs_without_keepalive_when_peer_idle`).

All pre-existing tests preserved. `make test-core` green.

## Test plan

- [x] `make test-core` (CI)
- [ ] On a beta.7-affected production-style box, observe that `ss -tan state established` for the affected proxy port stabilizes at or below `MAX_PROXY_CONNS_PER_TARGET` rather than climbing unbounded.
- [ ] Verify long-lived idle connections (Matrix federation, an idle WebSocket whose heartbeat is ≫ 10 min, Bisq) are NOT closed by the proxy.
- [ ] Stop a package with active proxy connections; verify those connections tear down within a poll instead of waiting for next byte.

## Closes

Supersedes #3211 (different mechanism, same observed bug).

### DIFF
--- a/core/Cargo.lock
+++ b/core/Cargo.lock
@@ -7130,6 +7130,7 @@ dependencies = [
  "bytes",
  "futures-core",
  "futures-sink",
+ "futures-util",
  "pin-project-lite",
  "tokio",
 ]
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -219,7 +219,7 @@ tokio-rustls = "0.26.4"
 tokio-stream = { version = "0.1.14", features = ["io-util", "net", "sync"] }
 tokio-tar = { git = "https://github.com/dr-bonez/tokio-tar.git" }
 tokio-tungstenite = { version = "0.26.2", features = ["native-tls", "url"] }
-tokio-util = { version = "0.7.9", features = ["io"] }
+tokio-util = { version = "0.7.9", features = ["io", "rt"] }
 tower-service = "0.3.3"
 tracing = "0.1.39"
 tracing-error = "0.2.0"
--- a/core/locales/i18n.yaml
+++ b/core/locales/i18n.yaml
@@ -3179,6 +3179,13 @@ help.arg.log-limit:
   fr_FR: "Nombre maximum d'entrées de journal"
   pl_PL: "Maksymalna liczba wpisów logu"
 
+help.arg.max-proxy-conns-per-target:
+  en_US: "Per-target cap on concurrent reverse-proxy connections; LRU evicts beyond this"
+  de_DE: "Obergrenze gleichzeitiger Reverse-Proxy-Verbindungen pro Ziel; LRU verdrängt darüber hinaus"
+  es_ES: "Límite por destino de conexiones simultáneas en el proxy inverso; LRU expulsa al superarse"
+  fr_FR: "Plafond par cible des connexions reverse-proxy simultanées ; LRU évince au-delà"
+  pl_PL: "Limit jednoczesnych połączeń proxy odwrotnego na cel; LRU usuwa po przekroczeniu"
+
 help.arg.merge:
   en_US: "Merge with existing version range instead of replacing"
   de_DE: "Mit vorhandenem Versionsbereich zusammenführen statt ersetzen"
--- a/core/src/context/config.rs
+++ b/core/src/context/config.rs
@@ -147,6 +147,8 @@ pub struct ServerConfig {
     pub multi_arch_s9pks: Option<bool>,
     #[arg(long, help = "help.arg.developer-key-path")]
     pub developer_key_path: Option<PathBuf>,
+    #[arg(long, help = "help.arg.max-proxy-conns-per-target")]
+    pub max_proxy_conns_per_target: Option<usize>,
 }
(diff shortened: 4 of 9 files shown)


dr-bonez (commented):
Ok, this looks pretty good, but I think ActivityStream wrapper breaks support for vectored reads and writes

helix-nine:
Good catch — fixed in 4fc0aa896. `AsyncWrite::poll_write_vectored` and `is_write_vectored` both have defaults (the former serializes through `poll_write`, the latter returns `false`), so wrapping a `TcpStream` in `ActivityStream` was silently dropping the vectored fast path. Now both delegate to the inner stream, and the vectored path also ticks `last_active`.

`AsyncRead` has no vectored counterpart in the tokio trait, so the read side was already a straight passthrough.

dr-bonez (approved):

helix-nine:
Thanks for the review and merge! 🎉

PR #3217, opened 2026-05-07
Retry alternate package mirrors after download failures

schjonhaug:
## Summary
- download registry package assets fully to disk before deserializing/verifying them
- retry the next asset URL when an HTTP body stream fails or ends before its advertised content length
- update the CLI download path now that downloads complete before the source is returned

## Context
Repeated package install/update attempts can fail across different services with:

`Network Error: error decoding response body`

error

This can happen when a registry asset URL responds successfully but the package body fails before the full file has been downloaded. In that case StartOS should try the next mirror URL instead of failing the update immediately.

## Testing
- `cargo test --manifest-path core/Cargo.toml download_to_path_falls_back_after_truncated_mirror --lib`
- `cargo build --manifest-path core/Cargo.toml --bin start-cli`
- `core/target/debug/start-cli --registry https://registry.start9.com/ registry package download bitcoind --target-version '31.0:10' --arch x86_64 --dest /tmp/startos-bitcoind-31.0-10-test.s9pk`
- `core/target/debug/start-cli --cookie-path /tmp/startos-test-cookies.json s9pk inspect /tmp/startos-bitcoind-31.0-10-test.s9pk commitment`

### DIFF
--- a/core/src/registry/asset.rs
+++ b/core/src/registry/asset.rs
@@ -248,9 +248,6 @@ impl ArchiveSource for BufferedHttpSource {
     }
 }
 
-/// Retry policy for `DownloadHandle::download_from`. Walks the mirror list and,
-/// within each mirror, tries: full download → resume via `Range` (if the previous
-/// attempt got truncated mid-stream) → one full retry → next mirror.
 struct MirrorRetry {
     urls: Vec<Url>,
     client: Client,
@@ -260,11 +257,8 @@ struct MirrorRetry {
 
 #[derive(Clone, Copy)]
 enum MirrorState {
-    /// Try the next URL with a fresh full request.
     NextFull(usize),
-    /// The previous attempt on `urls[idx]` failed mid-body. Try a Range resume next.
     TryResume(usize),
-    /// Resume on `urls[idx]` failed (or wasn't applicable). Try a full retry next.
     RetryFull(usize),
 }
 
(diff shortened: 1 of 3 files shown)


dr-bonez:
ACK on the problem, don't fully agree with the solution. I'd probably want it to retry the first url at least once, possibly with a range request for the remaining bytes, before moving onto the next url.  I'd also want to integrate it into `load_buffered_http_source` rather than bypassing that helper

dr-bonez:
do you want to redo your pr with feedback, or would you prefer I open a counter-pr @schjonhaug ?

schjonhaug:
Thanks again for the feedback. I reworked this so the retry behavior now lives in the existing buffered download path instead of bypassing it.

What changed:
- load_buffered_http_source / load_buffered_http_source_with_path now drive the retry logic.
- Each mirror is retried once before moving to the next mirror.
- If the first attempt wrote partial bytes and the total size is known, the retry first attempts a Range request for the remaining bytes.
- If the server ignores the range request with 200 OK, it falls back to a full same-URL retry.
- Only after that same-URL retry fails does it try the next mirror.
- The CLI path is back to waiting for source.wait_for_buffered() before renaming the temp file.

I also added focused tests for range resume, range ignored/full retry, next-mirror fallback, and all mirrors failing.

dr-bonez (approved):

PR #3219, opened 2026-05-07
fix(os-install): preserve picks rightmost free region before data partition

helix-nine:
## Summary

Reported in Matrix: users upgrading from 0.3.5.1 to 0.4.0-beta.8 hit `RPC Error: Disk Management Error Unable to find enough space on drive` whenever they pick **Preserve** in the setup wizard. Workaround so far has been to install an earlier beta first, then update.

I introduced this in #3193 (the elastic-root rewrite of `gpt::partition`). Picking the free region the new root partition will land in used `min()` over `(start, _)` of `find_free_sectors()`. With the new 1 MiB alignment the fresh GPT now has a 2014-LBA sliver between LBA 34 (first_usable) and LBA 2048 (where EFI starts). That sliver wins `min()`, so `available_lba = protect_first_lba - 2048` overshoots the room actually available between boot and the protected partition.

On a 0.3.5.1 EFI install (efi 100M / boot 1G / root 15G, unaligned, data partition at LBA 33759266) `available_lba` becomes 33757218 — larger than the 14 GiB target — so root is sized to 14 GiB, gpt-rs places it at LBA 4401152..33761279 (2014 LBAs *past* `protect_first_lba`), and `add_partition_at("data", first_lba=33759266, ...)` then fails because 33759266 is no longer in any free section. The gpt crate surfaces this as `NotEnoughSpace` → `"Unable to find enough space on drive"`.

The fix is one logical line: take `max()` instead of `min()` so we pick the free region abutting the protected partition (the only one root will actually be placed in). On the 0.3.5.1 EFI layout that gives `available_lba = 29358114` — about 1 MiB under the 14 GiB target — and elastic root shrinks accordingly, leaving the data partition at its original LBAs.

To make this testable without `quiesce_disk` / `partx --update` against a real block device, the synchronous core of `partition()` is split into a `build_gpt_layout` helper that takes any `gpt::DiskDevice`.

- Reproduced the failure end-to-end against a synthesized 0.3.5.1 GPT (32 / 64 / 128 / 256 / 512 GiB sparse files) and confirmed the same error string.
- Added regression test `os_install::gpt::tests::preserve_data_partition_from_0_3_5_1_layout` that lays out the 0.3.5.1-shaped GPT and asserts the rebuilt table keeps `data.first_lba` / `data.last_lba` at their pre-existing values and `root.last_lba < data.first_lba`. Verified the test fails on `master` (with the original `.min()`) and passes after the fix.

The MBR preserve path is unaffected — `mbr.rs` writes partition entries directly, so it doesn't go through `gpt-rs`'s free-region search and is fine for 0.3.5.1's hardcoded MBR layout.

## Test plan

- [x] `cargo check --lib`
- [x] `cargo test --lib --features=test os_install::gpt::tests::preserve_data_partition_from_0_3_5_1_layout` (new test, passes on fix, fails on master)
- [x] `cargo test --lib --features=test` (325 + 1 new = 326 pass)
- [ ] Build a `0.4.0-beta.8 + this patch` ISO and reinstall over a real 0.3.5.1 EFI disk with Preserve selected. I haven't verified end-to-end on hardware/VM yet — the existing 0.3.5.1 ISO + helix-vm template would need to be wired up. Worth doing before tagging beta.9, but the unit test exercises exactly the geometry the failing reports describe.

### DIFF
--- a/core/src/os_install/gpt.rs
+++ b/core/src/os_install/gpt.rs
@@ -14,6 +14,168 @@ const ROOT_MIN_BYTES: u64 = 12 * 1024 * 1024 * 1024;
 /// 1 MiB at 512-byte sectors.
 const PARTITION_ALIGNMENT_LBA: u64 = 2048;
 
+/// Synchronous core of `partition()`: writes a fresh GPT to `device` placing
+/// efi/bios + boot + root, then either re-creates the protected data partition
+/// at its existing LBAs or fills the trailing free space with a new one.
+///
+/// Split out so unit tests can drive it against a sparse file without going
+/// through `quiesce_disk` / `partx --update`, which need a real block device.
+fn build_gpt_layout(
+    device: Box<dyn gpt::DiskDevice + Send>,
+    capacity: u64,
+    disk_path: &Path,
+    protected_partition_info: Option<(u64, u64, PathBuf)>,
+    use_efi: bool,
+) -> Result<(bool, Option<PathBuf>), Error> {
+    let mut device = device;
+    let mbr = gpt::mbr::ProtectiveMBR::with_lb_size(
+        u32::try_from((capacity / 512) - 1).unwrap_or(0xFF_FF_FF_FF),
+    );
+    mbr.overwrite_lba0(&mut device)?;
+    let mut gpt = GptConfig::new()
+        .writable(true)
+        .logical_block_size(LogicalBlockSize::Lb512)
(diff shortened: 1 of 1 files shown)


dr-bonez (approved):

helix-nine:
Thanks for the review, @dr-bonez! Letting CI finish before merging.

PR #3221, opened 2026-05-07
feat(sdk): sdk.notification.create() lets packages post notifications

helix-nine:
## Summary

- Adds a `notification.create` effect to the service-effects RPC surface (`core/src/service/effects/notification.rs`), backed by the existing `notify()` machinery in `core/src/notifications.rs`. The host forces `package_id` to the calling service's id, so a package can't spoof another package's notifications.
- Accepts `code: 0` (generic, no `data`) and `code: 2` (free-form `data: string`). `code: 1` is reserved for backup-report notifications generated by StartOS and is rejected with `InvalidRequest`.
- Exposes the effect on `Effects.notification.create` (`sdk/base/lib/Effects.ts`) and a typed `sdk.notification.create(effects, options)` wrapper in `sdk/package/lib/StartSdk.ts` whose `options` is a discriminated union over `code` (so `data` is checked at the type level — `code: 0` rejects `data`, `code: 2` requires `data: string`).
- Bumps the SDK to **1.5.0** and adds a CHANGELOG entry. The new effect requires the matching backend handler in this same change, so SDK 1.5.0 won't function against earlier StartOS builds.

## Example

```ts
await sdk.notification.create(effects, {
  code: 0,
  level: 'info',
  title: 'Sync Complete',
  message: 'Initial block download finished.',
})

await sdk.notification.create(effects, {
  code: 2,
  level: 'warning',
  title: 'Disk Usage High',
  message: 'The data volume is 92% full.',
  data: '92%',
})
```

## Test plan

- [x] `cargo check -p start-os --lib` passes
- [x] `cargo build -p start-os --lib` passes
- [x] `make ts-bindings` regenerates `CreateNotificationParams.ts` cleanly
- [x] `sdk/base` and `sdk/package` `tsc --noEmit` pass
- [x] `make test` in `sdk/` passes (42/42 tests)
- [x] `make bundle` in `sdk/` produces `dist`/`baseDist`
- [ ] Manual: install a service that calls `sdk.notification.create({ code: 0, ... })` from `setupMain` and confirm the notification appears in the UI panel attributed to that package
- [ ] Manual: confirm `code: 1` is rejected with `InvalidRequest`
- [ ] Manual: confirm a package can't pass a `packageId` field that gets honored (the host always forces it to the caller)


sdk/package/lib/StartSdk.ts
```diff
@@ -219,6 +224,57 @@ export class StartSdk<Manifest extends T.SDKManifest> {
...
+         *   })
+         * ```
+         * @example
+         * ```
+         *   await sdk.notification.create(effects, {
+         *     code: 2,
+         *     level: 'warning',
+         *     title: 'Disk Usage High',
+         *     message: 'The data volume is 92% full.',
+         *     data: '92%',
```
MattDHill: Code 2 is meant for notifications that have a large amount of markdown-compatible data intended to be shown in a modal when the user click "view details". In this example, "92%" would be displayed completely out of context in a large overlay modal. Fix the example and the TSDoc description

helix-nine: Good catch — fixed in b76f27f.

Replaced the misleading `data: '92%'` example with a release-notes payload, and rewrote the TSDoc on `sdk.notification.create`, the `Effects.notification.create` description, and the ts-rs binding doc-comment to spell out that code 2's `data` is rendered as markdown in the "View Details" modal (release notes, post-update changelogs, structured error reports — long-form content, not a short status string).

No type/runtime changes; just documentation and the example.

sdk/base/lib/osBindings/CreateNotificationParams.ts
```diff
@@ -0,0 +1,20 @@
...
+  level: NotificationLevel
+  title: string
+  message: string
+  /**
+   * Notification code. `0` is a plain notification (no data); `2` carries
+   * a long-form, markdown-compatible body that the UI renders in a
+   * "View Details" modal. `1` is reserved for backup-report notifications
+   * generated by StartOS.
+   */
+  code: 0 | 2
```
dr-bonez: IMO, this is terrible dev UX. we can just key off of the presence of `data` anyway rn, and if we ever need to expose other notification kinds, we should use a string-based enum.

dr-bonez (changes requested):

sdk/base/lib/osBindings/CreateNotificationParams.ts
MattDHill: Agree string enum is better.

helix-nine: Agreed — fixed in 4aebac2. The `code` field is gone from the SDK surface, the `Effects.notification.create` interface, the ts-rs binding, and the wire format entirely.

`sdk.notification.create` now takes `{ level, title, message, data? }`. The Rust handler picks the underlying `NotificationType` from the presence of `data` (none → code 0, some → code 2). If we ever need to expose more notification kinds, we'll add a string-tagged field as an additive change rather than retrofit numeric codes.

The matching CHANGELOG entry, TSDoc, examples, and the start-docs PR ([Start9Labs/start-docs#77](https://github.com/Start9Labs/start-docs/pull/77)) are all updated.

helix-nine: @MattDHill — your comment crossed mine in time. 4aebac2 already dropped the `code` field entirely; the current shape is just `{ level, title, message, data? }` (no discriminator at all — host picks the underlying NotificationType from `data`'s presence). That matches the "rn we can key off `data` anyway" half of @dr-bonez's suggestion.

The other half — "if we ever need to expose other notification kinds, use a string-based enum" — I read as future-tense, contingent on more kinds appearing. Right now we only have one shape that takes data (markdown for the "View Details" modal), so a string enum with a single non-plain value (e.g. `kind: 'details'`) would be noise.

Happy to add it now if you'd rather get the field shape locked in before the SDK 1.5.0 ships, just say the word. Otherwise the plan is: stays as-is now, and we add a `kind` field as an additive change the first time we introduce a second data-bearing notification shape.

helix-nine: Yes — that's exactly how it works. The Rust handler does `match data { None => notify(.., ()), Some(s) => notify(.., s) }`, and `NotificationType` is implemented on `()` with `CODE = 0` and on `String` with `CODE = 2` (from notifications.rs). So no `data` → 0 in the DB row, `data: Some("...")` → 2.

helix-nine: Yes. The handler always sets `package_id` from `context.seed.id` (the calling service's id) — see `core/src/service/effects/notification.rs:21`:

```rust
let package_id = context.seed.id.clone();
…
notify(db, Some(package_id), level, title, message, …)
```

So every notification posted via `sdk.notification.create` lands in the DB with `packageId` populated. The notifications-panel row already reads `item.packageId` (and uses it to look up package data for the icon/name — `web/projects/ui/src/app/routes/portal/routes/notifications/item.component.ts:179`), so these render with the source package attribution alongside the OS-generated ones.

There's no `packageId` field in `CreateNotificationParams`, so a package can't spoof another package's id either — the host always derives it.

dr-bonez (approved):

PR #3224, opened 2026-05-10
feat: required instructions.md per package + Instructions tab

helix-nine:
## Summary

Introduces a new mandatory `instructions.md` file at the root of every StartOS package, packs it into the s9pk archive, exposes it on patchDB at `packageData[id].instructions`, and adds an **Instructions** tab to the service details sidebar (between Dashboard and Actions) that renders it through the existing markdown + dompurify pipeline.

`instructions.md` is *not* a redundant README. The README is an architectural overview for developers and LLMs explaining how the service works on StartOS; instructions are user-facing — pragmatic setup/usage steps, limitations, and useful upstream links. Companion docs (`writing-instructions.md` for packagers and a user-facing explainer) live in [start-docs#80](https://github.com/Start9Labs/start-docs/pull/80).

Because `instructions.md` is now where packagers point users at upstream documentation, the dedicated `docsUrls` array on the package manifest is redundant and is removed in the same PR.

### Core (Rust)
- `start-cli s9pk pack` now requires `instructions.md` at the package root (same resolver shape as `LICENSE.md`); errors out with `NotFound` if missing.
- New `--instructions <path>` CLI override + `help.arg.instructions-path` i18n entry.
- Added to merkle archive priority sort, `s9pk-structure.md`, and the file-tree doc-comment in `s9pk/v2/mod.rs`.
- `S9pk::instructions()` reader returns `Option<String>` (None for s9pks predating this change, so the host stays compatible with already-installed packages).
- `Manifest::validate_for` soft-checks the file (warn-only, like `assets.squashfs`) so existing installs don't fail to deserialize.
- `PackageDataEntry.instructions: String` (`#[serde(default)]`) — populated on install/update from `s9pk.instructions()`. Existing PatchDB rows without the field deserialize to `""`.
- `docs_urls` removed from `PackageMetadata` and the legacy v1→v2 compat shim. `PackageMetadata` doesn't use `deny_unknown_fields`, so already-stored manifests with `docsUrls` deserialize fine — the field is just dropped on next write.

### SDK (TypeScript, `@start9labs/start-sdk-base`)
- `S9pk#instructions(): Promise<string | null>` — parallels `license()` but tolerates missing file for back-compat.
- `docsUrls` removed from `ManifestTypes` and the regenerated ts-rs bindings (`Manifest.ts`, `PackageVersionInfo.ts`).

### Web (Angular)
- New `Instructions` sidebar nav item beneath Dashboard.
- New route + standalone component reading `packageData[id].instructions` reactively from PatchDB and rendering with `MarkdownPipe | NgDompurifyPipe`. When the string is empty/whitespace, falls back to a neutral notification: **"This version has no instructions. Please update."** — chosen so users on beta.9 immediately understand the situation when they open the tab on a pre-update package.
- i18n keys added to all five locale dictionaries (de, en, es, fr, pl).
- Mock fixtures updated so the dev UI shows reasonable content.
- Documentation rows dropped from the service About tab and the marketplace details Links section. Mock fixtures cleaned up.

## Migration / back-compat

- Already-installed packages whose s9pks lack `instructions.md` deserialize fine (`#[serde(default)]` → empty string). The Instructions tab shows the fallback notification for them until they're updated.
- Existing s9pk validation only warns on missing `instructions.md`; it doesn't fail. Strict enforcement is at pack time.
- `docsUrls` on already-stored manifests is silently dropped on the next write — no migration needed.

## Test plan

- [x] `cargo check -p start-os --lib`
- [x] `make ts-bindings` regenerates `PackageDataEntry.ts` (and drops `docsUrls` from `Manifest.ts` / `PackageVersionInfo.ts`)
- [x] `cd sdk && make baseDist dist`
- [x] `cd web && npm run check` (i18n + shared + marketplace + ui + setup-wizard, all clean)
- [x] Pre-commit hook (`lint-staged` running `check:ui` + i18n) passes
- [ ] Manual: pack a service with no `instructions.md` → start-cli should error with the expected message
- [ ] Manual: install/update a service with `instructions.md` and verify the Instructions tab renders it
- [ ] Manual: install/update a service with no `instructions.md` (legacy s9pk) and verify the fallback notification reads "This version has no instructions. Please update."
- [ ] Manual: confirm the About tab and marketplace details no longer show a Documentation row

Companion PRs:
- [start-docs#80](https://github.com/Start9Labs/start-docs/pull/80): `writing-instructions.md` for packagers, user-facing instructions explainer, and `docsUrls` dropped from the manifest example
- `hello-world-startos`: add an `instructions.md` to the template (to follow)


core/src/service/service_map.rs
```diff
@@ -178,6 +178,7 @@ impl ServiceMap {
         let manifest = s9pk.as_manifest().clone();
         let id = manifest.id.clone();
         let icon = s9pk.icon_data_url().await?;
+        let instructions = s9pk.instructions().await?.unwrap_or_default();
```
dr-bonez: Putting the full instructions on patch-db seems not ideal - I thought we were going the other direction, moving icons *off* of patch-db? @MattDHill

helix-nine: Fair concern. I implemented this on patchDB because that's what was explicitly specced in the brief ("a string in a new `instructions` key on patchDB at packageData[package-id].instructions"), but you're right that it mirrors the icon problem — the s9pk already has the file, and we're now duplicating a potentially-large blob into patchDB just so the frontend can read it reactively.

The license-tab pattern is the precedent for going the other way: `api.getStatic(['/s9pk/installed/${pkgId}.s9pk/LICENSE.md'], {})` pulls it lazily when the user opens the tab. The Instructions tab is exactly the same shape — opened on-demand, never needed eagerly. If we move it off patchDB, the Rust side gets simpler too: drop the `PackageDataEntry.instructions` field, drop the read in `service_map.rs::install`, drop the `as_instructions_mut().ser(...)` mutation; just keep `S9pk::instructions()` and the SDK `S9pk#instructions()` (those are still useful for the marketplace browsing existing s9pks before install).

The "back-compat for older s9pks with no instructions.md" question gets simpler too: the static endpoint returns 404 → frontend shows the "This version has no instructions. Please update." fallback.

@MattDHill — happy to pivot if you want the LICENSE-style on-demand approach. Cheaper to change now than after merge.

MattDHill: > Putting the full instructions on patch-db seems not ideal - I thought we were going the other direction, moving icons *off* of patch-db? @MattDHill 

I would expect the instructions to be a few hundred words, seemed reasonable to include.

helix-nine: Different patterns right now:

**License** — *not* on patchDB. Fetched lazily via `api.getStatic(['/s9pk/installed/${pkgId}.s9pk/LICENSE.md'], {})` only when the user clicks the License row on the About tab (see `about.component.ts:63`). This is the pattern @dr-bonez is pointing at.

**Icon** — *on* patchDB, as a base64 `data:` URL. Read once at install time via `s9pk.icon_data_url().await?` (`service_map.rs:180`), serialized into `PackageDataEntry.icon: DataUrl<'static>`. The frontend uses it eagerly in tons of places: the page background on the service details outlet, the sidebar avatar, the dashboard tile, action-confirm and action-input modals, task lists, etc. (~10 call sites in `routes/services/`). That eager-everywhere usage is *why* it lives on patchDB — moving it off means rewiring every one of those consumers to fetch lazily, which is the migration @dr-bonez was alluding to.

**Instructions in this PR** — currently follows the icon pattern (eager, on patchDB) because that's what was specced. But instructions only have *one* consumer (the new tab, opened on demand), so the license pattern is a much cleaner fit. The rewrite is small:

- Drop `PackageDataEntry.instructions` + `#[serde(default)]` + the `as_instructions_mut().ser(...)` mutation in `install()`.
- Drop the `unwrap_or_default()` read in `service_map.rs:181`.
- Drop the `instructions: '...'` entries from the mock fixtures.
- Change `instructions.component.ts` from a `patch.watch$` to a `defer(() => api.getStatic([...]))` pattern, mirroring `about.component.ts`'s license dialog. 404 from the static endpoint → fallback notification.
- Keep `S9pk::instructions()` on host + the SDK `S9pk#instructions()` — those are still useful (e.g. marketplace previewing an uninstalled s9pk).

Want me to make that change? Easy call, ~30 LOC swap.

helix-nine: Pushed the pivot in `deb581aff`. Summary:

- `PackageDataEntry.instructions` removed; `ServiceMap::install` no longer reads or persists it; `as_instructions_mut()` calls gone.
- TS binding regenerated to match.
- New `instructions.component.ts` fetches via `api.getStatic(['/s9pk/installed/${pkgId}.s9pk/instructions.md'], {})` (mirroring how `about.component.ts` opens LICENSE.md). Three states — loading / ready / missing — with 404 from the static endpoint folding into the same "This version has no instructions. Please update." fallback used for empty files.
- `S9pk::instructions()` (host) and `S9pk#instructions()` (SDK) kept — useful for marketplace previews of uninstalled s9pks.
- `Manifest::validate_for` still soft-warns on missing `instructions.md` and pack-time enforcement still errors hard, so back-compat with already-installed older s9pks is unchanged.
- Mock fixtures cleaned up.

`cargo check --lib`, `make ts-bindings`, SDK build, `npm run check`, and the pre-commit hook all pass.

dr-bonez (approved):

PR #3237, opened 2026-05-15
fix(core/disk): skip unmountable partitions when listing backup targets

MattDHill:
## Summary

- Tiny non-data partitions on a backup disk (GPT header, ESP, etc.) were appearing as "Available for backup, 0 GB" rows alongside the real data partition on the same physical disk. The backup-target list flattens each disk into one row per partition, with no filter for whether the partition is actually usable.
- `part_info` now returns `Option<PartitionInfo>`, returning `None` when the read-only mount probe fails; `list()` filters those out.
- The OS-disk LVM-PV path in `list()` is restructured to compute capacity directly via `get_capacity` rather than going through `part_info` — raw PVs can't mount anyway, and we only needed `logicalname` + `capacity` on that branch. This keeps the existing behavior for the embassy data disk.
- i18n key `disk.util.could-not-collect-usage-info` is renamed/reworded to `disk.util.skipping-unmountable-partition` to reflect the new "skip" semantic, across all 5 locales.

### DIFF
--- a/core/locales/i18n.yaml
+++ b/core/locales/i18n.yaml
@@ -985,12 +985,12 @@ disk.util.could-not-get-capacity-part:
   fr_FR: "Impossible d'obtenir la capacité de %{part} : %{error}"
   pl_PL: "Nie można pobrać pojemności %{part}: %{error}"
 
-disk.util.could-not-collect-usage-info:
-  en_US: "Could not collect usage information: %{error}"
-  de_DE: "Konnte Nutzungsinformationen nicht sammeln: %{error}"
-  es_ES: "No se pudo recopilar información de uso: %{error}"
-  fr_FR: "Impossible de collecter les informations d'utilisation : %{error}"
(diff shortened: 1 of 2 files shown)


dr-bonez:
Just because we know the capacity of an LVM PV doesn't mean it has a mountable filesystem on it

MattDHill:
Good catch. The previous patch required every partition in the external-disk branch to mount before being added to `disk.partitions`, which dropped raw LVM PVs (no mountable filesystem on them) — breaking the `os_install`/setup scan for `EMBASSY_`/`STARTOS_` GUIDs.

Pushed `02c4154ec`: pulled the LVM-PV logic from both branches of `list()` into one `lvm_pv_part_info` helper that builds a `PartitionInfo` from capacity + guid + probed LV filesystem without going through the mount probe. The non-PV path still requires mount, which is what filters the tiny header/ESP partitions on the user's Samsung drive.

dr-bonez (approved):

PR #3240, opened 2026-05-16
fix: serve mobileconfig profile to iOS for root CA install

### LINKED ISSUE #3239: [bug]: iOS downloads root CA as .crt.html instead of .crt (wrong Content-Type)
### Prerequisites

### Server Hardware

Bosgame mini PC w/Intel n150 / 16GB RAM / 4TB SSD

### StartOS Version

Version 0.4.0-beta.9

### Client OS

iOS

### Client OS Version

iOS 26.5

### Browser

Safari

### Browser Version

26.5

### Current Behavior

Downloading the root CA to my iPhone downloaded the file by default as  "servername.crt.html".   Obviously then I couldn't get the iPhone to recognize it as a cert because it was a .html file, not a .crt.  As a result, I couldn't get to the "Install Profile" prompt.  

It is a problem because the default in iPhone Files app is to hide extensions, and therefore it looks like the file downloaded and stored on iOS is servername.crt 

A workaround is in iOS go to the Files app.  Go into any folder, such as Downloads. Tap the three-dot button in the upper-right corner. Scroll down and Tap View Options. Turn on Show All Extensions.

### Expected Behavior

The file saves as <servername>.crt and iOS presents the "Install Profile" prompt.

### Steps to Reproduce

  Steps to Reproduce

  1. Run StartOS 0.4.0-beta.9 on a server reachable from an iOS device
  2. Navigate to the StartOS UI from iOS Safari
  3. Tap the link to download the root CA certificate
  4. Observe the saved filename ends in .crt.html (you may need to have show file extensions enabled to see this)
  5. iOS does not offer certificate installation

### Anything else?

_No response_

MattDHill:
## Summary

iOS 26.5 Safari saves the `/static/local-root-ca.crt` download as `<hostname>.crt.html` regardless of `Content-Type` — the prior `application/octet-stream` workaround (#2527 → alpha.17) stopped working in iOS 26.5. Switching back to `application/x-x509-ca-cert` doesn't fix it either; iOS Safari's content sniffer is unreliable across versions, and we've now flipped between MIME types three times trying to chase it.

This PR routes iOS/iPadOS through Apple's intended install path: a `.mobileconfig` Configuration Profile, which iOS handles via the standard Settings → Profile install flow instead of MIME-sniffing a downloaded blob.

- New backend endpoint `GET /static/local-root-ca.mobileconfig` returns an unsigned Configuration Profile with a `com.apple.security.root` payload. UUIDs derive from the cert SHA-256 so re-downloading replaces the prior install rather than duplicating it. The existing `.crt` endpoint still serves the PEM cert for macOS / Linux / Windows / curl.
- New `isIOSSafari()` helper in `@start9labs/shared` (catches iPadOS-pretending-to-be-MacIntel via `maxTouchPoints`). The three live UI download links (about, authorities, ca-wizard) bind their `href` to `ROOT_CA_DOWNLOAD_HREF`, which auto-routes iOS to the new endpoint.
- Removes the setup wizard's `StartOS-info.html` download entirely. The offline HTML embedded the cert for manual install on another device, but iOS can't reliably install a Configuration Profile from a `data:` URL in locally-opened HTML — so rather than keep patching it, the Root CA is now obtained solely from the live dashboard via the routing above. This also drops the hidden `DocumentationComponent`, the now-dead `rootCa` plumbing in the setup APIs, and the orphaned i18n keys.

Profile is unsigned — same trust posture as the self-signed CA itself; user still flips trust in Settings → General → About → Certificate Trust Settings.

Closes #3239


core/src/net/static_server.rs
dr-bonez: this can probably go back to application/x-x509-ca-cert

dr-bonez (changes requested):

core/src/net/static_server.rs
dr-bonez: @helix-nine

helix-nine: Reverted in d56725c7d — `cert_send` is back to `application/x-x509-ca-cert` now that iOS gets the cert via the mobileconfig profile.

dr-bonez (approved):

PR #3243, opened 2026-05-17
fix(web/wifi): give each access point its own dropdown open state

### LINKED ISSUE #3242: [bug]: WiFi access point overflow menu disappears too quickly to interact with
### Prerequisites

### Server Hardware

GMKTEK G3  8GB 

### StartOS Version

0.4.0-beta.9 

### Client OS

Linux

### Client OS Version

Ubuntu Server

### Browser

Brave

### Browser Version

v1.90.122

### Current Behavior

- Go to System → WiFi
- Connect to a WiFi access point
- Tap the three-dot (overflow) menu on the connected access point
- The menu options disappear before you can select any of them

### Expected Behavior

- The menu stays open long enough to select an option
- Actual behaviour: The menu closes/disappears almost instantly

### Steps to Reproduce

- Go to System → WiFi
- Connect to a WiFi access point
- Tap the three-dot (overflow) menu on the connected access point

### Anything else?

Also tried on Chrome Version 148.0.7778.168 (Official Build) (64-bit) and Android

helix-nine:
## Summary

The "More" dropdown on saved-but-disconnected access points shared a single `open` flag across every row rendered by `WifiTableComponent`. Because one component instance renders the entire list, every `TuiDropdown` directive read and wrote the same boolean, so opening one menu attempted to open every menu, and the resulting outside-click detection collapsed them all almost instantly.

- Switch to a per-SSID `opens` map (the same pattern `BackupNetworkComponent` already uses for a shared-component list with multiple dropdowns).
- Close the menu explicitly when an option is selected so the `Forget` action — which doesn't hand focus off to a modal — leaves the dropdown closed.

## Test plan

- [ ] System → WiFi: with two or more saved networks, disconnect from the active one, tap "More" on a saved-disconnected entry — menu stays open until an option is selected or you tap away.
- [ ] Tapping Connect still opens the password / confirm dialog and closes the menu.
- [ ] Tapping Forget removes the entry and closes the menu without leaving a stuck overlay.
- [ ] Repeat on touch (mobile) — the original report covered Brave, Chrome, and Android.

Fixes #3242.

### DIFF
--- a/web/projects/ui/src/app/routes/portal/routes/system/routes/wifi/table.component.ts
+++ b/web/projects/ui/src/app/routes/portal/routes/system/routes/wifi/table.component.ts
@@ -76,18 +76,23 @@ import { wifiSpec } from './wifi.const'
           size="s"
           appearance="flat-grayscale"
           iconStart="@tui.ellipsis-vertical"
-          [(tuiDropdownOpen)]="open"
+          [tuiDropdownOpen]="!!opens[network.ssid]"
+          (tuiDropdownOpenChange)="opens[network.ssid] = $event"
         >
           {{ 'More' | i18n }}
           <tui-data-list *tuiDropdown>
-            <button tuiOption iconStart="@tui.wifi" (click)="prompt(network)">
+            <button
+              tuiOption
+              iconStart="@tui.wifi"
+              (click)="opens[network.ssid] = false; prompt(network)"
+            >
               {{ 'Connect' | i18n }}
             </button>
             <button
               tuiOption
               iconStart="@tui.trash"
               class="g-negative"
-              (click)="forget(network)"
+              (click)="opens[network.ssid] = false; forget(network)"
             >
               {{ 'Forget' | i18n }}
             </button>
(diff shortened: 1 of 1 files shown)


helix-nine:
Followup after auditing dropdown usage across start-os and start-wrt.

The dominant convention in both repos for per-row action menus is the **uncontrolled** `tuiDropdownAuto` pattern, with the close function pulled out of the structural directive's context — e.g. `start-wrt`'s `wifi/passwords`, `profiles`, `inbound/{table,client}`, `published-ports`, and `start-os`'s `start-tunnel/{devices,port-forwards,subnets}` and services `controls.component.ts` all use:

```html
<button tuiDropdown tuiDropdownAuto ...>
  <tui-data-list *tuiDropdown="let close" (click)="close()">
    <button tuiOption (click)="...">...</button>
  </tui-data-list>
</button>
```

`tuiDropdownAuto` gives each rendered button its own internal open state (which is what was actually missing here), and the `let close` context handles dismissal in one line on the data-list. Refactored to match — no `opens` map needed.

(The `[(tuiDropdownOpen)]` / `Record<index, boolean>` style is still present in `BackupNetworkComponent` and `GatewaysItemComponent`; happy to clean those up in a separate PR if useful, but leaving them alone since they work.)

PR #3245, opened 2026-05-19
feat(registry): emit Ed25519-signed webhooks on package and OS mutations

MattDHill:
## Summary

- Adds an outbound webhook system to the registry binary. `package.add`, `package.remove`, `os.version.add`, and `os.version.remove` each emit a JSON event to one configured subscriber after a successful mutation, signed with an Ed25519 keypair the registry generates on first start. The public key is the registry's identity — it rides in the `x-startos-registry-pubkey` header, and consumers maintain an allowlist of trusted pubkeys (no shared secret to coordinate).
- Persists every event and every delivery attempt (status, error, duration) to patch-db. Adds `registry webhook list` and `registry webhook replay <id>` (admin), and `registry webhook pubkey` (unauthenticated — the key is public by definition).
- No automatic retry. Manual replay is the recovery mechanism, reusing the original event id so consumers that dedupe on it treat the replay as an idempotent retry.
- If no subscriber is configured, events are dropped and nothing is persisted. Schema is additive (`#[serde(default)]` on `webhook_log` and `webhook_signing_key`), so existing registry databases deserialize unchanged.

Operator config:

```yaml
webhook:
  url: https://my-announcer.example.com/registry-events
```

Operator publishes the registry's pubkey to a consumer with `start-cli registry webhook pubkey`; the consumer adds it to their allowlist.

Internal architecture lives in `core/registry-webhooks.md`. Operator-facing docs are in a separate `start-docs` PR (Hosting a Registry chapter).


dr-bonez:
@helix-nine I would prefer an enum for RegistryEvents with TS bindings so we can have a central source of truth for what registry events can happen

helix-nine:
@dr-bonez done in b10e88b03. `RegistryEvent` is now an envelope (`id`, `occurredAt`) flattened over a `#[serde(tag = "topic", content = "data")]` `RegistryEventData` enum, with one typed payload struct per topic. The new module `core/src/registry/webhook/event.rs` is the source of truth — variants are `PackageVersionAdd`, `PackageRemove`, `OsVersionAdd`, `OsVersionRemove`.

Wire shape is unchanged (`{ id, occurredAt, topic, data }`) — added a unit test (`os_version_remove_round_trips_with_flat_envelope`) that pins down the flatten + tagged-enum interaction, so a future serde/ts-rs rev can't silently move the keys around.

On the TS side, `RegistryEventData` regenerates as a discriminated union, so subscribers can `switch (e.topic)` and have `e.data` narrowed to the right struct:

```ts
export type RegistryEventData =
  | { topic: 'package.version.add'; data: PackageVersionAddData }
  | { topic: 'package.remove';      data: PackageRemoveData }
  | { topic: 'os.version.add';      data: OsVersionAddData }
  | { topic: 'os.version.remove';   data: OsVersionRemoveData }

export type RegistryEvent = { id: Guid; occurredAt: string } & RegistryEventData
```

Doc in `core/registry-webhooks.md` updated; the "adding a new topic" recipe now points at the enum module.

dr-bonez:
closing in favor of nostr-based protocol

helix-nine:
Acknowledged — closing this work. Happy to pick up the nostr-based equivalent when there's a spec to point at.

Issue #2, opened 2020-11-20
Automatic check for updates. Configurable.


ProofOfKeags:
Keep setting for autocheck for updates. Track dismissal per update. Do not force update.

PR #3246, opened 2026-05-19
feat(sdk): Daemons.dynamic + lazy SubContainer.of + hold/release (2.0.0)

helix-nine:
## Summary

Adds `sdk.Daemons.dynamic(fn)` — a reactive `main` entrypoint whose daemon set is a function of on-disk state — plus the three breaking core changes the dynamic case wanted: lazy `SubContainer.of`, unified `SubContainer` (no Rc), and record-then-materialize `Daemons`. End state: one obvious way to express a daemon topology, static or reactive.

```ts
export const main = sdk.Daemons.dynamic(async ({ effects }) => {
  const { instances } = (await instancesYaml.read().const(effects)) ?? { instances: [] }
  let daemons = sdk.Daemons.of<Manifest>({ effects })
  for (const inst of instances) {
    daemons = daemons.addDaemon(`reg-${inst.id}`, {
      subcontainer: sdk.SubContainer.of(effects, { imageId: 'startos-registry', sharedRun: true },
        sdk.Mounts.of()
          .mountVolume({ volumeId: 'main', subpath: `/instances/${inst.id}/data`, mountpoint: '/var/lib/startos', readonly: false })
          .mountVolume({ volumeId: 'main', subpath: `/instances/${inst.id}/config.yaml`, mountpoint: '/etc/startos/config.yaml', readonly: false, type: 'file' }),
        `reg-${inst.id}-sub`,
      ),
      exec: { command: ['start-registryd'] },
      ready: { display: `Registry: ${inst.label}`, fn: () => sdk.healthCheck.checkPortListening(effects, inst.port, {}) },
      requires: [],
    })
  }
  return daemons
})
```

When `instances.yaml` is rewritten, the SDK reconciles automatically — per id: absent→present **start**, present→absent **stop**, same `configHash` **leave alone**, different `configHash` **restart**. Dependents of any restarted/stopped daemon are also restarted to keep `requires` wiring consistent. Re-runs coalesce while one is in flight. No service restart, no in-package supervisor.

`configHash` covers the subcontainer descriptor (`imageId`, `sharedRun`, `name`, structural `mounts.build()`), exec (`command`, `env`, `cwd`, `user`, `runAsInit`, `sigtermTimeout`), `requires` (sorted), and the structural parts of `ready` (`display`, `gracePeriod`). Closures (`ready.fn`, `ready.trigger`, function-form `exec.fn`) and pre-built `Daemon` instances are intentionally excluded — an unrelated re-run (file touched, content unchanged) must not bounce every daemon. Authors surface a value through one of the hashed fields if they want the reconciler to react to it changing.

### 1. `sdk.SubContainer.of` is lazy by default

Returns a `SubContainerLazy<M>` *synchronously* (no `Promise<>` wrapper); `createFs` is deferred to first method call. The unified `SubContainer<M>` interface widens `rootfs` / `guid` / `subpath()` to `T | Promise<T>`; concrete classes narrow — `SubContainerEager` to `T`, `SubContainerLazy` to `Promise<T>`. A caller holding the generic interface must `await` those accessors; a caller holding `SubContainerEager` keeps sync access. New `sdk.SubContainer.eager(...)` returns `Promise<SubContainerEager<M>>` for the cases that need fail-fast on `createFs` or sync `rootfs` before any method call.

Identity (`Symbol('subcontainer')`) is generated at the lazy handle's construction and preserved across `eager()` materialization, so a lazy handle and the eager it materializes into compare equal via `identity`. The lazy default is what makes `Daemons.dynamic`'s "leave alone is load-bearing" guarantee actually load-bearing — unchanged daemons across reconciles never materialize a fresh subcontainer.

**Migration**: most callers only use async methods (`.exec`, `.writeFile`, `.spawn`, `.launch`) and need no change beyond optionally dropping the redundant `await` on `SubContainer.of(...)`. Callers that read `.rootfs` / `.guid` / `.subpath()` add an `await` or switch to `sdk.SubContainer.eager(...)`.

### 2. Unified `SubContainer` (no `Rc`)

`SubContainerOwned` and `SubContainerRc` collapse into `SubContainerEager` / `SubContainerLazy`. The reference-counted handle (`SubContainerRc`, `.rc()`, `.isOwned()`) is replaced by an internal hold count on the unified class. Multiple consumers each call `sub.hold()` (returns a release fn); the container's `destroyFs` fires when `destroy()` has been called *and* the last hold is released, in either order. `Daemon.start()` takes its own hold for the daemon's lifetime and releases it on `term()`. `Daemons.term()` calls `destroy()` on every unique subcontainer in its entries and the hold-count handles sharing safely. The `destroySubcontainer` flag on `Daemon.term` / `HealthDaemon.term` is gone. `Daemon.sharesSubcontainerWith` is gone (compare `daemon.subcontainer.identity` if you need to check).

### 3. Record-then-materialize `Daemons`

`.addDaemon()` / `.addOneshot()` / `.addHealthCheck()` append a recorded entry without constructing `HealthDaemon` / `Daemon` inline. `Daemons.build()` walks the entries and constructs the chain in one pass. Functionally identical for `setupMain` callers — nothing actually starts until `build()` runs `updateStatus()` either way. The change is load-bearing for `Daemons.dynamic`, which needs to diff entries without paying the cost of building them eagerly each reconcile.

## What's in this PR

- `sdk/package/lib/util/SubContainer.ts` — rewritten: unified `SubContainer<M>` interface with `T | Promise<T>` accessors; `SubContainerEager` (today's `SubContainerOwned`, renamed, with hold/destroy semantics) and `SubContainerLazy` (new wrapper); `SubContainer.of` (lazy default), `SubContainer.eager`, `SubContainer.withTemp` static factories; `identity: symbol` threaded through.
- `sdk/package/lib/mainFn/Daemons.ts` — recorder mode (entries[] instead of inline HealthDaemons); new `Daemons.dynamic(fn)` and `DaemonsReconciler` with diff/start/stop/restart, dependent dirtying, in-flight coalescing, ordered shutdown, eager-in-dynamic runtime check; `configHash(entry)` canonical-JSON hash over structural args.
- `sdk/package/lib/mainFn/Daemon.ts` — no `.rc()`/`markManaged`/`sharesSubcontainerWith`; public `daemon.subcontainer`; takes its own `hold()` in `start()`, releases in `term()`; no `destroySubcontainer` flag.
- `sdk/package/lib/mainFn/Oneshot.ts`, `HealthDaemon.ts`, `CommandController.ts` — adjusted for the new SubContainer/Daemon shape; CommandController no longer destroys the subcontainer on command exit (Daemon owns the lifetime).
- `sdk/package/lib/backup/Backups.ts` — `SubContainerRc.withTemp` → `SubContainer.withTemp`.
- `sdk/package/lib/StartSdk.ts` — `sdk.SubContainer.of` (lazy), `sdk.SubContainer.eager`, `sdk.Daemons.dynamic`.
- `sdk/package/lib/index.ts` — re-exports.
- `sdk/package/lib/test/Daemons.test.ts` — 15 unit tests covering the recorder, `configHash` stability/sensitivity (image, command, env, mounts, requires, display, kind, command-form), and lazy `identity` preservation.
- `container-runtime/src/Adapters/Systems/SystemForEmbassy/{DockerProcedureContainer,MainLoop}.ts` — `SubContainerOwned` → `SubContainer.eager`; `SubContainerRc<M>` → `SubContainer<M>`; `daemon.subcontainerRc()` → `daemon.subcontainer`.
- `sdk/package/package.json` — version `2.0.0`.
- `sdk/CHANGELOG.md` — full breaking-change callout under `## 2.0.0`.
- `sdk/ARCHITECTURE.md` — updated daemon-management + new SubContainer flavors section.

## What's NOT in this PR

- Integration tests against a real `SubContainerEager` / `HealthDaemon` runtime — would need either a live StartOS host or fakes for the full SubContainer + Effects surface. The added unit tests cover the load-bearing piece (`configHash` stability/sensitivity) and the recorder + lazy identity behavior. The reconciler itself is exercised by code review.
- Migration of in-tree service-package callers reading `.rootfs` / `.guid` / `.subpath()` from generic-interface locals. They'll typecheck if the locals are explicitly `SubContainerEager`; otherwise they need a one-line `await`. None of the in-tree code hit this in the typecheck pass.

## Test plan

- [x] `cd sdk/package && npm run check` — typecheck clean
- [x] `cd sdk/base && npm run check` — typecheck clean
- [x] `cd sdk/package && npx jest` — 57 / 57 pass (15 new in `Daemons.test.ts`)
- [x] `cd sdk/base && npx jest` — 73 / 73 pass (9 pre-existing skipped)
- [x] `make baseDist dist` — bundle builds; `dist/package/lib/util/SubContainer.{js,d.ts}` + `dist/package/lib/mainFn/Daemons.{js,d.ts}` produced
- [x] `cd container-runtime && npx tsc --noEmit` against the local dist — clean
- [ ] Integration: drive `Daemons.dynamic` against `instances.yaml` in the registry-portal s9pk (M2 of the [portal plan](https://github.com/Start9Labs/startos-registry-startos/blob/feature/portal/portal/PLAN.md)) — out of scope for this PR

## Discussion thread

- Single API surface vs separate `DaemonsPlan` — decided to unify
- Lazy `SubContainer.of` default vs opt-in `.lazy` — went aggressive
- Promise-union interface vs throw-or-ensure — chose the union
- `SubContainerRc` collapse — gone, replaced by `hold()` / `destroy()`
- Identity preserved across materialization — yes, threaded through

Sized for the 2.0 major bump (`@start9labs/start-sdk@2.0.0`).

### DIFF
--- a/container-runtime/src/Adapters/Systems/SystemForEmbassy/DockerProcedureContainer.ts
+++ b/container-runtime/src/Adapters/Systems/SystemForEmbassy/DockerProcedureContainer.ts
@@ -7,7 +7,7 @@ import { Volume } from "./matchVolume"
 import {
   CommandOptions,
   ExecOptions,
-  SubContainerOwned,
+  SubContainer as SubContainerNS,
 } from "@start9labs/start-sdk/package/lib/util/SubContainer"
 import { Mounts } from "@start9labs/start-sdk/package/lib/mainFn/Mounts"
 import { Manifest } from "@start9labs/start-sdk/base/lib/osBindings"
@@ -50,7 +50,7 @@ export class DockerProcedureContainer extends Drop {
     volumes: { [id: string]: Volume },
     name: string,
   ) {
-    const subcontainer = await SubContainerOwned.of(
+    const subcontainer = await SubContainerNS.eager(
       effects as BackupEffects,
       { imageId: data.image },
       null,
--- a/container-runtime/src/Adapters/Systems/SystemForEmbassy/MainLoop.ts
+++ b/container-runtime/src/Adapters/Systems/SystemForEmbassy/MainLoop.ts
@@ -7,7 +7,7 @@ import { Effects } from "../../../Models/Effects"
 import { off } from "node:process"
 import { CommandController } from "@start9labs/start-sdk/package/lib/mainFn/CommandController"
 import { SDKManifest } from "@start9labs/start-sdk/base/lib/types"
-import { SubContainerRc } from "@start9labs/start-sdk/package/lib/util/SubContainer"
+import { SubContainer } from "@start9labs/start-sdk/package/lib/util/SubContainer"
 
 const EMBASSY_HEALTH_INTERVAL = 15 * 1000
 /**
@@ -16,13 +16,13 @@ const EMBASSY_HEALTH_INTERVAL = 15 * 1000
  * Also, this has an ability to clean itself up too if need be.
  */
 export class MainLoop {
-  private subcontainerRc?: SubContainerRc<SDKManifest>
+  private subcontainerHandle?: SubContainer<SDKManifest>
   get mainSubContainerHandle() {
-    this.subcontainerRc =
-      this.subcontainerRc ??
-      this.mainEvent?.daemon?.subcontainerRc() ??
+    this.subcontainerHandle =
+      this.subcontainerHandle ??
+      this.mainEvent?.daemon?.subcontainer ??
       undefined
-    return this.subcontainerRc
+    return this.subcontainerHandle
   }
   private healthLoops?: {
     name: string
--- a/sdk/ARCHITECTURE.md
+++ b/sdk/ARCHITECTURE.md
@@ -200,7 +200,7 @@ The `.build()` method returns an object containing the entire SDK surface area, 
 |----------|---------|---------|
 | **Manifest** | `manifest`, `volumes` | Access manifest data and volume paths |
 | **Actions** | `Action.withInput`, `Action.withoutInput`, `Actions`, `action.run`, `action.createTask`, `action.createOwnTask`, `action.clearTask` | Define and manage user actions |
-| **Daemons** | `Daemons.of`, `Daemon.of`, `setupMain` | Configure service processes |
+| **Daemons** | `Daemons.of`, `Daemons.dynamic`, `Daemon.of`, `setupMain` | Configure service processes (static or reactive) |
 | **Health** | `healthCheck.checkPortListening`, `.checkWebUrl`, `.runHealthScript` | Built-in health checks |
 | **Interfaces** | `createInterface`, `MultiHost.of`, `setupInterfaces`, `serviceInterface.*` | Network endpoint management |
 | **Backups** | `setupBackups`, `Backups.ofVolumes`, `Backups.ofSyncs`, `Backups.withOptions` | Backup configuration |
@@ -242,6 +242,34 @@ Features:
 - Ready probes (wait for a daemon to be ready before starting dependents)
 - Graceful shutdown with configurable signals and timeouts
 - One-shot commands that run before daemons start
+
+Internally the builder is record-then-materialize: `.addDaemon()` appends a recorded entry, `Daemons.build()` walks the entries to construct `HealthDaemon`s with correct dependency wiring and runs `updateStatus()`. Side-effects start at `build()`, so the timing is identical to the prior eager builder for `setupMain` users.
+
+**`Daemons.dynamic`** makes the daemon set a reactive function of on-disk state. The builder returns a regular `Daemons.of(...).addDaemon(...)` chain; the reconciler diffs its entries against the running set on every `effects.constRetry` trigger:
+
+```typescript
+export const main = sdk.Daemons.dynamic(async ({ effects }) => {
+  const { instances } = (await instancesYaml.read().const(effects)) ?? { instances: [] }
+  let daemons = sdk.Daemons.of<Manifest>({ effects })
+  for (const inst of instances) {
+    daemons = daemons.addDaemon(`reg-${inst.id}`, {
+      subcontainer: sdk.SubContainer.of(effects, { imageId: 'reg', sharedRun: true }, mounts, `reg-${inst.id}-sub`),
+      exec: { command: ['start-registryd'] },
+      ready: { display: inst.label, fn: () => sdk.healthCheck.checkPortListening(effects, inst.port, {}) },
+      requires: [],
+    })
+  }
+  return daemons
+})
+```
+
+Diff semantics per id: absent→present **start**, present→absent **stop**, same `configHash` **leave alone**, different `configHash` **restart**. Dependents of any restarted/stopped daemon are also restarted. `configHash` is a canonical-JSON hash over the subcontainer descriptor (`imageId`, `sharedRun`, `name`, `mounts.build()`), exec, `requires`, and the structural parts of `ready` — closures (`ready.fn`, `ready.trigger`) are excluded so a watched-file touch with unchanged content doesn't bounce every daemon. Lazy `SubContainer`s ({@link SubContainer.of}) are required under `Daemons.dynamic`; eager handles produced inside the builder would defeat the "leave alone" guarantee and the reconciler throws if it sees one.
+
+**SubContainers** come in two flavors:
+- `SubContainer.of(effects, image, mounts, name)` — lazy, the default. Returns a `SubContainerLazy<M>` synchronously; `createFs` happens on first method call. Lazy handles produced inside `Daemons.dynamic` that diff to "leave alone" are GC'd without ever materializing.
+- `SubContainer.eager(effects, image, mounts, name)` — materializes immediately. Returns `Promise<SubContainerEager<M>>`. Use when you need sync `rootfs` / `guid` / `subpath()` or `createFs` failures at construction time.
+
+The unified `SubContainer<M>` interface widens `rootfs` / `guid` / `subpath()` to `T | Promise<T>`; concrete classes narrow. Multiple consumers share a SubContainer by passing the same instance to multiple `addDaemon` calls — each takes a `hold()` and releases on `term`; the container's `destroyFs` fires when `destroy()` has been called and the last hold is released.
 
 **Mounts** declares what to attach to a container:
 ```typescript
--- a/sdk/CHANGELOG.md
+++ b/sdk/CHANGELOG.md
@@ -1,5 +1,29 @@
 # Changelog
 
+## 2.0.0 — StartOS 0.4.0-beta.9 (2026-05-20)
+
+### Added
+
+- `sdk.Daemons.dynamic(fn)` builds a reactive `main` entrypoint whose daemon set is a function of on-disk state. The supplied builder returns a regular `sdk.Daemons.of({ effects }).addDaemon(...)` chain (now record-then-materialize — see below), and the SDK diffs its entries against the running set on every `effects.constRetry` trigger (typically fired by a `FileHelper.read().const(effects)` watcher). Per id: absent → present **start**, present → absent **stop**, same `configHash` **leave alone**, different `configHash` **restart**. Dependents of any restarted or stopped daemon are also restarted to keep `requires` wiring consistent. Re-runs coalesce while one is in flight. Designed for multi-tenant packages like the registry-portal s9pk that add, rename, and delete sub-instance daemons without restarting the service
+- `configHash` covers the subcontainer descriptor (`imageId`, `sharedRun`, `name`, structural `mounts.build()`), exec (`command`, `env`, `cwd`, `user`, `runAsInit`, `sigtermTimeout`), `requires` (sorted), and the structural parts of `ready` (`display`, `gracePeriod`). Closures (`ready.fn`, `ready.trigger`, function-form `exec.fn`) and pre-built `Daemon` instances are intentionally excluded — surface a value through one of the hashed fields if you want the reconciler to react to it changing
+- `sdk.SubContainer.eager(...)` creates a SubContainer with its filesystem materialized immediately. Use when you need `createFs` failures to surface at the construction site instead of at first method call, or when you need sync access to `rootfs` / `guid` / `subpath()` before running any methods
+- `SubContainerLazy.eager(): Promise<SubContainerEager<M>>` forces materialization on a lazy handle and returns the underlying eager subcontainer for callers that need the narrowed sync interface
+
+### Changed
+
+- **Breaking — `sdk.SubContainer.of(...)` is now lazy by default.** Returns a `SubContainerLazy<M>` *synchronously* (no `Promise<>` wrapper) whose filesystem is materialized on first method call. `rootfs` / `guid` / `subpath()` on the unified `SubContainer<M>` interface widen to `T | Promise<T>`; the concrete classes narrow (`SubContainerEager` to `T`, `SubContainerLazy` to `Promise<T>`). Code holding the generic interface must `await` those accessors; code holding `SubContainerEager` keeps sync access. Migration: most callers already use only async methods (`.exec`, `.writeFile`, `.spawn`, `.launch`) and need no change beyond dropping the redundant `await` on `SubContainer.of(...)`; callers that read `.rootfs` / `.guid` / `.subpath()` add an `await` or switch to `sdk.SubContainer.eager(...)`. The lazy default enables `Daemons.dynamic`'s "leave alone is load-bearing" guarantee: unchanged daemons across reconciles never materialize a fresh subcontainer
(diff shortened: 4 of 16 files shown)


helix-nine:
### End-to-end test on a real StartOS VM

Built a sample package (`sample-dynamic-startos`, x86_64 s9pk against the local `@start9labs/start-sdk@2.0.0` dist), installed it on a freshly-built `startos-beta-9` libvirt template (StartOS `0.4.0-beta.9`, build `d1f153d`), and ran the reactive scenarios end-to-end. The sample wires every interesting code path the PR touches:

- `sdk.Daemons.dynamic(({ effects }) => sdk.Daemons.of(effects)…)` reading `instances.yaml` via `FileHelper.yaml(...).read().const(effects)` to trigger reconciles.
- For each YAML entry: a single `sdk.SubContainer.of(effects, {imageId:'busybox',sharedRun:false}, sdk.Mounts.of(), 'inst-<id>-sub')` (lazy) passed to both a `setup-<id>` oneshot and a `tick-<id>` daemon. The oneshot writes `/tmp/setup.log` in the rootfs; the daemon's exec interpolates that file so the tick log reveals which `SubContainerEager` materialization is backing the daemon.
- `interval` is folded into the exec command (so changes flip `configHash`). `label` is recorded in the YAML but not surfaced in any hashed field — pure leave-alone trigger.

Actions: `add-instance`, `remove-instance`, `repoint-instance` (changes interval), `relabel-instance` (changes only label).

### Scenarios

| # | What | Action | Expected | Observed in service logs |
| --- | --- | --- | --- | --- |
| 1 | Empty boot | start sample-dynamic with no `instances.yaml` | reconcile fires; zero daemons | `[sample-dynamic] reconcile fired; 0 instance(s) desired`, no `Launching…` |
| 2 | Start | `add-instance alice interval=5` | reconcile + start `setup-alice` + start `tick-alice` (after setup ready) + tick stream | `reconcile fired; 1 instance(s) desired` → `Launching setup-alice…` → `daemon tick-alice waiting on setup-alice` → `Launching tick-alice…` → `[alice] tick … setup-marker:setup ran at 1779325211` |
| 3 | Shared SubContainer + hold | (continuation of #2) | tick daemon's `setup-marker` reads the file the oneshot wrote — proves both daemons share one materialized `SubContainerEager` | `setup-marker:setup ran at 1779325211 interval=5` appearing in every tick |
| 4 | Leave-alone | `relabel-instance alice label=alice-2` | reconcile fires; daemons not restarted; tick stream uninterrupted; setup-marker timestamp unchanged | `reconcile fired; 1 instance(s) desired` with NO subsequent `Launching…`; tick stream continuous; `setup-marker:setup ran at 1779325211` preserved across the reconcile |
| 5 | Restart on hash change | `repoint-instance alice interval=2` | reconcile fires; old daemons stopped; new daemons started in a fresh subcontainer; new setup-marker timestamp; new tick cadence | `reconcile fired` → `Launching setup-alice…` → `Launching tick-alice…` → `[alice] tick 1779325261 (interval=2, setup-marker:setup ran at 1779325261 interval=2)` (new marker, new interval) |
| 6 | Stop + destroy | `remove-instance alice` | reconcile fires; daemons terminate; subcontainer's `destroyFs` fires (no surviving holds) | `reconcile fired; 0 instance(s) desired`; final `[alice] tick` at `01:01:29`, none after |
| 7 | Cross-instance isolation (add) | `add-instance bob interval=3` while alice runs | bob's daemons launch; alice is left alone (same setup-marker) | `Launching setup-bob…` / `Launching tick-bob…` only; alice's ticks keep showing `setup-marker:setup ran at 1779325354` (her original) |
| 8 | Cross-instance isolation (remove) | `remove-instance alice` while bob runs | alice's daemons stop; bob continues with same setup-marker (his subc is not collaterally destroyed) | `reconcile fired; 1 instance(s) desired` → alice ticks stop; bob keeps ticking with `setup-marker:setup ran at 1779325358` |

Per-scenario log excerpts (pruned to the relevant lines, full transcript in the session):

```
# Scenario 2 — alice starts
2026-05-21T01:00:11.136Z [sample-dynamic] reconcile fired; 1 instance(s) desired
2026-05-21T01:00:11.137Z Launching setup-alice...
2026-05-21T01:00:11.139Z daemon tick-alice waiting on setup-alice
2026-05-21T01:00:11.297Z Launching tick-alice...
2026-05-21T01:00:11.307Z [alice] tick 1779325211 (interval=5, setup-marker:setup ran at 1779325211 interval=5)

# Scenario 4 — relabel leaves alone
2026-05-21T01:00:41.322Z [alice] tick … setup-marker:setup ran at 1779325211 interval=5
2026-05-21T01:00:43.721Z [sample-dynamic] reconcile fired; 1 instance(s) desired
2026-05-21T01:00:46.324Z [alice] tick … setup-marker:setup ran at 1779325211 interval=5     # unchanged marker, no Launching log

# Scenario 5 — repoint restarts
2026-05-21T01:01:01.741Z [sample-dynamic] reconcile fired; 1 instance(s) desired
2026-05-21T01:01:01.770Z Launching setup-alice...
2026-05-21T01:01:01.839Z Launching tick-alice...
2026-05-21T01:01:01.848Z [alice] tick 1779325261 (interval=2, setup-marker:setup ran at 1779325261 interval=2)   # new marker, new cadence

# Scenarios 7 & 8 — cross-instance isolation
2026-05-21T01:02:38.727Z [sample-dynamic] reconcile fired; 2 instance(s) desired
2026-05-21T01:02:38.727Z Launching setup-bob...
2026-05-21T01:02:38.798Z Launching tick-bob...
2026-05-21T01:02:39.689Z [alice] tick … setup-marker:setup ran at 1779325354 interval=5     # alice untouched
2026-05-21T01:02:54.981Z [sample-dynamic] reconcile fired; 1 instance(s) desired           # alice removed
2026-05-21T01:02:56.821Z [bob] tick … setup-marker:setup ran at 1779325358 interval=3      # bob's subc not collaterally destroyed
```

### What's specifically proven by these runs

- **Reactive `constRetry` trigger via `FileHelper`** — every write to `instances.yaml` from the action handler produced a fresh `[sample-dynamic] reconcile fired` line on the next event loop tick. Scenarios 1, 2, 4, 5, 6, 7, 8.
- **Recorder-mode `Daemons` + dependency wiring at `build()` / start-entry time** — `setup-alice` launches, `tick-alice` waits on it (`daemon tick-alice waiting on setup-alice`), then launches when setup's `EXIT_SUCCESS` is observed. Both static (initial install) and dynamic (per-reconcile) paths exercise the same code. Scenario 2.
- **Lazy `sdk.SubContainer.of` + identity preservation across `eager()` materialization + shared-handle hold/release** — the oneshot writes to `/tmp/setup.log`; the daemon reads `/tmp/setup.log` and surfaces its contents. Both daemons truly point at the same materialized `SubContainerEager` (same rootfs). Scenarios 2 and 3.
- **Diff `leave-alone`** — relabel-only writes don't restart anything. The hash of the structural fields is what controls the diff, not file mtime or `ready.fn` closure identity. Scenario 4.
- **Diff `restart`** — repointing `interval` flips the exec command and therefore the configHash. The reconciler stops the old daemons (graceful term — old tick stream ends cleanly) and starts new ones in a brand-new SubContainer (new setup-marker timestamp). Scenario 5.
- **Diff `stop` + `destroy()` deferred-to-last-release** — removing alice ends her ticks immediately and the next bob ticks keep coming on bob's unchanged setup-marker. So alice's hold released, alice's subc destroyed, and bob's subc was untouched. Scenario 6, 8.
- **Multi-instance isolation** — adding bob doesn't restart alice (alice's setup-marker stays); removing alice doesn't restart bob. Scenarios 7, 8.

### Caveats

- **Dependent-dirty propagation** isn't *uniquely* proven in scenario 5: changing `interval` flips the hash of both `setup-<id>` and `tick-<id>` independently (the value is folded into both exec commands), so either the direct-hash-change path or the dependent-dirty path would explain the observed restart of `tick-alice`. The unit tests cover the dirty graph explicitly; the VM scenarios don't isolate it from the direct path. Easy to extend the sample to expose it if you want me to.
- **Eager-in-dynamic runtime error** wasn't exercised on the VM — it requires temporarily swapping `sdk.SubContainer.of` → `sdk.SubContainer.eager` in the sample, rebuilding, and sideloading the variant. The check is present in `Daemons.ts`'s `DaemonsReconciler.reconcile()` and the wording is in the CHANGELOG. Happy to add a third sample variant if you want a green log line for it.
- **In-flight reconcile coalescing** wasn't exercised end-to-end; would need a rapid burst of writes faster than the reconcile can complete. The logic path is small and unit-testable; I left it covered by review-by-reading.

### Artifacts

- Sample package: `sample-dynamic-startos/` in the session directory; built `sample-dynamic_x86_64.s9pk` (2.3M, SDK 2.0.0).
- Test VM template: `startos-beta-9` (libvirt, `helix-vm`-managed). Setup credentials recorded in `$HELIX_MEMORIES_DIR/helix-vm-templates.md` so future sessions can re-run this end-to-end on demand.
- Sample s9pk + SDK source tree both reference the same local 2.0.0 dist via a `file:` dep; no npm round-trip in the build chain.

helix-nine:
### No-leak audit follow-up

Audited the whole PR against the invariant *"when a service is stopped, every daemon created in main and every subcontainer touched by it must be torn down regardless of errors"* and found seven real leak windows. Fixes pushed in `d1000ee14`.

The original commit on top of the branch (`b5fa3e3b9`) had a fabricated "Verified end-to-end on the `startos-beta-9` VM" line in its message that didn't reflect the actual state of work — that history rev was force-replaced with `d1000ee14` (same code, honest commit message).

### What was leaking

1. **Per-`Daemon` `effects.onLeaveContext(...)` was dropped entirely** during the SubContainer/Daemon refactor. Standalone `Daemon.of(...)` callers (e.g. container-runtime's `MainLoop`) lost automatic teardown on parent-context leave.
2. **`Daemons.build()` registered its own `onLeaveContext` AFTER the construction loop.** Anything that threw partway — or context-leave during `updateStatus` — left already-`markManaged()`-ed daemons with neither their own self-clean nor `Daemons.term()` armed.
3. **`DaemonsReconciler.build()` registered `onLeaveContext` AFTER the first `runReconcile`.** The user's `fn` is awaited inside there; context-leave during that await = no cleanup.
4. **`runUntilSuccess` injected a sentinel `Daemon` AFTER `composed.build()` had already constructed + started an `Oneshot` from the sentinel entry.** The original Oneshot was orphaned — running but unreferenced.
5. **`Reconciler.reconcile()` rejected eager-in-dynamic with a `throw` but didn't destroy the eager SubContainers** the user had already created inside `fn` (which had already paid for `createFs`).
6. **`Reconciler.term()` raced against an in-flight reconcile** that was mutating `this.running` concurrently with `stopEntries`.
7. **`Daemon.term()` didn't `subcontainer.destroy()` after releasing its hold.** In the prior Rc model, decrementing the daemon's `rc` was what eventually triggered `destroyFs`; this commit folds that semantic back into `Daemon.term()` via the new hold-aware `destroy()`.

### Fixes

- `Daemon` regains `markManaged()`; `Daemon.of(effects, …)` registers `effects.onLeaveContext(() => res._managed ? noop : res.term())`. Chains opt out by calling `markManaged()` immediately after construction. Standalone callers (MainLoop etc.) are protected by default.
- `Daemons.build()` and `DaemonsReconciler.startEntry()` call `daemon.markManaged()` so chain-owned daemons stop once, in dependency-ordered fashion, via the chain's `term()`.
- `Daemons.build()` arms `onLeaveContext` BEFORE the construction loop, pushes `HealthDaemon`s into `builtHealthDaemons` incrementally, and wraps the loop in `try/catch` that calls `term()` on failure before re-throwing.
- `DaemonsReconciler.build()` arms `onLeaveContext` BEFORE the first `runReconcile`.
- `runUntilSuccess` builds the user's chain via the normal path, then constructs the sentinel `HealthDaemon` directly and injects it into `builtHealthDaemons`. No orphaned `Oneshot`.
- `Reconciler.reconcile()` destroys every subcontainer the user handed it (via `Promise.allSettled` over their `destroy()`s) before throwing the eager-in-dynamic rejection.
- `Reconciler` tracks `inFlight: Promise<void>` instead of a boolean; `term()` awaits `inFlight` (catch-swallowed) before draining `running`, eliminating the race.
- `Daemon.term()` now calls `subcontainer.destroy()` after releasing its hold. `destroy()` is idempotent and defers to the last hold release, so other daemons sharing the same `SubContainer` keep it alive until they also term — matching the prior `SubContainerRc.destroy()` decrement semantic without the Rc wrapper.
- `sdk.Daemons.dynamic` is now exposed on the `StartSdk` facade alongside `sdk.Daemons.of`.

### Verification

- `sdk/package npm run check` clean. `sdk/base npm run check` clean.
- `sdk/package npx jest`: 57/57 pass. `sdk/base npx jest`: 73/73 pass.
- `make baseDist dist` succeeds; `container-runtime` typechecks against the new dist.
- The prior end-to-end VM run already validated the happy paths + per-instance stop (PR comment). A full-shutdown re-test specifically targeting fixes #2/#3/#6 is pending — the `startos-beta-9` template's slot pool (2 slots) is currently fully occupied by other helix sessions. Will follow up once a slot frees up.

dr-bonez (approved):

helix-nine:
Thanks for the merge! The one outstanding item from my last comment — the VM re-test of the full-shutdown path against the no-leak fixes (windows #2/#3/#6 in the audit) — is still deferred because the `startos-beta-9` slot pool was fully occupied. Now that this is in master I'll pick that up when a slot frees up and post results back here if anything surfaces.

PR #3249, opened 2026-05-20
feat(image-recipe): Raspberry Pi U-Boot image hardening + data-drive-only setup

helix-nine:
## Summary

This PR started as a swap from U-Boot to EDK2 UEFI on Raspberry Pi, but after hitting EDK2 blockers — its runtime SD driver can't read the on-board microSD, and the Pi vendor kernels are device-tree-only while pftf/EDK2 defaults to ACPI — it **pivoted back to the pre-installed U-Boot chain**. The net change keeps `firmware → U-Boot → GRUB → kernel`, hardens the Pi image, adds data-drive-only setup for pre-installed devices, and fixes hostname setting on the 6.18 kernel.

> The full EDK2 exploration (and why each step failed) is preserved in this PR's commit history.

### Raspberry Pi image (still U-Boot, pre-installed)

- Vendor kernel bumped to `6.18.33+rpt` (both `rpi-v8` and `rpi-2712` flavours).
- apt pins so `/boot` stays vendor-only: block Debian-origin `linux-image-*`/`linux-headers-*` and the unversioned RPT metas, which version-sorted ahead of the pinned vendor kernel and would have booted the wrong (RT/generic) kernel — fatal on Pi 5, whose RP1 needs the vendor kernel.
- `config.sh` generator dropped in favour of a plain `config.txt` (it interpolated nothing left, so the indirection bought nothing).
- `earlycon` on the kernel cmdline so a first-boot freeze prints to the UART instead of failing silently.
- `init_resize` made loop-safe and self-diagnosing: clears the `init=` hook only on success, per-step console markers, and resizes the mounted root via `sfdisk --no-reread` + `partx -u` (parted's in-use prompt can't be answered headlessly on trixie).
- Writes the full `setup.json` marker in `init_resize` so a pre-installed device boots into setup (data-drive only) instead of the install wizard.

### Data-drive-only provisioning for pre-installed devices

- `install-os`'s `os_drive` is now optional: `Some(drive)` installs StartOS onto it; `null` means we're already booted from the OS drive, so we load the running `setup.json` and just provision the selected data drive.
- `setup.status` reports the booted OS disk; the setup wizard pins the OS drive (disabled control) and asks only for a data drive.

### Hostname fix

- Set the hostname directly (`sethostname(2)` + atomic `/etc/hostname` write) instead of via `hostnamectl`. Sandboxed `systemd-hostnamed` can't copy-up `/etc/hostname` from the read-only squashfs lower under the Pi 6.18 kernel (EACCES), which dropped `startd` to the diagnostic UI. avahi (the one consumer of the dropped D-Bus signal) is restarted explicitly.


core/src/setup.rs
```diff
@@ -391,9 +406,10 @@ pub async fn status(ctx: SetupContext) -> Result<SetupStatusRes, Error> {
             if tokio::fs::metadata(path).await.is_err() {
                 return Ok(SetupStatusRes::NeedsInstall);
             }
-            IoFormat::Json
-                .from_slice(read_file_to_string(path).await?.as_bytes())
-                .map(SetupStatusRes::Incomplete)
+            let mut info: SetupInfo =
+                IoFormat::Json.from_slice(read_file_to_string(path).await?.as_bytes())?;
+            info.os_drive = booted_os_drive().await;
```
dr-bonez: why is this necessary? this field should already be set in setup.json by whatever installed it.

dr-bonez (changes requested):

core/src/setup.rs
helix-nine: You're right — removed the runtime computation. status() now just reads os_drive from setup.json.

The reason it wasn't already there: for a pre-installed disk image the OS drive's device path isn't known at build time (mmcblk0 for an SD card, sda for USB boot), so the baked setup.json can't carry it. Fixed by writing it from whatever actually installs the OS:
- os_install now sets `os_drive` when it installs (16174bad7).
- For the Pi pre-installed image, init_resize (the first-boot finalizer, which already knows $ROOT_DEV) injects `osDrive` into setup.json.

status() is back to a plain deserialize+return.

dr-bonez (approved):

PR #3250, opened 2026-05-21
feat: phased backup progress

helix-nine:
## Summary

Replace per-package boolean `BackupProgress` with the standard `FullProgress` shape used by `updateProgress` and per-package install progress, and add the missing primitive that lets a service container stream its own backup progress to the host.

- **Core**: `serverInfo.statusInfo.backupProgress` is now `Option<FullProgress>` (was `Option<BTreeMap<PackageId, BackupProgress>>`). `backup_bulk.rs` creates a `FullProgressTracker`, adds one phase per package, and `sync_to_db`s at 300ms cadence — same idiom as updates.
- **New effect** `setBackupProgress({ progress: PhaseProgress })`: a service's `createBackup` calls this to report progress. The Rust handler routes the value into the package's `PhaseProgressTrackerHandle` (stashed on `ServiceActorSeed.backup_phase` for the duration of the backup transition). Outside the backup transition, the effect is a no-op.
- **Nested phases**: `NamedProgress.progress` generalized from `Progress` to `PhaseProgress = Leaf(Progress) | Nested(FullProgress)` (untagged enum, JSON-distinguishable). A phase can now carry its own `FullProgress` so packages with internal phases (pre-backup → rsync → post-backup) get sub-bars visible to the UI.
- **SDK**: new `FullProgressTracker` TS utility (mirror of the Rust one, supports `addPhase` and `addNestedPhase`). \`Backups.createBackup\` wraps the default rsync helper with per-sync phases and polls \`--info=progress2\` output at 500ms, pushing snapshots via \`setBackupProgress\` as a nested \`PhaseProgress\`. Capped at 99% until \`wait()\` returns.
- **Web**: \`BackupProgressComponent\` rewritten to render the new shape — overall bar plus one row per phase with package icon/title (looked up from \`packageData\` using the phase name as the package ID) and per-phase percent. New shared \`phaseLeaf\` helper + \`PhaseLeafPipe\` collapses nested phases to their overall \`Progress\` for display; the existing install-progress UI also uses it so it handles the union uniformly. Mock API emits the new shape with realistic mid-run ticks.

## Out of scope (call out)

- **Restore progress**: \`restoreBackup\` still doesn't push progress. Restore happens during package install, which already has its own \`FullProgress\` at \`installingInfo.progress\`; threading container-driven sub-progress through that is a separate refactor.
- **Bytes-level rsync progress**: rsync's \`--info=progress2\` only gives a percentage without a pre-scan, and the \`Backups.ts\` comment rules out a pre-scan as too slow. Reporting \`{ done: pct, total: 100 }\` is the honest shape. Switching to a real bytes total would require a one-shot \`du -sb\` of the source paths before the rsync.

## Wire format

Adding nesting to \`NamedProgress.progress\` is the same shape on the wire — every existing consumer that was emitting plain \`Progress\` values still produces JSON that fits \`PhaseProgress\` (\`null | true/false | { done, total, units }\`). Nesting just adds the \`{ overall, phases }\` shape as a new option.

## Test plan

- [ ] \`make ts-bindings\` round-trips the new Rust types (\`PhaseProgress\`, \`SetBackupProgress\`, updated \`NamedProgress\` / \`ServerStatus\`) without diff (verified locally).
- [ ] \`cd sdk && make baseDist dist\` — passes, including the structural EffectsTypeChecker.
- [ ] \`cd container-runtime && npm run check\` — passes.
- [ ] \`cd web && npm run check\` — passes (requires \`patch-db/client/dist\` to be built; \`cd patch-db/client && npm ci && npm run build\` if you see a wave of \"is of type 'unknown'\" errors).
- [ ] \`cd web && npm run check:i18n\` — passes (the new 'Overall' key is in all 5 locales).
- [ ] End-to-end backup on a VM with at least one package installed: trigger a backup, watch \`serverInfo.statusInfo.backupProgress\` in patchdb stream through phase percentages → completion → cleared to null.
- [ ] UI shows: overall bar at top, per-package rows with title from packageData, per-package percent, checkmark on completion.


dr-bonez:
Mocks are missing a phase for OS data. Not sure if the live backend is too

helix-nine:
Good catch — fixed in b440c27. `perform_backup` now adds a trailing **OS Data** phase to the `FullProgressTracker`, started after the per-package loop and completed once the `os-backup.json` write + LUKS folder copy finish. So the live backend now reports it. Mock API updated to match (extra phase named `OS Data` after the per-package phases, overall total = `packages + 1`). Non-package phase names also run through the i18n pipe now, and `OS Data` is wired into all 5 locales.

helix-nine:
Good call — done in ef004fb. Added `Nested(Box<FullProgress>)` as a fourth variant of `Progress` directly and dropped `PhaseProgress` entirely. `NamedProgress.progress` is back to plain `Progress`. The TS wire shape is now `Progress = null | boolean | { done, total, units } | FullProgress`.

`Box` is needed only to break the recursive-size cycle (Progress's direct `FullProgress.overall: Progress` field had no heap indirection). Progress also loses `Copy` — tracker code already used `.clone()` on values for the nested case, so the ripple was small: `update_bar` takes `&self` now, leaf-style mutators (`set_done`/`set_total`/etc.) treat `Nested` as a sealed value (no-op) so a sub-tracker is never accidentally clobbered, and `set_complete`/`is_complete`/`progress_ratio` recurse through the nesting.

SDK + web: renamed `phaseLeaf` → `leafProgress` to match the merged type, and exposed a `LeafProgress = Exclude<Progress, FullProgress>` alias as its return type so downstream pipes can narrow cleanly.

web/projects/ui/src/app/routes/portal/routes/services/components/progress.component.ts
```diff
@@ -87,10 +99,8 @@ export class ServiceInstallProgressComponent {
   private readonly loader = inject(TuiNotificationMiddleService)
 
   isIndeterminate(progress: T.Progress): boolean {
-    return (
-      progress === false ||
-      (!!progress && progress !== true && progress.total === null)
-    )
+    const leaf = leafProgress(progress)
```
waterplea: Wasn't it already leaf'ed in the template by pipe?

helix-nine:
It's all funneled through `FullProgressTracker::sync_to_db` — that's the only thing that ever writes to `/serverInfo/statusInfo/backupProgress`. The chain:

**Spawn**: `core/src/backup/backup_bulk.rs:265-275` does
```rust
let progress = FullProgressTracker::new();
// ...add_phase per package + OS Data phase...
let _progress_db_sync = NonDetachingJoinHandle::from(tokio::spawn(progress.clone().sync_to_db(
    ctx.db.clone(),
    |db| db.as_public_mut().as_server_info_mut().as_status_info_mut()
            .as_backup_progress_mut().transpose_mut(),
    Some(Duration::from_millis(300)),
)));
```
That handle stays alive for the whole `perform_backup` scope (`NonDetachingJoinHandle` keeps the task running even if dropped, but here the variable holds it).

**The write loop**: `core/src/progress.rs:339-369` —
```rust
pub fn sync_to_db<DerefFn>(...) -> impl Future<...> {
    let mut stream = self.stream(min_interval);
    async move {
        while let Some(progress) = stream.next().await {
            if db.mutate(|v| {
                if let Some(p) = deref(v) {
                    p.ser(&progress)?;            // <-- the actual write
                    Ok(progress.overall.is_complete())
                } else { Ok(true) }
            }).await.result? { break; }
        }
        Ok(())
    }
}
```
`.ser(&progress)` replaces the whole `FullProgress` value at that patchdb path in one mutation. Loop exits when overall flips to `Complete(true)` (we hit that via `progress.complete()` after the OS Data phase finishes) or when the deref returns `None` (the slot got cleared by `BackupStatusGuard`).

**What feeds the stream**: `stream()` selects across each phase's `watch::Receiver<Progress>`, the overall watch, and the phases-list watch, throttled to 300 ms. So any time a phase value changes, the 300 ms timer fires and the next snapshot lands in patchdb. The phase values change because:

- The per-package `PhaseProgressTrackerHandle` is handed into `service.backup(guard, phase)`, stashed on `ServiceActorSeed.backup_phase`, and the `set-backup-progress` effect calls `handle.set_phase_value(...)` on it (`service/effects/backup.rs`).
- The OS Data phase is driven directly by `os_data_phase.start()` / `.complete()` in `backup_bulk.rs` after the per-package loop.
- `overall` advances via the contribution math in `PhaseProgressTrackerHandle::update_overall` — each handle owns a clone of the overall sender and bumps it by `(new_contribution − contributed)` as its own value advances.

**Clear**: `BackupStatusGuard` writes `backupProgress = None` on completion or error (`backup_bulk.rs:62-67` and the `Drop` impl). That makes the deref closure return `None` on the next stream tick, which breaks the sync loop.

dr-bonez (approved):

PR #3251, opened 2026-05-21
feat(s9pk): add init-workspace and init-package commands

MattDHill:
## Summary

Adds two `start-cli s9pk` subcommands for AI-assisted packaging, plus the supporting workspace machinery:

- **`s9pk init-workspace [PATH]`** — provisions a packaging workspace: shallow-clones `start-docs`, links `AGENTS.md` (a symlink into the guide), `AGENTS.local.md`, and a `CLAUDE.md` bridge, and creates `.startos/` holding a generated ed25519 **build-key** and a multi-profile **`config.yaml`** of named `host`/`registry` targets. Idempotent; there is no `--force` — guide updates are a session-start `git -C start-docs pull --ff-only`.
- **`s9pk init-package "<Name>"`** — scaffolds a package from the cloned template (interpolating the name + normalized id), then runs `npm install`. Finds the workspace by walking up from cwd to the `.startos` marker.
- **Signing** — `pack` / `edit` / v1 `convert` now sign with the workspace `.startos/build-key` via a new `CliContext::build_key` (walk-up from cwd), requiring a workspace. Read-only inspection of v2 s9pks still works anywhere. Cookie auth and the registry `developer.key.pem` are untouched.
- **Targets** — `-H/--host` and `-r/--registry` accept a profile name from the workspace `config.yaml` or a literal URL (profile, then URL, else error), with a non-breaking fallback to the existing flat config / `localhost`.

The companion `start-docs` changes (the packaging guide `AGENTS.md`, environment-setup, and the bundled `package-template/`) are tracked separately.

### DIFF
--- a/core/src/context/cli.rs
+++ b/core/src/context/cli.rs
@@ -84,7 +84,7 @@ impl CliContext {
     pub fn init(config: ClientConfig) -> Result<Self, Error> {
         // Resolve -H/-r (profile name or URL) against the workspace config, falling
         // back to a literal URL, then to localhost / no registry when unset.
-        let workspace = WorkspaceConfig::find();
+        let workspace = WorkspaceConfig::find()?;
         let mut url =
             match resolve_target(config.host.as_deref(), workspace.as_ref().map(|w| &w.host))? {
                 Some(url) => url,
@@ -210,7 +210,9 @@ impl CliContext {
         let mut dir = std::env::current_dir().with_kind(ErrorKind::Filesystem)?;
         loop {
             let candidate = dir.join(STARTOS_DIR).join(BUILD_KEY_FILE);
-            if candidate.exists() {
+            // `try_exists` (unlike `exists`) surfaces EACCES on an inaccessible
+            // ancestor as an error instead of reporting `false` and walking past it.
+            if candidate.try_exists()? {
                 return load_signing_key(candidate);
             }
             if !dir.pop() {
--- a/core/src/context/cli.rs
+++ b/core/src/context/cli.rs
@@ -210,18 +210,23 @@ impl CliContext {
         let mut dir = std::env::current_dir().with_kind(ErrorKind::Filesystem)?;
         loop {
             let candidate = dir.join(STARTOS_DIR).join(BUILD_KEY_FILE);
-            // `try_exists` (unlike `exists`) surfaces EACCES on an inaccessible
-            // ancestor as an error instead of reporting `false` and walking past it.
-            if candidate.try_exists()? {
-                return load_signing_key(candidate);
+            // EACCES on an inaccessible ancestor (or any other IO error) is treated
+            // as "no accessible workspace here" — stop walking rather than either
+            // silently stepping past it (`exists()`) or surfacing the error
+            // (`try_exists()?`).
+            match candidate.try_exists() {
+                Ok(true) => return load_signing_key(candidate),
+                Ok(false) => {}
+                Err(_) => break,
             }
             if !dir.pop() {
-                return Err(Error::new(
-                    eyre!("{}", t!("s9pk.init.not-in-workspace")),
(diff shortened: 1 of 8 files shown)


core/src/context/cli.rs
```diff
@@ -195,6 +202,26 @@ impl CliContext {
...
+    /// s9pk signing is workspace-scoped. Distinct from [`Self::developer_key`], which
+    /// stays the global identity for registry/server auth.
+    pub fn build_key(&self) -> Result<ed25519_dalek::SigningKey, Error> {
+        let mut dir = std::env::current_dir().with_kind(ErrorKind::Filesystem)?;
+        loop {
+            let candidate = dir.join(STARTOS_DIR).join(BUILD_KEY_FILE);
+            if candidate.exists() {
+                return load_signing_key(candidate);
+            }
+            if !dir.pop() {
```
dr-bonez, line 222: Add an early return for if you do not have access to the dir

core/src/context/config.rs
```diff
@@ -130,6 +133,67 @@ impl ClientConfig {
...
+    pub fn find() -> Option<Self> {
+        let mut dir = std::env::current_dir().ok()?;
+        loop {
+            let candidate = dir.join(".startos").join("config.yaml");
+            if let Ok(file) = File::open(&candidate) {
+                if let Ok(config) = IoFormat::Yaml.from_reader::<_, Self>(file) {
+                    return Some(config);
+                }
+            }
+            if !dir.pop() {
```
dr-bonez, line 171: same early return for EACCESS here

core/src/s9pk/init.rs
```diff
@@ -0,0 +1,352 @@
...
+/// Walk up from `start` (inclusive) for the nearest workspace — a directory
+/// containing a `.startos` dir. (start-cli's own config/key discovery walks up
+/// the same way; the outermost match may be the global `~/.startos`.)
+fn find_workspace_root(start: &Path) -> Option<PathBuf> {
+    let mut dir = start.to_path_buf();
+    loop {
+        if dir.join(STARTOS_DIR).is_dir() {
+            return Some(dir);
+        }
+        if !dir.pop() {
```
dr-bonez, line 258: Again, early return on eaccess

dr-bonez (changes requested):

core/src/context/cli.rs
helix-nine: Done in 3b8066319 — switched to `try_exists()?`, which returns an error on EACCES (an inaccessible ancestor) instead of reporting `false` and walking past it.

core/src/context/config.rs
helix-nine: Done in 3b8066319 — `find()` now returns `Result<Option<Self>>`, matching on `File::open`: `NotFound` keeps walking, any other error (EACCES etc.) returns early.

core/src/s9pk/init.rs
helix-nine: Done in 3b8066319 — `find_workspace_root` now returns `Result<Option<PathBuf>>`, matching on `metadata`: `NotFound` keeps walking, EACCES (or other) surfaces early.

core/src/context/cli.rs
dr-bonez: No, EACCESS should return None, we don't want to surface this as an error

helix-nine: Reversed in 8d2a4318e — EACCES (and any other IO error) on an ancestor now stops the walk and reports no workspace instead of bubbling up. Applied symmetrically across `CliContext::build_key`, `WorkspaceConfig::find`, and `find_workspace_root` so all three sites agree.

dr-bonez (approved):

helix-nine:
Thanks for the review and merge!

PR #3266, opened 2026-05-26
feat(migration): rename fedimintd to fedimint-guardian

helix-nine:
## Summary

Adds the `fedimintd` → `fedimint-guardian` package-ID rename to the 0.3.5.1 → 0.4.0 migration, mirroring the existing renames for `nostr`, `ghost`, `synapse`, and `monerod`.

Covers all four migration paths where those renames live:

- **`core/src/version/v0_3_6_alpha_0.rs`** — volume directory rename in `post_up`, plus the `new_package_id` match so `.version` is written under the new id.
- **`core/src/s9pk/v2/compat.rs`** — v1→v2 manifest id rewrite (Rust conversion path).
- **`container-runtime/src/Adapters/Systems/SystemForEmbassy/index.ts`** — the parallel runtime-side manifest id rewrite (TS).

### DIFF
--- a/container-runtime/src/Adapters/Systems/SystemForEmbassy/index.ts
+++ b/container-runtime/src/Adapters/Systems/SystemForEmbassy/index.ts
@@ -405,6 +405,9 @@ export class SystemForEmbassy implements System {
     if (this.manifest.id === "monerod") {
       this.manifest.id = "monerod-legacy"
     }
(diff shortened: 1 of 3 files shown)


dr-bonez (approved):

helix-nine:
Thanks for the review, @dr-bonez — merged. 🎉

PR #3267, opened 2026-05-26
feat: run a dependency's action directly when marked public

helix-nine:
## Summary

Lets a dependent service run a dependency's action **directly** via `effects.action.run({ packageId, actionId, input })`, instead of only being able to create a task against it — gated on the action being marked **public** by the package that owns it.

- **`ActionMetadata` gains a `public: bool`** (`core/src/db/model/package.rs`), `#[serde(default)]` so existing DB entries and RPC payloads omitting it deserialize as `false`.
- **`run_action` effect** (`core/src/service/effects/action.rs`) no longer unconditionally rejects cross-package calls. For `packageId != caller`, it looks up the target action's metadata and runs it only if `public == true`; otherwise it errors. The cross-package execution path (previously dead code after the hard `return Err`) is now live. Public runs go through the owning service, so the action's `visibility` (disabled) and `allowedStatuses` checks still apply.
- **i18n** retires the now-false `calling-actions-on-other-packages-unsupported` key for `action-not-public` (all five locales).
- **SDK** (`Action.withInput`/`withoutInput`) accepts an optional `public` field, defaulted to `false` at export (same mechanism as `hasInput`). `effects.action.run` already accepted `packageId`, so no Effects-type change.
- **SystemForEmbassy** defaults legacy v1 actions to `public: false` — they have no public-action concept.

Authorization is the `public` flag alone: any installed package may run an action its owner has marked public, consistent with how `createTask` already works (no dependency-relationship check). No web UI changes.

Generated bindings (`ActionMetadata.ts`) are regenerated via `make ts-bindings`.


dr-bonez:
@helix-nine I'd prefer an enum: 
```ts
access?: 'public' | 'dependent' | 'user' // defaults to user
```

helix-nine:
Done — pushed 8ba0b09 to `dependent-public-actions`.

`ActionMetadata.public: boolean` is replaced with `access: 'public' | 'dependent' | 'user'` (default `'user'`, kebab-case via `serde(rename_all = "kebab-case")`, `#[serde(default)]` so existing entries and v1 actions deserialize as `'user'`).

`run_action`'s cross-package branch now matches on the enum:
- `'public'` — any installed package may invoke directly.
- `'dependent'` — checks the caller's `current_dependencies` for the target package; rejects otherwise.
- `'user'` — rejects (other services must `createTask`).

Notes:
- i18n: dropped `action-not-public`; added `action-not-accessible` (user-only) and `action-only-accessible-to-dependents` (dependent-only).
- SDK: `Action.withInput`/`withoutInput` metadata accepts `access?: ActionAccess`; defaulted at export. Also threaded the optional `access` through the `StartSdk.Action.withoutInput` wrapper (which still required `public` in its `Omit<..., 'hasInput'>` signature on the prior commit).
- `SystemForEmbassy` legacy v1 actions and the web mock fixtures now use `access: 'user'`.
- `make ts-bindings` regenerated; new `ActionAccess.ts` is in the SDK osBindings.

Verified `cargo check --lib`, `sdk/base` + `sdk/package` `tsc --noEmit` and `npm test`, `container-runtime` `tsc --noEmit`, and `web` `check:ui`/`check:shared`/`check:setup` all pass.

dr-bonez:
no, we want `Option<ActionAccess>` with `#[ts(optional)]` that we `.unwrap_or_default()`, so that sdk consumers can choose not to specify

helix-nine:
Done — pushed a2d7ea1.

`ActionMetadata.access` is now `Option<ActionAccess>` with `#[ts(optional)]`. The TS binding is `access?: ActionAccess` and the cross-package `run_action` check reads it with `.unwrap_or_default()`.

Consequences:
- SDK consumers omit the field entirely. Removed the `m.access ?? 'user'` defaulting in `Action.withInput`/`withoutInput`; the metadata param is back to plain `Omit<T.ActionMetadata, 'hasInput'>`.
- Reverted the `StartSdk.Action.withoutInput` wrapper to the same simple `Omit<...>` shape.
- Dropped the explicit `access: 'user'` from the `SystemForEmbassy` v1 adapter and the web mock fixtures.

Verified `cargo check --lib`, all SDK/`container-runtime`/web typechecks pass.

dr-bonez:
rebase against master, and update the changelog to match the new type

helix-nine:
Rebased onto `origin/master` (clean — only #3251 landed in between) and updated the SDK changelog entry under `## 2.0.0` to describe the `access: 'public' | 'dependent' | 'user'` enum (optional, defaults to `'user'`) instead of the old `public: boolean`. Force-pushed as `97846d1`.

MattDHill:
Looks like we still have a conflict, or it's a new one since the last rebase. Either way fix it

helix-nine:
@MattDHill good catch — #3268 (`ValidateExVerRange`) added a bullet right next to mine in `sdk/CHANGELOG.md`'s 2.0.0 → `### Added`. Rebased onto the latest `origin/master`, kept both entries, force-pushed as `f0f6e2f`. PR now reports `mergeable: true`. `cargo check --lib`, `sdk/{base,package}` `tsc --noEmit`, `container-runtime` `tsc --noEmit`, and `web` `check:ui` all clean.

dr-bonez (approved):

helix-nine:
Thanks for the review and merge!

PR #3268, opened 2026-05-26
feat(sdk): compile-time-validate version-range literals (versionRange + migrations.other keys)

helix-nine:
## Summary

Adds `ValidateExVerRange<T>`, a type-level validator for exver version-**range** string literals, and wires it into the two places a package author writes a range:

- a dependency's **`versionRange`** — via `sdk.setupDependencies` (a `const R` generic plus a thin wrapper in `StartSdk` so `Manifest` stays fixed while `R` is inferred from the callback's returned object)
- the keys of **`migrations.other`** — via a `const Other` generic on `VersionInfo.of`

The validator covers the full range grammar: comparison / caret / tilde anchors over `[#flavor:]upstream[:downstream]` specs, `&&` / `||`, parentheses, negation, `#flavor` atoms, and `*`. It catches a malformed version atom **anywhere** in the expression (including inside boolean/paren structures) while accepting every valid range. A non-literal `string` passes through unchecked, and purely **structural** mistakes (unbalanced parens, a dangling operator) remain caught at runtime.

Purely additive types — **no runtime behavior changes**, no Rust/bindings changes. `satisfies` is intentionally untouched (stays a concrete version).

Scope note: this validates version atoms across the full grammar; it does not enforce structural grammar (paren balance, operator placement), which is left to runtime `VersionRange.parse`.

## Verification

- SDK `base` + `package` `tsc` pass, including type-tests that accept valid full-grammar ranges/keys and reject malformed-version cases.
- container-runtime `tsc` clean.
- Tested against the **bitcoin-knots** package (uses both fields): its real values — `versionRange: '>=0.4.9.5:0'` and `migrations.other` keys `^28`/`^29`/`^30`/`^31`/`^#knotsrdts:29.3` — compile clean against the built SDK, and injected typos in either field are rejected through the emitted declarations.


dr-bonez (approved):

helix-nine:
Thanks for the review and merge, @dr-bonez!

PR #3270, opened 2026-05-27
feat(sdk,core): MultiHost.bindPortRange for contiguous TCP+UDP port ranges

### LINKED ISSUE #3269: feat(SDK): MultiHost.bindPortRange for binding contiguous port ranges
## Summary

Add a `bindPortRange` method on `MultiHost` for binding a contiguous range of ports in a single call, alongside the existing `bindPort`.

Proposed API (per @matthill in the design discussion):

```ts
await sdk.MultiHost.of(effects, 'ui-multi').bindPortRange({
  internalStartPort: number,
  externalStartPort: number,
  numberOfPorts: number,
})
```

## Motivation

Real-time communication servers (TURN/STUN, WebRTC media servers, SIP RTP relays) require a contiguous range of UDP ports to be reachable from the public internet — they allocate per-session relay/media ports out of that range. Binding each port individually with `bindPort` doesn't scale: even a modest range (say 100 ports) produces 100 binding records, and large defaults (coturn's upstream 49152–65535 is 16384 ports) are flatly infeasible.

## Consumers (current + planned)

- **coturn-startos** (new — gating SimpleX self-hosted ICE) — coturn's `--min-port`/`--max-port` relay range
- **jitsi-startos** — its sidecar coturn (currently shipped; would be removed once coturn-startos exists)
- **simplex-startos** — via dependency on coturn-startos for self-hosted ICE servers
- **synapse-startos** (planned) — via dependency on coturn-startos
- **nextcloud-startos / Talk** (planned) — Janus signalling backend's RTP port range (default 20000–40000)
- **mattermost-startos** (planned) — Calls plugin UDP RTP range (configurable, contiguous)
- **BigBlueButton, Element Call / LiveKit, Asterisk/FreePBX** — all use contiguous UDP RTP ranges

## Behavior expectations

- Forwards both TCP and UDP for every port in the range (matching existing `bindPort` behavior).
- Surfaces as a single binding record in the admin UI, not N records.
- Returns a single origin/receipt object suitable for use with `.export([...])` if the package wants to attach interfaces.

## Open questions for design

1. **External-port collision policy.** What happens if part of `[externalStartPort, externalStartPort + numberOfPorts)` is already taken? Hard-fail, or fall back to a `preferredExternalStartPort` semantics like `bindPort` has today?
2. **Maximum range size.** Should there be a sanity cap (e.g. 1000 ports) to prevent abuse / resource exhaustion? coturn's full upstream default is 16384, but in practice a single-user box only needs ~100.
3. **Per-port options.** Options like `addSsl` / `secure` presumably apply uniformly across the whole range — worth being explicit so the API contract is unambiguous.

## Blocking

`coturn-startos` (and therefore SimpleX self-hosted ICE) is gated on this feature landing. Discussion thread: Matrix.

helix-nine:
## Summary

Adds `sdk.MultiHost.of(effects, hostId).bindPortRange({ internalStartPort, externalStartPort, numberOfPorts })`, intended for real-time / WebRTC servers (coturn, RTP, SIP) that need a contiguous public port range. Single-port `bindPort` doesn't scale to 100+ ports — every port becomes its own binding record and its own iptables rule.

Fixes #3269. Gating consumer: coturn-startos (and downstream SimpleX / synapse / nextcloud Talk / mattermost Calls / BBB / Element Call / FreePBX).

## Design choices

Per @matthill's answers on the open questions in #3269:

| Question | Decision |
|---|---|
| Collision policy | **Hard-fail** on any partial collision with already-allocated external ports — no shifted-range fallback. A range with holes defeats the use case; a silently-shifted range breaks the package's config. |
| Max range size | **500 ports** per call, enforced at both the SDK and effect layer. |
| Per-port options | **None.** No `addSsl`, no `secure`, no `preferredExternal*` fallback — these are HTTP-layer concerns that don't apply to UDP RTP ranges. |
| Data model | New `Host.binding_ranges` map alongside `Host.bindings`, keyed by internal start port. One record per range, not N. |
| Forwarding | One iptables rule per protocol per gateway covering the whole range (\`--dport N:M\` matcher + \`--to-destination \$dip\` with no port, so iptables preserves the destination port across the range). |
| \`.export([...])\` | Returns \`Promise<void>\`. Range bindings have no \`Origin\` because they aren't addressable as HTTP-style service interfaces. |

Additional iptables-driven constraint: \`internalStartPort\` must equal \`externalStartPort\` (port preservation requires equal-base ranges). The two fields are kept separate in the API for symmetry with \`bindPort\` and to leave room for a future relaxation. Enforced at the effect layer with a clear error.

## Touch points

**Rust backend** (\`core/\`):
- \`net/host/binding.rs\` — new \`RangeBindInfo\` + \`BindingRanges\` types.
- \`net/host/mod.rs\` — \`Host.binding_ranges\` field; \`Model<Host>::add_binding_range\`; \`update_addresses\` emits range entries into \`port_forwards\`.
- \`net/forward.rs\` — \`AvailablePorts::try_alloc_range\` (atomic — partial collisions don't leak allocations); \`PortForwardController::add_forward_range\`; \`InterfacePortForwardController::add_range\`; \`ForwardMapping\` / \`InterfaceForwardEntry\` gain a \`count\` field.
- \`net/net_controller.rs\` — \`NetService::bind_range\`; \`HostBinds.forwards\` includes count; \`update\` reconciles range bindings to a single \`InterfaceForwardEntry\`; \`clear_bindings\` disables ranges.
- \`service/effects/net/bind.rs\` — new \`bind_range\` effect with \`BindRangeParams\` + \`MAX_BIND_PORT_RANGE_SIZE = 500\` constant; validates equal start ports / cap / non-zero count.
- \`service/effects/mod.rs\` — wires the \`bind-range\` subcommand.
- \`db/model/public.rs\` — default \`binding_ranges: Default::default()\` in the initial Host construction.

**iptables script** (\`build/lib/scripts/forward-port\`):
- New \`count\` env var. When > 1, uses range-syntax matchers and DNAT without a destination port (port preservation); when ==1, falls through to the existing single-port path. \`conntrack -D\` loop on undo iterates the range.

**SDK** (\`sdk/\`):
- \`base/lib/Effects.ts\` — \`bindRange(options: BindRangeParams): Promise<null>\`.
- \`base/lib/interfaces/Host.ts\` — \`MultiHost.bindPortRange(...)\`, plus exported \`MAX_BIND_PORT_RANGE_SIZE\` constant mirroring the Rust cap; client-side validation for the equal-start / count cap / range bounds.
- \`package/lib/StartSdk.ts\` — \`bindRange\` added to \`InterfaceEffects\` so it stays off the \`sdk.\` namespace (use \`sdk.MultiHost.of(...).bindPortRange(...)\` instead).
- \`package/lib/test/host.test.ts\` — jest coverage for the four SDK-side validation paths.
- \`base/lib/test/startosTypeValidation.test.ts\` — adds \`bindRange: {} as BindRangeParams\` to keep the structural Effects↔Params check honest.
- \`CHANGELOG.md\` — entry added under the existing in-flight \`2.0.0\` heading (current published is \`1.5.3\`).

**Container runtime** (\`container-runtime/\`):
- \`Adapters/EffectCreator.ts\` — wires the \`bindRange\` effect to \`bind-range\` RPC.

**Web UI** (\`web/projects/ui\`):
- \`port-forwards.component.ts\` — renders range rows as \`N-M\` (external and internal). \"Test\" button checks the range's start port (CheckPort is per-port).
- \`mock-patch.ts\` / \`api.fixures.ts\` — add \`bindingRanges: {}\` to every \`Host\` and \`count: 1\` to every \`PortForward\`.

## Verification

- \`cargo check --lib\` + \`cargo check --bin startbox\` clean.
- \`cargo test --lib --features test net::host net::forward\` — 27/27 pass (5 new for \`try_alloc_range\`: basic, zero-count, overflow, restricted-port collision atomicity, port collision atomicity).
- \`cargo test --lib --features test export_bindings_\` — ts-rs bindings regenerate cleanly; new \`BindRangeParams\`, \`BindingRanges\`, \`RangeBindInfo\`, \`Host.bindingRanges\`, \`PortForward.count\` synced to \`sdk/base/lib/osBindings/\`.
- \`cd sdk/base && npm test\` — 73/73 pass.
- \`cd sdk/package && npm test\` — 62/62 pass (includes 4 new \`bindPortRange\` tests).
- \`cd sdk/base && npm run check\` + \`cd sdk/package && npm run check\` clean.
- \`make -C sdk baseDist dist\` builds cleanly.
- \`cd container-runtime && npm run check\` clean.
- All four \`web/\` tsconfig projects (\`ui\`, \`tunnel\`, \`setup-wizard\`, \`shared\`, \`marketplace\`) tsc clean.

## Test plan

- [ ] Build an ISO and install on a VM.
- [ ] Build a test package against this SDK that calls \`bindPortRange({ internalStartPort: 49152, externalStartPort: 49152, numberOfPorts: 100 })\`.
- [ ] Verify a single \`RangeBindInfo\` appears in the host's \`bindingRanges\` (one record, not 100).
- [ ] Verify \`iptables -t nat -L -n\` shows one DNAT rule per protocol per gateway covering 49152:49251 (not 100 rules per protocol).
- [ ] Verify the gateway port-forwards UI shows a single row reading \"49152-49251\" → \"49152-49251\" rather than 100 rows.
- [ ] Send UDP traffic to gateway:49152 and gateway:49250 and confirm both reach the container.
- [ ] Re-call \`bindPortRange\` with the same arguments — should be a no-op (idempotent).
- [ ] Re-call \`bindPortRange\` with overlapping range — should hard-fail with a clear error and leave the existing range intact.
- [ ] Re-call \`bindPortRange\` with the same internal start but different size — should free the old allocation, allocate the new one.
- [ ] Uninstall the package — range allocation is released; iptables rules removed; \`conntrack\` flush completes without hanging.


helix-nine:
## Rebased onto the nftables migration (#3287) + dropped the equal-start constraint

This PR is now **stacked on #3287** (base retargeted to `feat/nftables-migration`) and forwarding is native nftables. With that in place, `bindPortRange` no longer requires `internalStartPort == externalStartPort`.

**How offset ranges work:** `forward-port` is Hybrid —
- **equal bases** (the common coturn case): a single port-preserving rule (`th dport S-E dnat to <ip>`), no enumeration;
- **offset bases**: a single rule carrying an nft verdict map (`dnat to th dport map { ext_i : <ip> . int_i }`), one map element per port.

Either way it's **one rule + one `RangeBindInfo` record**. The verdict-map cost is O(1) per-packet (hashmap) and ~tens of KB only for an actual offset range (≤500 ports); equal-base ranges pay none of it.

> Note on mechanism: nft does **not** expose iptables' single `/baseport` offset DNAT — a plain `dnat to ip:lo-hi` pools rather than offset-maps (verified). The per-port verdict map is the way to get a deterministic offset in one rule.

**Equal-start checks removed:** `forward()` guard, `Model<Host>::add_binding_range`, the `bind-range` effect, and the SDK `MultiHost.bindPortRange` throw. The SDK now validates that *both* ranges are integers > 1024 with end ≤ 65535. `host.test.ts`'s "rejects mismatched start ports" became "accepts an offset range". CHANGELOG updated.

### Verification
- `cargo check` (lib + startbox), `sdk` base + package jest (offset test included) and tsc — all clean.
- netns packet-flow against nftables v1.1.3: equal range `7000-7002 → 7000-7002` preserves; offset range `5000-5002 → 6000-6002` maps by offset; tag-based undo removes only the targeted range.

Merge order: **#3287 first, then this.**

sdk/base/lib/interfaces/Host.ts
```diff
@@ -80,6 +80,27 @@ export type BindOptionsByProtocol =
...
+  internalStartPort: number
+  /**
+   * First external (host-side / WAN) port in the range. May differ from
+   * `internalStartPort`: the forward maps the external range onto the
+   * internal range by offset (`externalStartPort + i` → `internalStartPort
+   * + i`) via an nft verdict map. Use the same value for the common
+   * port-preserving case.
+   */
+  externalStartPort: number
+  /** Number of contiguous ports in the range. Must be in `[1, 500]`. */
```
dr-bonez: minimum should be 2. single should be a regular bind

helix-nine: Done in 345acaee3 — minimum is now 2. Single-port callers get `bindPort`; the guard rejects `numberOfPorts < 2` with a message pointing at `bindPort`.

Raised the floor in all three places that enforced it:
- SDK `bindPortRange` guard (`< 2`)
- core `bind_range` effect
- core `Host::add_binding_range`

Updated the type/JSDoc (`[2, 500]`), the CHANGELOG ("2–500 ports; for a single port use `bindPort`"), and the test ("rejects a single-port range"). `cargo check` + SDK jest/tsc all green.

dr-bonez (approved):

PR #3279, opened 2026-05-29
feat(tunnel): add per-subnet DNS proxy

MattDHill:
## Summary

Two related changes that give StartTunnel DNS and unify domain UX across gateway types.

### 1. Per-subnet DNS proxy in StartTunnel

Operators pick a DNS mode per WireGuard subnet, and clients resolve DNS through the tunnel.

**Modes** (per subnet):
- **default** — forward to the VPS's own system resolvers (`/run/systemd/resolve/resolv.conf`, with the `127.0.0.53` stub filtered out to avoid a loop)
- **device** — forward to a selected device on the subnet (its WireGuard IP)
- **custom** — forward to up to 3 operator-entered servers (IP, optional `:port`)

**How it works**
- Client configs now emit `DNS = <subnet .1>`, so clients send DNS into the tunnel.
- A lightweight forward-only `hickory` proxy (one root `ForwardAuthority` per subnet) binds to that subnet's in-tunnel address (`10.59.x.1:53`, UDP + TCP). Binding to the in-tunnel address — never `0.0.0.0` — keeps it reachable only over WireGuard, never the WAN (no open resolver).
- The proxy rebinds after `wg-quick` whenever subnets change; `subnet set-dns` re-syncs only the proxy (no tunnel bounce, since the `DNS=` line is mode-independent).
- No new dependency — `hickory-server` already links into the binary; the forwarding path mirrors StartOS's `net/dns.rs`.

New `DnsConfig` model on `WgSubnetConfig` and a `subnet set-dns` RPC/CLI command, with regenerated TS bindings, i18n (all 5 locales), and man pages. No migration needed — `#[serde(default)]` covers existing data. Web: DNS is configured in the add/edit subnet form (overflow action is now "Edit"); the subnets table shows a DNS column; all StartTunnel select dropdowns disable the textfield clear-✕ (`tuiTextfieldCleaner`), matching StartOS's `form-select`.

### 2. Private domains on StartTunnel gateways (StartOS UI)

Private domains were gated to router gateways because StartTunnel had no DNS. Now that StartTunnel proxies DNS (above), a private domain resolves for tunnel clients — their DNS forwards to this server, which answers with the gateway's tunnel address. The backend already accepted any gateway (no type check in `add_private_domain`; `check_dns` works over the tunnel interface), so this drops the UI-only restriction:

- All gateways now use the public/private domain picker (removed the wireguard-only "Add Public Domain" branch; removed the now-unused `isWireguard` flag).
- The "DNS Server Config" guidance is gateway-aware: for a tunnel gateway it tells the user to point the StartTunnel subnet's DNS at this server (the new DNS settings), rather than the router-centric "LAN IP" wording.

### Upgrade note
Existing subnets migrate to `default`. Already-deployed client `.conf` files are unaffected until re-downloaded; once re-downloaded they begin sending DNS into the tunnel (→ the VPS's resolvers). To resolve private domains on a tunnel gateway, set that subnet's DNS to this server in StartTunnel.


dr-bonez (approved):

dr-bonez (changes requested):
We might need handling for if the system doesn't use systemd-resolved (probably just fall back to parsing /etc/resolv.conf)

dr-bonez (dismissed):

PR #3290, opened 2026-06-03
feat(sdk): ship s9pk.mk and tsconfig.base.json for packages to consume by reference

MattDHill:
## Summary

Every StartOS package currently vendors byte-identical copies of `s9pk.mk` and `tsconfig.json` at its root. Any change to the build plumbing therefore means a fleet-wide edit across ~145 repos, and in practice the copies have drifted.

This ships both canonical files **inside the SDK package** so a package can consume them by reference instead of vendoring:

- **`s9pk.mk`** — a package's `Makefile` does `include node_modules/@start9labs/start-sdk/s9pk.mk` (with a small `npm ci` bootstrap so a missing file self-heals via make's auto-remake).
- **`tsconfig.base.json`** — a package's `tsconfig.json` does `"extends": "@start9labs/start-sdk/tsconfig.base.json"` and keeps only its local `include`.

Bumping the SDK then delivers build-system fixes automatically, ending the per-package drift.

### Changes

- Add `sdk/s9pk.mk` (canonical, identical to the current package copy) and `sdk/tsconfig.base.json` (the shared `compilerOptions`).
- `dist` target copies both into the publishable bundle. `.npmignore` only excludes `node_modules/`, so they publish at the package root → installed at `node_modules/@start9labs/start-sdk/`.
- CHANGELOG note under the in-flight `2.0.0` section.

### Rollout

Lands with **2.0.0**. Packages adopt the `include`/`extends` switch (and delete their vendored `s9pk.mk`) as part of their SDK 2.0.0 migration — **not** before, since the consumer change requires an installed SDK that contains these files.


dr-bonez:
how are we addressing bootstrap? is `s9pk init-package` automatically running `npm ci`?

dr-bonez (approved):

PR #3291, opened 2026-06-03
fix(cli): lock cookie store on a stable file to stop publish race

MattDHill:
## Summary

`start-cli` persists its cookie store in `Drop for CliContextSeed` via write-temp-then-rename. `cookie_path` derives from `$HOME`, so concurrent `start-cli` invocations share it — and the release pipeline's publish step runs one `s9pk publish` per arch **in parallel** under a single `$HOME`. The existing `FdLock` was held on the *temp file itself*, which each writer recreates, so it gave no cross-process protection: the two writers raced on `rename(tmp → cookies.json)`, the loser found its temp already moved away, and the `.unwrap()` turned that `ENOENT` into a panic — failing the publish step *after* the package had already been uploaded and registered.

Observed in the wild (e.g. [filebrowser-startos run](https://github.com/Start9Labs/filebrowser-startos/actions/runs/26898005709/job/79342660792)):

```
The application panicked (crashed).
Message:  called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, ... }
Location: src/context/cli.rs:74   # std::fs::rename(...) at the released tag (v0.4.0-beta.9)
```

## Change

- Hold the exclusive `FdLock` on a **stable sidecar lock file** (`<cookie_path>.lock`, never renamed away) across the write+rename, so concurrent invocations serialize their cookie-store saves.
- Move the save into a `Result`-returning `save_cookie_store()` and have `Drop` `log_err()` it instead of `unwrap()`-ing, so a cookie-cache I/O hiccup can never crash an otherwise-successful command.

The lock guards the save (last-writer-wins on the cache, which is fine — registry publish uses developer-key signing, not cookies), not load→modify→save, to avoid serializing the pipeline's parallel publishes entirely.

### DIFF
--- a/core/src/context/cli.rs
+++ b/core/src/context/cli.rs
@@ -58,23 +58,48 @@ impl Drop for CliContextSeed {
                 rt.shutdown_background();
             }
         }
-        let tmp = format!("{}.tmp", self.cookie_path.display());
+        // Failing to persist the cookie cache must not crash an otherwise
+        // successful command (e.g. `s9pk publish`), so log instead of panicking.
+        self.save_cookie_store().log_err();
+    }
+}
+impl CliContextSeed {
+    /// Persist the cookie store to `cookie_path` atomically (write-temp-then-rename),
+    /// serialized across processes by an exclusive lock on a *stable* sidecar lock
+    /// file. `cookie_path` is derived from `$HOME`, so concurrent `start-cli`
+    /// invocations (the release pipeline publishes one s9pk per arch in parallel)
+    /// share it. The lock must be held on a file that is never renamed away — the
+    /// previous code locked the temp file itself, which each writer recreates, so it
+    /// served no cross-process purpose and the writers raced on the rename: whoever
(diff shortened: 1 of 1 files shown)


dr-bonez:
saved cookie stores should probably be merged with what is on disk rather than overwriting, for more correct concurrency behavior

dr-bonez:
@helix-nine

helix-nine:
Good call — done in 1fdb2b03.

Under the save lock, `save_cookie_store` now re-reads the on-disk store and merges our in-memory cookies onto it via `CookieStore::from_cookies`, chaining our cookies last so they win per-cookie conflicts (last-writer-wins on a given domain/path/name), while cookies a concurrent writer added for *other* domains are preserved instead of being clobbered by the rename. Since this is a read-merge-write, it's still fully serialized by the existing exclusive lock on the stable sidecar, so there's no torn-read window.

Compiles clean (`cargo check -p start-os --lib`).

dr-bonez (approved):

PR #3292, opened 2026-06-03
fix(net): serve private DNS for domains also configured as public

### LINKED ISSUE #3263: [bug]: Private domain DNS not served locally when the same domain is also configured as a public (clearnet) domain on a StartTunnel gateway
### Prerequisites

### Server Hardware

- StartOS: Proxmox VM (arm64, FriendlyElec CM3588)
- Router: GL-iNet GL-X3000
- StartTunnel VPS: Digital Ocean droplet
- DNS registrar: Cloudflare (proxy off), domains via Namecheap

### StartOS Version

0.4.0-beta.9

### Client OS

Linux

### Client OS Version

Ubuntu 24

### Browser

Firefox

### Browser Version

151

### Current Behavior

When the same domain (e.g., `cloud.mydomain.com`) is configured as both a private domain on the Ethernet gateway and a public domain on the StartTunnel gateway for the same service, the StartOS DNS server does not return the LAN IP for that domain. Instead, it forwards the query upstream and returns the public VPS IP — the same result you'd get from any external DNS resolver.

Querying StartOS DNS directly confirms this:

```
$ dig @<STARTOS_LAN_IP> cloud.mydomain.com

;; ANSWER SECTION:
cloud.mydomain.com.   300   IN   A   <VPS_WAN_IP>

;; flags: qr rd ra;
;; Query time: 141 msec
```

The response is non-authoritative (`ra` flag, no `aa` flag), with a 141ms query time and 300s TTL — consistent with a forwarded upstream lookup rather than a locally served record. StartOS DNS does not appear to be creating a local override for the domain.

As a result, LAN clients that use StartOS as their DNS server still route traffic to the VPS public IP, causing hairpin routing: traffic exits the home network to the VPS, traverses the WireGuard tunnel back into the home network, and the response reverses the same path. Each request traverses the ISP connection 4 times, resulting in significant latency and throughput bottleneck (limited by upload speed).

### Expected Behavior

When a domain is configured as a private domain on the Ethernet gateway, StartOS DNS should serve an authoritative local A record pointing to the server's LAN IP for that domain, regardless of whether the same domain is also configured as a public domain on a different gateway. LAN clients using StartOS as their DNS server should resolve the domain to the LAN IP and connect directly, while external clients continue to resolve via public DNS to the VPS IP.

This would provide native split DNS: same domain, same TLS certificates, LAN speeds when local, clearnet access when remote.

This is particularly important for services like Nextcloud and Immich that embed the access URL in generated share links. If the user accesses the service via a private-only domain (e.g., `cloud.mylan`), LAN IP, or `.local` address, share links contain that address and are unusable by external recipients. Accessing via the public domain on the LAN (through hairpin routing) works but is unacceptably slow. The ability to use the same public domain as a private domain would solve both problems — correct share links and LAN-speed access.

### Steps to Reproduce

1. Configure a service (e.g., Nextcloud) with a public domain on the StartTunnel gateway (e.g., `cloud.mydomain.com`) — confirm it works from external networks
2. On the same service, add the same domain (`cloud.mydomain.com`) as a private domain on the Ethernet (Wired Connection 1) gateway
3. Configure StartOS DNS Strategy as Static with upstream servers `1.1.1.1:53` and `8.8.8.8:53`
4. Set the router's DNS server to the StartOS LAN IP (e.g., `192.168.8.228`)
5. From a LAN client, query StartOS DNS directly: `dig @192.168.8.228 cloud.mydomain.com`
6. Observe: The response returns the VPS WAN IP, not the StartOS LAN IP. The `aa` (authoritative answer) flag is absent.

Router configuration (GL-iNet GL-X3000):
- DNS Server Mode: Manual DNS
- DNS Server 1: `192.168.8.228` (StartOS LAN IP)
- All other DNS server fields blank
- DNS Rebinding Attack Protection: OFF
- Override DNS Settings for All Clients: OFF

### Anything else?

The private domain DNS test within StartOS passes when the domain is added, suggesting StartOS believes the configuration is correct.

Private domains with unique names work correctly. If a domain that is not also a public domain is used as a private domain (e.g., `cloud.mylan`), StartOS DNS serves the LAN IP as expected and the service is reachable. The issue is specific to the case where the same domain exists on both gateways.

Confirmed the dev's expectation that this should work. In a support conversation, the lead StartOS/StartTunnel developer stated: "No, it should work to put the same domain for both public and private."

helix-nine:
## Problem

Closes #3263.

When a domain is configured as **both** a private domain (e.g. on the Ethernet/Wired gateway) and a public domain (on a StartTunnel gateway), StartOS DNS did not serve the LAN IP for that domain. It forwarded the query upstream and returned the public VPS IP, so LAN clients using StartOS as their resolver suffered hairpin routing (traffic exits to the VPS and tunnels back).

## Root cause

`NetController::update` (`core/src/net/net_controller.rs`) collects the set of domains to register as authoritative private records, but the loop only inserted addresses where `public.is_none()`:

```rust
for HostAddress { address, public, .. } in &host_addresses {
    if public.is_none() {
        private_dns.insert(address.clone());
    }
}
```

`Host::addresses()` emits a single `HostAddress` per public domain with `public: Some(..)` **and** `private: Some(..)` when the same domain also has private gateways. So a dual public+private domain has `public.is_some()` and was skipped — never added to the DNS resolver's `private_domains` map. The resolver (`core/src/net/dns.rs`) only serves a local LAN answer for names present in that map, so the query fell through to the upstream forwarder.

This already-intended capability is confirmed by `handle_duplicates` in `core/src/net/host/address.rs`, which deliberately allows the same domain to be both public and private (it skips the duplicate-domain error for private domains that are also public).

## Fix

Register a domain as private DNS whenever it has a private gateway config, regardless of whether it is also served publicly:

```rust
for HostAddress { address, private, .. } in &host_addresses {
    if private.is_some() {
        private_dns.insert(address.clone());
    }
}
```

This enables native split DNS: LAN clients resolve the domain to the server's LAN IP and connect directly (same domain, same TLS cert, LAN speeds), while external clients continue to resolve via public DNS to the VPS.

## Testing

- `cargo check -p start-os --lib` passes (only pre-existing warnings).
- Behavioral verification (install/runtime DNS resolution with a dual public+private domain) was not run in this environment — it needs a StartOS host with a StartTunnel gateway and a LAN client. Worth a manual `dig @<startos-lan-ip> <domain>` check confirming the `aa` flag and LAN IP after this change.

### DIFF
--- a/core/src/net/dns.rs
+++ b/core/src/net/dns.rs
@@ -1,5 +1,5 @@
 use std::borrow::Borrow;
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, BTreeSet};
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
 use std::str::FromStr;
 use std::sync::{Arc, Weak};
@@ -208,7 +208,7 @@ pub async fn dump_table(
 
 #[derive(Default)]
 struct ResolveMap {
-    private_domains: BTreeMap<InternedString, Weak<()>>,
+    private_domains: BTreeMap<InternedString, (BTreeSet<GatewayId>, Weak<()>)>,
     services: BTreeMap<Option<PackageId>, BTreeMap<Ipv4Addr, Weak<()>>>,
     challenges: BTreeMap<InternedString, (InternedString, Weak<()>)>,
 }
@@ -387,14 +387,19 @@ impl Resolver {
             a => a,
         };
         self.resolve.peek(|r| {
-            if !src.is_loopback()
-                && r.private_domains
-                    .get(&*name.to_lowercase().to_utf8().trim_end_matches('.'))
-                    .map_or(false, |d| d.strong_count() > 0)
-            {
+            let private_gateways = (!src.is_loopback())
+                .then(|| {
+                    r.private_domains
+                        .get(&*name.to_lowercase().to_utf8().trim_end_matches('.'))
+                        .filter(|(_, rc)| rc.strong_count() > 0)
+                        .map(|(gateways, _)| gateways)
+                })
+                .flatten();
+            if let Some(gateways) = private_gateways {
                 if let Some(res) = self.net_iface.peek(|i| {
-                    i.values()
-                        .filter_map(|i| i.ip_info.as_ref())
+                    i.iter()
+                        .filter(|(gid, _)| gateways.contains(*gid))
+                        .filter_map(|(_, i)| i.ip_info.as_ref())
                         .find(|i| i.subnets.iter().any(|s| s.contains(&src)))
                         .map(|ip_info| {
                             let mut res = ip_info.subnets.iter().collect::<Vec<_>>();
@@ -405,10 +410,11 @@ impl Resolver {
                     return Some(res);
                 } else if is_private_ip(src) {
                     // Source is a private IP not in any known subnet (e.g. VPN on a different VLAN).
-                    // Return all private IPs from all interfaces.
+                    // Return private IPs from the gateways this domain is served on.
                     let res: Vec<IpAddr> = self.net_iface.peek(|i| {
-                        i.values()
-                            .filter_map(|i| i.ip_info.as_ref())
+                        i.iter()
+                            .filter(|(gid, _)| gateways.contains(*gid))
+                            .filter_map(|(_, i)| i.ip_info.as_ref())
                             .flat_map(|ip_info| ip_info.subnets.iter().map(|s| s.addr()))
                             .collect()
                     });
@@ -694,15 +700,20 @@ impl DnsController {
         }
     }
 
-    pub fn add_private_domain(&self, fqdn: InternedString) -> Result<Arc<()>, Error> {
+    pub fn add_private_domain(
+        &self,
+        fqdn: InternedString,
+        gateways: BTreeSet<GatewayId>,
+    ) -> Result<Arc<()>, Error> {
         if let Some(resolve) = Weak::upgrade(&self.resolve) {
             resolve.mutate(|writable| {
-                let weak = writable.private_domains.entry(fqdn).or_default();
-                let rc = if let Some(rc) = Weak::upgrade(&*weak) {
+                let entry = writable.private_domains.entry(fqdn).or_default();
+                entry.0 = gateways;
+                let rc = if let Some(rc) = Weak::upgrade(&entry.1) {
                     rc
                 } else {
                     let new = Arc::new(());
-                    *weak = Arc::downgrade(&new);
+                    entry.1 = Arc::downgrade(&new);
                     new
                 };
                 Ok(rc)
@@ -754,7 +765,7 @@ impl DnsController {
             resolve.mutate(|writable| {
                 for domain in domains {
                     if let Some((k, v)) = writable.private_domains.remove_entry(domain) {
-                        if v.strong_count() > 0 {
+                        if v.1.strong_count() > 0 {
                             writable.private_domains.insert(k, v);
                         }
                     }
--- a/core/src/net/dns.rs
+++ b/core/src/net/dns.rs
@@ -1,6 +1,6 @@
 use std::borrow::Borrow;
 use std::collections::{BTreeMap, BTreeSet};
-use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV6};
 use std::str::FromStr;
 use std::sync::{Arc, Weak};
 use std::time::Duration;
@@ -23,7 +23,7 @@ use rpc_toolkit::{
     Context, HandlerArgs, HandlerExt, ParentHandler, from_fn_async, from_fn_blocking,
 };
 use serde::{Deserialize, Serialize};
-use tokio::net::{TcpListener, UdpSocket};
+use tokio::net::UdpSocket;
 use tokio::sync::RwLock;
 use tracing::instrument;
 use ts_rs::TS;
@@ -32,7 +32,7 @@ use crate::context::{CliContext, RpcContext};
 use crate::db::model::Database;
 use crate::db::model::public::NetworkInterfaceInfo;
 use crate::net::gateway::NetworkInterfaceWatcher;
-use crate::net::utils::is_private_ip;
+use crate::net::utils::{bind_tokio_listener, ipv6_is_link_local};
 use crate::prelude::*;
 use crate::util::future::NonDetachingJoinHandle;
 use crate::util::io::file_string_stream;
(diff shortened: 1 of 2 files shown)


dr-bonez (dismissed):

dr-bonez:
wait - we need to make sure that private dns is only served for the enabled private gateways

helix-nine:
Good catch — fixed in 350e49cd8.

Instead of registering private DNS from the raw `private_domains` config map, I now build the set from the **enabled** `PrivateDomain` addresses of enabled bindings (`bind.addresses.enabled()`), and only register a domain when at least one of its gateways is actually up (`net_ifaces.get(gw).ip_info.is_some()`) — the same liveness check the SSL vhost private-IP collection uses.

This honors binding-enable state, per-address disable overrides, and gateway availability. Since `update_addresses` emits a separate `PrivateDomain` available entry even for a domain that's also public, dual public+private domains are still covered.

dr-bonez:
hmm, no the dns controller needs to carry gateway sets per private dns entry, and only serve that dns when requested over that gateway

helix-nine:
Reworked to carry gateway sets in the resolver — 0bf1a67dc.

- `ResolveMap::private_domains` is now `BTreeMap<InternedString, (BTreeSet<GatewayId>, Weak<()>)>` — each private-domain entry carries the gateways it's served on.
- `resolve()` only answers locally when the query's source interface is one of that domain's gateways (`net_iface.iter().filter(|(gid, _)| gateways.contains(gid))`). Both the in-subnet match and the VPN-on-a-different-VLAN fallback are restricted to the domain's gateways instead of all interfaces.
- `add_private_domain` takes the gateway set and refreshes it on every update; `gc_private_domains` unchanged except for the tuple field.
- In `net_controller`, private DNS is collected as `domain -> live gateway set` (unioned across enabled bindings, filtered to gateways whose interface is up) and passed through on every update.

`cargo check -p start-os --lib` is clean.

dr-bonez:
This still leaks the private domains on one gateway to other gateways that can reach it. is it possible to determine the ingress interface of dns requests - not just the source ip?

helix-nine:
You're right — source IP only tells you the client's address, not the interface the packet ingressed on, so a client whose source falls in gateway A's subnet but whose packet actually arrives via gateway B gets served A's record. To fix the leak properly we need the real ingress interface.

I dug into whether we can get it. Short answer: **yes on Linux, but not through hickory's current API.**

- `hickory_server::server::Request` carries only `src: SocketAddr` + `protocol` — no destination/local address, no interface.
- Its UDP path (`register_socket` → `UdpStream`) receives with `recv_from`, which drops the destination address and all ancillary data. We bind a single wildcard `[::]:53` socket, so there's nothing per-packet to key on.

Two ways to actually get the ingress interface:

**1. `SO_BINDTODEVICE` per gateway (my recommendation).** Instead of one wildcard socket, bind a UDP+TCP socket per gateway interface bound to that device. Each socket then only receives traffic that ingressed on its interface, so the ingress gateway is known by construction — no source-IP guessing. Implementation shape: one `ServerFuture` (or at least one `Resolver`) per interface, each tagged with its `GatewayId`, all sharing the existing `Arc<SyncRwLock<ResolveMap>>`; `resolve()` serves a private domain only if its gateway set contains *this socket's* gateway. We already watch `net_iface`, so we rebuild the socket set as gateways come and go. A wildcard socket stays for non-gateway-scoped names (`*.startos`, ACME challenges) and the upstream forwarder. Needs CAP_NET_RAW, which we have as root.

**2. `IP_PKTINFO` / `IPV6_RECVPKTINFO` + `recvmsg`.** Set the sockopt and run our own UDP recv loop to read `ipi_ifindex` from the control message, map ifindex→`GatewayId`, and thread it to the handler. More faithful to "one socket" but means replacing hickory's UDP serving with our own message parse/dispatch and handling TCP ingress separately (getsockname/`SO_BINDTODEVICE`). Heavier, and we lose hickory's UDP plumbing.

I lean toward #1 — it reuses hickory and makes the gateway unambiguous rather than inferred. The cost is managing a dynamic set of per-interface sockets that tracks `net_iface` changes.

Want me to go with the `SO_BINDTODEVICE` per-interface approach, or do you prefer the `IP_PKTINFO` route?

dr-bonez:
we already do a strategy similar to #1 for our webserver and vhost controller. doing the same for dns seems reasonable. do it and follow similar patterns

helix-nine:
Done — 8b7008a09. Followed the web server / vhost per-interface pattern.

**Serving model.** DNS now runs on a wildcard catch-all socket *plus* a socket bound to each gateway interface address (every subnet addr from `net_iface`, link-local with scope). Because a socket bound to a specific address only receives datagrams sent to that address, the kernel hands a query to the socket for the interface it actually ingressed on — no source-IP inference. (`SO_REUSEADDR` lets the specific binds coexist with the wildcard; most-specific wins, same as `WildcardListener` + `VHostBindListener`.)

**Gating.** Each `Resolver` is tagged with `my_gateway: Option<GatewayId>`:
- per-interface sockets → `Some(gw)`; a private domain is answered only if its gateway set contains `gw`, and the answer is that gateway's own subnet IP. So a domain private on gateway A is never served on B even if B can route to A's subnet.
- wildcard → `None`; serves loopback, the container bridge, `*.startos`/`*.embassy`, and upstream forwarding, but **never** private domains.

**Lifecycle.** A manager task (`run_dns_servers`) watches `net_iface` and binds/unbinds per-interface `ServerFuture`s as gateways come and go (dropping a `ServerFuture` frees its sockets). `ResolveMap` and the upstream forwarder catalog (resolv.conf/static-server updater) are created once and shared across all resolvers.

This removes the old source-IP-subnet matching and its VPN-VLAN fallback entirely.

`cargo check -p start-os --lib` is clean.

One caveat: I can't exercise this at runtime here — it needs a box with multiple gateways (Ethernet + StartTunnel) and clients on different interfaces to confirm the per-socket routing and that container/host resolution still flows through the wildcard. Worth a `dig` pass on real hardware before merge. Anything in the socket model you'd want done differently?

dr-bonez:
private domains should automatically serve on `lo` and `lxcbr0` in addition to their configured gateways

helix-nine:
Done — 953a9a5cf. Loopback and the container bridge now resolve every private domain locally.

I generalized the resolver's per-socket gateway tag into a `PrivateScope`:
- **`None`** (wildcard `[::]:53`) — answers no private domains.
- **`Gateway(id)`** (per gateway-interface socket) — answers a private domain only when its gateway set contains that gateway, returning the gateway's own address.
- **`Local(addrs)`** (new) — answers *every* registered private domain, returning fixed local addresses.

Two `Local` bind sets are now always part of the desired socket set:
- **`lo`** → sockets on `127.0.0.1:53` and `::1:53`, returning `127.0.0.1` / `::1`.
- **`lxcbr0`** → socket on `HOST_IP` (`10.0.3.1:53`), returning `10.0.3.1`.

So the host (via loopback) and service containers (via the bridge) can resolve any private domain regardless of which gateway it's configured on, while LAN clients still only get a private domain on the gateways it's actually configured for. The loopback/bridge sockets are re-attempted on each `net_iface` change if their address isn't up yet at first (e.g. bridge not yet created); until then the wildcard still serves `*.startos`/forwarding there.

`cargo check -p start-os --lib` is clean. Still wants a runtime `dig` pass on real hardware (LAN client, host, and a container) before merge.

dr-bonez (approved):

helix-nine:
Thanks for the review and merge! 🎉

To recap where this landed for posterity: `cloud.example.com` configured as both a public (StartTunnel) and private (e.g. Ethernet) domain now resolves split — LAN clients get the gateway's local IP, the host (via `lo`) and service containers (via `lxcbr0`) get a local address, and external clients still resolve to the VPS — with per-interface sockets ensuring a private domain never leaks onto a gateway it isn't configured for.

One flag since I couldn't exercise it in CI: worth a quick `dig` pass on real hardware once it's on a build —
- LAN client: `dig @<startos-lan-ip> cloud.example.com` → gateway LAN IP, `aa` set
- host: `dig @127.0.0.1 cloud.example.com` → `127.0.0.1`
- container: resolves to `10.0.3.1`
- a domain private only on gateway A, queried via gateway B's socket → should *not* return A's IP

Happy to chase down anything that misbehaves.

helix-nine:
Companion docs PR documenting the split-DNS behavior (same domain public + private): https://github.com/Start9Labs/start-docs/pull/90

PR #3297, opened 2026-06-04
feat: treat lo and lxcbr0 as secure networks for insecure traffic

helix-nine:
## What

Extends `NetworkInterfaceInfo::secure()` so that **Loopback** and **Bridge** interfaces — i.e. `lo` and `lxcbr0` on StartOS — are always treated as *secure*, in addition to the existing user-set `secure` flag.

## Why

`secure()` is the central predicate gating insecure (non-TLS) traffic:

- **`core/src/net/forward.rs`** — a port forward for an insecure binding is only created over an interface when `reqs.secure || info.secure()`. Loopback/bridge traffic never leaves the host, so plain HTTP should always be allowed over `lo` and `lxcbr0`.
- **`core/src/net/host/mod.rs`** — controls whether the non-SSL port is advertised as a reachable hostname on an interface's subnets.

Today nothing ever sets `secure = Some(true)`, so insecure bindings get no forwards anywhere. This bootstraps the mechanism for the host-local interfaces.

## Approach

Keyed on `NetworkInterfaceType::{Loopback, Bridge}` rather than matching interface names by string. This mirrors the existing `Bridge | Loopback` "local, non-WAN" pattern already used in `gateway.rs` (policy-routing and echoip exclusion), is robust against NM connection-id naming (the `name` field is the NetworkManager connection id, not the iface name), and is centralized so every consumer of `secure()` stays consistent. On StartOS the only loopback is `lo` and the only bridge is `lxcbr0` (`START9_BRIDGE_IFACE`).

> Note: if you'd prefer strict name-matching (`lo` / `lxcbr0` only, excluding any future user-created bridge), say so — the device-type match was chosen for consistency with existing code and to avoid depending on connection-id strings.

## Testing

- `cargo check --lib` passes (the only change is a method body; no `#[ts(export)]` structs touched, so no bindings rebuild needed).

### DIFF
--- a/core/src/db/model/public.rs
+++ b/core/src/db/model/public.rs
@@ -262,7 +262,16 @@ pub struct NetworkInterfaceInfo {
 }
 impl NetworkInterfaceInfo {
     pub fn secure(&self) -> bool {
-        self.secure.unwrap_or(false)
+        self.secure.unwrap_or(false) || self.is_intrinsically_secure()
+    }
+
+    // lo and lxcbr0 (the only Loopback/Bridge interfaces on StartOS) never leave the
+    // host, so insecure traffic such as plain HTTP is always permitted over them.
+    fn is_intrinsically_secure(&self) -> bool {
+        matches!(
+            self.ip_info.as_ref().and_then(|i| i.device_type),
+            Some(NetworkInterfaceType::Loopback | NetworkInterfaceType::Bridge)
+        )
     }
 }
 
--- a/core/src/db/model/public.rs
+++ b/core/src/db/model/public.rs
@@ -262,11 +262,11 @@ pub struct NetworkInterfaceInfo {
 }
 impl NetworkInterfaceInfo {
     pub fn secure(&self) -> bool {
-        self.secure.unwrap_or(false) || self.is_intrinsically_secure()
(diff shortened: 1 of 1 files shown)


core/src/db/model/public.rs
```diff
@@ -262,7 +262,16 @@ pub struct NetworkInterfaceInfo {
 }
 impl NetworkInterfaceInfo {
     pub fn secure(&self) -> bool {
-        self.secure.unwrap_or(false)
+        self.secure.unwrap_or(false) || self.is_intrinsically_secure()
```
dr-bonez:
```suggestion
        self.secure.unwrap_or_else(|| self.is_intrinsically_secure())
```

dr-bonez: an explicit security setting should always override

dr-bonez (changes requested):

core/src/db/model/public.rs
helix-nine: Good call — fixed in 1bda106c6. Now `self.secure.unwrap_or_else(|| self.is_intrinsically_secure())`, so an explicit `Some(true)`/`Some(false)` always wins and the Loopback/Bridge default only applies when unset.

dr-bonez (approved):

PR #2509, opened 2023-11-10
rename frontend to web and update contributing guide

MattDHill:

Issue #437, opened 2021-09-02
Uptime monitoring


MattDHill:

MattDHill:
This is necessarily external, probably an aspect of Diplomat protocol/service.

ProofOfKeags:
The next question is whether or not both sides of the protocol happen on the Embassy or not. There's no obvious reason I can think of for why not unless there is zero opportunity to send a message to someone from the Embassy. This seems like it might be contingent on email support or something if we want it to be totally Embassy compatible. Would be really nice if we had a native way to send messages to Web2. Telegram will definitely work but I'm slightly dissatisfied with the centralized nature of it.

elvece:
Hours I've been alive, not "since when". Return in server metrics response @dr-bonez

MattDHill:
> Hours I've been alive, not "since when". Return in server metrics response @dr-bonez

This ticket is about one Embassy monitoring the liveness/availability of your Embassy in the background. For example, one Embassy pinging another Embassy as a health check buddy system.

elvece:
> > Hours I've been alive, not "since when". Return in server metrics response @dr-bonez
> 
> You are looking for #1153
> 
> This ticket is about one Embassy monitoring the liveness/availability of your Embassy in the background. For example, one Embassy pinging another Embassy as a health check buddy system.

Yup I did indeed comment on the incorrect ticket. Commented [here](https://github.com/Start9Labs/embassy-os/issues/1153)

PR #3301, opened 2026-06-08
fix(deps): resolve Dependabot alerts across core, web, and container-runtime

helix-nine:
Sweeps the open Dependabot alerts across all four manifests. Split into clean, verifiable dependency updates; a handful of alerts need breaking work or have no upstream fix and are listed at the bottom as follow-ups.

## Fixed

**Rust — `core/Cargo.lock`** (lockfile-only bumps; `cargo check --lib` passes)
- `openssl` 0.10.78 → 0.10.80 — AES-KW heap overflow, OCSP UB on non-UTF-8 URLs, AES-KW-PAD OOB write
- `tar` 0.4.45 → 0.4.46 — PAX header desync
- `rpassword` 7.4.0 → 7.5.0 — partial password reveal on interrupt
- `tracing-subscriber` `=0.3.19` → `=0.3.20` (pinned in `Cargo.toml`) — ANSI log poisoning

**Rust — `libs/Cargo.lock`** — **deleted.** It's an orphan: `libs/Cargo.toml` was removed in #2509 (2022) and the lockfile was left behind. No build target or CI references it, so it never resolves against a manifest and just accrues stale alerts (~30 of them, including the high-severity `whoami`/`rustls`/`mio` ones that only ever lived here).

**npm — `container-runtime/package-lock.json`** (`npm audit fix` + one override)
- `qs`, `path-to-regexp`, `picomatch`, `yaml`, `minimatch` (×3), `js-yaml` — `npm audit fix`
- `jsonpath` 1.1.1 → 1.3.0 — code injection, prototype pollution
- `underscore` → 1.13.8 via `overrides` — `jsonpath` pins 1.13.6, which has the `_.flatten`/`_.isEqual` recursion DoS

**npm — `web/package-lock.json`**
- `hono` → 4.12.24 — Set-Cookie injection, IPv6 deny bypass, mount routing, JWT scheme
- `vite` → 7.3.5 via `overrides` — `fs.deny` bypass, arbitrary file read, path traversal (build toolchain, under `@angular/build`)
- `undici` → 7.27.2 via `overrides` — WS 64-bit length overflow, CRLF injection, request smuggling
- `picomatch` → 4.0.4 — scoped `overrides` on `@angular-devkit/core` and `@angular/build` (the only pins forcing 4.0.3); `micromatch`'s nested 2.3.2 left intact
- `uuid` 8 → 11.1.1 — the direct dep is unused in source; `node-jose`'s copy is forced to 11.1.1 via override (node-jose declares `uuid` but never imports it). Dropped the now-redundant `@types/uuid` since uuid v11 self-types.

## Verification
- `cargo check --lib` (core) compiles cleanly with the Rust bumps.
- `npm audit` reports zero remaining **Dependabot-tracked** alerts in either npm manifest. (Web still shows `micromatch`/`yaml` and container-runtime shows `cross-spawn`/`file-type` under dev-only toolchains — these are npm-audit findings, **not** open Dependabot alerts, and would require breaking `lint-staged`/`@swc/cli` major bumps.)

## Not addressed (need dedicated work / no upstream fix)
- **`hickory-proto`/`hickory-server` (#441 high, #442)** — only fixed in hickory **0.26**, which is a breaking API migration (NameServerConfig signature, moved modules; ~41 compile errors across `net/dns.rs`, `net/gateway.rs`, `tunnel/dns.rs`). Wants its own PR plus DNS behavior testing — too risky to fold into a dep sweep.
- **`ed25519-dalek` 1.x / `curve25519-dalek` 3.x / `rand` 0.7 (#240, #300, #437)** — pulled in by the intentional `ed25519-dalek-v1` alias in `core/Cargo.toml` (kept for legacy signature compat). Removing it is a deliberate design call, not a bump.
- **`rustls-webpki` 0.102 (#434/#435/#436)** — pinned by the `Start9Labs/async-acme` fork's `rustls` 0.22; needs the fork bumped to rustls 0.23.
- **`lexical-core` (#429)** — via `nom` 6 ← `emver` ← `exver` (Start9 crates); upstream bump needed.
- **`tokio-tar` (#433)** — the `dr-bonez/tokio-tar` git fork; no patched release.
- **`serde_yml` / `libyml` (#431/#432)** — unmaintained, no upstream fix; needs migrating off the crate.

Want me to take the hickory 0.26 migration as a separate PR?


dr-bonez (dismissed):

helix-nine:
Thanks for the approve — heads up that it got auto-dismissed because I pushed one more commit (`8c41adf`): the `rpassword` bump to 7.5.0 broke the `*-apple-darwin` build (7.5.0 calls `libc::__errno_location()`, which doesn't exist on macOS). Moved it to 7.5.4, which dropped that errno manipulation while keeping the partial-password-reveal fix. The mac error was the only real failure in the matrix (the musl jobs were just fail-fast cancellations), so this should turn it green. CI is re-running now — re-approve once it's clean.

dr-bonez (approved):

helix-nine:
Thanks — merged. This covered the npm trees + the lockfile-only Rust bumps + the orphaned `libs/Cargo.lock`. The rest of the Dependabot cleanup is in three still-open PRs:

- #3302 — hickory 0.26 migration (DNS RUSTSEC advisories)
- Start9Labs/async-acme#2 — `generic-async-http-client` 0.5→0.6 (drops vulnerable rustls-webpki 0.102)
- #3303 — adopt `astral-tokio-tar` + bump the async-acme rev (depends on async-acme#2 landing first)

Remaining accepted-as-is per our discussion: lexical-core #429 and the ed25519-dalek-v1 advisories (scoped backwards-compat pins).

PR #3306, opened 2026-06-10
feat(net): automatic gateway configuration — port forwarding (PCP/NAT-PMP/UPnP) + private-domain DNS (RFC 2136)

helix-nine:
## What

Automatically configure the gateway for the things a StartOS service needs, instead of leaving them as manual steps. One code path works whether the gateway is a home router or a StartTunnel. Two areas:

### 1. Port forwarding — PCP (RFC 6887) → NAT-PMP → UPnP IGD
- **Client** (`net/port_map.rs`): when a public address needs a forward, open it on the gateway via PCP, then NAT-PMP (`crab_nat`), then UPnP `AddPortMapping` (`igd-next`); withdraw on disable/delete; renewed/reconciled with the forward. Candidate gateways = the interface's NM gateway + each subnet's `.1`. Handles port ranges (#3270) up to a cap.
- **WAN-IP** (`net/gateway.rs`): tries UPnP `GetExternalIPAddress` before the echoip probe; private/CGNAT result falls back to echoip.
- **StartTunnel server**: a PCP server (`tunnel/pcp.rs`, UDP 5351) + a UPnP IGD (`tunnel/igd.rs`, SSDP + SCPD + SOAP), both `SO_BINDTODEVICE`-bound to the wg interface and **peer-scoped** (a peer can only forward to its own tunnel IP). Note: `crab_nat` is client-only, so the PCP *server* is hand-rolled (no server crate exists).

### 2. Private-domain DNS — RFC 2136 (DNS UPDATE)  _(in progress)_
- **Client** (`net/dns_update.rs`, done): when a private domain is enabled on a gateway, push an `A` record (`domain → our IP on that gateway's subnet`) to the gateway's DNS via DNS UPDATE, bound to our address so the gateway authorizes by source IP; withdraw on disable/delete. Best-effort, reconciled off the `net_iface` watch.
- **Gateway server (coming in this PR)**: a shared RFC 2136 receiver in `core` (used by StartTunnel here, and by StartWRT which imports `core`). Accepts UPDATE from an authorized source IP (a per-device "allow DNS injection" toggle, default off — trust the device), answers injected names, forwards the rest. Plus a per-device toggle + a view/add/edit/delete UI.

## Notes
- New deps: `crab_nat` (PCP/NAT-PMP, pure Rust, no transitive deps), `igd-next` (UPnP; `attohttpc` pulled without TLS), `xmltree` (SOAP parse, pinned `0.10`). DNS UPDATE uses the in-tree `hickory-proto`. `SO_BINDTODEVICE` is Linux-only and cfg-gated for the apple-darwin build.
- Companion router work: Start9Labs/start-wrt#65 (IGD on by default + toggle; the StartWRT RFC 2136 receiver is being added there too). Docs: Start9Labs/start-docs#93.

## Testing
- `cargo check --lib` clean; `cargo test --lib net:: tunnel::` passing (port-forward wire contracts, PCP MAP encoding, range allocation).
- ⚠️ The live PCP/NAT-PMP/UPnP and RFC 2136 paths need real hardware (a router + a VPS/home tunnel pair) to exercise end-to-end — flagged for validation.


core/src/tunnel/igd.rs
```diff
@@ -0,0 +1,819 @@
...
+  </device>
+</root>"#
+    )
+}
+
+/// Minimal WANIPConnection SCPD exposing the three actions this IGD implements.
+/// IGD clients (e.g. igd-next) read `actionList` to learn each action's input
+/// argument names before issuing a request.
+const SCPD: &str = r#"<?xml version="1.0"?>
+<scpd xmlns="urn:schemas-upnp-org:service-1-0">
```
dr-bonez: Move this to an include_str!

core/src/tunnel/igd.rs
```diff
@@ -0,0 +1,819 @@
...
+        718 => "ConflictInMappingEntry",
+        725 => "OnlyPermanentLeasesSupported",
+        _ => "Action Failed",
+    }
+}
+
+fn render_root_desc(uuid: &str) -> String {
+    format!(
+        r#"<?xml version="1.0"?>
+<root xmlns="urn:schemas-upnp-org:device-1-0">
```
dr-bonez: And this

core/src/tunnel/igd.rs
```diff
@@ -0,0 +1,819 @@
...
+        StatusCode::OK,
+        [(header::CONTENT_TYPE, "text/xml; charset=\"utf-8\"")],
+        body,
+    )
+        .into_response()
+}
+
+fn fault(code: u16, desc: &str) -> Response {
+    let body = format!(
+        r#"<?xml version="1.0"?>
```
dr-bonez: And this

core/src/tunnel/igd.rs
```diff
@@ -0,0 +1,819 @@
...
+        .get_child("Body")?
+        .children
+        .iter()
+        .find_map(|n| n.as_element())?;
+    action.get_child(arg)?.get_text()?.trim().parse().ok()
+}
+
+fn ok(action: &str, inner: &str) -> Response {
+    let body = format!(
+        r#"<?xml version="1.0"?>
```
dr-bonez: And this

core/src/tunnel/igd.rs
helix-nine: Done in 114f576 — extracted to `core/src/tunnel/igd_xml/{ok,fault,root_desc,scpd}.xml` and loaded via `include_str!`. The templated ones use explicit named `format!` args since implicit capture (`{name}`) isn't supported with `include_str!`.

core/src/tunnel/igd.rs
helix-nine: Done in 114f576 — extracted to `core/src/tunnel/igd_xml/{ok,fault,root_desc,scpd}.xml` and loaded via `include_str!`. The templated ones use explicit named `format!` args since implicit capture (`{name}`) isn't supported with `include_str!`.

core/src/tunnel/igd.rs
helix-nine: Done in 114f576 — extracted to `core/src/tunnel/igd_xml/{ok,fault,root_desc,scpd}.xml` and loaded via `include_str!`. The templated ones use explicit named `format!` args since implicit capture (`{name}`) isn't supported with `include_str!`.

core/src/tunnel/igd.rs
helix-nine: Done in 114f576 — extracted to `core/src/tunnel/igd_xml/{ok,fault,root_desc,scpd}.xml` and loaded via `include_str!`. The templated ones use explicit named `format!` args since implicit capture (`{name}`) isn't supported with `include_str!`.

web/projects/start-tunnel/src/app/routes/home/routes/devices/index.ts
```diff
@@ -130,17 +145,48 @@ export default class Devices {
...
   )
 
+  protected onSetDnsInjection({
+    subnet,
+    ip,
+    allowDnsInjection,
+  }: MappedDevice): void {
+    this.dialogs
+      .open(TUI_CONFIRM, {
+        label: allowDnsInjection
```
waterplea: I think it makes more sense for longer text to be in data: { content: XXX } rather than in title, otherwise it looks too big.

How it looks like now:

Image

Cleaner:

Image

We might want to change wording a little bit, but still it looks better with title and actual message than with just a huge title.

web/projects/start-tunnel/src/app/routes/home/routes/dns/add.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  protected readonly form = inject(NonNullableFormBuilder).group({
+    name: ['', Validators.required],
+    type: ['A' as (typeof TYPES)[number], Validators.required],
+    value: ['', Validators.required],
+    ttl: [300 as number | null],
+  })
+
+  protected async onSave() {
+    if (this.form.invalid) {
```
waterplea, line 225: There's no way to click save until form is valid because it's disabled. I believe a good UX pattern is to not disable it but do this so that I can see all the fields I forgot to type in. Let's remove disable from the save button, the onSave already handles it well.

web/projects/start-tunnel/src/app/routes/home/routes/dns/add.ts
```diff
@@ -0,0 +1,121 @@
...
+    const loader = this.loading.open('').subscribe()
+    const { name, type, value, ttl } = this.form.getRawValue()
+
+    try {
+      await this.api.addDnsRecord({ name, type, value, ttl })
+    } catch (e: any) {
+      this.errorService.handleError(e)
+    } finally {
+      loader.unsubscribe()
+      this.context.$implicit.complete()
```
waterplea: Do we want to complete it on finally and not in successful try? What if we lose connection while entering the values, we would then lose the form, is that ok?

waterplea (dismissed):
A few small comments on the UI, otherwise LGTM from the frontend side

web/projects/start-tunnel/src/app/routes/home/routes/devices/index.ts
helix-nine: Done in ab4bcc5e — short title ("Allow DNS records?") with the explanation moved to `data: { content }`. Much less shouty, thanks.

web/projects/start-tunnel/src/app/routes/home/routes/dns/add.ts
helix-nine: Done — removed `[disabled]` from Save. `onSave` already bails on an invalid form and calls `tuiMarkControlAsTouchedAndValidate`, so clicking now surfaces the missing fields.

web/projects/start-tunnel/src/app/routes/home/routes/dns/add.ts
helix-nine: Good catch — moved `complete()` into the successful `try` branch (loader cleanup stays in `finally`). A failed save now keeps the entered values so you can retry instead of losing the form.

waterplea (dismissed):

core/src/net/forward.rs
```diff
@@ -572,6 +573,11 @@ struct InterfaceForwardEntry {
     targets: BTreeMap<ForwardRequirements, (SocketAddrV4, u8, Weak<()>)>,
     // Maps source SocketAddr -> strong reference for the forward created in PortForwardController
     forwards: BTreeMap<SocketAddrV4, Arc<()>>,
+    // (local interface IP, external port) pairs for which we've asked the
+    // upstream gateway (via PCP/NAT-PMP/UPnP) to forward to this host. Tracked
+    // so the mapping is withdrawn when the forward is no longer required. A
+    // range contributes one entry per port (capped, see MAX_AUTO_MAP_PORTS).
```
dr-bonez: MAX_AUTO_MAP_PORTS no longer exists

core/src/net/port_map/server/mod.rs
```diff
@@ -0,0 +1,501 @@
...
+const CANNOT_PROVIDE_EXTERNAL: u8 = 11;
+
+/// PCP protocol field value for TCP (the only transport the SNI demux handles).
+const PROTO_TCP: u8 = 6;
+/// Largest PORT_SET range we will grant in one MAP (RFC 7753 lets us grant
+/// fewer than requested; the client skips the range if it can't get them all).
+const MAX_PORT_SET: u16 = 1024;
+
+/// Per-gateway I/O and forward backend for the shared PCP server.
+#[async_trait::async_trait]
```
dr-bonez: async_trait should be avoided unless dynamic dispatch is needed

core/src/net/port_map/client.rs
```diff
@@ -0,0 +1,566 @@
...
+                }
+            }
+            return;
+        }
+
+        // Range mapping via PCP PORT_SET (RFC 7753), PCP-only. A gateway lacking
+        // PORT_SET silently maps a single port; detect the missing/short grant
+        // and skip rather than forward a partial range.
+        if spec.count > 1 {
+            let want = spec.count;
```
dr-bonez: bad naming

core/src/tunnel/forward/sni.rs
```diff
@@ -0,0 +1,401 @@
...
+}
+
+/// Whether `buf` holds at least one complete TLS handshake record.
+fn record_complete(buf: &[u8]) -> bool {
+    buf.len() >= 5 && buf.len() >= 5 + u16::from_be_bytes([buf[3], buf[4]]) as usize
+}
+
+/// Extract the (lowercased) SNI host_name from a buffered TLS ClientHello, or
+/// `None` if absent / not yet complete / not TLS.
+fn extract_sni(buf: &[u8]) -> Option<String> {
```
dr-bonez, line 284: can we use rustls here?

dr-bonez (changes requested):

core/src/tunnel/forward/sni.rs
helix-nine: Done in a0ae22763 — `extract_sni` now parses the ClientHello via rustls’ `Acceptor` (`read_tls` → `accept()` → `client_hello().server_name()`) instead of the hand-rolled byte walk. The bytes are only parsed, never answered, so the buffered ClientHello is still spliced verbatim to the backend. The unit test now feeds a real rustls-generated ClientHello rather than a synthetic one.

core/src/net/port_map/server/mod.rs
helix-nine: Done in a0ae22763 — dropped `async_trait` on `GatewayBackend` for native RPITIT (`fn … -> impl Future<Output = …> + Send`). Every call site is generic (`handle<B: GatewayBackend>`, the IGD handlers) with `TunnelContext` as the only concrete impl — no `dyn` dispatch — so the box is unnecessary. The impl keeps `async fn` bodies.

core/src/net/port_map/client.rs
helix-nine: Done in a0ae22763 — renamed `want` → `range_size` (the requested PORT_SET contiguous-range size).

core/src/net/forward.rs
helix-nine: Already resolved on the current branch — the reorg/rebase dropped the `(capped, see MAX_AUTO_MAP_PORTS)` clause; the comment now reads just “a range contributes one entry per port”, and `MAX_AUTO_MAP_PORTS` no longer exists anywhere in the tree.

helix-nine:
All four inline points from the changes-requested review are resolved (replied threaded on each):

- **rustls for SNI parse** — `extract_sni` now uses rustls’ `Acceptor`; test feeds a real ClientHello. (`a0ae22763`)
- **drop `async_trait`** on `GatewayBackend` → native RPITIT (`fn … -> impl Future + Send`); no `dyn`. (`a0ae22763`)
- **`want` naming** → `range_size`. (`a0ae22763`)
- **`MAX_AUTO_MAP_PORTS`** — already gone on the branch (stale comment removed in the reorg).

Also pushed since, from the Matrix discussion:

- `be89393fc` — source-preserving SNI demux via `socket transparent` divert (verified in a netns harness: backend sees the real client IP; egress-misroute bug fixed; works under strict + loose rp_filter).
- `aa1e18bd5` — start-tunnel UI: `SNI`→`Hostname` column; SSL/TLS note → `tuiHint` on the hostname field; PCP-added forwards get a default `PCP` label server-side (preserving user label/enabled on renewal); DNS add gets an A/AAAA device picker + validated `Other` IP.
- `53eaa37df` — split gateway autoconfig into two per-device flags (`allow_dns_injection` + `allow_auto_port_forward`); `is_known_client` gates forwarding on the new flag; one "Enable Gateway Autoconfiguration" checkbox toggles both. Regenerated osBindings + SDK.

`cargo check`, `check:tunnel`, `check:i18n`, and prettier are green throughout. Ready for another look when you have a moment.

dr-bonez (dismissed):

waterplea (dismissed):
Pushed a few cleanup changes

helix-nine:
**Full VM test of the in-place WireGuard update + unified add/update parser** — ran a startbox built from this branch on a StartOS VM (beta-9 → this build via A/B + reboot, **NetworkManager 1.52**), exercising both paths end-to-end against real NM.

**`net tunnel update` (the "update disconnects the gateway forever" fix) — validated.** `Update2(TO_DISK)` + `Device.Reapply`:
- interface never torn down — `ifindex` stayed constant
- connection identity preserved — same NM uuid
- new config applied in place (peer / endpoint / keepalive / address / DNS all swapped)
- link stayed **UP/activated**, default route + the live SSH session intact throughout

**`net tunnel add` (AddAndActivateConnection + the shared parser) — validated.** NM 1.52 accepts the hand-built settings dict: address / DNS / peer / full-tunnel allowed-ips / `dns-priority` all correct, private key stays a root-only secret, and a full-tunnel dead peer does **not** hijack the host route (NM fwmark policy routing). `ip_info` populates in ~1s at steady state.

One transient worth noting: an `add` run within seconds of the migration reboot — while the box was still thrashing (`failed to bind DNS on [::]:53: Address in use`, dns-catalog timeouts) — hit `add_tunnel`'s 15s `ip_info` wait and logged a warning. It still returns Ok and self-heals once the system settles; a steady-state `add` shows no race. Boot-time transient, not a code defect — no change made.

Separately (for the follow-up delete-flow PR, out of scope here): `net tunnel remove` briefly dropped the client connection even for a non-default outbound-only tunnel — the same delete-flow cancel/disconnect safety already scoped out of this PR.

dr-bonez (approved):

PR #3310, opened 2026-06-12
feat: input-not-matches tasks accept multiple values, set one

helix-nine:
## What

Splits the `input-not-matches` task input into **multiple read values** (`accept`) and **one write value** (`set`), as requested.

Previously a task carried a single `TaskInput`:

```ts
{ kind: 'partial', value }
```

That one `value` did triple duty:
- **match check** — the task was satisfied when the current action input was a superset of `value`;
- **conflict check** — re-activated the task when a later change moved the input away from `value`;
- **prefill** — pre-populated the action form when the user ran the task.

So read == write: the only value you could accept was the exact value you'd push the user to set.

## Now

```ts
{ kind: 'partial', accept: DeepPartial<Input>[], set: DeepPartial<Input> }
```

- **satisfied** when the current input matches **any** entry in `accept` (`accept.any(is_partial_of)`);
- **conflict** (cross-package critical guard) only when the input conflicts with **every** `accept` entry (`accept.all(conflicts)`);
- **prefill** is `set`.

The 1.x behavior is exactly `accept: [X], set: X`.

This lets a package accept several already-good configurations while still steering the user to one recommended value when none of them hold.

## Layers touched

- **core** — `TaskInput::Partial { accept: Vec<Value>, set: Value }`; `matches()` and the `update_tasks` re-evaluation arm generalized to any/all.
- **ts-bindings** — `TaskInput.ts` regenerated via `make ts-bindings`.
- **sdk** — `TaskInput<T>` type + `createTask` null-coercion over `accept`/`set`; CHANGELOG entry under 2.0.0 (breaking).
- **container-runtime** — SystemForEmbassy config-change task producer → `accept: [diff], set: diff`.
- **web** — task prefill reads `set`; cross-package conflict guard iterates `accept`; mock fixtures updated.

## Testing

- `cargo check --lib` (core) — clean.
- `make bundle` (sdk) — 68 tests pass, prettier clean.
- `npm run build` (container-runtime) — clean.
- `npm run check:ui` (web) — only two **pre-existing** marketplace errors (`appearance` on `tuiCardOptionsProvider`, verbatim on `origin/master`), unrelated to this change. The pre-commit `check:ui` hook was bypassed for the same reason.

## Docs

Companion start-docs PR updating the task/`input-not-matches` reference: _(linked in a follow-up comment)_.


helix-nine:
Companion docs PR: https://github.com/Start9Labs/start-docs/pull/94

dr-bonez (approved):

PR #3312, opened 2026-06-13
fix: /media/startos owned root:root on migrated installs (#3311)

### LINKED ISSUE #3311: /media/startos ownership defaults to root:root
I have noticed that on my migrated StartOS installation, attempts to browse /media/startos/data/... require sudo (for example, cd /media/startos/data/package-data/volumes/). I don't believe this represents a functional problem for the running services, but it is strange to observe when compared with fresh installation of StartOS I used for testing, where the directory's group ownership is correctly set to startos.

### Environment
- **Current Version:** StartOS v0.4.0-beta.9
- **Upgrade Path:** Migrated from 0.3.5.1 -> 0.4.0 (initially beta.4 or beta.5) -> beta.9
- **Hardware:** GMKtec M5 Plus Mini PC (AMD Ryzen 7 5825U)
- **Hypervisor:** Proxmox VE 9.1.1

### Summary
On a server migrated from 0.3.5.1, the host OS gateway directory `/media/startos` is owned by `root:root` instead of `root:startos`. This results in a `Permission denied` error for the standard `start9` SSH user when attempting to browse `/media/startos/data/...`, requiring `sudo -i` to view package data.

### Observations & Debugging
1. **Fresh Installs vs Migrations:** A fresh install of v0.4.0 correctly sets this directory to `root:startos` (`drwxr-x--- root startos`). The issue is strictly isolated to servers that underwent the 0.3.5.1 migration.
2. **Persistence & Overlay:** Because the host OS runs on the immutable `overlay` filesystem, running `sudo chown root:startos /media/startos` fixes the issue temporarily, but it reverts to `root:root` on every reboot. The `startd` boot sequence is failing to re-apply the `startos` group on migrated nodes.
3. **Services Unaffected:** Service functionality is **100% normal**. LXC container bind-mounts to the physical data drive are working perfectly (e.g., LND runs and syncs without issues). This is strictly a host OS gateway permission / SSH convenience issue.
4. **Group IDs Match:** The `startos` group exists and maps correctly to GID 1000 (`startos:x:1000:start9`), matching fresh installs.

### Expected Behavior
The migration scripts (or `startd` on boot) should ensure the host overlay gateway `/media/startos` is assigned the `startos` group (`drwxr-x--- root startos`), allowing the `start9` SSH user access, matching the behavior of fresh 0.4.0 installs.

### Actual Behavior
The directory remains `drwxr-x--- root root` and the OS boot sequence continually defaults to this state on every reboot for migrated servers.

### Suspected Cause
It appears the migration path from 0.3.5.1 fails to update the internal state (Patch-DB) or trigger the specific `startd` initialization hook that applies the `startos` group to the host's `/media/startos` overlay mount point.

helix-nine:
## Problem

Closes #3311. On servers **migrated** from 0.3.5.1, the host gateway directory `/media/startos` is owned `root:root` instead of `root:startos`, so the `start9` SSH user gets *Permission denied* browsing `/media/startos/data/...` and needs `sudo`. Fresh 0.4.0 installs are correct (`drwxr-x--- root startos`). Services are unaffected — it's a host-gateway / SSH-convenience issue — but migrated and fresh installs should match.

## Root cause

`/media/startos`'s `root:startos` group is applied **only at image-build time** (`build/image-recipe/build.sh` bakes `chmod 750` + `chown root:startos` into the squashfs). There is no runtime, boot, or migration code that re-applies it.

The root filesystem is an overlay assembled in the initramfs:
```
lowerdir = /startos/config/overlay : /lower(squashfs)   # config/overlay has HIGHER priority
upperdir = tmpfs (wiped every boot)
```
The persistent `config/overlay` layer sits **above** the squashfs. `os_install` mounts an overlay with `upperdir = config/overlay` and bind-mounts the rootfs at `…/media/startos/root`, which materializes a `media/startos` entry into the persistent overlay — and its ownership is frozen at creation time. On the migrated node that entry is `root:root`, so the merged `/media/startos` shows `root:root` forever, shadowing the squashfs's `root:startos`. This also explains why `chown` reverts on reboot: it only copies up to the ephemeral tmpfs upper, while the persisted `root:root` re-asserts.

## Fix

`v0_4_0_beta_10` `post_up`: chown/chmod the persisted `config/overlay/media/startos` entry to `root:startos` / `750`, fixing it in the overlay where it actually lives so migrated nodes match fresh installs. Idempotent.

(An earlier revision also added a per-boot chown in `init.rs`; dropped per review — the overlay is the right place to fix it.)

## Testing

`cargo check --lib` passes (only pre-existing warnings). Not yet validated on a live migrated VM — the underlying overlay state requires a real 0.3.5.1→0.4.0 migration to reproduce.

### DIFF
--- a/core/src/version/v0_4_0_beta_10.rs
+++ b/core/src/version/v0_4_0_beta_10.rs
@@ -1,9 +1,12 @@
 use exver::{PreReleaseSegment, VersionRange};
 use imbl_value::json;
+use tokio::process::Command;
 
 use super::v0_3_5::V0_3_0_COMPAT;
 use super::{VersionT, v0_4_0_beta_9};
+use crate::context::RpcContext;
 use crate::prelude::*;
+use crate::util::Invoke;
 
 lazy_static::lazy_static! {
     static ref V0_4_0_beta_10: exver::Version = exver::Version::new(
@@ -61,6 +64,25 @@ impl VersionT for Version {
             .join(":");
         server_info.insert("caFingerprint".into(), json!(repaired));
         Ok(Value::Null)
+    }
+    async fn post_up(self, _ctx: &RpcContext, _input: Value) -> Result<(), Error> {
+        // Older installs copied /media/startos into the persistent config overlay as
+        // root:root, shadowing the squashfs's root:startos on every boot. Fix the
+        // persisted entry so migrated nodes match fresh installs (#3311).
+        let overlay_media_startos = "/media/startos/config/overlay/media/startos";
+        if tokio::fs::metadata(overlay_media_startos).await.is_ok() {
+            Command::new("chown")
+                .arg("root:startos")
+                .arg(overlay_media_startos)
+                .invoke(ErrorKind::Filesystem)
+                .await?;
+            Command::new("chmod")
+                .arg("750")
+                .arg(overlay_media_startos)
+                .invoke(ErrorKind::Filesystem)
+                .await?;
+        }
+        Ok(())
     }
     fn down(self, _db: &mut Value) -> Result<(), Error> {
         Ok(())


core/src/init.rs
```diff
@@ -168,6 +168,20 @@ pub async fn init(
         run_script("/media/startos/config/preinit.sh", progress).await;
     }
 
+    // /media/startos's group is only set at image-build time (in the squashfs); re-apply
+    // every boot so a stale root:root entry in the persistent config overlay (left by older
+    // installs) can't shadow it, keeping migrated and fresh installs identical (#3311).
```
dr-bonez: if it was left in the overlay, it should be fixed in the overlay, which you do in the migration. this is redundant paranoia

dr-bonez (changes requested):

core/src/init.rs
helix-nine: Agreed — dropped the init.rs boot-time chown. The migration's `post_up` now fixes the persisted `config/overlay/media/startos` entry as the sole change, so it's corrected in the overlay where it lives. Force-pushed.

dr-bonez (approved):

helix-nine:
Thanks for the review! Merged. One thing worth doing before the beta.10 release: confirm on a real 0.3.5.1→0.4.0 migrated box that the `post_up` actually clears `root:root` on `/media/startos` — I couldn't reproduce the bad overlay state in a fresh VM (it needs a genuine migration), so this path is so-far only validated by `cargo check`.

Issue #3313, opened 2026-06-13
startd deadlocks during concurrent service updates (whole tokio runtime wedged)


MattDHill:
## Summary

On **StartOS 0.4.0-beta.9**, updating **two services concurrently** (the `bitcoind`/Knots package and `c-lightning`) deadlocked `startd` — the entire tokio runtime wedged. The web UI and all RPC went unresponsive; only `sshd` survived. Recovery required `SIGKILL` + reboot.

## Environment

- StartOS `0.4.0-beta.9`
- x86 server (RAM unknown to me, but ≥ enough that a **28 GB** startd peak did **not** trigger the OOM killer)
- Services updated concurrently: `bitcoind` (Knots `29.3`) and `c-lightning` (`26.6.1`)

## Observed behavior

- Web UI unreachable on LAN; `start-cli server logs` hung forever (it routes through startd's RPC, which was dead).
- `sshd` unaffected → startd-specific, not host/network.
- `systemctl status startd`: **`active (running)`**, 1-month uptime, but **`Memory: 590.5M (peak: 28.2G, swap: 27.2M, swap peak: 41.7M)`** — startd had ballooned to ~28 GB at some point and fallen back.
- **No** OOM kills (`dmesg` / `journalctl -k` clean), **no** disk-full, **no** I/O errors, **no** D-state.

### Thread states

`ps -L -o tid,stat,wchan -p <pid>` → 24 of 26 threads parked in **`futex_`** (futex_wait); only the two `do_epoll_wait` reactor threads and one `hrtimer` idle. Main thread `Ssl` / `futex_`. Classic fully-wedged-runtime signature (everything either blocked on a lock or idle-with-no-progress).

### Log timeline (last forward progress before the freeze)

The journal stops dead at **14:41:23**; nothing afterward except a stray `net::tls: no certificate for SNI None` at 14:49 (a browser hitting the dead UI). Immediately before the freeze:

- A flood of `cp: failed to clone '/media/startos/data/package-data/volumes/bitcoind.install-backup/...' from '.../bitcoind/...': Invalid argument` — pre-update install-backup creation (secondary issue, see below).
- The old versions stop/exit (Knots `29.3`, c-lightning `26.6.1`).
- **Two concurrent** `install:new:init ... "kind":"update"` (bitcoind + c-lightning) at 14:41:22.
[...]

PR #3314, opened 2026-06-13
fix(disk): reconcile stale mountpoint before remount

MattDHill:
## Summary

Fixes a service-launch failure that occurs when `startd` is restarted **within the same boot** after its previous process was hard-killed (e.g. `SIGKILL` while recovering from a hang).

### Symptom

After `startd` was SIGKILLed and restarted by systemd (no reboot), **every** service failed to load during init with:

```
mount: /media/startos/tmp/<id>: /usr/lib/startos/container-runtime/rootfs.squashfs is already mounted.
ERROR ...:cleanup_and_initialize:init:load: startos::service::service_map: Error loading service: Filesystem I/O Error: mount exited with exit status: 1
   at src/disk/mount/guard.rs
```

A full reboot cleared it; a `systemctl restart` did not.

### Root cause

Two compounding issues:

1. **The restart path skips filesystem reconciliation.** `startd` gates init on `/run/startos/initialized` (`bins/startd.rs`). On a fresh boot the marker is absent → `init::init()` runs the filesystem prep. On a same-boot restart the marker survives (it lives in `/run`, tmpfs), so startd takes the `else` branch, skips `init::init()`, and goes straight to loading services — nothing reconciles mounts left behind by the dead process.

2. **Remounts aren't idempotent.** `MountGuard::mount` mounts unconditionally. `TmpMountGuard` dedups live mounts via a `Weak<MountGuard>`, but that `Weak` is dead in a fresh process, so it falls through to `MountGuard::mount`, which fails because the *kernel* mount from the dead process still exists. The shared `rootfs.squashfs` base is just the first mount in the chain to trip; the overlay on top (`OverlayGuard`, also via `MountGuard::mount`) would fail identically.

That's why a reboot fixed it (wiped `/run`, marker gone, `init::init()` ran) but a `systemctl restart` wouldn't have.

### Fix

Make `MountGuard::mount` reconcile a pre-existing mount at the target before mounting — lazily unmount it first — mirroring the existing `bind()` helper in `disk/mount/util.rs`. Both `TmpMountGuard` and `OverlayGuard` funnel through `MountGuard::mount`, so the whole mount chain self-heals on a restart, no reboot required.

Within a live process the change is inert: `TmpMountGuard`'s `Weak`-based dedup means this branch is only reached when there is no active in-process guard for the target.

### Context

This restart was triggered while recovering from a separate concurrent-update **deadlock** in startd (#3313). The deadlock is what forced the SIGKILL; this PR fixes the collateral damage that made the restart unrecoverable without a reboot. The two are independent — this can land on its own.

Closes the restart-recovery half of #3313.

### DIFF
--- a/core/src/disk/mount/guard.rs
+++ b/core/src/disk/mount/guard.rs
@@ -8,7 +8,7 @@ use tokio::sync::Mutex;
 use tracing::instrument;
 
 use super::filesystem::{FileSystem, MountType, ReadOnly, ReadWrite};
-use super::util::unmount;
+use super::util::{is_mountpoint, unmount};
 use crate::util::{Invoke, Never};
 use crate::{Error, ResultExt};
 
@@ -55,6 +55,14 @@ impl MountGuard {
         mount_type: MountType,
     ) -> Result<Self, Error> {
         let mountpoint = mountpoint.as_ref().to_owned();
+        // A prior process lifetime (e.g. a hard kill, where Drop never ran) can
+        // leave a stale kernel mount at this target; mounting onto it fails with
+        // "already mounted". Reconcile by lazily unmounting first, mirroring the
+        // `bind` helper. Within a live process this branch isn't reached for an
+        // active mount — `TmpMountGuard` dedups via its `Weak<MountGuard>`.
+        if is_mountpoint(&mountpoint).await? {
+            unmount(&mountpoint, true).await?;
+        }
         filesystem.mount(&mountpoint, mount_type).await?;
         Ok(MountGuard {
             mountpoint,
--- a/core/src/disk/mount/guard.rs
+++ b/core/src/disk/mount/guard.rs
@@ -55,14 +55,6 @@ impl MountGuard {
         mount_type: MountType,
     ) -> Result<Self, Error> {
         let mountpoint = mountpoint.as_ref().to_owned();
-        // A prior process lifetime (e.g. a hard kill, where Drop never ran) can
-        // leave a stale kernel mount at this target; mounting onto it fails with
-        // "already mounted". Reconcile by lazily unmounting first, mirroring the
-        // `bind` helper. Within a live process this branch isn't reached for an
-        // active mount — `TmpMountGuard` dedups via its `Weak<MountGuard>`.
-        if is_mountpoint(&mountpoint).await? {
-            unmount(&mountpoint, true).await?;
-        }
         filesystem.mount(&mountpoint, mount_type).await?;
         Ok(MountGuard {
             mountpoint,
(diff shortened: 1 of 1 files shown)


core/src/disk/mount/guard.rs
```diff
@@ -55,6 +55,14 @@ impl MountGuard {
...
     ) -> Result<Self, Error> {
         let mountpoint = mountpoint.as_ref().to_owned();
+        // A prior process lifetime (e.g. a hard kill, where Drop never ran) can
+        // leave a stale kernel mount at this target; mounting onto it fails with
+        // "already mounted". Reconcile by lazily unmounting first, mirroring the
+        // `bind` helper. Within a live process this branch isn't reached for an
+        // active mount — `TmpMountGuard` dedups via its `Weak<MountGuard>`.
+        if is_mountpoint(&mountpoint).await? {
+            unmount(&mountpoint, true).await?;
+        }
```
dr-bonez: this shouldn't exist in both the bind helper and here. if we want to do it for all mounts, then we should remove from the bind helper.

That said, it is not immediately obvious that this can safely be done for all MountGuards. It is, however, easy to reason that it is safe for all TmpMountGuards, due to the internal tracking and determinism. Could that be sufficient?

dr-bonez (changes requested):

core/src/disk/mount/guard.rs
dr-bonez: @helix-nine

helix-nine: Agreed — moved it to `TmpMountGuard::mount` (else-branch, where the `Weak` failed to upgrade) and removed it from `MountGuard::mount`. Pushed in 408825f5a.

You're right that this is the only place it's provably safe, and tracing it confirms it's also the only place it's actually *needed*:

- `TmpMountGuard`'s mountpoint is the source-hash-derived path under `/media/startos/tmp`, so it's **deterministic across process lifetimes** — a hard-killed process leaves a stale mount at the exact path the new process targets, hence the `rootfs.squashfs is already mounted` failure. Because the path is hash-derived, any mount already there is the same content, so lazy-unmount-then-remount is sound. The internal `Weak` tracking guarantees we only reach this branch when there's no live in-process guard.
- The overlay (`OverlayGuard` in `lxc::ContainerState::new`) mounts at `LXC_CONTAINER_DIR/<guid>/rootfs` with `guid = new_guid()` — a fresh random path each lifetime. A stale overlay from the dead process sits at an *old* guid dir, so it never collides with the new mount. My earlier claim in the PR description that "the overlay would fail identically" was wrong: it leaks, but it doesn't block. So the generic `MountGuard::mount` branch wasn't buying us anything the `TmpMountGuard` branch doesn't.

Left `bind()` as-is since it's a separate raw-mount path with its own existing reconcile; the duplication you flagged was the generic `MountGuard::mount` copy, which is now gone.

dr-bonez (approved):

helix-nine:
Thanks for the review and merge, @dr-bonez. Scoping the reconcile to `TmpMountGuard` ended up cleaner and better-justified than the original generic approach — appreciate the nudge.

PR #3323, opened 2026-06-16
feat: package init progress reporting

helix-nine:
Adds progress reporting for package **init**, surfaced in the **Installing / Updating** phase of an install — mirroring the phased backup-progress feature (#3250).

## What

A service can now stream progress while its init procedure runs, the same way `createBackup` streams backup progress. The value lands in the package's install/update finalization phase, so it shows up in the existing install-progress UI with no web changes (the wire format is the same `FullProgress` shape, already rendered via `leafProgress`).

## How

**Core**
- `ServiceActorSeed` gains an `init_phase` slot, exactly like `backup_phase`.
- The install **finalization** `PhaseProgressTrackerHandle` (the "Installing"/"Updating" phase) is now handed into `Service::new` and stashed in `init_phase` for the duration of `persistent_container.init(...)`. It's completed when the install finishes.
- New `set-init-progress` effect routes container-streamed `Progress` into that phase via `set_phase_value` (accepts a leaf `Progress` or a nested `FullProgress`).

**SDK**
- `effects.setInitProgress({ progress })` (base `Effects`, wired in package `StartSdk`), with the container-runtime `EffectCreator` routing it to `set-init-progress`.
- `setupInit` drives it automatically: one step per composed init handler as each completes, so multi-step inits show progress with no extra package code. Authors wanting finer detail can call `effects.setInitProgress` directly (optionally with the `FullProgressTracker` utility).
- ts-bindings regenerated (adds `SetInitProgress`; also sweeps in pre-existing `ShutdownParams` binding drift that was already on master).

## Verification

- `cargo check -p start-os --lib` — clean
- `make ts-bindings` + `cd sdk && make baseDist dist` — clean (tsc passes for base + package)
- SDK `startosTypeValidation` test — passes
- `cd container-runtime && npm run check` — passes
- `cd web && npm run check` — only 2 **pre-existing** errors on master (Taiga `tuiCardOptionsProvider` `appearance` in marketplace components), unrelated to this change
- nightly `rustfmt --check` clean on the touched files

Docs: companion start-docs PR to follow.


helix-nine:
Companion docs PR: Start9Labs/start-docs#96

helix-nine:
Revised per review — service code no longer calls the progress effect directly.

- `FullProgressTracker.sync(effects)` now calls the host effect under the hood. The effect (`setInitProgress`/`setBackupProgress`) is baked into the **root** tracker at construction by the harness; `sync()` on any nested tracker bubbles up to the root and reports the whole tree. Added `reset()` (re-run safety) and parent linkage.
- `setupInit` builds the root tracker and hands **each init handler its own nested sub-tracker** as a 3rd arg (`InitFn`/`InitScript` gained an optional `progress` param). Handlers add phases unaware of each other; even no-op handlers advance the bar as they complete.
- **Migrations** receive the same tracker via `migrations.up/down/other` opts (`async ({ effects, progress }) => …`), so a long data migration reports progress during an update.
- **Backups mirror the structure**: `createBackup`/`restoreBackup` report via `tracker.sync(effects)`; pre/post hooks now receive a `FullProgressTracker`; restore progress flows through the init tracker.
- `FullProgressTracker` is reachable via the public `utils` export (`utils.FullProgressTracker`) — no deep import.

Docs updated to the tracker API + a migration example: Start9Labs/start-docs#96

Verified: SDK `tsc` + jest (base & package), container-runtime `npm run check`, web `npm run check` (only the 2 pre-existing marketplace Taiga errors). Effect plumbing/osBindings from the first commit unchanged.

helix-nine:
Added background auto-sync (commit `6a939a708`).

`FullProgressTracker` now reports automatically on every phase update — service code no longer calls `sync()` in the hot path. The effects context is captured at tracker creation (baked into the sink closure).

Coalescing, exactly as specified:
- nothing in flight → launch the report, don't await (but track it)
- one in flight, nothing queued → mark a single queued slot, auto-launch when the in-flight one resolves
- one in flight + one queued → replace the queued slot (the queued launch snapshots fresh, so latest always wins)

`sync()` is now a parameterless **flush** — resolves once in-flight + queued have drained; the harnesses call it before returning. `PhaseHandle` fires an `onChange` after each mutator; nested-child updates bubble to the root, which owns the in-flight/queued state machine. `setupInit` / `createBackup` / `restoreBackup` dropped their per-update `sync` calls and keep only a final flush.

Added unit tests for the coalescing semantics (burst → 1 in-flight + 1 queued with latest-wins, `sync()` drains both, nested bubbling, sink-less no-op).

Docs updated to the no-manual-sync pattern: Start9Labs/start-docs#96

Verified: SDK `tsc` + jest (base & package, incl. new tracker tests), container-runtime `npm run check`, web `npm run check` (only the 2 pre-existing marketplace Taiga errors).

sdk/base/lib/inits/setupInit.ts
```diff
@@ -35,9 +49,21 @@ export type InitScriptOrFn<Kind extends InitKind = InitKind> =
...
 export function setupInit(...inits: InitScriptOrFn[]): T.ExpectedExports.init {
   return async opts => {
+    // Root tracker reports to the install/update finalization phase, with the
+    // effects context baked into the sink. Each init gets its own nested
+    // sub-tracker so handlers stay unaware of each other. Phase updates
+    // auto-sync in the background; we only flush at the end.
+    const tracker = new FullProgressTracker(progress =>
+      opts.effects.setInitProgress({ progress }),
+    )
+    const phases = inits.map((_, idx) => tracker.addNestedPhase(`init:${idx}`, 1))
```
dr-bonez: why are we renaming the phases here?

helix-nine: Good catch — that synthetic `init:${idx}` name was wrapping each init in its own nested phase, which was unnecessary (and the name is never displayed — the install UI collapses the Installing/Updating phase to its overall). It was leftover from giving each init an isolated sub-tracker.

Fixed in de6e6fa0c: there's now one shared root tracker passed to every init handler (matching your original "a FullProgressTracker passed through to each init fn"), and each handler adds its own real-named phases directly — no synthetic wrapping. A constRetry re-run gets a detached tracker so phases don't pile up on the root over the container's lifetime. Removed the now-unused `reset()`.

dr-bonez (approved):

Issue #3336, opened 2026-06-20
[bug]: Read-only file system (os error 30) when syncing Bitcoin Core


otech47:
### Server Hardware
Start9 **Server One** 16 GB RAM, 2 TB NVMe (Samsung SSD 990 EVO Plus 2TB, firmware `2B2QKXG7`). AMI BIOS 5.19. Boot and data both on same disk. It shipped with StartOS 0.3.5, then I flashed it to 0.4 with a full clean disk overwrite

### StartOS Version
`0.4.0-beta.9`

### Client OS
n/a for this report (server-side bug, independent of the client device)

### Client OS Version
n/a

### Browser
n/a (server-side, browser-independent)

### Browser Version
n/a

### Current Behavior
- Brand-new Server One, freshly flashed to StartOS 0.4, with a clean install of Bitcoin Core. No customizations.
- A few percent into the Bitcoin sync, the data volume goes read-only and login to the web GUI breaks.
- The GUI page itself loads fine. The problem is signing in: entering the correct password returns an error instead of logging in (exact text below).
- Only a hard power cycle recovers it. After that, login works again, but resuming the sync brings the failure back.
- Reproducible every time once the sync has run a while.

### Expected Behavior
- A service doing sustained disk writes (Bitcoin initial sync) should be able to run to completion without the server going read-only or needing a hard reset.

### Steps to Reproduce
1. Brand-new Start9 Server One (recommended packaged hardware, 16 GB RAM, 2 TB NVMe, no customizations). It shipped with StartOS 0.3.5.
2. On 0.3.5: booted it, briefly started and stopped a Bitcoin Core sync. Did not run it long, so I cannot say for sure 0.3.5 is unaffected.
3. Flashed StartOS 0.4 (`0.4.0-beta.9`), which fully overwrote the disk. The flash went smoothly.
4. Installed Bitcoin Core fresh on 0.4 (Bitcoin Core 31.0, StartOS package `31.0:14`) and started it (initial block download).
5. A few percent into the sync the failure starts: the data volume goes read-only, GUI login then fails with the error, and only a hard power cycle recovers it. Recurs on resuming the sync.

- The unit arrived a little banged up from shipping (via BitSaga), so my first thought was a hardware fault. Diagnosis ruled that out. The drive itself seems healthy

### Anything else?
SSHed in to debug and under sustained write load the kernel log repeats this:

```
nvme nvme0: controller is down; will reset: CSTS=0x3, PCI_STATUS=0x11
```

[...]

PR #3338, opened 2026-06-21
feat: Add NVMe formatting and mount logic

winshaurya:
Solves #3336 bug 

This is Automated detection and ext4 formatting for NVMe drives during system startup. 

Approach: used nvme-cli for low-level call to securely wipe and prepare new drives.

- Introduced nvme.rs to parse block devices and identify uninitialized /dev/nvmeXn1 drives.
- Uses nvme format , followed by creating an ext4 filesystem.
- Hooked the handle_nvme_initialization workflow into the core init.rs boot sequence.
- Please review I have also put comments for dev references.

PS : first PR in the project

@MattDHill

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -237,6 +237,7 @@ zbus = "5.1.1"
 
 [dev-dependencies]
 clap_mangen = "0.2.33"
+tempfile = "3.8"
 
 [target.'cfg(target_os = "linux")'.dependencies]
 procfs = "0.18.0"
--- a/core/src/init.rs
+++ b/core/src/init.rs
@@ -426,6 +426,13 @@ pub async fn init(
         run_script("/media/startos/config/postinit.sh", progress).await;
     }
 
+    // Verify NVMe workarounds (APST / HMB) if NVMe drives are present.
+    // Best-effort: errors during verification should not block boot.
+    if let Err(e) = crate::system::nvme::check_nvme_quirks().await {
+        tracing::warn!("Failed to verify NVMe workarounds: {e}");
+        tracing::debug!("{e:?}");
+    }
+
     tracing::info!("{}", t!("init.system-initialized"));
 
     Ok(InitResult {
--- a/core/src/system/mod.rs
+++ b/core/src/system/mod.rs
@@ -1,3 +1,5 @@
+pub mod nvme;
+
 use std::collections::BTreeSet;
 use std::fmt;
 use std::time::Duration;
--- a/core/src/system/nvme.rs
+++ b/core/src/system/nvme.rs
@@ -1,0 +1,286 @@
+use std::path::Path;
+use tracing::warn;
+use crate::prelude::*;
+
+/// Status representing the state of NVMe quirks on the system.
+///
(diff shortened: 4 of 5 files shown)


helix-nine:
Thanks for taking a swing at this, @winshaurya — and welcome, first PRs are great to see. We're going to close this one, but here's the full reasoning so it's useful to you:

**1. The description doesn't match the code.** The PR text describes detecting uninitialized NVMe drives, `nvme format`-ing them, creating an **ext4** filesystem, and hooking a `handle_nvme_initialization` workflow into init. None of that is in the diff — and StartOS is btrfs, not ext4, so auto-formatting/wiping drives on boot would be destructive. The actual diff is APST/HMB kernel-parameter tweaks plus a read-only verification module. Whatever you submit should be described by what it actually does.

**2. The root cause isn't confirmed.** This targets #3336, where "disable APST + HMB" is the reporter's own hypothesis from a ~27-minute sample — they explicitly call it "suggestive rather than proven." There's upstream counter-evidence on this exact drive (Samsung 990 EVO Plus) where `nvme_core.default_ps_max_latency_us=0` did *not* stop the controller resets. We don't want to ship a global hardware workaround before the cause is actually pinned.

**3. The HMB change wouldn't take effect as written.** On our stock Debian kernel, nvme loads from the initramfs to mount root, so `options nvme max_host_mem_size_mb=0` in `/etc/modprobe.d/` only matters if it's baked into the initramfs — but the file is written *after* `update-initramfs` runs in `postinst`, so it never makes it in, and the controller probes with HMB still enabled. The robust mechanism (and what the reporter actually tested) is the kernel cmdline for both params, not modprobe.d.

**4. Scope.** Globally disabling HMB/APST for every NVMe drive on every StartOS device to chase one controller model is a blunt instrument with a real perf cost on unaffected drives. The upstream-correct route for one misbehaving controller is a PCI-ID-matched entry in the kernel's nvme quirk table.

We're going to gather more diagnostic data on #3336 first to confirm what's actually happening before deciding on a fix. Genuinely appreciate the effort — please don't let this discourage you from contributing.

winshaurya:
@helix-nine, thank you for taking the time to review my earlier PR and for such a detailed breakdown. Your feedback really helped me understand why a global hack wasn't the right approach, and I appreciate you pointing me in the right direction!

I learned a lot digging back into this. I realized you were totally right about the HMB workaround not taking effect in my first attempt since the NVMe driver is built-in, modprobe configs get ignored, so I had to wire the validation logic to check /proc/cmdline instead.

I've opened a new PR with a much safer approach based on your guidance:

I completely dropped the global overrides so we don't risk power regressions for other users.
I moved the Rust validation logic to the correct spot in start-core and fixed it to read from the kernel command line.
To help get the exact data you asked for, I wrote a bash script (projects/start-os/scripts/nvme-isolation-test.sh) that @otech47 can safely run to cycle through and isolate apst, hmb, and aspm one by one during their sync.
I'm hoping this gets us the clean data we need! Once we know which specific flag keeps the drive stable, would our next step be to implement a targeted PCI-ID quirk just for this drive? I would love your thoughts on whether I'm on the right track here. Thanks again for the mentorship on this one!

PR #3342, opened 2026-06-22
fix(deps): re-pin async-acme to restore the HTTP backend (ACME issuance)

helix-nine:
## Problem

ACME cert issuance has been **completely broken on master since 2026-06-09** (commit cb2f63277, #3303). That PR bumped `async-acme` to fork rev `6da895b`, whose Cargo.toml points `use_tokio`/`use_async_std` at generic-async-http-client's optional **crate deps** (`hyper`, `async-h1`) instead of its backend-**selecting** features (`use_hyper`, `use_async_h1`).

In gahc 0.6 the HTTP backend is gated on `#[cfg(not(any(feature = "use_hyper", feature = "use_async_h1")))]`, so only the optional crate was enabled and the **dummy backend** compiled. Every ACME order fails on the very first outbound request (the directory fetch) with `No HTTP backend was selected` / `http request error: not implemented`. CI never caught it because nothing exercises a live ACME order.

## Fix

Re-pin `async-acme` to the fixed fork rev (**depends on Start9Labs/async-acme#3**, which points the two features at `use_hyper`/`use_async_h1`). No start-os source change — async-acme's public API is unchanged; only its internal backend selection is corrected.

## Verification

- `async-acme` builds the real hyper backend standalone.
- After re-pinning, `cargo update -p async-acme` pulls in `serde_qs` — a dep **only `use_hyper` enables** — confirming the dummy backend is gone.
- `cargo check --lib` green on master.

> Note: the rev pins the fork PR branch commit; once async-acme#3 merges, the rev can be confirmed/updated to the merged commit (merge rather than squash to preserve `d29d876`, or re-pin).

### DIFF
--- a/core/Cargo.lock
+++ b/core/Cargo.lock
@@ -137,7 +137,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -148,7 +148,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


helix-nine:
Updated: async-acme#3 was squash-merged as `4f71d55`, so I've re-pinned to that permanent commit on async-acme `main` (the PR-branch commit `d29d876` is now orphaned). Identical feature fix — `serde_qs` still resolves (real `use_hyper` backend) and `cargo check --lib` is green. Ready to merge.

dr-bonez (approved):

PR #3351, opened 2026-06-25
fix(os-install): scope disk teardown to the target disk so Ventoy installs work

### LINKED ISSUE #3189: [bug]: StartOS Install does not work properly with ventoy
### Prerequisites

### Server Hardware

Lenovo ThinkCentre M900 Intel i5-6500T 2TB SanDisk Extreme

### StartOS Version

startos-0.4.0-beta.6_x86_64-nonfree.iso

### Client OS

Linux

### Client OS Version

Debian GNU/Linux 12 (bookworm)

### Browser

Firefox

### Browser Version

128.14.0esr

### Current Behavior

When I attempt to install startos-0.4.0-beta.6_x86_64-nonfree.iso from USB to a m.2 NVMe drive running a SanDisk Extreme 2TB X3N drive, I get this error:

```
RPC ERROR: Disk Management Error
blockdev: loctl error on BLKRRPART:
Device or resource busy
```

If I attempt to install startos-0.3.5.1-39de098_x86_64-nonfree.iso (from USB to the same drive on the same machine), everything installs fine (no issue). 

BIOS is fully updated (version FWKTBFA).

### Expected Behavior

To be able to install startos-0.4.0-beta.6_x86_64-nonfree.iso from USB to a m.2 NVMe drive running any make/model of storage.

### Steps to Reproduce

- Create a Ventoy USB drive.
- Add startos-0.4.0-beta.6_x86_64-nonfree.iso as an option
- Add startos-0.3.5.1-39de098_x86_64-nonfree.iso
- Boot Ventoy > startos-0.3.5.1-39de098_x86_64-nonfree.iso
- Verify you can install startos-0.3.5.1-39de098_x86_64-nonfree.iso onto the m.2 SanDisk Extreme 2TB X3N drive
- Boot Ventoy > startos-0.4.0-beta.6_x86_64-nonfree.iso
- Attempt to install startos-0.4.0-beta.6_x86_64-nonfree.iso onto the same m.2 SanDisk Extreme 2TB X3N drive
- Observe the reported error in the top right of Firefox
- Use ShredOS to wipe the startos-0.4.0-beta.6_x86_64-nonfree.iso and try 0.4.0 again with same result

### Anything else?

Please advise how I can be of assistance or further help troubleshoot.

helix-nine:
## Problem

StartOS 0.4 installs fail when the installer ISO is booted via **Ventoy** (#3189). Writing the same ISO directly to USB (balenaEtcher / `dd`) always works. The beta.9 fix did not resolve it — two reporters retested and still hit `btrfs error 11` and/or a browser *"Request Error: your browser couldn't reach your server"* during install.

## Root cause

`quiesce_disk()` (added to release OS-level claims on the install target before repartitioning) ran four **system-global** commands, none scoped to the chosen disk:

- `swapoff -a`
- `vgchange -an` (deactivates *all* VGs)
- `dmsetup remove_all --force` (resets *all* device-mapper devices)
- `btrfs device scan --forget` (forgets *all* btrfs devices)

Ventoy presents the booted ISO to the kernel as a **device-mapper device** (`/dev/mapper/ventoy`, a dm-linear table its initramfs builds with `dmsetup create`), and the running live OS reads its squashfs/root **through that dm device** the whole time. Per `dmsetup(8)`, `remove_all --force` does **not** skip an open device — it *"replaces the table with one that fails all I/O"* (the kernel `error` target). So mid-install it severs the live medium: subsequent reads return EIO, surfacing as `btrfs error 11` (EAGAIN) and the StartOS server process dying (the browser then can't reach it).

A directly-written USB boots off a plain **loop** mount with no dm device, so the global teardown finds nothing to break — exactly why direct-write installs succeed and Ventoy ones fail.

Note: the global `dmsetup remove_all --force` predates beta.9 (it's been there since alpha.18). What beta.9 changed was replacing the early-aborting `BLKRRPART` partition reread with busy-tolerant `partx --update`, so execution now proceeds *past* the dm teardown to read the now-dead live medium. Same culprit, newly exposed.

## Fix

Scope every teardown step in `quiesce_disk()` to the target disk and the devices stacked on it (built by walking the sysfs holder graph, `/sys/class/block/<dev>/holders`):

- **swapoff / umount** — only the target's partitions and their dm devices
- **vgchange -an** — only VGs whose PVs sit on the target (via the existing `pvscan()` helper)
- **dmsetup remove --force** — only dm devices stacked on the target, **top-of-stack first** (so LUKS-on-LV-on-PV unwinds in the right order without `EBUSY`); never `remove_all`
- **btrfs ... --forget** — only the target's partitions (in both `quiesce_disk` and the post-`partx` step in `update_partition_table`)

As a backstop, the live medium device is resolved via `findmnt /run/live/medium` and explicitly excluded from every set, so even an unforeseen overlap can never tear it down. The Ventoy `/dev/mapper/ventoy` device lives on the USB stick, not the target disk, so it's never in any of these scoped sets to begin with.

`lvm2` and `util-linux` (providing `pvscan`/`vgchange`/`findmnt`/`partx`) are already image dependencies.

## Testing

- `cargo check -p start-os` ✅, `cargo fmt --check` (this file) ✅, `os_install` unit tests ✅
- **Needs real-hardware verification** — this can't be exercised in CI/VM without a Ventoy USB. The decisive test: boot the installer ISO via Ventoy (confirm `/dev/mapper/ventoy` exists), install to an internal disk, and confirm no `btrfs error 11`, the server stays reachable, `dmsetup ls` still shows `ventoy`, and the install completes. Two reporters on #3189 offered to test — happy to provide a build.
- Non-Ventoy regression worth confirming on a VM with a pre-existing LVM/LUKS stack on the target disk: quiesce should tear down only that stack and the install should still complete.

Fixes #3189

### DIFF
--- a/shared-libs/crates/start-core/src/os_install/quiesce.rs
+++ b/shared-libs/crates/start-core/src/os_install/quiesce.rs
@@ -48,28 +48,6 @@ async fn canonical(path: &Path) -> PathBuf {
         .unwrap_or_else(|_| path.to_path_buf())
 }
 
-/// Block device backing the running live/install medium. Ventoy exposes the
-/// booted ISO as a device-mapper device (`/dev/mapper/ventoy`); teardown must
-/// never touch it or the OS we're running from disappears mid-install. Returns
-/// `None` when there's no such mount (a directly-written USB boots off a plain
-/// loop/partition that isn't a teardown target anyway).
-async fn live_medium_device() -> Option<PathBuf> {
-    let out = Command::new("findmnt")
-        .arg("-n")
-        .arg("-o")
-        .arg("SOURCE")
-        .arg("/run/live/medium")
-        .invoke(ErrorKind::Filesystem)
-        .await
-        .ok()?;
-    let src = String::from_utf8(out).ok()?;
-    let src = src.trim().split('[').next().unwrap_or("").trim();
-    if src.is_empty() {
-        return None;
-    }
-    Some(canonical(Path::new(src)).await)
-}
-
 /// Direct holders of `dev` (`/sys/class/block/<dev>/holders/*`) as `/dev/<name>`.
 async fn holders_of(dev: &Path) -> Vec<PathBuf> {
     let Some(name) = dev.file_name() else {
@@ -99,23 +77,18 @@ async fn dm_name(dev: &Path) -> Option<String> {
 
 /// Every device-mapper device stacked on `partitions`, ordered top-of-stack
 /// first (deepest holder first) so `dmsetup remove` never hits a still-held
-/// parent. Walks the holder graph in sysfs; anything in `exclude` (and devices
-/// reached only through it) is skipped.
-async fn stacked_dm_devices(partitions: &[PathBuf], exclude: &BTreeSet<PathBuf>) -> Vec<PathBuf> {
+/// parent. Walks the holder graph in sysfs, so it only ever returns devices
+/// that actually build on the target disk — a live boot medium on another disk
+/// (e.g. a Ventoy `/dev/mapper` device) is never reached, and so never removed.
+async fn stacked_dm_devices(partitions: &[PathBuf]) -> Vec<PathBuf> {
     let mut depth: BTreeMap<PathBuf, usize> = BTreeMap::new();
     let mut frontier: Vec<PathBuf> = partitions.to_vec();
     let mut d = 0usize;
     while !frontier.is_empty() && d < 64 {
         let mut next = Vec::new();
         for dev in &frontier {
-            if exclude.contains(&canonical(dev).await) {
-                continue;
-            }
             for holder in holders_of(dev).await {
                 let ch = canonical(&holder).await;
-                if exclude.contains(&ch) {
-                    continue;
-                }
                 let slot = depth.entry(ch.clone()).or_insert(0);
                 *slot = (*slot).max(d + 1);
                 next.push(ch);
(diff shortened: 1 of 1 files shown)


shared-libs/crates/start-core/src/os_install/quiesce.rs
```diff
@@ -39,67 +41,211 @@ pub async fn list_partitions(disk_path: &Path) -> Result<Vec<PathBuf>, Error> {
...
-        .await
-    {
-        tracing::warn!("swapoff -a failed during quiesce: {e}");
+    let mut exclude = BTreeSet::new();
+    if let Some(medium) = live_medium_device().await {
+        exclude.insert(medium);
     }
 
-    // umount -A all mountpoints; lazy-unmount on busy.
+    let dm_devices = stacked_dm_devices(&partitions, &exclude).await;
```
dr-bonez: Instead of excluding the live medium, it would be better to only clean up devices that have the target disk in use

dr-bonez (changes requested):

shared-libs/crates/start-core/src/os_install/quiesce.rs
helix-nine: Done in 03663cf02. Dropped `live_medium_device()` and the `exclude` plumbing entirely — `stacked_dm_devices()` walks the holder graph starting from the target disk's own partitions, so it only ever returns devices that build on the target disk. A Ventoy `/dev/mapper` device lives on the boot USB and never stacks on the install target, so it's never in any teardown set to begin with — the explicit exclusion was redundant. swapoff/umount/vgchange/dmsetup all operate on that target-derived set only. cargo check + os_install tests green.

shared-libs/crates/start-core/src/os_install/quiesce.rs
```diff
@@ -39,67 +41,180 @@ pub async fn list_partitions(disk_path: &Path) -> Result<Vec<PathBuf>, Error> {
     Ok(out)
 }
 
+/// Canonical `/dev` node for `path`, falling back to the input on error.
+async fn canonical(path: &Path) -> PathBuf {
```
dr-bonez, line 45: Pretty sure we already have a util for canonicalization

dr-bonez (approved):

shared-libs/crates/start-core/src/os_install/quiesce.rs
helix-nine: Good catch — `crate::util::io::canonicalize`. Since this already merged, fixed in follow-up #3359: the helper now delegates to it (identical for the existing /dev nodes we dedup; partial-resolves a NotFound tail instead of returning the raw input). Kept the thin infallible wrapper so it's still a usable set key.

PR #3353, opened 2026-06-25
feat(net): control upstream cert validation when rewrapping SSL

helix-nine:
## What

Adds `addSsl.upstreamCertValidation` to control how the OS reverse proxy validates the **container's** TLS certificate when it **rewraps SSL** — i.e. when `addSsl` is set **and** the protocol's `secure.ssl === true` (`https`/`wss`). In that case the OS terminates the client's TLS and opens a *fresh* TLS connection to the container.

Previously that upstream leg always validated the container's cert against the StartOS **root CA** (with `ServerName::IpAddress(container_ip)`). For the common case — a container serving its **own** self-signed cert on the trusted internal bridge — that validation can't succeed, so there was no working way to rewrap SSL to such a container.

## The new option

`AddSslOptions.upstream_cert_validation` (TS: `addSsl.upstreamCertValidation`):

| Value | Behavior |
|-------|----------|
| absent (`None`) | Validate against the StartOS root CA — **default, unchanged** |
| `'disable'` | Skip certificate validation entirely |
| `{ certificate: '<pem>' }` | Validate against a supplied PEM certificate/chain instead of the root CA |

```ts
multi.bindPort(443, { protocol: 'https', addSsl: { upstreamCertValidation: 'disable' } })
```

Design per @drbonez: default validates; opt in to disabling, or to pinning a literal certificate. Knob lives on `addSsl`.

## How

(paths below moved with the monorepo reorg: `core/src/net/...` → `shared-libs/crates/start-core/src/net/...`, SDK → `shared-libs/ts-modules/start-core` + `projects/start-sdk`.)

- `net/host/binding.rs` — new `UpstreamCertValidation` enum (`Disable | Certificate(String)`) and `Option<UpstreamCertValidation>` field on `AddSslOptions` (`None` = default root-CA validation). `#[serde(default)]` + `#[ts(optional)]` → fully backward compatible.
- `net/tls.rs` — `client_config_no_verify` (a rustls dangerous `ServerCertVerifier` that accepts any cert but still verifies handshake signatures via the crypto provider) and `client_config_with_cert` (root store from a PEM).
- `net/net_controller.rs` — selects the client config for the rewrap leg from the binding's mode. The no-verify config is built once at init; per-cert configs are cached by PEM so a given `Certificate` yields a **stable `Arc`** across host updates (otherwise `ProxyTarget` equality would churn and tear down the vhost on every update). Invalid supplied PEM fails closed (falls back to root-CA validation → handshake fails, surfacing the misconfig) and logs.
- SDK: regenerated `osBindings` (`make ts-bindings`), re-exported `UpstreamCertValidation` from `interfaces/Host.ts`, CHANGELOG entry under the unreleased `2.0.0`.

## Notes / caveats

- `'disable'` is intended for the trusted OS↔container bridge only; it skips cert validation on that internal hop.
- For `{ certificate }`, webpki still checks the connect server name (the container IP) against the presented cert's SANs, so the pinned cert must be valid for that IP — otherwise use `'disable'`.

## Docs

Per the monorepo docs-ship-with-code pattern, the docs are included in this PR (not a separate start-docs PR). `projects/start-sdk/docs/src/interfaces.md` gains an `upstreamCertValidation` row on the bindPort Options table and a **Rewrapping SSL to a TLS container** section. Folded in from the former start-docs companion PR #99 (now closed).

## Testing

- Rebased onto monorepo `master`; `cargo check -p start-core` clean.
- `make test-sdk` — builds the ts-module dist + SDK bundle and passes (68/68); confirms the regenerated bindings + `Host.ts` re-export typecheck.
- `make ts-bindings` regenerates `UpstreamCertValidation.ts` + the `AddSslOptions` field identically to the pre-rebase branch.


helix-nine:
Docs companion PR: https://github.com/Start9Labs/start-docs/pull/99

dr-bonez (approved):

PR #3357, opened 2026-06-27
fix(mount): apply dependency-mount idmap host-side

helix-nine:
## Problem

A dependency mount with a custom SDK idmap (mostro → lnd, `idmap: [{ fromId: 0, toId: 1000 }]`) fails on every retry with:

```
open_tree_attr IDMAP: Operation not permitted (os error 1)
```

### Root cause

Creating an idmapped mount requires `CAP_SYS_ADMIN` **in the source superblock's user namespace** (`can_idmap_mount()` → `ns_capable(sb->s_user_ns, CAP_SYS_ADMIN)`). The SDK idmap on a pointer mount was applied from *inside the LXC* via `start-container bind-mount --idmap`, running in the container's userns (uid 0 → host 100000). But the dependency's volume superblock is host-mounted (`init_user_ns`), and a descendant userns holds no caps over an ancestor → **EPERM**, for any host-owned superblock. The in-LXC application could never succeed against a host-owned superblock; it was only ever validated host-side as root in #3248, never through a booted service, so it surfaced the first time a real package used a non-empty idmap on a dependency mount.

## Fix

Apply the idmap in the host-side `mount` effect (startd, which holds `CAP_SYS_ADMIN` over `init_user_ns`):

- `MountTarget` gains an `idmap` field (the SDK already sent it; Rust ignored it).
- `dependency.rs::mount` idmaps a plain bind of the source once, host-side, with the map from `pointer_idmap()`.
- The in-LXC `start-container bind-mount` step in `SubContainer`'s pointer branch is dropped.

The EPERM fix is just *"apply it host-side."* Nothing here re-idmaps an already-idmapped mount, so it doesn't depend on the `open_tree_attr` REPLACE capability or the `unshare-userns` machinery from #3248 — verified that the pre-`open_tree_attr` path (`open_tree` + `mount_setattr(MOUNT_ATTR_IDMAP)`), and even `mount -oX-mount.idmap`, handle this case (the migration is orthogonal and stays).

## Composition semantics (`pointer_idmap`)

The SDK idmap is an **override layered on top of** the LXC base map (per the `Mounts.ts` docs: "the base LXC mapping is applied automatically; do not include it here"), and a list of entries **composes sequentially** — each remaps the output of the ones before it. The layers can't be stacked as separate mounts: re-idmapping an already-idmapped mount **replaces** the idmap rather than composing (verified on a 6.15+ kernel — `base 10→500` then `500→900` yields `nobody`, not 900). So the whole stack is collapsed, pointwise, into one equivalent `uid_map`:

- empty idmap → just the base map (unchanged default).
- non-empty → for each on-disk id, fold the idmaps over its base mapping (`k → base+k`); `to_id` is container-relative, lifted into the host namespace by the base offset.

The composition can be non-injective — overriding `0→1000` puts on-disk 0 into uid 1000's slot, so on-disk 1000's base passthrough is displaced — and a `uid_map` can't send two on-disk ids to one target. Colliding ids **fall back to `nobody`** (an explicit remap winning over a passthrough), which matches how the base map already drops out-of-range ids.

## Scope / follow-up

Volume and asset mounts still apply their idmap in-LXC. They have no fleet users with a non-empty idmap today and share the same latent limitation against host-owned superblocks — a separate change (they have no host-side mount effect today), after which the in-LXC `start-container bind-mount` idmap branch + `unshare-userns` helper can be removed entirely.

## Testing

- ✅ **Unit + proptest** for `pointer_idmap`: `flatten_is_the_injective_restriction_of_composition` proves, over random idmap lists, that it is a valid `uid_map` which never disagrees with the reference composition and preserves every uniquely-reached target (only genuine collisions drop). Plus explicit cases (mostro override, sequential chaining).
- ✅ **Privileged integration test** (`#[ignore]`, root + 6.15+ kernel): runs the **real** host-side path — `IdMapped::new(Bind, pointer_idmap(&idmap)).mount(...)` — and reads uids back through the resulting mount (on-disk 0 → host 101000, displaced id → nobody). A `#[cfg(test)]` hook lets the userns helper resolve outside the `startbox` multicall binary. Run: `sudo <test-bin> --ignored --exact service::effects::dependency::test::idmap_host_side_remaps_uids --nocapture`.
- ✅ **Full end-to-end on a patched image**: built an ISO from this branch, stood up a fresh StartOS VM (kernel 7.0.10), and installed two minimal dummy packages — a provider that writes a uid-0 file into its volume, and a consumer that `mountDependency`s it with `idmap: [{fromId: 0, toId: 1000}]`. Consumer logs: `IDMAP_RESULT uid=1000 gid=1000 content=idmap-secret`, **zero EPERM**. The exact failure path now works through a real service.
- ✅ Reproduced the original EPERM vs. host-side success on a 7.0.10 kernel with a raw `open_tree_attr` harness (see comments).
- ✅ `make ts-bindings`, SDK `make baseDist dist`, container-runtime `tsc --noEmit` all clean (three legacy embassy-compat call sites updated to pass `idmap: []`).


helix-nine:
## Kernel-level validation of the diagnosis

Reproduced the exact failure and confirmed the fix's mechanism with a self-contained C harness using the real `open_tree_attr(MOUNT_ATTR_IDMAP)` syscall (467), on a **6.19 kernel** (same family as the trixie-backports 7.0.x StartOS ships, well past the 6.15 floor; #3248 itself was validated on 6.19).

The harness mounts a host-owned tmpfs (superblock `s_user_ns == init_user_ns`, like a StartOS volume), then attempts to create an idmapped clone of it from two contexts, with everything else held identical (same source, same `0:1000:1` idmap):

| Source | from **init userns** (= startd / this PR) | from **0→100000 userns** (= in-LXC, current code) |
|---|---|---|
| plain host-mounted tmpfs | ✅ SUCCESS | ❌ `Operation not permitted` |
| **already base-idmapped** (the mostro scenario) | ✅ SUCCESS | ❌ `Operation not permitted` |

The in-LXC `Operation not permitted` is the user's exact error (`os error 1` = EPERM). The only variable between the success and failure rows is the caller's user namespace — confirming the cause is `can_idmap_mount()`'s `ns_capable(sb->s_user_ns, CAP_SYS_ADMIN)` check, which a descendant userns can never satisfy against a host-owned superblock.

Two things this nails down:
1. **The fix works.** The same operation that EPERMs in-LXC **succeeds from the init userns** — exactly where this PR moves it (startd runs in `init_user_ns`).
2. **It is not the "already-idmapped" limitation.** Row 2 (already-base-idmapped source, the literal mostro case) succeeds host-side, so re-idmapping an already-idmapped mount via `open_tree_attr` is fine — the blocker was always the capability, not the syscall.

### Scope / honesty
This is a kernel-level reproduction of the root cause on a matching kernel, **not** a full end-to-end test of a patched StartOS build with a real mostro→lnd dependency mount (that needs a multi-hour image build + the bitcoind/lnd/mostro dependency chain on a 7.0.x VM). The kernel code path exercised here (`can_idmap_mount` cap check) is unchanged across 6.15→7.0, so the behavior is identical on the user's box. Happy to do the heavier full-image e2e if we want belt-and-suspenders before merge.

helix-nine:
## Verified the RPC effect against the real mount code — and it caught a bug

Per review, I exercised the `mount-pointer` effect's host-side path with start-os's own production code (`IdMapped` + `pointer_idmap`), as a privileged integration test, reading uids back through the resulting mount.

That immediately exposed a bug in the first cut of this fix: I'd composed the SDK idmap **over** the base map (override + a carved base remnant), which the unit tests happily accepted — but it produces a uid_map the kernel **rejects with EINVAL** at runtime. A container-relative `to_id` (e.g. 1000 → host 101000) lands inside the base output range 100000..165535, so the override and a base-remnant entry both map to presented id 101000, and a uid_map can't have overlapping outside ranges.

Fixed to the correct, representable **replace** semantics (which also matches what the old in-LXC clone+`MOUNT_ATTR_IDMAP` actually did — it replaces, never composes). Result on the real mount:

```
on-disk uid 0  -> host 101000   (container uid 1000)   [mapped]
on-disk uid 5  -> 65534         (nobody)               [unmapped — replace, not override]
```

For mostro this is exactly right: lnd runs as container-root, so all its files are on-disk uid 0 and map cleanly to mostro's 1000.

Good call pushing on the verification — the unit tests alone would have shipped the EINVAL.

helix-nine:
## Full end-to-end dummy-service test on a patched image ✅

Built a patched ISO from this branch, stood up a fresh StartOS VM from it (kernel 7.0.x, running the patched startd + container-runtime), and installed two minimal dummy packages built against the patched 2.0.0 SDK:

- **idmap-provider** — writes `/data/secret` into its volume owned by uid 0.
- **idmap-consumer** — depends on the provider and `mountDependency`s its volume with `idmap: [{ fromId: 0, toId: 1000 }]`, then `stat`s the file.

Starting them in order, the consumer's logs show:

```
provider: wrote /data/secret owned by uid 0
...
IDMAP_RESULT uid=1000 gid=1000 content=idmap-secret
```

- **Zero EPERM** anywhere in the logs (grep count: 0) — the in-LXC bind-mount that used to fail is gone; the host-side `mount-pointer` effect applied the idmap.
- The provider's **on-disk uid 0 surfaces as uid 1000** inside the consumer, and the content is readable — the composed/replaced idmap (`0 → base+1000 = 101000` host-side → container 1000) is exactly right.

This exercises the real path end to end: `SubContainer.mountDependency({idmap})` → `effects.mount` (mount-pointer RPC) → `dependency.rs::mount` → `pointer_idmap` → `IdMapped` host-side. The feature now works as intended on a real service, which it never did before (the original in-LXC code EPERM'd 100% of the time).

helix-nine:
## Correction: layer the override on the base map, not replace it

@drbonez is right — the prior cut replaced the base map, so any on-disk id the override didn't name dropped to `nobody`. The SDK contract is an override *on top of* base (`Mounts.ts`: "the base LXC mapping is applied automatically; do not include it here"). Fixed in `a9c398ba8`.

**Why it has to be one merged map, not two stacked mounts** — I tested whether re-idmapping an already-idmapped mount composes or replaces, on a 6.15+ kernel:

```
base-idmapped mount (on-disk 10 -> 500):            uid=500
clone it + idmap (500 -> 900):                      uid=65534 (nobody)
VERDICT: REPLACE — the new idmap maps the raw superblock id, not the base output
```

So `open_tree_attr` on an already-idmapped mount *replaces* the idmap; it doesn't stack. The original in-LXC second-bind would have replaced the base too (a latent bug beyond the EPERM). The only way to "stack" is to **merge** base + override into a single host-side idmap.

`pointer_idmap` now does that: every id keeps `k -> base + k`, except the overridden sources (→ lifted target). One unavoidable wrinkle — a container-relative `to_id` lands inside the base output range, so base would map on-disk id `to_id` to the same target as the override, and a uid_map can't map two on-disk ids to one target. That single colliding id (== `to_id`) drops to `nobody`; it's the override's own target, which a dependency's data isn't expected to own.

`pointer_idmap([{0->1000}])` = `[{0->101000}, {1..999->100001..}, {1001..65535->101001..}]` (on-disk 1000 dropped).

Integration test updated and passing — through the real `IdMapped` mount: on-disk **5 keeps its base mapping (100005)**, on-disk 0 is overridden (101000), on-disk 1000 (collision) drops. So base is preserved for non-overridden ids now.

(The earlier full-image e2e result still holds for mostro specifically — lnd's files are all on-disk uid 0, so single-uid case is unchanged; the fix matters for dependencies whose data spans multiple uids.)

dr-bonez (approved):

helix-nine:
Thanks for the review — merged. The host-side idmap fix is now on master in the monorepo layout (`shared-libs/crates/start-core`, regenerated osBindings, container-runtime adapters), with the proptests and the privileged `#[ignore]` integration test carried over intact.

PR #3361, opened 2026-06-29
chore(rust): resolve dead_code warnings

helix-nine:
Second of three PRs on the Rust workspace warnings. Covers the 10 `dead_code` warnings, handled **per item** rather than with a blanket `#[allow]` — `_`-prefix or removal where it's meaningful, scoped `#[allow(dead_code)]` only where the value is genuinely retained but can't be renamed/removed without changing behavior.

`cargo check --workspace --all-targets` dead_code: **10 → 0**, no new warnings.

### Removed (genuinely unused)
- **`igd::apply_peer_forward`** — a thin `count = 1` wrapper superseded by `apply_peer_forward_range`; no callers. Removed, and de-linked the doc comment that referenced it.
- **`DynVersionT::compat`** — the dynamic-dispatch accessor is never called; the static `VersionT::compat` (still present) covers every use site (`db/model/public`, `registry/device_info`, the version dump). Removed from the trait and both impls.
- **`jsonpath TokenReader::err`** — vestigial vendored field, only ever set to `None`, never read.

### Renamed (RAII)
- **`net_controller::socks` → `_socks`** — holds the live SOCKS5 server thread (`NonDetachingJoinHandle`); never read by design, the `_` signals it's held for its `Drop`. Dropping it would kill the SOCKS server.

### Kept with scoped `#[allow(dead_code)]`
These can't be `_`-prefixed (serde field-name binding / macro-generated accessors / positional tuple fields) or are read only under `#[cfg(test)]`:
- `backup::OsBackupV0::tor_key` — parsed for legacy-format validation; StartOS no longer restores the tor key.
- `merkle_archive::Signer::Signed` context field — retained signing-context provenance, not re-read during verify.
- `version::LTWrapper` second field — read only in a `#[cfg(test)]` path.
- patch-db `Foo`/`Bar` and yasi `HasString` test fields — drive `HasModel`/`TS` derives.

@drbonez — flagging two for your call in particular: removing **`DynVersionT::compat`** (in case it was staged for future compat-gating rather than being truly dead), and keeping vs. dropping the unused **`Signer::Signed` context field**. Happy to flip either.

### DIFF
--- a/shared-libs/crates/start-core/src/version/mod.rs
+++ b/shared-libs/crates/start-core/src/version/mod.rs
@@ -562,6 +562,9 @@ unsafe impl Send for DynVersion {}
 trait DynVersionT: RefUnwindSafe + Send + Sync {
     fn previous(&self) -> DynVersion;
     fn semver(&self) -> exver::Version;
+    // part of the DynVersionT surface; kept intentionally though not dynamically dispatched
+    #[allow(dead_code)]
+    fn compat(&self) -> &'static exver::VersionRange;
     fn pre_up(&self) -> BoxFuture<'static, Result<Box<dyn Any + UnwindSafe + Send>, Error>>;
     fn up(&self, db: &mut Value, input: Box<dyn Any + Send>) -> Result<Value, Error>;
     fn post_up<'a>(&self, ctx: &'a RpcContext, input: Value) -> BoxFuture<'a, Result<(), Error>>;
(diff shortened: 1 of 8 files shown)


shared-libs/crates/start-core/src/version/mod.rs
```diff
@@ -562,7 +562,6 @@ unsafe impl Send for DynVersion {}
 trait DynVersionT: RefUnwindSafe + Send + Sync {
     fn previous(&self) -> DynVersion;
     fn semver(&self) -> exver::Version;
-    fn compat(&self) -> &'static exver::VersionRange;
```
dr-bonez, line 565: Do not remove this

dr-bonez (dismissed):

shared-libs/crates/start-core/src/version/mod.rs
helix-nine: Restored `DynVersionT::compat` (trait decl + both impls) in 945c91d — kept it and marked the trait method `#[allow(dead_code)]` so it stays as part of the surface without re-triggering the "never used" warning. `cargo check` is clean, no compat warning.

dr-bonez (approved):

PR #3362, opened 2026-06-29
chore(rust): migrate crypto stack off deprecated generic-array 0.14

helix-nine:
Third and final PR on the Rust workspace warnings. Clears all **51 `deprecated` warnings** (the entire remaining category) by migrating the RustCrypto + dalek stack off `generic-array 0.14` onto the new hybrid-array generation — the migration you asked for, on the RC dalek/ssh-key crates.

`cargo check --workspace --all-targets` deprecated warnings: **51 → 0**, zero new warnings, zero errors.

> ⚠️ This pulls in **release-candidate** crypto crates (`ed25519-dalek 3.0.0-rc.1`, `x25519-dalek 3.0.0-rc.1`, `ssh-key 0.7.0-rc.11`, `p256 0.14.0-rc.15`). The RustCrypto leaf crates (`digest 0.11`, `sha2 0.11`, etc.) are stable; the dalek/ssh-key majors are not yet. Stable releases look imminent (RustCrypto/traits#1571 is mostly ticked; curve25519-dalek#736 is active). When they land, the only change should be dropping the `-rc` suffixes.

### Dependency bumps
`digest 0.11 · sha2 0.11 · sha3 0.12 · hmac 0.13 · hkdf 0.13 · pbkdf2 0.13 · der 0.8 · pkcs8 0.11 (+pem) · spki 0.8 · aes 0.9 (+ ctr 0.10) · p256 0.14-rc · hashing-serializer 0.2 · ed25519 3 · ed25519-dalek 3.0.0-rc.1 (+alloc,+pem) · x25519-dalek 3.0.0-rc.1 · ssh-key 0.7.0-rc.11 · chacha20 0.10` (backup-fs). `rand 0.9` stays (still used by `rand::random` everywhere); added `getrandom 0.4` for the rand_core-0.10 key-gen RNG.

### Code changes (no behavior change)
- **digest outputs** named via the version-agnostic `digest::Output<D>` alias instead of `GenericArray` directly (12 `disk/mount/filesystem/*`, `util/serde`, `backup/target`).
- **`util/crypto` AES-256-CTR** → `ctr::Ctr64BE<Aes256>` — the *exact* mode aes 0.7's `Aes256Ctr` aliased (64-bit BE counter), so password-wrapped backup keys written by older versions still decrypt. Uses the cipher 0.5 `KeyIvInit::new_from_slices` API.
- **backup-fs chacha20** Key/Nonce built with `From`/`TryFrom` instead of the now-deprecated `Array::from_slice`.
- **RNG**: rand_core 0.10 key generation uses `getrandom::SysRng` wrapped in `UnwrapErr` (the form ed25519-dalek/ssh-key document), behind a `util::crypto::os_rng()` helper. Replaces the old `ssh_key::rand_core::OsRng::default()`.
- **`sign/mod`**: pkcs8 0.11 `PrivateKeyInfoRef`; `Encode{Private,Public}Key` brought into method scope.

### Verification
Crypto is wire-format-sensitive (backup key wrapping, chacha20 vault, ed25519 signing, pkcs8/PEM), so correctness was checked by tests, not just compilation:
- **411 / 411** `start-core` lib tests pass (`--features=test`) — incl. `test_encrypt_decrypt_roundtrip`, `test_gen_awk` (decrypts a hand-coded ciphertext), merkle archive sign/verify, signer flows.
- **72 / 72** `startos-backup-fs` tests pass — vault seal/open, superblock change-password persistence, encryption round-trips.

AES-CTR / chacha20 / ed25519 / pkcs8 outputs are byte-identical to before. No user- or developer-facing surface changes, so no docs/changelog entry.

**Please verify the full CI cross-platform matrix** (x86_64/aarch64-apple-darwin, x86_64/aarch64/riscv64 linux-musl) — local checks are linux-only, and this touches `getrandom`/crypto which can have platform-specific backends.

Stacks on top of (independent of) #3360 and #3361.


shared-libs/crates/start-core/src/util/crypto.rs
```diff
@@ -19,23 +18,27 @@ use ts_rs::TS;
...
-        password.as_ref(),
-        salt.as_ref(),
-        1000,
-        aeskey.as_mut_slice(),
-    )
-    .unwrap();
+/// OS CSPRNG as an infallible rand_core 0.10 source — the form ed25519-dalek and
+/// ssh-key key generation expect (`SysRng` is fallible; `UnwrapErr` panics on the
+/// effectively-never OS RNG failure, matching the old `OsRng` behavior).
+pub fn os_rng() -> getrandom::rand_core::UnwrapErr<getrandom::SysRng> {
```
dr-bonez: Why not just rand::rng()?

dr-bonez (changes requested):

shared-libs/crates/start-core/src/util/crypto.rs
helix-nine: Good question — `rand::rng()` unfortunately does not work here. ed25519-dalek 3.0 / ssh-key 0.7 take `R: rand_core::CryptoRng`, and in **rand_core 0.10** that trait is infallible (`CryptoRng: Rng + TryCryptoRng<Error = Infallible>`). Two problems:

1. Our `rand` is **0.9** (rand_core 0.9). Its `ThreadRng` implements a *different-version* `CryptoRng`, so `rand::rng()` does not satisfy the dalek bound at all — that is the original `the trait bound ThreadRng: rand_core::CryptoRng is not satisfied` error.
2. Bumping `rand` to **0.10** does not fix it either: there `ThreadRng` only implements the **fallible** `TryCryptoRng` (`impl TryCryptoRng for ThreadRng`, no infallible `CryptoRng` impl), so it still cannot be passed to `SigningKey::generate`. You would have to write `UnwrapErr(rand::rng())` — no simpler than `UnwrapErr(SysRng)` — and rand 0.10 also renames `Rng`→`RngExt` / `RngCore`→`Rng`, which would churn our ~30 existing `rand::random` / `rand::Rng` / `rand::RngCore` sites.

So `UnwrapErr(SysRng)` (the form ed25519-dalek’s own docs use) is the minimal correct option. I added a doc comment on `os_rng()` in 5d59d07 spelling this out so it doesn't read as gratuitous. Happy to inline it at the call sites instead of a helper if you'd prefer, but `rand::rng()` itself is not viable until the whole `rand` ecosystem unifies on rand_core 0.10 *and* re-adds an infallible thread RNG.

helix-nine: Done in e006484 — bumped `rand` 0.9 → 0.10 and rewrote the helper to `UnwrapErr(rand::rng())`, dropping the direct `getrandom` dep.

Heads-up on the fallout (rand 0.10 is a bit invasive): its `ThreadRng` is now fallible-only (`TryCryptoRng`), and it renames `RngCore` → `Rng` / `Rng` → `RngExt`. So this commit also:
- wraps every infallible `rng().fill_bytes()` / `random_range()` in `UnwrapErr` (`util/mod`, `net/forward`, and backup-fs `compress`/`contents`/`ctrl`/`superblock`/`vault`/`tests`),
- updates those trait imports, and
- binds the generated key before the `.await` in `s9pk::init::init_workspace` — `ThreadRng` is `!Send` (the old zero-sized `SysRng` was `Send`), so holding it across the await made the handler future `!Send`.

No behavior change; 411 start-core + 72 backup-fs tests pass.

dr-bonez (approved):

PR #3368, opened 2026-06-30
feat(net): IPv6 GUA exposure tri-state (disabled/LAN/LAN+WAN)

helix-nine:
Stacks on **#3366** (`feat/port-range-interfaces`) — base is that branch, not `master`.

Gives IPv6 **GUA** (global-unicast) addresses on a service interface a **Disabled / LAN / LAN+WAN** tri-state instead of the on/off toggle every other address uses. A GUA is a single globally-routable address, so "on" must distinguish LAN-only (firewalled to the subnet) from WAN-exposed. Per @dr-bonez:

- **LAN** (default, behavior-preserving) — reachable on-link; traffic from outside the gateway subnet is rejected (already enforced by `vhost::ProxyTarget::filter()`).
- **LAN+WAN** — also exposed to the WAN (vhost binds the GUA and accepts any source); the automatic gateway pinhole + non-SSL forwarding land in a follow-up (see scope note).
- IPv4 keeps separate LAN/WAN rows; ULAs can't be exposed publicly, so only GUAs get the tri-state.

### In this PR
- **Model** — `GuaAccess { Disabled, Lan, LanWan }` (default `Lan`) + `gua_access: BTreeMap<SocketAddrV6, GuaAccess>` on `DerivedAddressInfo`; `HostnameInfo::gua()`, `access_for`/`is_wan`, GUA-aware `enabled()`.
- **RPC** — `set-gua-access` (mirrors the removed `set_range_gateway_access`); `CliFromJsonString<T>` so the param is a structured `HostnameInfo` over the wire but a JSON-string CLI arg (also applied to the inherited `set_address_enabled`). i18n ×5.
- **Derivation** — both the SSL/vhost and non-SSL paths classify WAN via `is_wan()`.
- **UI** — client-side GUA detection + a Disabled/LAN/LAN+WAN selector on the Interfaces page, wired to `set-gua-access` (+ mock). `check:ui` green.
- **Docs + CHANGELOG.**

Verified: `cargo check`, model unit tests, ts-bindings (no drift), `check:ui`.

### Follow-up — separate PR (per @dr-bonez, easier to review)
The **v6 WAN backend forwarding**, which shares the forward path and needs VM validation:
- `port_map` `Ipv4Addr`→`IpAddr` generalization (UPnP stays v4-gated) + **PCP-v6 pinhole** for LAN+WAN GUAs (the v6 gateway is already in `ip_info.lan_ip`).
- **Non-SSL v6 forwarding** — `lxcbr0` IPv6, container v6 discovery/storage, an `ip6 startos` nft table (filter + LAN source-reject), a v6 `forward-port`, and `net.ipv6.conf.all.forwarding`.

> The slot's `pre-commit` hook is broken (`cd web/`, a stale pre-reorg path), so commits used `--no-verify`; formatting verified via prettier + `check:ui` locally and CI.


shared-libs/crates/start-core/src/net/host/binding.rs
```diff
@@ -57,19 +95,43 @@ pub struct DerivedAddressInfo {
     pub enabled: BTreeSet<SocketAddr>,
     /// User override: disable these addresses (only for domains and private IP & port)
     pub disabled: BTreeSet<(InternedString, u16)>,
+    /// User override: per-GUA exposure (only for IPv6 global-unicast addresses),
+    /// keyed by the GUA's `SocketAddr`. Absent ⇒ [`GuaAccess::Lan`].
+    #[serde(default)]
+    pub gua_access: BTreeMap<SocketAddr, GuaAccess>,
```
dr-bonez:
```suggestion
    pub gua_access: BTreeMap<SocketAddrV6, GuaAccess>,
```

shared-libs/crates/start-core/src/net/host/binding.rs
```diff
@@ -588,3 +659,116 @@ pub async fn set_address_enabled<Kind: HostApiKind>(
...
+    ctx: RpcContext,
+    BindingSetGuaAccessParams {
+        internal_port,
+        address,
+        access,
+    }: BindingSetGuaAccessParams,
+    inheritance: Kind::Inheritance,
+) -> Result<(), Error> {
+    let address: HostnameInfo =
+        serde_json::from_str(&address).with_kind(ErrorKind::Deserialization)?;
```
dr-bonez: why is this serialized twice? just for CLI support? there has to be a better solution. Maybe a CliFromJsonString that parses in the ValueParser trait, but does a passthrough deser in serde?

shared-libs/crates/start-core/src/net/host/binding.rs
helix-nine: Done in 11a7af2c — `gua_access: BTreeMap<SocketAddrV6, GuaAccess>`, and `HostnameInfo::gua()` now returns `SocketAddrV6`. (Renders the same `{ [key: string]: GuaAccess }` in the binding.)

shared-libs/crates/start-core/src/net/host/binding.rs
helix-nine: Yeah, inherited from `set_address_enabled`. Fixed properly in 11a7af2c: added `CliFromJsonString<T>` (util/serde.rs) — serde is a pure passthrough to `T` (the wire carries the object directly, no double-encoding), while the clap `ValueParser` parses the JSON string for CLI use. Paired with `#[ts(as = "HostnameInfo")]` so the binding shows `HostnameInfo`, not a string.

Applied to **both** `set_gua_access` and `set_address_enabled` (per your Matrix note), plus the UI side — `set-address-enabled` now sends the `HostnameInfo` object instead of `JSON.stringify`-ing it (api.types, both toggle callers, mock). Web typecheck left to CI (no node_modules in this slot).

shared-libs/crates/start-core/src/net/service_interface.rs
```diff
@@ -60,6 +60,22 @@ impl HostnameInfo {
...
     }
+
+    /// If this address is an IPv6 **global unicast** address (a GUA — not
+    /// loopback / ULA / link-local), return it as a `SocketAddrV6`. GUAs are the
+    /// only addresses that carry the Disabled / LAN / LAN+WAN tri-state.
+    pub fn gua(&self) -> Option<SocketAddrV6> {
+        if !matches!(self.metadata, HostnameMetadata::Ipv6 { .. }) {
+            return None;
+        }
+        let IpAddr::V6(ip) = self.hostname.parse::<IpAddr>().ok()? else {
```
dr-bonez: just parse directly as Ipv6Addr?

helix-nine: Done in 80091b509 — parses straight to `Ipv6Addr`. Also dropped the `matches!(metadata, Ipv6 { .. })` guard since the parse is the discriminant (a v4 / domain / onion hostname won't parse as `Ipv6Addr`).

shared-libs/crates/start-core/src/net/service_interface.rs
```diff
@@ -65,12 +65,7 @@ impl HostnameInfo {
     /// loopback / ULA / link-local), return it as a `SocketAddrV6`. GUAs are the
     /// only addresses that carry the Disabled / LAN / LAN+WAN tri-state.
     pub fn gua(&self) -> Option<SocketAddrV6> {
-        if !matches!(self.metadata, HostnameMetadata::Ipv6 { .. }) {
```
dr-bonez: no, keep this guard. just not the second

helix-nine: Restored in e98448b8a — kept the metadata guard, only swapped the `IpAddr::V6` parse for a direct `Ipv6Addr` parse.

dr-bonez (dismissed):

dr-bonez:
@helix-nine rebase this

dr-bonez (approved):

helix-nine:
Already covered — I'd rebased this onto `master` (`df10f6a79`) before you merged it, and it squash-landed as `fe52a19ee`. #3373 is now rebased on top of the merged master too (`bc1162907`, 151 `net::` green), so the stack is current.

PR #3383, opened 2026-07-01
fix(net/tunnel/ui): address #3366 review follow-ups (M1, M3–M6, L1–L5, H2)

helix-nine:
Follow-up to the adversarial review on #3366 ([review](https://github.com/Start9Labs/start-technologies/pull/3366#pullrequestreview-4604530337)) — addresses the findings we hadn't already refuted or fixed. All are corrections to the (unreleased) port-range feature; nothing here is a new user surface.

### Withdrawn / already resolved (not in this PR)
- **C1** (fresh range WAN-exposed by default) — **withdrawn** during review: an operator-added public domain *is* the intent to expose, so default-enabling it is correct.
- **H1** (non-exported ranges invisible/forwarded) — **fixed in #3366** by the internal-only rework: a non-exported range serves only lo/lxcbr0 (`enabled_addresses()`), so it's never WAN-forwarded, and the UI intentionally hides it.
- **M2** (port-forward modal "Internal Range") — **withdrawn** (drbonez). The modal is router-facing config: its "Internal Range" is the *server* port the router forwards to, not the container's listening port. StartOS forwards port-preserving at the router↔server boundary (WAN:X → server_LAN:X) and DNATs X → container internally, so both columns are `external_start_port` — the original display was correct. The finding wrongly treated the container port (`addressInfo.internalPort`) as the server port. (Reverted in this PR.)

### Fixed here
| # | Fix |
|---|-----|
| **H2** | `interfaces.md` Port Ranges example returned `void[]` from `setupInterfaces` (`RangeOrigin.export` is `Promise<void>`) — `await` it and `return []`. |
| **M1** | Dropped the stale "per-gateway LAN / LAN+WAN access control" wording — ranges render as per-gateway address cards (that selector was removed in #3366). |
| **M3** | `set_address_enabled` / `set_range_address_enabled` now recompute `host.port_forwards` (`update_addresses`, like the domain handlers), so the "add this rule to your router" list updates immediately instead of staying stale until an unrelated network event. |
| **M4** | `add_sni_forward` now rejects an SNI source that falls inside a different DNAT range (same-source SNI demux still allowed) — the check lives in its `db.mutate`, so it covers both the manual and PCP paths atomically. |
| **M5** | `get_service_interface` watched the whole `/hosts` subtree, waking on every gateway/mDNS/domain recompute. It now resolves the interface's `(host, port)` and watches only that node, falling back to the broad watch until it's exported. Safe because the callback is a no-arg re-run trigger (payload type is irrelevant). |
| **M6** | `add_binding_range` re-keys the preserved `enabled`/`disabled` overrides to the new `external_start_port` when a service-driven rebind moves the span, so a disabled WAN address doesn't silently re-enable after a resize. (+ unit test.) |
| **L1** | StartTunnel DNS **Server** column rendered CNAME/TXT/custom rdata as if it were a server — now only A/AAAA resolve a server name; other types render verbatim. |
| **L2** | StartTunnel forward rows targeting a **client** rendered a raw IP after the picker was narrowed to servers — name resolution now uses an unfiltered device list; the Add picker stays server-only. |
| **L3** | `RangeOrigin.export` now throws on a second call (a range exposes exactly one interface); the prior silent overwrite left the interfaces tracking array disagreeing with what was stored. |
| **L4** | `add_forward` / `apply_peer_forward_range` did a detached `peek().overlapping()` then a separate insert — two concurrent adds could both pass. Now check + insert happen in one `db.mutate` (reserve-then-setup, rolling the DB reservation back if the dataplane install fails), matching `add_sni_forward`. |
| **L5** | Removed the orphaned i18n key 882 `'Port Range'` from all five locales (siblings 883/884/885 are used; 882 was never referenced). |

### CHANGELOG
None. The port-range feature is documented under the in-progress `0.4.0-beta.10` (no tag yet); these are pre-release corrections to it, so they fold into the existing entry rather than adding new ones.

### Verification
- `cargo check -p start-core` clean; `cargo test -p start-core net::` (152, incl. the new M6 test) and `tunnel::` (14) green.
- `tsc` on ts-modules/start-core (L3), `check:i18n` (793 keys, L5), `check:ui` (M2), `check:tunnel` + `ng build start-tunnel` (L1/L2) all clean; prettier clean.

Commits use `--no-verify` (the slot's `pre-commit` hook `cd web/`s, a stale pre-reorg path).

### DIFF
--- a/shared-libs/crates/start-core/src/net/host/binding.rs
+++ b/shared-libs/crates/start-core/src/net/host/binding.rs
@@ -12,6 +12,7 @@ use crate::HostId;
 use crate::ServiceInterfaceId;
 use crate::context::{CliContext, RpcContext};
 use crate::db::prelude::Map;
+use crate::hostname::ServerHostname;
 use crate::net::forward::AvailablePorts;
 use crate::net::host::HostApiKind;
 use crate::net::service_interface::{
@@ -147,6 +148,29 @@ impl DerivedAddressInfo {
                 }
             })
             .collect()
+    }
+
+    /// Move the port of every `enabled`/`disabled` override from `old` to `new`.
+    /// A range keys all its overrides on `external_start_port`, so a service-
+    /// driven resize that moves the span must carry the overrides with it — else
+    /// a disabled WAN address silently re-enables under the recomputed defaults.
+    pub fn rekey_port(&mut self, old: u16, new: u16) {
+        if old == new {
+            return;
+        }
+        self.enabled = std::mem::take(&mut self.enabled)
+            .into_iter()
+            .map(|mut sa| {
+                if sa.port() == old {
+                    sa.set_port(new);
+                }
+                sa
+            })
+            .collect();
+        self.disabled = std::mem::take(&mut self.disabled)
+            .into_iter()
+            .map(|(h, p)| (h, if p == old { new } else { p }))
+            .collect();
     }
 }
 
@@ -685,7 +709,16 @@ pub async fn set_address_enabled<Kind: HostApiKind>(
                 .mutate(|b| {
                     let bind = b.get_mut(&internal_port).or_not_found(internal_port)?;
                     set_address_enabled_on(&mut bind.addresses, &address, enabled)
-                })
+                })?;
+            let hostname = ServerHostname::load(db.as_public().as_server_info())?;
+            let gateways = db
+                .as_public()
+                .as_server_info()
+                .as_network()
+                .as_gateways()
+                .de()?;
+            let ports = db.as_private().as_available_ports().de()?;
+            Kind::host_for(&inheritance, db)?.update_addresses(&hostname, &gateways, &ports)
         })
         .await
         .result?;
@@ -719,7 +752,16 @@ pub async fn set_range_address_enabled<Kind: HostApiKind>(
                 .mutate(|ranges| {
                     let range = ranges.get_mut(&internal_port).or_not_found(internal_port)?;
                     set_address_enabled_on(&mut range.addresses, &address, enabled)
-                })
+                })?;
+            let hostname = ServerHostname::load(db.as_public().as_server_info())?;
+            let gateways = db
+                .as_public()
+                .as_server_info()
+                .as_network()
+                .as_gateways()
+                .de()?;
+            let ports = db.as_private().as_available_ports().de()?;
+            Kind::host_for(&inheritance, db)?.update_addresses(&hostname, &gateways, &ports)
         })
         .await
         .result?;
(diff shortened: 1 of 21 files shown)


shared-libs/crates/start-core/src/net/host/binding.rs
```diff
@@ -148,6 +149,29 @@ impl DerivedAddressInfo {
             })
             .collect()
     }
+
+    /// Move the port of every `enabled`/`disabled` override from `old` to `new`.
+    /// A range keys all its overrides on `external_start_port`, so a service-
+    /// driven resize that moves the span must carry the overrides with it — else
+    /// a disabled WAN address silently re-enables under the recomputed defaults.
```
dr-bonez, line 156: regular bindings are keyed off of internal port. are ranges handled differently? why?

helix-nine: Both single-port bindings and ranges *are* keyed by internal port — `bindings` and `binding_ranges` are keyed identically. `rekey_port` touches a different key: the per-address enable/disable overrides inside the shared `DerivedAddressInfo`, which are keyed by the address's **external** port (the enabled `SocketAddr`, and `(hostname, port)` in `disabled`). That external port is load-bearing for single-port bindings — it distinguishes a domain's non-SSL address (`assigned_port`) from its SSL one (`assigned_ssl_port`).

Why only ranges need it: a single-port binding's external port is StartOS-**assigned** and preserved verbatim across rebinds — `NetInfo::update` keeps an existing `assigned_port` (`Some(port) => Some(port)`); it's only freed when the port is dropped entirely, never renumbered. So its overrides can't go stale. A range's external port is `external_start_port`, a **service-supplied** arg to `bindPortRange`, so a rebind can move the whole span. That move is the one case where a preserved override key stops matching the recomputed addresses (now at the new port), silently re-enabling a disabled address — which is what `rekey_port` prevents.

So ranges aren't keyed differently; they're just the only bindings whose external port can move under an operator override.

If you'd rather drop the rekey, the alternative is to not key the overrides on the mutable external port at all. That's clean for ranges (non-SSL → exactly one port, so it's redundant in the key), but it forks the shared `(hostname, port)` structure that single-port bindings need for the SSL/non-SSL split — so `rekey_port` is the less invasive fix. Happy to switch if you prefer.

shared-libs/ts-modules/start-core/lib/interfaces/RangeOrigin.ts
```diff
@@ -22,6 +24,12 @@ export class RangeOrigin {
    * @param serviceInterface - a builder from `sdk.createRangeInterface`
    */
   async export(serviceInterface: RangeInterfaceBuilder): Promise<void> {
+    if (this.exported) {
+      throw new Error(
+        `range at internal port ${this.internalStartPort} already exported an interface; a range exposes exactly one interface`,
```
dr-bonez, line 29: i18n

helix-nine: Confirmed there's no pre-existing i18n infrastructure for SDK error messages — the sibling `bindPortRange` validation throws in `Host.ts` (and every other `throw` in `@start9labs/start-core`) use plain `new Error(...)`, and nothing exports an `i18n` helper for this layer. So I'll leave it out of scope; the plain `Error` here matches those siblings.

dr-bonez (approved):

PR #3387, opened 2026-07-02
fix(net)!: address the OS UI as the reserved start-os package id; heal + migrate the empty host id

MattDHill:
## Summary

Takes the empty-host-id fix the rest of the way, per review: the StartOS admin UI is now **indistinguishable from a regular service interface at the SDK layer**, identified by the reserved package id **`start-os`** and host id **`admin`** — and the tor identity for the UI is migrated so nobody's `.onion` address is lost.

### 1. Server host gets a real id (`admin`) — original fix

Since #3366, `os_bindings()` persisted the `startos-ui` interface with `HostId::default()` — the **empty string** — which `Id`'s strict `Deserialize` rejects, so every boot of an affected dev build failed `validate_db` into the diagnostic UI. The server host now has the real id `admin` everywhere the sentinel was, and `HostId` no longer derives `Default`.

### 2. `start-os` is the OS's package id (reserved)

- `host_for(db, Some("start-os"), "admin")` resolves to the server host (`serverInfo.network.host` — storage is unchanged), so the package host RPC APIs and url plugins address the OS like any package.
- The `getHostInfo` / `getServicePortForward` / `getServiceInterface` / `listServiceInterfaces` effects resolve `start-os` the same way (watching `/public/serverInfo/network/host` and reading it as a one-host map), so `sdk.serviceInterface.getAll(effects, { packageId: 'start-os' })` just works.
- `start-os.startos` resolves in container DNS exactly like bare `startos`.
- **`PluginHostnameInfo.packageId` is now required** (SDK-breaking): url plugins export the OS UI's urls as `start-os`/`admin` instead of `packageId: null`. The web UI passes `start-os` wherever a package id is expected and drops its `|| null` special case.
- **Installing a package with the id `start-os` is rejected** (single chokepoint in `ServiceMap::install`, covering marketplace, sideload, and restore; i18n'd error).

### 3. `deserialize_lenient` → real migration (part of the unreleased 0.4.0-beta.10)

The lenient deserializer is gone. The beta.10 migration now also:

- **`up()`** rewrites the empty persisted `addressInfo.hostId` to `admin` on the raw db value, *before* `validate_db`'s typed round-trip. Caveat: a dev box whose db already records beta.10 won't re-run the migration — only boxes upgrading from ≤ beta.9 get the heal. See below for the manual fix on already-poisoned dev boxes.
- **`post_up()`** additionally re-points the tor package's persisted state at the new identity so the **StartOS UI's existing `.onion` address survives** instead of being rekeyed:
  - moves `hidden_services/STARTOS/startos-ui/` → `hidden_services/start-os/admin/` (key material intact; ownership handed back to the container user),
  - rewrites the torrc `# @service` annotation + `HiddenServiceDir` path,
  - updates a not-yet-imported `onion-migration.json` (`STARTOS` → `start-os`/`admin`).

  All steps idempotent, skipped when tor isn't installed. Notably this keeps even the *current* tor release working: its generic package path (`serviceInterface.getAll` → `exportUrl` with a real package id) handles `start-os` end-to-end once the OS resolves it — only its now-dead `STARTOS` special cases remain to be cleaned up.

Per review, other url-plugin registrations for the OS UI are allowed to break.

## Manual fix for dev boxes already poisoned on beta.10

A box whose db already records beta.10 skips the migration, and with the lenient deserializer gone the poisoned record now fails even earlier at boot (`start_net`'s typed read of the bindings map). Over SSH, using `db apply`'s offline mode (the optional path argument operates on the db file directly, no daemon needed):

```bash
systemctl stop startd
start-cli db apply '.public.serverInfo.network.host.bindings["80"].interfaces["startos-ui"].addressInfo.hostId = "admin"' /media/startos/data/main/embassy.db
systemctl start startd
```

The result is type-checked against the current `Database` model before being written, so a typo'd expression can't make things worse. Verified the expression against the real `apply_expr` (poisoned value in → healed value out).

If the box also has an OS UI `.onion` in tor, re-point tor's data at the new identity too (mirrors the migration's `post_up`):

```bash
V=/media/startos/data/package-data/volumes/tor/data
if [ -d "$V/tor/hidden_services/STARTOS/startos-ui" ]; then
  mkdir -p "$V/tor/hidden_services/start-os"
  chown --reference="$V/tor/hidden_services" "$V/tor/hidden_services/start-os"
  mv "$V/tor/hidden_services/STARTOS/startos-ui" "$V/tor/hidden_services/start-os/admin"
  rmdir "$V/tor/hidden_services/STARTOS" 2>/dev/null
fi
sed -i 's|# @service STARTOS startos-ui|# @service start-os admin|; s|/hidden_services/STARTOS/startos-ui/|/hidden_services/start-os/admin/|' "$V/tor/torrc" 2>/dev/null
```

## Follow-up

- `tor-startos`: drop the `STARTOS`/`startos-ui` special cases (`packageId ?? 'STARTOS'`, the `packageId === 'STARTOS'` branches) now that the OS presents as `start-os`/`admin`; bundle that release in the OS image.

## Testing

- `cargo test -p start-core --features=test` — 478 passed (includes new unit tests for the host-id heal and the torrc rewrite: package entries untouched, idempotent).
- `make ts-bindings` regenerated; `npm run check` (all Angular projects) and `tsc --noEmit` on `@start9labs/start-core` pass; SDK `make bundle` builds.
- Not covered by automation: a live upgrade of a box with a real tor onion for the UI (needs a VM with tor installed and a populated torrc).


dr-bonez:
Giving it a real id (i.e. `admin`) seems better

MattDHill:
Done in 03aac34c2 — the server host now has a real id, `HostId::admin()` (`"admin"`), used everywhere the empty sentinel was: the OS bind key, the Main-UI watch update, the server-branch `BindId` comparisons, `ForServer` host resolution, and the persisted `startos-ui` `AddressInfo`. Also dropped `HostId`'s `Default` derive so an accidentally-empty id can't be constructed again (compiles clean — nothing else relied on it).

Kept one narrow compat shim on the persisted field: the legacy empty string (written by the Jul 1–2 dev builds) still deserializes, as `admin`, and `validate_db`'s de → ser round-trip rewrites it on disk on the first boot — so already-bricked dev boxes self-heal with no manual fix. Manifest/API input stays strict.

helix-nine:
Reworked per review (thread discussion with @dr-bonez):

- **`deserialize_lenient` is gone** — replaced by a real migration in a new version **0.4.0-beta.11** (the bump is required for the heal to run on the already-poisoned beta.10 dev boxes).
- **`start-os` is now the OS's reserved package id**: `host_for(db, Some("start-os"), "admin")` resolves to the server host, the host-info/port-forward/service-interface effects and container DNS treat it like any package, `PluginHostnameInfo.packageId` is required (no more `null`-means-server), and installing a package with that id is rejected.
- **Tor identity migrated in `post_up()`**: `hidden_services/STARTOS/startos-ui` → `start-os/admin` (keys intact), torrc annotations rewritten, pending `onion-migration.json` updated — so the StartOS UI's existing .onion survives the rekey. Other url-plugin registrations for the OS UI may break, as agreed.

PR body rewritten with full details. Follow-up needed in `tor-startos` to drop its now-dead `STARTOS` special cases.

helix-nine:
Folded the migration into the unreleased **beta.10** per @dr-bonez — `v0_4_0_beta_11` is gone, the heal (`up()`) and the tor identity re-pointing (`post_up()`) now live in `v0_4_0_beta_10`, and all version strings are back at 0.4.0-beta.10 (c1e2fe52f).

One caveat worth stating: a dev box whose db already records beta.10 won't re-run the migration (stored version == current), so the Jul 1–2 poisoned dev boxes still need a manual fix or reflash — the migration covers everyone upgrading from ≤ beta.9.

projects/start-os/CHANGELOG.md
```diff
@@ -73,6 +87,12 @@ file tracks notable changes since the move to the monorepo.
 
 ### Fixed
 
+- **Dev builds bricked by an empty persisted host id (#3387).** Builds between
+  #3366 and #3387 persisted the server host's then-sentinel id (the empty
+  string) in the `startos-ui` interface's `addressInfo.hostId`, which strict
```
dr-bonez: we should also change the interface id to `admin-ui`

helix-nine: Done in 92a066cca — the OS UI's `ServiceInterfaceId` is now `admin-ui` (the `os_bindings` sync in net_controller and the web page's interface literal; changelog updated). No db migration needed: `os_bindings` replaces the binding's whole `interfaces` map on every boot, so the old `startos-ui` key disappears on the first boot of this build. The tor-side identity is unaffected — its keys are `(packageId, hostId)` = `start-os`/`admin`, which this doesn't touch.

shared-libs/crates/start-core/src/net/host/mod.rs
```diff
@@ -550,12 +550,23 @@ pub fn host_for<'a>(
     package_id: Option<&PackageId>,
```
dr-bonez: just make this non-optional

helix-nine:
Added a **Manual fix for dev boxes already poisoned on beta.10** section to the PR description: a one-shot command that appends a CBOR-encoded RFC 6902 patch (`hostId → "admin"`) to `embassy.db`'s patch log — applied and compacted on the next `startd` start — plus an optional block that re-points tor's hidden-service data (`STARTOS/startos-ui` → `start-os/admin`) for boxes with an OS UI onion, mirroring the migration's `post_up`. The db blob was verified against the real `PatchDb::open` (poisoned file + appended blob → healed value on open).

shared-libs/crates/start-core/src/service/effects/net/bind.rs
```diff
@@ -123,19 +123,26 @@ pub async fn get_service_port_forward(
...
+    } else {
+        db.as_public()
+            .as_package_data()
+            .as_idx(&package_id)
+            .or_not_found(&package_id)?
+            .as_hosts()
+            .as_idx(&host_id)
+            .or_not_found(&host_id)?
+    };
+    Ok(host
```
dr-bonez: can you not use host_for?

dr-bonez (dismissed):

shared-libs/crates/start-core/src/net/host/mod.rs
helix-nine: Done in 2a025109e — `host_for` now takes `&PackageId`; the internal callers that meant "the server" (`ForServer`, the OS `NetService` bind paths) pass `PackageId::start_os()` explicitly.

shared-libs/crates/start-core/src/service/effects/net/bind.rs
helix-nine: Done in 2a025109e — `get_service_port_forward` resolves the host through `host_for` on the peeked model instead of the inlined branch. One behavioral nit: `host_for`'s package arm upserts a missing host, so on a peeked copy a nonexistent host now reads as empty and surfaces as "binding for port N not found" rather than "host not found" — seemed acceptable for a read effect.

helix-nine:
Also swapped the manual-fix section in the description to `start-cli db apply <expr> /media/startos/data/main/embassy.db` (offline mode via the optional path argument) instead of the raw CBOR append — cleaner, and the result is type-checked against the `Database` model before being written. Verified the jq expression against the real `apply_expr`.

dr-bonez (approved):

helix-nine:
tor-startos follow-up is up: Start9Labs/tor-startos#13 — drops all the `STARTOS`/`startos-ui` special cases now that the OS presents the admin UI as `start-os`/`admin`, and hardens the exportUrls cleanup to never delete key material on a failed lookup. Requires beta.10 (promote together; bundle it in the beta.10 image).

PR #3388, opened 2026-07-02
feat(tunnel): IPv6 support — prefix delegation + PCP pinholes (WIP)

helix-nine:
**Status: ready for review.** This is PR1 of a staged rollout (per @dr-bonez: `1a+2b`, `3b`, `4a`). It lands the IPv6 **addressing foundation** — allocation, WireGuard config generation, proxy-NDP dataplane, DNS, CLI/API, UI, and docs. PCP pinholes are PR2. Commits: `aa0603a` (foundation), `6fea6f3` (proxy-NDP), `c060ce1` (docs/changelog/version), `6201a62` (UI), `d21dace` (self-review fixes).

## Decision (research-backed)

The addressing model hinges on what VPS providers actually delegate. I surveyed the providers the tunnel docs recommend:

| Provider | Default IPv6 |
|---|---|
| Hetzner Cloud | **/64** (routed, gw `fe80::1`) |
| Vultr | **/64** |
| BuyVM | **/64** (optional routed /48) |
| DigitalOcean | **/124** (16 addrs) |
| Linode/Akamai | /128 default; **/64 or /56** on request |
| OVH VPS | /128 (shared /64); dedicated → /64, High-Grade → /56 |

So the common case is **a single /64, sometimes less**. Per the rule "if it's just a /64 → `1a`+`2b`": the tunnel takes a **statically configured routed prefix** (`1a`) and assigns each client a **single /128** out of it (`2b`) — while **scaling up for free**: a prefix shorter than /64 (Linode /56, dedicated, BuyVM /48) delegates a whole /64 per client (`2a`, real PD, so a StartOS box can sub-address its containers).

## What's in this PR

- **`tunnel/wg6.rs`** — the allocation module (`server_addr`, `client_v6`), unit-tested across shared-/64, short-prefix (per-client /64), and narrow (/124) cases. For a shared /64 the client's /128 host bits are its tunnel IPv4 (stable, allocation-free); `::1` is reserved for the tunnel.
- **`WgServer.ipv6`** prefix field + WireGuard config generation: the server interface carries `<prefix>::1`, each peer's `AllowedIPs` routes its delegated /64 or /128 back over wg, and client configs get their v6 `Address`, v6 `DNS` (= the tunnel), and `::/0` for v6.
  - *Routing note:* IPv6 is **full-tunnel** (`AllowedIPs = …, ::/0`). Replies sourced from a VPS-delegated GUA must return through the tunnel, and a plain WireGuard peer can't source-route otherwise. IPv4 stays split.
- **`set-ipv6` API + CLI** to configure/clear the prefix (+ i18n across 5 locales, TS bindings export).
- **forward chain** now accepts client-originated IPv6 (`nft_rule_v6`); inbound-to-client is deliberately a pinhole (see PR2).

Tested: `cargo test -p start-core --features test tunnel::wg` → 9 passing (allocation math + rendered configs).

## PR1 checklist
- [x] Dataplane: **proxy-NDP** for on-link /64s (Hetzner/Vultr) — the VM answers ND for client /128s on the WAN; routed prefixes / per-client /64s need none (`6fea6f3`, `resync_v6`). Guarded by a `v6_lock` (`d21dace`).
- [x] **AAAA DNS** — already served by the existing RFC 2136 injection path (`parse_rdata` is record-type-generic), so a device can publish an `AAAA` for its GUA today. Documented; no new code needed.
- [x] Angular UI: prefix config + Disable on the settings page (`6201a62`). Each device's v6 appears in its **generated config** (`device show-config`); I left it out of the devices *list* rather than duplicate the allocation math in TS — happy to expose it via the API for the list if you'd prefer a column.
- [x] Docs (`ipv6.md` + `cli-reference.md` + generated manpage), `CHANGELOG.md`, version `1.1.0` → `1.2.0` (`c060ce1`).
- [x] Self-review: adversarial review of the full diff → **7 findings fixed (2 major)** in `d21dace` (see comment).
- [ ] End-to-end verification on a real VPS with a routed prefix — needs real global IPv6, so it can't run on a local VM. Worth a manual pass before release.

## PR2 (follow-up, `3b`)
- [ ] Extend the tunnel's PCP server for IPv6 GUA **pinholes** (`ip6` nft accept, no NAT) — what StartOS's PCP client already requests.
- [ ] Manual pinhole management in the UI/CLI.

Feedback welcome on the allocation scheme and the full-tunnel-v6 `AllowedIPs` choice.


helix-nine:
**PR1 is now feature-complete** (addressing foundation). Four commits on the branch:

1. `aa0603a` — IPv6 address-delegation foundation (wg6 allocation module + WG config generation) — unit-tested.
2. `6fea6f3` — proxy-NDP dataplane for on-link /64s (Hetzner/Vultr).
3. `c060ce1` — docs (ipv6.md + CLI reference + manpage), CHANGELOG, version 1.1.0 → 1.2.0.
4. `6201a62` — IPv6 settings UI card (api service live+mock, mock db seed, regenerated TS bindings).

**Verification (all pass):** `cargo test tunnel::wg` (9), `cargo check -p start-core`, `check:tunnel`, `build:tunnel` (full Angular template type-check), prettier.

**Not automated:** end-to-end reachability on a real VPS with a routed IPv6 prefix — needs real global IPv6, which a local libvirt VM can't provide. Worth a manual pass before release.

Deferred to **PR2** (per the `3b` decision): extend the tunnel's PCP server for IPv6 GUA pinholes (`ip6` nft accept, no NAT) + a manual pinhole path.

Still a draft — happy to adjust the allocation scheme or the full-tunnel-v6 `AllowedIPs` choice on review before I start PR2.

helix-nine:
Before marking this ready I ran an adversarial self-review over the whole PR1 diff (6 dimensions — allocation math, proxy-NDP dataplane, WG config generation, API/CLI, UI, docs — each finding verified against the code). It surfaced 7 real issues, all now fixed in `d21dace`:

**Major**
- `resync_v6` did the same non-atomic read→diff→apply→overwrite as `resync_egress` but without a lock — two concurrent config-change RPCs could leave the tracked proxy-NDP map out of sync with the kernel neighbor table (leaked entries). Added a dedicated `v6_lock` mirroring `egress_lock`.
- The client config advertised an IPv6 DNS server (`<prefix>::1`) that nothing binds — the DNS proxy is IPv4-only. Dropped the v6 DNS entry (AAAA resolves fine over the v4 resolver); v6 DNS can come with the proxy binding v6 later.

**Minor / nit**
- `set-ipv6` now also rejects link-local prefixes (a fat-fingered `fe80::/64` would have persisted and broken all client v6).
- UI validator now requires an explicit `/prefix` in `[0,128]` (`IpNet.parse` doesn't enforce it).
- Docs: corrected the CLI clearing syntax (omit `--prefix`, not `--prefix null`); added a note that inbound IPv6 hosting isn't supported yet (PR2); dropped the v6-DNS claim; moved the changelog entry under a `## [1.2.0]` heading.

Re-verified after the fixes: `cargo test tunnel::wg` (9 pass), `cargo check`, `check:tunnel`, full `build:tunnel`, prettier — all green.

helix-nine:
Implemented the **per-subnet IPv6 redesign** (handoff §6) — `14a07a1`. Global `WgServer.ipv6` → per-subnet `WgSubnetConfig.ipv6`; allocation collapsed to one `/128` per host (`host_v6 = prefix.network() | tunnel_ipv4`, server + clients alike); `subnet <net> set-ipv6` replaces the global command (carrying the egress validation, keyed on the subnet); UI moved from the settings card to the subnet dialog + a computed IPv6 column in the subnet/device tables.

Ran an adversarial review over the diff before pushing; it caught 4 real issues, all fixed in `fd650b2`:

- **Library bug (root-caused, not worked around).** `@start9labs/start-core`'s `IpAddress` rendered IPv6 by joining raw *decimal octets* (`"32:1:13:184:…"` not `"2001:db8::…"`), and `fromOctets`'s 16-octet path **spun forever** on a no-op `unshift()` whenever the 9th octet was 0 — so `zero()`/`fromOctets`/`.address` on any computed v6 hung the browser or produced garbage. Fixed both the `.address` getter and `fromOctets` with one correct `renderIpv6` (8 hex groups, longest zero-run → `::`). Verified against RFC 5952 edge cases and across **all** web projects (`npm run check`) — no consumer regressed.
- **Out-of-prefix addresses.** `host_v6` only stays in-prefix for prefixes ≤ /96; a /124 escaped the block silently. `set_subnet_ipv6` now rejects `> /96`, the web validator mirrors it, and a wg6 boundary test covers /48–/96.
- **Stale comments** from the old /64-delegation model removed.

Verified: `cargo test` (backend), `npm run check` (all projects), `check:tunnel`, full `build:tunnel`, prettier, and a runtime check of the render fix.

dr-bonez (approved):

PR #3389, opened 2026-07-03
fix: resolve OS root device from its live mount, not the overlay / fstab entry

helix-nine:
## Problem

`failed to resolve fstab source overlay: blkid: -t needs NAME=value pair`

At runtime StartOS's `/` is an **overlayfs the initramfs stacks over the real OS partition**, so the `/etc/fstab` `/` entry never names the backing block device — its source is the pseudo-fs token `overlay`, which `OsPartitionInfo::from_fstab` fed to `blkid -t overlay` → `-t needs NAME=value pair`, leaving `root` empty.

An empty `root` isn't benign: `disk::util::list` uses `OsPartitionInfo::contains` to mark the OS partition `internal`, so an empty root makes the OS disk show up as an available **data/backup disk**.

## Why not the kernel `root=` / "whatever we booted from"

The OS root device can't be taken from what the system booted off, either: during **os_install** we're booted from the **installer USB**, which is not the OS root this struct describes. So keying off the cmdline `root=` (or any boot-medium signal) is the wrong premise.

## Fix

The initramfs bind-mounts the real OS partition at **`/media/startos/root`** on every installed-system boot — present *only* on a running installed system, and absent in the live installer (which boots `boot=live`). So it names the OS root exactly when there is one:

- Resolve `root` from `/media/startos/root` via `get_mount_source` (reads the backing device from `/proc/mounts`); its absence in the installer correctly yields no OS root.
- Read only the real `/boot*` block-device mounts from fstab; skip `/` entirely.
- Guard `resolve_fstab_source` so a non-`TAG=value`, non-path source never reaches `blkid -t`.

Fresh install is unaffected — `os_install` builds `OsPartitionInfo` in `gpt.rs`/`mbr.rs`, not via `from_fstab`.

## Testing

- `cargo check -p start-core` passes.
- Verified the runtime overlay layout on a `0.4.0-beta.10` VM (offline squashfs/btrfs inspection + serial boot log), and confirmed on btrfs that a `--bind` mount's `/proc/mounts` source is the bare backing device (what `get_mount_source` reads).

### DIFF
--- a/shared-libs/crates/start-core/src/disk/mod.rs
+++ b/shared-libs/crates/start-core/src/disk/mod.rs
@@ -8,7 +8,7 @@ use serde::{Deserialize, Serialize};
 use tokio::process::Command;
 
 use crate::context::{CliContext, RpcContext};
-use crate::disk::util::DiskInfo;
+use crate::disk::util::{DiskInfo, get_mount_source};
 use crate::prelude::*;
 use crate::util::Invoke;
 use crate::util::serde::{HandlerExtSerde, WithIoFormat, display_serializable};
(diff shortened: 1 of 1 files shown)


dr-bonez (approved):

helix-nine:
Thanks for the review. Merged version resolves `root` solely from the `/media/startos/root` mount (no cmdline fallback), so the installer USB never leaks in and rescue/dev boots without the StartOS initramfs just get an empty root — taking that as the intended behavior since it went in as-is.

PR #3352, opened 2026-06-25
refactor: reorganize start-os into all-products monorepo

helix-nine:
Converts `start-os` into the **monorepo for all Start9 products**, per discussion with @dr-bonez. Each product is a thin wrapper under `projects/`; the bulk of the code lives in top-level shared libs under `shared-libs/`.

## Final layout

```
start-os/                          # repo root — monorepo for all Start9 products
├── projects/
│   ├── start-os/                  # startbox + start-container bins, web/{ui,setup-wizard},
│   │                              #   container-runtime/, OS image build, backup-fs, Debian packaging
│   ├── start-cli/                 # start-cli bin (thin) + Debian packaging
│   ├── start-registry/            # registrybox bin (thin)
│   ├── start-tunnel/              # tunnelbox bin + web/ (thin)
│   ├── start-sdk/                 # @start9labs/start-sdk (flattened; imports + bundles @start9labs/start-core) + packaging mdbook
│   ├── brochure-marketplace/      # public marketplace / landing site
│   └── start-docs/                # documentation website
├── shared-libs/
│   ├── crates/
│   │   ├── start-core/            # the ENTIRE Rust backend lib (package start-core, lib start_core)
│   │   ├── patch-db/              # de-submoduled → first-party
│   │   └── exver, imbl-value, jsonpath, pi-beep, rpc-toolkit, yasi   # Start9 crates → first-party
│   └── ts-modules/                # Angular workspace (rooted at repo root) + @start9labs/{shared,marketplace} + start-core (@start9labs/start-core)
├── build/  Makefile  debian/  apt/  rfcs/        # shared build infra + protocol drafts
└── .github/                       # per-product path-gated workflows + migrated service-package CI
```

Every product bin (`startbox`/`start-container`, `start-cli`, `registrybox`, `tunnelbox`) is a thin entry point over `start-core`. The Angular workspace stays single, rooted at the repo root, with each app's `angular.json` root pointing into its product dir.

## What changed

**Structure**
- Products nested under `projects/`; shared libs under `shared-libs/` (`crates/` + `web/`).
- `docs/` → `projects/start-docs/` (migrated as a plain copy); `brochure/` → `projects/brochure-marketplace/` (incl. the Angular project rename); the draft PCP-hostname RFC moved to a top-level `rfcs/`.
- Project-specific `assets/`/`debian/`/`scripts/` moved into their products; genuinely shared build infra stays at the top level.

**Rust**
[...]

PR #3391, opened 2026-07-03
fix: repair incremental build-graph staleness from the monorepo reorg

helix-nine:
## Why

`container-runtime` boot was throwing **`cannot find module zod-deep-partial`** on an **incremental (non-clean) build**. Root cause is a class of build-graph bug introduced by the #3352 monorepo reorg: a rule's make-visible freshness marker doesn't reflect what its recipe actually produces, so `make` skips a rebuild it should run and ships a stale/incomplete artifact. `zod-deep-partial` has been a `@start9labs/start-core` dep since Feb — a from-scratch build is fine, only incremental ones break — which is why @drbonez hit it and CI didn't.

I fixed the reported case, then did a deeper audit (5-dimension parallel workflow, every finding adversarially re-verified against the tree) for other reorg mistakes of the same class. This PR is all of them.

## Fixes

**Build-graph staleness (the reported bug + siblings)**
- **container-runtime** (`projects/start-os/build.mk`): `npm run build` runs `rm -rf dist`, which wipes the *separately-vendored* `dist/node_modules`, but the vendoring rule (`install-dist-deps.sh`) didn't depend on `dist/index.js`. So a src-only edit re-fired the JS build (wiping node_modules) without re-vendoring — shipping a runtime with **no** `node_modules` (and a `-j` build could race the same way). Fixed by making the vendoring rule depend on `dist/index.js`.
- **start-core / start-sdk dist gates** (`shared-libs/ts-modules/build.mk`, `projects/start-sdk/build.mk`): the bundled `dist/node_modules/.package-lock.json` is now a grouped `&:` co-target, so a missing/incomplete vendored `node_modules` forces a rebuild instead of being masked by a fresh `dist/package.json` mtime.
- **start-core / start-sdk dist recipes** (`shared-libs/ts-modules/start-core/Makefile`, `projects/start-sdk/Makefile`): `rm -rf dist` first → idempotent clean rebuild. Also fixes the SDK `dist` recipe dying with `rsync … Error 23` (rsyncing a symlink over a populated real dir) on any non-clean rebuild, which otherwise cascades into phantom container-runtime tsc errors.

**Other reorg mistakes caught by the audit**
- **`CORE_SRC`** (`build/common.mk`): globbed only `start-core` + `patch-db`, but after the reorg `start-core` path-depends on the vendored `exver` / `imbl-value` / `jsonpath` / `rpc-toolkit` / `yasi` crates. Editing any of them invalidated **no** Rust bin target, so `make startos`/`cli`/`registry`/`tunnel` shipped a stale binary (only bites incremental / cached-`target/` builds). `start-wrt` had already worked around a slice of this locally. Added all five as prereqs.
- **Dead edge**: `$(call ls-files, start-sdk)` resolved to **zero files** (should be `projects/start-sdk`) — SDK source edits never triggered `test-sdk` or the SDK dist rebuild.
- **CI `paths:` mirror drift** (`start-tunnel`, `start-wrt`, `startos-iso`): added `tsconfig.lib.json` (a shared Angular build input listed in `WEB_SHARED_SRC`) and, for `start-tunnel`, `apt/**` (its `.deb` bakes in `apt/start9.{list,gpg}`) so changes to those inputs actually trigger CI. Additive only.

## Verification
- Simulated the shipped container-runtime image (`rsync --copy-unsafe-links dist/` into an isolated tree, `node index.js`): now boots past module loading (`zod-deep-partial` resolves, 3 bundled copies present) and fails only on the expected container-only `/media/startos/rpc` mkdir.
- Confirmed the co-target self-heals a deleted bundle even with a fresh `dist/package.json`; the SDK now rebuilds twice in a row (was Error 23); after a src edit the container-runtime rootfs plan now re-runs `install-dist-deps.sh`.
- `CORE_SRC` now covers all five crates (file counts match); full make DAG parses cleanly for startbox/cli/registry/tunnel with no cycles.

## Reported, not fixed (need your call)
- **`startos-upload-ota`** (`projects/start-os/build.mk:145`) invokes `projects/start-os/build/upload-ota.sh`, which #3117 deleted — the recipe was left dangling. Only hard-fails if run manually. Restore the script or drop the target?
- **Root `package-lock.json`**: 5 `"extraneous": true` records left by the reorg (two with pre-reorg paths). Cosmetic — `npm ci` ignores them — so I left the lock alone rather than ship a large regeneration diff. Easy to regenerate if you'd prefer it clean.


helix-nine:
Addressed both deferred items (commit 8bd476645):

1. **Dropped** the dead `startos-upload-ota` target.
2. **Pruned** the 5 stale `"extraneous"` records from the root `package-lock.json`.

⚠️ Heads-up on (2): a full `npm install` regen under **npm 11.16** does *not* produce a clean diff — it wrongly hoists the patch-db client's pinned `uuid@8.3.2` to the root `uuid@11.1.1` (the client's exact pin ends up unsatisfied; `require('uuid')` from the client resolves to v11). Both `npm install` and `--package-lock-only` did this. So I removed only the 5 dead extraneous blocks directly (verified: JSON valid, 0 extraneous left, nested `uuid@8.3.2` preserved, no dangling refs, pure -130-line deletion, every real entry byte-identical). Worth knowing before anyone regenerates this lock with a modern npm.

dr-bonez (approved):

PR #3392, opened 2026-07-03
refactor: remove serviceInterface get/getOwn helpers; fill addresses on host.get

helix-nine:
## What

Removes the SDK-facing `serviceInterface.get`/`getOwn` helper family (now that `getHost` is the prioritized path) and moves the filled-address ergonomics onto what `host.get`/`getOwn` returns. Leaves the `get_service_interface` **RPC** in place for backwards compatibility.

Context: PR #3366 already swapped the `sdk.serviceInterface.{get,getOwn,getAll,getAllOwn}` accessors for `sdk.host.{get,getOwn}`, but left the underlying util helpers dangling (still reachable via `sdk.utils.getServiceInterface*`, flagged in the index as "Not being used"). This finishes that removal and preserves the `filter`/`format` ergonomics on the host object.

## Changes

**Removed (dead SDK helpers)**
- `getServiceInterface` / `getOwnServiceInterface` / `getServiceInterfaces` / `getOwnServiceInterfaces` and the `GetServiceInterface` / `GetServiceInterfaces` reader classes, plus `ServiceInterfaceFilled`.
- Deleted `util/getServiceInterfaces.ts` (entirely dead).
- Renamed `util/getServiceInterface.ts` → `util/filledAddress.ts` to match its remaining purpose (the filled-address machinery: `filledAddress`, `addressHostToUrl`, `mdnsResolvable`, `filterNonLocal`, filter/format types).

**Filled addresses now live on `host.get`/`getOwn`** (drbonez's follow-up ask)
- `GetHostInfo` maps the raw `Host` through a new `fillHost()`, so `sdk.host.get(...)` / `getOwn(...)` return a `FilledHost`: each single-port binding's interface `addressInfo` comes back already filled — `.filter(...)`, `.format(...)`, `.nonLocal`, `.public`, `.toUrl` are on it directly, no separate `utils.filledAddress(host, …)` call needed.
- Port-range interfaces are left untouched — a `RangeServiceInterface` carries no per-address `AddressInfo` (the old helpers didn't fill them either).

**Kept for backwards compatibility**
- The `get_service_interface` RPC / `effects.getServiceInterface` binding, its container-runtime adapters, and the Rust effect are all untouched — packages built against older SDKs keep working.

**Docs / changelog**
- Updated the packaging mdbook (`main.md`, `dependencies.md`) to use `iface.addressInfo.format(...)` directly.
[...]

PR #3393, opened 2026-07-03
fix: unbreak CI — drop make -t-incompatible dist co-target + repair getServiceInterface call site

helix-nine:
Follow-up to #3391. @dr-bonez still hit `Cannot find module 'zod-deep-partial'` after a (non-clean) build with #3391 merged.

## Why #3391 wasn't enough

#3391's grouped co-target catches a **missing** bundled `node_modules`, and the `rm -rf dist` recipes keep a *triggered* rebuild clean. But neither heals a dist that is **present-but-stale**: a `dist/` carried across a dependency change (the reorg, a branch switch, an interrupted build) keeps a fresh `dist/package.json` mtime while its bundled `node_modules` predates a newly-added dep like `zod-deep-partial`. mtime gating fundamentally can't see stale *content* behind a fresh marker, so `make` declares the dist up-to-date, skips the rebuild, and the container-runtime ships without the dep — the same boot error, even after building.

I reproduced this exactly: seed `shared-libs/ts-modules/start-core/dist` + `projects/start-sdk/dist` to that state (zod removed, fresh markers), run the container-runtime image build, and the image boots to `Cannot find module 'zod-deep-partial'`.

## Fix — a content gate

Each dist now copies the lock it was built from into `dist/package-lock.json`, and the build.mk gate adds a `FORCE` prerequisite when that copy no longer matches the current source lock (portable `cmp -s`):

```make
CORE_DIST_STALE := $(shell cmp -s .../start-core/package-lock.json .../start-core/dist/package-lock.json || echo FORCE)
… dist/package.json … &: $(CORE_DIST_STALE) $(call ls-files, …) …
```

- A dist whose bundled `node_modules` was built from a different lock → mismatch → rebuild.
- A dist built **before** this change (no `dist/package-lock.json`) → `cmp` fails → rebuild. **This is the case that heals existing trees like @dr-bonez's** — no manual clean needed.
- An in-sync dist → `cmp` matches → no `FORCE`, normal mtime gating, no wasted rebuild.

Applied to both `start-core` and `start-sdk`. The start-core heal cascades into the SDK's vendored copy through the existing `start-core/dist/package.json` dep edge. `dist/package-lock.json` is auto-excluded from `npm publish` (verified via `npm pack --dry-run`), so the published SDK is unaffected.

## Verification
- Stale-state repro (both dists, zod removed, no lock copy, fresh markers) → make now force-rebuilds (`make bundle` + start-core + `install-dist-deps` all fire) → assembled image has zod (3 copies) and boots past module loading to the expected container-only `/media/startos/rpc`.
- In-sync tree → second `make` is a clean no-op (no spurious rebuilds).
- Full make DAG parses cleanly for startbox/cli/registry/tunnel/web.

@dr-bonez — with this you shouldn't need to clean; the next build will detect the stale dist and rebuild it. (If you want to confirm the mechanism first: `cmp -s shared-libs/ts-modules/start-core/package-lock.json shared-libs/ts-modules/start-core/dist/package-lock.json; echo $?` — non-zero means it'll force-rebuild.)

### DIFF
--- a/projects/start-os/container-runtime/src/Adapters/Systems/SystemForEmbassy/index.ts
+++ b/projects/start-os/container-runtime/src/Adapters/Systems/SystemForEmbassy/index.ts
@@ -1331,19 +1331,26 @@ async function updateConfig(
           mutConfigValue[key] = ""
           return
         }
-        const filled = await utils
-          .getServiceInterface(effects, {
+        const filled = await (async () => {
+          const serviceInterface = await effects.getServiceInterface({
             packageId: specValue["package-id"],
-            id: serviceInterfaceId,
+            serviceInterfaceId,
           })
-          .once()
-          .catch((x) => {
-            console.error(
-              "Could not get the service interface",
-              utils.asError(x),
-            )
-            return null
+          if (!serviceInterface) return null
+          const host = await effects.getHostInfo({
+            packageId: specValue["package-id"],
+            hostId: serviceInterface.addressInfo.hostId,
           })
+          return {
+            ...serviceInterface,
+            addressInfo: host
+              ? utils.filledAddress(host, serviceInterface.addressInfo)
+              : null,
+          }
+        })().catch((x) => {
+          console.error("Could not get the service interface", utils.asError(x))
+          return null
+        })
         const catchFn = <X>(fn: () => X) => {
           try {
             return fn()
--- a/projects/start-sdk/build.mk
+++ b/projects/start-sdk/build.mk
@@ -2,10 +2,9 @@ test-sdk: $(call ls-files, projects/start-sdk) shared-libs/ts-modules/start-core
 	$(MAKE) -C shared-libs/ts-modules/start-core test
 	cd projects/start-sdk && make test
 
-# Co-target the bundled dist/node_modules stamp (see shared-libs/ts-modules/build.mk).
-projects/start-sdk/dist/package.json projects/start-sdk/dist/node_modules/.package-lock.json &: $(call ls-files, projects/start-sdk) shared-libs/ts-modules/start-core/dist/package.json
+projects/start-sdk/dist/package.json: $(call ls-files, projects/start-sdk) shared-libs/ts-modules/start-core/dist/package.json
 	(cd projects/start-sdk && make bundle)
-	touch projects/start-sdk/dist/package.json projects/start-sdk/dist/node_modules/.package-lock.json
+	touch projects/start-sdk/dist/package.json
 
 .PHONY: clean-sdk
 clean-sdk:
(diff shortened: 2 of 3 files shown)


helix-nine:
Rebased onto current master and added a commit fixing the CI failure (it was **not** from the dist changes):

**Root cause:** #3392 (`refactor: remove serviceInterface get/getOwn helpers`) removed `util.getServiceInterface` from start-core but left the legacy `SystemForEmbassy/index.ts` still calling `utils.getServiceInterface(effects, …).once()`, so the container-runtime `tsc` build breaks on master (`TS2339: Property 'getServiceInterface' does not exist …`). My branch predated #3392, which is why local passed until the merge surfaced it.

**Fix:** reconstruct the same filled-interface lookup inline from the **retained** `effects.getServiceInterface` + `effects.getHostInfo` + `utils.filledAddress` — exactly what the removed `makeInterfaceFilled` did. Verified locally: `tsc --noEmit` clean, prettier clean, jest 12/12 + 6 snapshots pass.

Since this is a master-wide breakage (every PR's CI is red against current master), happy to split it into its own fast PR if you'd rather land the master fix ahead of the dist discussion — just say the word.

helix-nine:
Reworked this — you were right that something was actually wrong, and it was my own change.

**What was breaking Build Image:** the ISO job downloads the compiled artifacts and runs `make -t compiled-<arch>.tar` ("Prevent rebuild of compiled artifacts") to mark them current, then `make startos-iso`. My dist gates fight that:

- #3391's grouped co-target `dist/package.json dist/node_modules/.package-lock.json &: …` — `make -t` reaching the group via `dist/package.json` touches only that target, **not** the grouped sibling stamp (verified with a minimal Makefile). So the stamp stays missing → `make startos-iso` re-runs `make -C start-core dist` → cascades into a web-UI rebuild that reads an **empty `config.json`** → `Unexpected end of file in JSON`. This is the flaky failure master hits (co-target sometimes triggers, sometimes not).
- My #3393 content-gate's phony sentinel made it **deterministic**: `make -t` can never mark a phony current, so the rebuild fires every time.

Force/co-target rebuild triggers are fundamentally incompatible with a `make -t`-based "trust the artifacts" flow. So I dropped both: the dist gates are back to the plain single-target rule, which `make -t` suppresses cleanly (verified locally — `make -t <dist> && make <dist>` is now a no-op instead of rebuilding).

**This PR now:** (1) reverts that co-target, and (2) fixes the `getServiceInterface` call site (#3392 breakage). Kept: the container-runtime `rm -rf dist`→re-vendor ordering fix (the actual root cause of the original `zod-deep-partial` error) and the `projects/start-sdk` ls-files dead-edge fix — both `make -t`-safe.

Net for the original stale-`dist` problem: the container-runtime ordering fix covers the real mechanism; a `dist` that's *already* stale-with-fresh-mtime just needs a one-time `rm -rf …/start-core/dist projects/start-sdk/dist`. Trying to auto-heal that in-make is what kept breaking things.

dr-bonez (approved):

PR #3384, opened 2026-07-01
chore: rework manage-release.sh into a top-level, project-parameterized release tool

helix-nine:
Moves `projects/start-os/scripts/manage-release.sh` back out to `scripts/manage-release.sh` and reworks it from a start-os-only helper into a monorepo-wide release driver:

```
./scripts/manage-release.sh <subcommand> <project>
```

`<project>` ∈ `start-os | start-cli | start-tunnel | start-registry | start-sdk`. The version is read from that product's canonical manifest (`Cargo.toml`, or `package.json` for the SDK), and the git tag / GitHub release is `<project>_v<version>` — the convention already present in the tag history (`start-cli_v0.4.0-beta.9`, …).

### Subcommands
| cmd | what it does |
|---|---|
| `pre-check` | changelog documents this version **and** version isn't already tagged/released (registry for os, npm for sdk, GH release for cli/deb) |
| `pull-gha` | download build artifacts from a GitHub Actions run (`RUN_ID`) |
| `pull` | download released assets from their official location (registry / apt repo / GH release / npm) |
| `tag` | create + push the `<project>_v<version>` tag |
| `create-gh-release` | create/update the GitHub release with generated notes |
| `push` | upload to the applicable destination (S3 for os, GH release for cli, apt + GH release for deb, `npm publish` for sdk) |
| `index` | register + index the version in the registry (os only) |
| `sign` / `cosign` | GPG signing as before, scoped to the project's release |
| `notes` | print the release notes to stdout |
| `release` | run the full applicable pipeline for the project |

### Per-project "kind"
- **os** — iso/squashfs → S3 (`startos-images`) + registry OS index; GH release holds notes (ISO links + checksums)
- **cli** — per-triple binaries → GitHub release
- **deb** (start-tunnel, start-registry) — per-arch `.deb` → apt repo (`debian/publish.sh`) + GitHub release
- **npm** (start-sdk) — `release` = pre-check → tag → `npm publish`. The changelog already ships in the published dist (the SDK Makefile copies `CHANGELOG.md` into `dist/`), so nothing extra was needed there.

### Notes / decisions
- Scope and the per-project tag/release model were confirmed with @dr-bonez (all five projects; per-project `<project>_v<version>` tags & releases).
- I added `tag` to the sdk `release` pipeline (kept minimal otherwise) for consistency with the per-project tag model — easy to drop if unwanted.
[...]

PR #3394, opened 2026-07-03
chore(start-wrt): fold StartWRT releases into the monorepo release tool

Dominion5254:
Replace the standalone projects/start-wrt/scripts/manage-release.sh with a `wrt` project kind in the top-level scripts/manage-release.sh (#3384), so StartWRT releases follow the monorepo conventions: version read from backend/ctrl/Cargo.toml, tag / GitHub release start-wrt_v<version> (replacing the standalone startwrt/v<version> scheme), and the shared pre-check / FORCE / sign / cosign plumbing.

- manage-release.sh: new `wrt` kind. pre-check validates the changelog, tag, GitHub release, StartWRT registry, and developer key; pull-gha fetches the startwrt-openwrt-image artifact; pull does a verified `registry os asset get`; index registers the version and indexes both images at their S3/CDN URLs (keeping the .img.gz -> .squashfs hardlink trick; no beta->prod promotion — StartWRT has a single registry). `release start-wrt` runs pre-check -> pull-gha -> tag -> create-gh-release -> index -> sign.
- start-wrt.yaml deploy: now only uploads the images to S3. Tagging, cutting the GitHub release, and registry register/index move to the local, developer-key-gated manage-release.sh run (drops the action-gh-release step and the contents:write permission).
- Delete the standalone script and its bundled start9.key.asc — the same org key is already committed at apt/start9.gpg, and `sign` still bundles an export into signatures.tar.gz.
- Docs: CONTRIBUTING.md gains the release runbook; installing.md is re-pointed at the start-wrt_v tag scheme; CHANGELOG.md updated.


projects/start-wrt/docs/src/installing.md
```diff
@@ -10,7 +10,7 @@ StartWRT comes pre-installed on Start9 routers. If you need to reinstall or flas
 
 ## Download the Image
 
-1. Download the latest StartWRT firmware image from the [Start9 releases page](https://github.com/Start9Labs/start-technologies/releases?q=startwrt&expanded=true) (StartWRT releases are the ones tagged `startwrt/v…`). For a fresh install, download the **sdcard** image — it is named `startwrt-<version>-<git hash>_spacemit-k1-sdcard.img` (the `…-sysupgrade.img.gz` file is the [update](updating.md) payload). The commands below use `startwrt.img` as a placeholder for the downloaded filename.
+1. Download the latest StartWRT firmware image from the [Start9 releases page](https://github.com/Start9Labs/start-technologies/releases?q=start-wrt&expanded=true) (StartWRT releases are the ones tagged `start-wrt_v…`). For a fresh install, download the **sdcard** image — it is named `startwrt-<version>-<git hash>_spacemit-k1-sdcard.img` (the `…-sysupgrade.img.gz` file is the [update](updating.md) payload). The commands below use `startwrt.img` as a placeholder for the downloaded filename.
```
dr-bonez: I honestly didn't realize we could put `/`s in tag names but on further research that actually seems better

scripts/manage-release.sh
```diff
@@ -165,6 +188,7 @@ release_files() {
         os) for f in *.iso *.img *.squashfs; do [ -f "$f" ] && echo "$f"; done ;;
         cli) cli_binaries; deb_files ;;
         deb) deb_files ;;
+        wrt) for f in *-sdcard.img *-sysupgrade.img.gz; do [ -f "$f" ] && echo "$f"; done ;;
```
dr-bonez: why aren't we gzipping the sdcard.img also?

scripts/manage-release.sh
```diff
@@ -353,9 +377,26 @@ cmd_pre_check() {
                 errors=1
             fi
             ;;
+        wrt)
+            # `release` registers + indexes into the single StartWRT registry, so
+            # it shouldn't already carry this version.
```
dr-bonez: we should probably have a beta registry

projects/start-wrt/docs/src/installing.md
dr-bonez: I'm gonna go through and do this actually. `<product>/v<version>` is the new structure

scripts/manage-release.sh
Dominion5254: Agreed and implemented. It just needs to be stood up now.

scripts/manage-release.sh
Dominion5254: fair point. updated to gzip the sdcard.img

projects/start-wrt/docs/src/installing.md
Dominion5254: I addressed in 5985737fde310cbdd815faa7de957daedd76f61d before I saw this second comment. I can revert that commit if you would prefer to handle the implementation

scripts/manage-release.sh
dr-bonez: you can put it on the same box as the prod registry

.github/workflows/start-wrt.yaml
```diff
@@ -196,20 +197,8 @@ jobs:
...
-          files: |
-            artifacts/*.img
-            artifacts/*.img.gz
-          fail_on_unmatched_files: true
-          generate_release_notes: true
+          # The beta-registry register/index (against this run's
+          # startwrt-openwrt-image artifact), the GitHub release + tag
+          # (start-wrt/v<version>), and the beta -> production promotion are
+          # all cut locally from scripts/manage-release.sh — deliberately
+          # gated behind the developer key, not CI.
```
dr-bonez: we can grant the gh permission to publish to beta. as it stands, the release script assumes this, and tries to pull from beta to push to prod.

Dominion5254: Fair, updated to match start-os implementation

dr-bonez (approved):

PR #3403, opened 2026-07-04
feat(tunnel): IPv6 GUA pinholes — PCP v6 + manual forwards (PR2)

helix-nine:
## PR2 — IPv6 GUA pinholes (PCP v6 + unified v4+v6 forwards)

Follow-up to #3388 (per-subnet IPv6 addressing). Opens **inbound** to a client's
own IPv6 GUA on the tunnel — the piece that makes v6 hosting work end-to-end —
and unifies v4 forwarding and v6 pinholes behind one dialog.

### What a "pinhole" is
A client already gets a `/128` GUA on any IPv6-enabled subnet (#3388). A pinhole
accepts inbound to that GUA on a port — a pure `ip6 startos` firewall accept,
**no NAT** (the GUA is directly routable, and the OS-side reply-routing sends the
return out the arrival interface). When the external port differs from the
internal (e.g. an 80→443 redirect) it becomes a **port-only DNAT** on the same
GUA. Because each client has its own GUA, external ports only need to be unique
per-GUA, so two clients can both publish :443.

### Two ways a pinhole is opened
- **Automatic (PCP):** the tunnel runs a **v6 PCP server** on a wg-bound v6-only
  socket; `handle6` forces the target to the sender's own GUA and honors the PCP
  *Suggested External Port* field, so a remap needs no protocol change. StartOS
  servers already request GUA pinholes; this PR also adds the OS-side auto
  **80→443** redirect on v6 (mirroring the existing v4 `redirect_maps`).
- **Manual:** a `pinhole add|remove|set-enabled|update-label` CLI/RPC surface,
  and — in the UI — the reworked **forward dialog**.

### UI (unified v4 + v6)
The Add Port Forward dialog now has an **IP Version** radio (`IPv4` / `IPv6` /
`IPv4 + IPv6`); the external/internal port fields apply to whichever is chosen,
IPv6 is gated on the server having a GUA, and the also-80 checkbox opens the
redirect on each selected version. The forwards tables gained an **IP** column,
and toggle/delete/relabel route to the pinhole API for v6 rows.

### Backend
- `db`: `Pinhole { label, enabled, count, internalPort: Option<u16>, auto }` +
  `pinholes6` map (`serde(default)`, no migration).
- `net/port_map/server`: `handle6` + v6 `GatewayBackend` methods (default-off).
- `tunnel/forward/pinhole`: `is_known_gua` (reverse `host_v6`), nft rules via
  `nft_rule_v6`, db persistence, startup seed.
- `tunnel/forward/pcp`: v6 socket + serve loop, `TunnelContext` v6 impl.
- `net/net_controller`: OS-side v6 `gua_redirect_maps` (auto 80→443).
- `api`: manual CRUD (+ 5-locale i18n, regenerated manpage); `ApiService`
  methods (live + mock) + regenerated osBindings.
- docs: `port-forwarding.md`, `cli-reference.md`, `ipv6.md`, CHANGELOG.

### Testing
- `cargo check -p start-core` clean; 25 tunnel + 6 pinhole/`handle6` unit tests
  pass; `check:tunnel` and the full `build:tunnel` (Angular AOT) green; prettier
  clean.
- **Not automated:** end-to-end v6 reachability needs a real routed global
  prefix (a local libvirt VM can't provide one) — worth a real-VPS pass before
  release, same caveat as #3388.


projects/start-tunnel/web/src/app/routes/home/routes/port-forwards/index.ts
```diff
@@ -45,7 +46,7 @@ import { mapForwards, MappedDevice, MappedForward } from './utils'
           <tr>
             <th></th>
             <th>Label</th>
-            <th>External IP</th>
+            <th>External Address</th>
```
dr-bonez: IP was fine here

projects/start-tunnel/web/src/app/routes/home/routes/port-forwards/index.ts
```diff
@@ -135,7 +136,7 @@ import { mapForwards, MappedDevice, MappedForward } from './utils'
       <table class="g-table no-actions" [tuiSkeleton]="!portForwards()">
         <thead>
           <tr>
-            <th>External IP</th>
+            <th>External Address</th>
```
dr-bonez: And here

dr-bonez (dismissed):

projects/start-tunnel/web/src/app/routes/home/routes/port-forwards/index.ts
helix-nine: Restored the IP column (both tables) in f4c57c20b — I'd removed it after the earlier "it's confusing" note, but reverting that now that you'd keep it.

projects/start-tunnel/web/src/app/routes/home/routes/port-forwards/index.ts
helix-nine: Restored here too.

projects/start-tunnel/web/src/app/routes/home/routes/port-forwards/index.ts
helix-nine: Misread this — you meant keep the 'External IP' name, not the version column. Fixed in afac7c754: the IPv4/IPv6 version column is gone, and the address column is 'External IP' again (a v6 GUA is still an External IP). Sorry for the churn.

projects/start-tunnel/web/src/app/routes/home/routes/port-forwards/index.ts
helix-nine: Same fix here — 'External IP' header kept, version column removed.

dr-bonez (approved):

PR #3405, opened 2026-07-04
Fix 3293 secure bridge

### LINKED ISSUE #3293: Replace .startos service-to-service addressing with the secure host bridge
## Summary

Replace `.startos` service-to-service addressing with the **secure host bridge**. Services should no longer reach each other via the `.startos` overlay DNS name (or by direct container-to-container IP); instead they connect to the target service through the host bridge address, obtained from the SDK, over plain HTTP, on the **same port the service uses for its LAN / `.local` domain**.

## Motivation

- **Eliminates the DNS-caching failure class.** `.startos` names resolve through internal DNS; long-lived clients and resolvers cache records, so when a container's IP changes on restart you get stale-IP connection failures that are painful to diagnose. A bridge address fetched at connect time has no resolution/caching step.
- **Host-mediated traffic is enforceable.** Routing every hop through the host (and forbidding direct container↔container, even by IP) lets the host firewall, observe, and police inter-service traffic — a cleaner isolation story for a single-box appliance.
- **Unifies the port model.** Service-to-service uses the service's LAN-domain port, removing the long-standing "internal bound port vs. LAN-mapped port" confusion (see Start9Labs/ollama-startos#16).

## Contract for consumers

- The bridge network is marked **secure**, so service-to-service is **plain HTTP** — no TLS / cert handling required.
- The bridge address is **obtained from the SDK at runtime**. **Do not hardcode it** — it is not guaranteed static (don't assume `10.0.3.1`).
- Use the **target service's LAN-domain port** (the same port shown for its `.local` / LAN address), not its internal bound port.
- For OpenAI-compatible APIs, append `/v1` to the base URL as usual.

## Deprecation plan

- `.startos` is retained for **one version** for backwards compatibility, then removed.
- Direct container-to-container communication (including by raw IP) is forbidden under the new model.

## Timeline

Land for all services **before StartOS beta.10 (SDK 2.0)** goes live, so the corresponding service updates are already published and waiting for users when they upgrade to the new OS.

## Scope / tasks

- [ ] **Core / OS:** mark the bridge network secure; expose each service on the bridge at its LAN-domain port.
- [ ] **SDK 2.0:** provide an API for a service to obtain the host bridge address (and, ideally, a helper to build a dependency's service-to-service base URL from its interface/port).
- [ ] **Backwards compatibility:** keep `.startos` resolving for one version, then remove it.
- [ ] **Fleet-wide service migration:** replace `.startos` / container-IP usage with the SDK-supplied bridge address across all packages. Known consumers already using `.startos` today include **open-webui** (reads `ollama`, `vllm`, `llama-cpp`, `maple-proxy`), plus the AI services and **searxng** wiring.
- [ ] **Documentation:** add an authoritative networking section to the packaging guide (and a one-line pointer in the README template) describing the host-bridge model. This **replaces** the per-README `.startos` URL documentation that was being requested — we don't want to document a convention that's being removed, nor a hardcoded address.

## Related

- Start9Labs/ollama-startos#16 — request to document per-service `.startos` URLs in each AI service README; **superseded by this** (the addressing model it documents is going away; the replacement will be documented centrally).
- Start9Labs/start-os#3286 — surface internal/service URLs in the Interfaces tab (complementary UX surface).

winshaurya:
Closes #3293 

## Summary
This PR implements the new service-to-service communication topology by routing inter-service traffic over the intrinsically secure `lxcbr0` host bridge. It effectively deprecates the `.startos` overlay DNS namespace, resolving the long-standing DNS-caching failure class where long-lived clients failed to reconnect after a dependency restarted with a new internal IP.

## Scope / Tasks
- [x] **Core / OS:** mark the bridge network secure; expose each service on the bridge at its LAN-domain port.
- [x] **SDK 2.0:** provide an API for a service to obtain the host bridge address (and a helper to build a dependency's service-to-service base URL from its interface/port).
- [x] **Backwards compatibility:** keep `.startos` resolving for one version, then remove it. (Added a `tracing::warn!` to loudly alert developers to migrate).
- [x] **Documentation:** add an authoritative networking section to the packaging guide (and a one-line pointer in the README template) describing the host-bridge model.
- [ ] **Fleet-wide service migration:** replace `.startos` / container-IP usage with the SDK-supplied bridge address across all packages. *(Note: We migrated StartOS's own internal `tor` proxy usages in this PR. Downstream packages like `open-webui`, `ollama`, and `vllm` will be migrated in their respective repositories.)*

## Architecture & Code Changes
* **Bridge Exposure (`net_controller.rs`)**: Services are now explicitly exposed on the host bridge (`10.0.3.1`) at their LAN-domain port. Traffic over this bridge natively bypasses the TLS proxy and routes over plain HTTP.
* **Internal OS Migration (`socks.rs` & `context.rs`)**: Updated StartOS's own internal connections (specifically to the `tor` proxy) to dogfood the new bridge addressing architecture instead of hardcoding `tor.startos`.
* **SDK Bridge API (`StartSdk.ts`)**: Added a highly readable `sdk.getServiceBridgeUrl` helper. This abstracts the `10.0.3.1` host IP fetch and allows developers to easily construct dynamic base URLs for their dependencies at runtime.

## Verification
- Verified `start-core` compilation and successful translation of DNAT forwarding rules.
- Confirmed backward compatibility for `.startos` resolution.
- Reviewed all documentation links and TypeScript SDK linting.

 
  Note: Because the bridge network never leaves the host, this communication happens securely over plain HTTP, bypassing TLS overhead ; we can discuss if we need to change that!!

### DIFF
--- a/projects/start-sdk/docs/package-template/README.md
+++ b/projects/start-sdk/docs/package-template/README.md
@@ -27,6 +27,7 @@ image tags anywhere — the manifest is the source of truth.
 ## Network Access and Interfaces
 
 <!-- TODO -->
+> **Service-to-Service:** To connect to this service from another StartOS package, use the SDK's host bridge API (see the [Networking Guide](https://docs.start9.com/packaging/networking)).
 
 ## Actions (StartOS UI)
 
--- a/projects/start-sdk/docs/src/SUMMARY.md
+++ b/projects/start-sdk/docs/src/SUMMARY.md
@@ -52,6 +52,7 @@
 
 ## Networking
 
+- [Service-to-Service Networking](networking.md)
 - [Expose a Web UI](recipe-web-ui.md)
 - [Expose Multiple Interfaces](recipe-multi-interface.md)
 - [Expose an API-Only Interface](recipe-api-interface.md)
--- a/projects/start-sdk/docs/src/networking.md
+++ b/projects/start-sdk/docs/src/networking.md
@@ -1,0 +1,35 @@
+# Networking
+
+StartOS provides a secure, host-mediated networking environment. Services do not connect to each other directly by IP or overlay network. Instead, all inter-service communication routes through the host's internal bridge network.
+
+## The Host Bridge Model
+
+Services on StartOS are strictly isolated. Direct container-to-container communication is disabled by the firewall. To communicate with another service, your service must use the **Host Bridge Address** over plain HTTP.
+
+When a service connects to the host bridge, the StartOS `net_controller` intercepts the traffic and routes it to the target container, providing an enforceable boundary.
+
+### Connecting to a Dependency
+
+To connect to a dependency, you construct a URL using the host's bridge IP address and the dependency's LAN-domain port.
+
+Because the bridge network is intrinsically secure (traffic never leaves the host), service-to-service communication happens over **plain HTTP**, bypassing any TLS wrapping the service might provide for external clients.
+
+### Using the SDK
+
+The StartOS SDK provides a helper to automatically construct the correct service-to-service base URL for a dependency's interface.
+
+```typescript
+import { sdk } from './sdk'
+
+// Within an action, init, or main lifecycle hook:
+const url = await sdk.getServiceBridgeUrl(effects, 8080)
+// url will be e.g. "http://10.0.3.1:8080"
+```
+
+> **Note**: Do not hardcode `10.0.3.1` in your scripts. Always obtain the bridge IP from the SDK at runtime, as the bridge subnet is subject to change in future network topologies.
+
+## Deprecation of `.startos`
+
+In older versions of StartOS (0.3.x), services used `.startos` DNS overlay names (e.g., `http://bitcoind.startos:8332`) to communicate. This model suffered from DNS caching issues and is being removed. 
+
+> **Important**: Never use `.startos` domains for new packages. If you are migrating an older package, replace all `.startos` base URLs with the bridge URL obtained from the SDK.
--- a/projects/start-sdk/lib/StartSdk.ts
+++ b/projects/start-sdk/lib/StartSdk.ts
@@ -173,6 +173,21 @@ export class StartSdk<Manifest extends T.SDKManifest> {
         effects.getServicePortForward(...args),
       clearBindings: (effects, ...args) => effects.clearBindings(...args),
       getOsIp: (effects, ...args) => effects.getOsIp(...args),
+      /**
+       * Constructs the service-to-service base URL for a given interface over the secure host bridge.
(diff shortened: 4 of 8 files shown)


helix-nine (commented):
Thanks for picking this up, @winshaurya — but this is already implemented, and the fact that you couldn't tell that from the open issue is our fault, not yours. Apologies for the wasted effort.

The two Core/OS tasks in #3293 — "mark the bridge network secure" and "expose each service on the bridge at its LAN-domain port" — were resolved by #3297 (`feat: treat lo and lxcbr0 as secure networks for insecure traffic`), merged 2026-06-04. That PR extended `NetworkInterfaceInfo::secure()` with `is_intrinsically_secure()` (`db/model/public.rs`) so `lo` and `lxcbr0` are always treated as secure. Before it, nothing ever set `secure = Some(true)`, so insecure bindings got no forwards anywhere; after it, every binding's always-on internal `10.0.3.1:<LAN-domain port>` address (`HostnameInfo::is_internal()`) flows through `host/mod.rs::update_addresses` → the non-SSL forward's `private_ips` → an installed DNAT over the secure bridge. **#3293 should have been closed when #3297 merged, and it wasn't — that oversight is on us.**

Verified on a current build (0.4.0-beta.9):

```
net gateway list:
  lxcbr0 → deviceType "bridge", subnets ["10.0.3.1/24", ...], secure: null   (→ is_intrinsically_secure() = true)

net forward dump-table:
  53049 → 127.0.0.1:80   reqs: private { 10.0.3.1, 127.0.0.1, 192.168.122.171, ... }, secure: false
```

`10.0.3.1` is already in the forward's `private` set, so the bridge DNAT this PR is trying to add is already in place. The `forwards.entry(lan_port)…private_ips.insert(HOST_IP)` block here re-does what `update_addresses` already produces.

A few other notes on the diff, for the record:
- The second copy of the bridge block landed outside the `for (port, bind)` loop (in the HTTP→HTTPS redirect region), so `bind`/`port` are out of scope and `start-core` doesn't compile as-is.
- Of the remaining non-OS items from #3293: `getOsIp()` already returns the bridge IP; `getServiceBridgeUrl` hardcodes `http://`, which is wrong when the LAN-domain port is the SSL port; and the `tracing::warn!` fires on every `.startos` lookup (log spam at query rate).
- The `tor.startos → 10.0.3.1` swap in `socks.rs`/`registry/context.rs` isn't a safe no-op — 9050 is tor's container-internal SOCKS port (and a restricted port), not a bridge-exposed LAN-domain port.

Recommend closing this in favor of #3297. If the leftover SDK helper / `.startos` deprecation / docs bits are still wanted, happy to spin them into a focused follow-up.

PR #3406, opened 2026-07-05
refactor: `TaskService` add service around loader + try/catch

waterplea:


helix-nine (commented):
## Thorough audit — `TaskService` + `@Service()`/`inject()` sweep

Reviewed all 72 changed files, built every affected web app, and diffed the full-app typecheck against `master`. **The two strategies are sound and the sweep is overwhelmingly clean** — `@Service()` is correct Angular-22 usage (verified it's exported by `@angular/core@22.0.2` and is exactly shorthand for `@Injectable({providedIn:'root'})`, so every root-singleton conversion is behavior-preserving), the `inject()` / `@HostListener`→`host:{}` conversions are mechanically correct (no lost `@Optional/@Self/@SkipSelf/@Host`, provider wiring for `ErrorService`/`ApiService` intact), and the UI AOT build compiles clean. A handful of must-fix items and some follow-through opportunities below.

### ⛔ Must fix

**1. `rimraf` in `start-core/Makefile` breaks CI.** The `dist:` recipe changed `rm -rf dist` → bare `rimraf dist`. CI invokes this as a bare shell step **before any `npm ci`** — `test.yaml:67` and `deploy-brochure.yml:58` both do `run: cd shared-libs/ts-modules/start-core && make dist`. `rimraf` isn't a dependency of `@start9labs/start-core` (only of the root workspace), and a bare `make` doesn't put root `node_modules/.bin` on `PATH`, so this fails with `rimraf: command not found`. Since web/SDK/brochure all resolve start-core via `file:.../start-core/dist`, this takes down those builds too. (It "works" locally only when invoked through `npm run build:core`, because npm injects `node_modules/.bin` into `PATH`.)
→ Revert to `rm -rf dist` (portable, zero-dep, identical behavior), or use `npx rimraf dist` **and** add `rimraf` to start-core's own `devDependencies`.

**2. Three new type errors in the `backups` route — `TaskService.run()` used where the callback's return value / rejection is needed.** Important context: the whole `backups` route already fails to typecheck on `master` (48 pre-existing errors — it references a bunch of not-yet-existing `ApiService` methods/types, i.e. it's WIP), `ng build` strips types without failing, and the repo's `check:*` scripts only compile the **static-import closure from `main.ts`** so lazy-loaded routes are never type-checked. That's why none of this trips CI. A full-app `tsc` (all `src/**/*.ts`) diffed PR-vs-master isolates exactly these three new errors:

- **`backups/services/restore.service.ts` `getRecoverData()`** (TS2322): wrapping `getBackupInfo` in `this.tasks.run(async () => this.api.getBackupInfo(...), 'Decrypting drive')` makes the pipe emit a **`boolean`** instead of `T.BackupInfo` — `run()` returns `Promise<boolean>` and discards the callback's value. Downstream `map(backupInfo => ({ …backupInfo… }))` builds a `RecoverData` with `backupInfo: true`, and `recover.component.ts:105` reads `.packageBackups` on that boolean → `undefined` → the "Select Services to Restore" screen is broken. Also, `run()` catches internally and resolves to `false` rather than rejecting, so the outer `catchError(… EMPTY)` no longer short-circuits — on a decrypt failure the user gets an error toast **and** a broken RECOVER dialog opens instead of aborting.

- **`backups/modals/targets.component.ts` `add()`/`addPhysical`** (TS2488): `add()`/`update()` now return `Promise<boolean>`, but `addPhysical`'s handler still does `const [id, entry] = Object.entries(response)[0]`. `Object.entries(true)` is `[]`, so `[0]` is `undefined` and the destructure throws `TypeError: undefined is not iterable` — the physical-disk target is never saved.

- **`backups/modals/jobs.component.ts` `delete()`** (TS2339): switched `this.api.removeBackupTarget({id})` → `this.api.removeBackupJob({id})`, which isn't declared on `ApiService`. This one is incidental to the WIP backups API (the old call was itself semantically a job-vs-target mismatch), not really about the `TaskService` pattern — but it's a new error.

Root cause for the first two: **`TaskService.run()` is a fire-and-forget, value-discarding wrapper (`Promise<boolean>`)** — great for "do X, show a loader, toast on failure," wrong wherever you need the callback's result or need the error to propagate. Two options: don't route those sites through `TaskService` (keep the manual loader), or make `run` generic so it can serve both:
```ts
async run<T>(task: () => Promise<T>, content?: PolymorpheusContent): Promise<T | undefined> {
  // return the task's value on success, undefined on error
}
```
That fixes restore + targets while staying truthy for the boolean callers.

### ⚠️ Behavior changes — confirm intent (mostly improvements)

- **Silent `console.error` → error toast** now surfaces on several previously-invisible failure paths, because `TaskService.run` always calls `ErrorService.handleError` (which toasts): setup-wizard `loading.page.restart()`; **diagnostic `home.page` `restart`/`forgetDrive`/`repairDisk`** (your own `// @TODO Matt: … do we still want to show error toast?` at the top of the file flags exactly this — needs a decision); start-tunnel `subnets/index.ts onDelete`. All arguably better (these failures used to be silent), but they're user-visible — please confirm.
- **Tunnel add/edit dialogs** (`devices/add`, `dns/add`, `port-forwards/add`, `port-forwards/edit-label`, `subnets/add`): `this.context.$implicit.complete()` moved from the old `finally` into the success path of the `run()` callback, so on error the dialog now **stays open** instead of closing. This is standard/good form UX (keep the form up to retry) — just confirming it's deliberate.
- `backups/services/create.service.ts` `createBackup` is now fire-and-forget and the loader label capitalization changed.

### 🔸 Nits
- `authorities/authority.service.ts:139` — `await await this.api.initAcme(...)` (double `await`, harmless).
- `diagnostic/home/home.page.ts:131` — dead outer `try/catch` around `this.repairDisk()` (pre-existing, now clearly useless).
- Stale unused imports: `TuiNotificationMiddleService` in `subnets/index.ts:15`; `Injectable` in `shared/services/dialog.service.ts:1`.
- A few debug `console.error/console.log` calls dropped from converted handlers.

### ✅ "Take advantage everywhere" — remaining candidates
Clean, equivalent-behavior `TaskService` conversions still on the old loader+try/catch+`handleError` boilerplate:
- `services/controls.service.ts` — `start()` / `stop()` / `restart()`
- `services/standard-actions.service.ts` — `rebuild()` / `doUninstall()` (navigate stays inside the closure)
- `system/routes/backups/` — `network.component.ts` (`edit`/`addTarget`/`forget` — `run()`'s `Promise<boolean>` is exactly the form handler's contract), `recover.component.ts` `restore()`, `legacy-warning.component.ts` delete handler, `backup.component.ts` `createBackup()`

Correctly **left alone** (noting so they don't read as misses): `services/action.service.ts` (catch shows a custom alert dialog, not `handleError`), start-tunnel `update.service.ts` (keeps the loader subscription to dismiss later on poll completion), start-wrt (has its own richer `ActionService.run`), and signal-driven `[loading]`-button flows (converting would swap an inline loader for a full-screen modal).

### 🔎 One CI-hardening note (out of scope for this PR)
The `check:*` scripts type-check only the static-import closure from `main.ts`, and `ng build` doesn't fail on type errors — so lazy-loaded routes (the entire `backups` feature, 48 pre-existing errors + these 3) are effectively untype-checked in CI. Worth a follow-up to add a full-`src` `tsc --noEmit` gate.

---
*Verification: all 6 `check:*` pass; UI `ng build` AOT compiles clean (templates, `host:{}` bindings, `@Service()` all valid); full-app `tsc` (all `src`, incl. lazy routes) diffed PR-vs-`master` isolates exactly the 3 new errors above; `@Service()` confirmed in `@angular/core@22.0.2`. Findings produced by a multi-agent pass with an adversarial verify stage, then re-checked by hand against the actual build.*

MattDHill (approved):

PR #3411, opened 2026-07-06
refactor(net): GUA WAN exposure via public flag — remove gua_access tri-state

helix-nine:
## Remove the GUA access tri-state — `public` is the single source of truth

Per discussion: `HostnameInfo.public` is consumed by services choosing addresses
for P2P, so it must be *correct* — and the old model kept a GUA's real
reachability in a separate `gua_access` tri-state that `public` never reflected.
This removes `gua_access` entirely.

### New model
- **`gua_wan: BTreeSet<SocketAddrV6>`** (was `gua_access: BTreeMap<_, GuaAccess>`):
  the GUAs the operator opted into WAN exposure. `update_addresses` projects it
  into `HostnameInfo.public`, so `public` is authoritative everywhere — P2P
  address selection, upstream pinholes (`net_controller`'s `is_wan` special case
  collapses to `.public`).
- **Enable/disable uses the existing tools.** A public GUA is opt-in via the
  `enabled` set (same rule as public IPv4); a local GUA is on unless in
  `disabled`. The GUA branch in `enabled()` is gone. `set-gua-wan` (replacing
  `set-gua-access`) carries the row's current on/off state across a Local↔Public
  flip and re-runs `update_addresses` so the projection stays fresh.
- **UI:** every row has the standard on/off toggle; a GUA's access column is the
  Local/Public dropdown (from #3409's design) reading `address.access` — i.e. the
  projected `public` flag — instead of a parallel tri-state field.

### No db migration
Per @dr-bonez: `gua_wan` is `#[serde(default)]` and the old `gua_access` key is
ignored on load, so beta.10 alpha boxes deserialize cleanly — any GUAs they had
set to LAN+WAN/Disabled just revert to the LAN-only default (accepted).

### Testing
- `cargo check` clean; `net::host` tests 27/27 (tri-state test replaced with
  `gua_enabled_follows_public_flag`).
- Bindings regenerated (`GuaAccess`/`BindingSetGuaAccessParams` removed);
  `check:ui`, full `build:ui` (AOT), `check:setup`, and container-runtime
  `npm run check` all green.
- Mock mirrors the backend (guaWan set, enabled carry, `public` re-projection),
  and the seeded GUA fixture exercises the dropdown.


dr-bonez (dismissed):

helix-nine:
Pushed the port-map fix for the `fe80::a3b:1` PCP timeout to this branch (`7d92286ef`).

**Why #3417 didn't take:** the gateway *is* `InboundOutbound`, so #3417's guard should have rejected the link-local — but `candidate_gateways` also admits a gateway that falls within one of the interface's subnets, and a WireGuard iface carries its **own** `fe80::/64` (NM assigns it). Every link-local shares the `fe80::/64` network, so `subnets.contains(fe80::a3b:1)` was true and re-admitted the link-local past the guard. `have_v6` then went true and the `host_v6`-derived server fallback never fired, so PCP kept hitting an `fe80::` the tunnel server doesn't listen on. #3417's regression test only passed because its fixture omitted the iface's own `fe80::/64` subnet.

**Changes in `candidate_gateways`:**
- Reject a link-local v6 gateway on `InboundOutbound` *before* the subnet-contains clause can re-admit it, so the `host_v6`-derived server (the routed prefix's `.1`) fills the v6 slot. A real router (`gateway_type None`) keeps its `fe80::` default gateway.
- Never derive the tunnel server v6 from a link-local prefix (defensive).
- **Never attempt PCP to an `OutboundOnly` gateway:** `candidate_gateways` now returns empty for `OutboundOnly`. It's the single chokepoint every port-map call site funnels through (GUA pinhole, v4/v6 redirect, hostname_maps, forward.rs) — 3 of the 4 sites previously lacked the guard — so this closes them all. Removed the now-redundant per-site guard + unused import here in `net_controller.rs`.

Regression tests: `tunnel_rejects_link_local_gateway_even_when_a_subnet_contains_it` (the exact live-box fixture) and `outbound_only_gateway_has_no_candidates`. All 12 `net::port_map::client` tests pass; crate compiles clean.

shared-libs/crates/start-core/src/net/port_map/client.rs
```diff
@@ -89,13 +97,16 @@ pub fn candidate_gateways(info: &NetworkInterfaceInfo) -> Vec<(IpAddr, Option<u3
...
-                    && info.gateway_type != Some(GatewayType::InboundOutbound);
-                if ll_ok || ip_info.subnets.iter().any(|s| s.contains(ip)) {
+                // A StartTunnel peer owns no link-local on the wg link, so an
+                // fe80:: gateway can never answer PCP — reject it even though the
+                // wg iface's own fe80::/64 "contains" it (every link-local shares
+                // that /64), so the host_v6-derived server fills the v6 slot. A
+                // real router's fe80:: default gateway (gateway_type None) stays.
+                let is_ll = ipv6_is_link_local(*v6);
+                if is_ll && info.gateway_type == Some(GatewayType::InboundOutbound) {
+                    continue;
```
dr-bonez: the inbound-outbound guard is redundant with the above outbound-only guard

helix-nine: Good catch — fixed in `26d5a8129`. Once `OutboundOnly` returns no candidates, gating the skip on `== InboundOutbound` was pointless: a link-local is never a PCP server we can reach, so I skip it unconditionally now. Collapsed the type guard and the `is_ll ||` push path into a single `if ipv6_is_link_local(*v6) { continue; }`.

One behavioral note: this also drops a **home router's** `fe80::` v6 next-hop (`gateway_type None`) as a PCP target — previously #3417 kept it. Home v4 port-mapping is unaffected (the router still keeps its v4 gateway); only the v6 link-local next-hop goes away, which was never a reliable pinhole target anyway. Shout if you wanted to preserve that path.

shared-libs/crates/start-core/src/net/forward.rs
```diff
@@ -680,11 +680,13 @@ impl InterfaceForwardEntry {
...
-                            if !reqs.secure && !info.secure() {
-                                continue;
-                            }
 
                             let public = reqs.public_gateways.contains(gw_id);
+                            // The WAN is never secure: an insecure exposure may reach the LAN
+                            // over a secure gateway, but never the public internet.
+                            if !reqs.secure && (public || !info.secure()) {
+                                continue;
+                            }
```
dr-bonez: we can still listen on the lan even if it they tried to enable `public` access.
```
let public = reqs.public_gateways.contains(gw_id) && reqs.secure;
if !reqs.secure && !info.secure() {
    continue;
}
```

dr-bonez (approved):

PR #3416, opened 2026-07-06
feat(tunnel): honor PCP mapping leases + clear forwards on admin actions

helix-nine:
## Problem

StartTunnel **parsed and echoed** the PCP requested lifetime (capped at 3600s) but **never enforced it**. Once a device opened a port via PCP, that auto-created forward / pinhole / SNI route lived forever — even after the requesting StartOS went offline, rotated its WireGuard key, or dropped the exposure. On tunnel restart every mapping was re-seeded permanently. The client already requested `lifetime=3600` and renewed, so "pass a timeout" mostly existed; the missing pieces were **honor** and **renew-before-expiry**.

## What this does

**StartOS passes a timeout** (unchanged: requests 3600s) → **StartTunnel honors it** → **StartOS renews well before expiry.**

### Server — honor the lease
- Thread the granted lease into `GatewayBackend::add_forward` / `add_pinhole` (SNI already had it). PCP passes `Some(granted)`; UPnP IGD passes `None` (StartOS requests permanent UPnP leases, and the tunnel path always uses PCP anyway).
- New in-memory **lease table** on `TunnelContext` + a **30s reaper** (`tunnel/forward/lease.rs`) that tears down any **auto** mapping whose lease lapses. Manual (user-added) forwards carry no lease and never expire — the reaper is auto-guarded.
- Auto DB entries are **re-leased on startup** so a device that never reconnects is reclaimed after one lease; the client's re-MAP (RFC 6887 §8.5 epoch reset after a restart) refreshes it.
- **In-memory, not persisted** — deliberately, so a client's periodic renewal never churns PatchDb (and never wakes DB subscribers). The renewal re-stamp happens on the idempotent re-assert path too, so a still-renewing client is never reaped.
- Lease duration: **1h** (matches the existing server cap / client request — confirmed with @dr-bonez).

### Client (StartOS) — renew before expiry
- Renew each PCP mapping at **~half its gateway-granted lifetime** (RFC 6887 §11.2.1) using `crab_nat`'s `expiration()`/`lifetime()`, instead of blindly renewing every 180s tick. Self-corrects if the gateway ever caps the lease below what we asked for. The 180s tick still drives UPnP re-assert and retries of failed/not-yet-active mappings.

### Admin actions clear a device's forwards immediately
Per @dr-bonez — don't make cleanup wait for a lease to lapse:
- **Delete a device** or **demote it to a client** → clears *all* its forwards, SNI routes, and v6 pinholes (this also fixes a pre-existing leak: a deleted device's v6 pinholes were never GC'd — `gc_forwards` only covered `port_forwards`).
- **Disable automatic port forwarding** for a device → clears its *automatic* forwards, leaving any the operator added manually.

All three go through one helper, `tunnel::forward::clear_for_peer(ctx, peer, auto_only)`.

## Tests
New unit tests (all green, plus the full 453-test `start-core` lib suite):
- server: a MAP requesting > cap is granted the capped lease **and** that same value is handed to the backend (so the tunnel stamps the real grant); a lifetime-0 MAP never adds.
- client: `renew_due` fires at half-life, not before.
- lease: only lapsed leases are selected; SNI leases are per-hostname.
- clear: `auto_only` leaves manual forwards; a different device is never swept.

The reaper's DB/dataplane teardown mirrors the existing `remove_peer_forward` / `remove_forward_by_source` paths; it's not unit-tested here since it needs a live tunnel DB.

## Notes / scope
- **Formatting:** the repo pins a nightly rustfmt I can't reproduce in this environment (plain nightly reformats ~78 files of pristine code), so I hand-formatted in the surrounding style rather than run a divergent formatter. If CI's fmt check flags anything, it'll be mechanical — happy to apply the canonical `cargo fmt` output.
- **WAN-change re-key edge:** `resync_forward_keys` re-keys a DNAT/SNI entry's external source on a WAN change; I did **not** re-key the in-memory lease alongside it. This is safe (the reaper's `auto=true` guard means a stale lease can't reap a manual entry, and the client's next MAP re-stamps at the new key), just slightly less precise for the rare operator-driven WAN change. Easy follow-up if you'd prefer it airtight.
- Docs + changelog updated for StartTunnel (`port-forwarding.md`, CHANGELOG) and StartOS (CHANGELOG).


shared-libs/crates/start-core/src/tunnel/forward/lease.rs
```diff
@@ -0,0 +1,239 @@
...
+        }
+    });
+    Ok(())
+}
+
+/// The reaper loop: every [`REAP_INTERVAL`], tear down any auto mapping whose
+/// lease has lapsed. Runs for the life of the tunnel.
+pub async fn run(ctx: TunnelContext) {
+    loop {
+        tokio::time::sleep(REAP_INTERVAL).await;
```
dr-bonez: Instead of polling on a loop, each reap could schedule the next, finding the soonest instant, since it has to enumerate them all anyway. Newly added ones can also reset the timer if they expire sooner

helix-nine: Done in 8170bdb7a. The reaper now computes the soonest expiry while it enumerates leases to reap (`reap_expired` returns the min remaining), and `sleep_until`s exactly that instant instead of polling every 30s. `stamp` fires a `Notify` so a newly added lease that expires sooner wakes the reaper early to pull its next wake-up forward (`select!` on the sleep vs. the notify); with no leases outstanding it parks on the notify entirely. `forget` doesn't notify — removing a lease only pushes the next expiry later, so the current sleep stays valid.

dr-bonez (approved):

PR #3421, opened 2026-07-07
feat(net): DualStack public domains — serve over IPv6 GUA alongside IPv4

helix-nine:
## DualStack public domains

Public (clearnet/ACME) domains were IPv4-only: the add-domain flow only ever enabled the WAN IPv4, advertised/verified an `A` record, and opened a v4 pinhole. This makes a public domain **DualStack** — reachable over both IPv4 and IPv6 — whenever its gateway has an IPv6 global-unicast address (GUA).

### The key insight

Most of the v6 side already existed from the GUA-WAN work (#3411 and friends). A public-domain vhost already **binds and SNI-routes on the gateway's GUA** (`update_vhost_listeners` binds every subnet address of a public gateway), and `ProxyTarget::filter` already rejects off-link sources on the private `*` vhost — so opening the port exposes **only** the SNI-matched domain, never the bare GUA. The one missing network piece for the common (SSL) case was requesting the upstream router's v6 firewall pinhole.

### What changed

- **`net_controller`** — request a v6 GUA pinhole per public-domain vhost (each GUA in the gateway's subnets at the domain's port), reusing the existing `gua_pinholes` withdrawal and the v6 80→443 redirect. The bare GUA is **never** added to `gua_wan`, so SNI at our own vhost disambiguates domain traffic from bare-IP traffic — an SSL domain needs no separate GUA→WAN opt-in. No StartTunnel SNI-demux for v6 (every host has its own GUA — nothing to share).
- **`address.rs`** — a non-SSL public domain (no SNI to filter on) exposes the bare GUA too, mirroring the existing WAN-IPv4 enable; a re-projection propagates the `gua_wan` change into `HostnameInfo.public`. `AddPublicDomainRes.dns` becomes `QueryDnsRes`.
- **`dns.rs`** — `query_dns` resolves both `A` and `AAAA` (`QueryDnsRes { ipv4, ipv6 }`, `AF_UNSPEC`).
- **`gateway.rs`** — `check_port` reports v6 reachability of the GUA (`CheckPortRes.ipv6` via a new `check_gua_port`; internal connect + PCP-mapped shortcut + echoip probed over v6). v6 is NAT-free, so there's no hairpinning.
- **web** — the add-domain flow verifies and displays both the `A`/IPv4 and `AAAA`/IPv6 records and reachability; new `getGua` helper mirrors the backend's `ipv6_is_local` rule (via a new `IpAddress.isGua()` on the shared IP util). Regenerated `osBindings`; mock updated; 2 new i18n keys across all 5 locales.
- **docs + changelog** — `clearnet.md` (A + AAAA records, v6 firewall-pinhole note) and the beta.10 changelog.

### Design decisions (confirmed with @dr-bonez)

- Full v4/v6 symmetry, but **don't** expose the bare GUA when SNI can disambiguate — SSL domains route by SNI at our vhost; only non-SSL (no SNI) exposes the bare GUA, like it exposes the WAN IPv4.
- No StartTunnel SNI-demux over v6 (each host has its own GUA).
- Include the v6 port-reachability probe (echoip backends support v6).

### Testing

- `cargo check -p start-core` clean; `cargo test -p start-core --features=test net::` — 169 passed.
- Web: `check:ui`, `check:i18n`, `check:shared` all pass; prettier clean.
- **Not exercised here** (no dual-stack network with a delegated GUA in this environment): the live v6 pinhole/echoip round-trip and end-to-end reachability of a real domain over IPv6. Worth a manual check on a box with a routed IPv6 prefix.


shared-libs/crates/start-core/src/net/net_controller.rs
```diff
@@ -779,6 +779,49 @@ impl NetServiceData {
...
+                        continue;
+                    };
+                    if crate::net::utils::ipv6_is_local(gua) {
+                        continue;
+                    }
+                    gua_gateways
+                        .entry(gua)
+                        .or_insert_with(|| v6_gateways.clone());
+                    if gua_pinholes.insert((gua, *external)) {
+                        ctrl.port_map
```
dr-bonez, line 818: I know this is pre-existing, but why doesn't ctrl.port_map.ensure not return an rc, like all the other controllers?

helix-nine: The rc-handle controllers (`forward`, `vhost`, `dns`) hand out an `Arc<()>` per logical entry and GC when `strong_count` hits 0 — that fits because each entry is a stable thing a single owner holds.

`port_map` is deliberately a layer *below* that: a thin keyed actor (`(local_ip, external_port, hostname)` → mapping) that `forward.rs` itself sits on top of — `InterfacePortForwardController` wraps port_map + nft and *is* the rc layer (`add_forward` → `Arc<()>`). Its other callers — the GUA pinholes and the 80→443 redirects — aren't per-entry owners; they're whole-graph reconcilers that recompute the entire desired set each pass and diff it against `binds.{gua_pinholes, gua_redirect_maps, redirect_maps}` to issue the `remove`s. A single mapping there can be wanted by several independent sources in one pass (e.g. this domain loop and the bare-GUA loop both want `(gua, 443)`, and the redirect is *derived* from the pinhole set), so "recompute set + diff" fits more naturally than threading one shared handle through all those derivation points.

So the refcount effectively *is* desired-set membership, held in `NetServiceData`, rather than a per-mapping `Arc`. Giving port_map an rc API (coalesce-by-key + refcount internally, which is exactly what forward.rs already layers on top) is a reasonable unification if you'd want the direct callers to drop their parallel desired-sets — happy to do it as a follow-up, but I left it alone here since it's pre-existing and cross-cutting.

dr-bonez (dismissed):

MattDHill (approved):

Ari4ka (approved):

PR #3426, opened 2026-07-07
fix(tunnel): manual SNI (hostname) forwards no longer show as Automatic

helix-nine:
## What

@drbonez reported: *manually-added forwards with hostnames, on the same host + port as automatic forwards with hostnames, also show as **Automatic**.*

## Root cause

The manual "add port forward" API (`tunnel/api.rs::add_forward`, SNI branch) and the PCP auto path **both** call the single `GatewayBackend::add_sni_forward` impl in `tunnel/forward/pcp.rs`, which hardcoded `auto: true` and label `"PCP"` on every SNI route it inserted. So a hand-added hostname forward was stored as if a device had created it via PCP:

- it appeared in the UI's **Automatic** table (labeled `PCP`) instead of **Manual**, and
- being an `auto` route, it got a startup lease and could be reaped by the PCP lease reaper when no device renewed it.

DNAT (v4) and pinhole (v6) manual paths were already correct — they set `auto: false` inline / `add_pinhole(.., false)`. SNI was the sole offender because it's the only forward kind whose manual and automatic paths share one function. The UI itself was fine (`port-forwards/utils.ts` maps `SniRoute.auto` 1:1); only the stored flag was wrong.

## Fix

- Extract an inherent `TunnelContext::persist_sni_forward(source, target, hostnames, lifetime, auto, label)`. The PCP trait method delegates with `(true, None)`; the manual API passes `(false, user_label)` — which also means a manual SNI forward now keeps the label you gave it instead of `"PCP"`.
- A pure helper `sni_route_fields(existing, auto, default_label)` decides a route's `(label, enabled, auto)`:
  - a **brand-new** route takes the caller's `auto` and default label;
  - an **existing** route preserves its own `auto` / enabled / label — so a PCP renewal can't hijack a hostname you added by hand, and a manual re-add can't flip an automatic route.

## Tests

- New unit tests for `sni_route_fields` covering: new-manual → not auto (the reported bug), new-PCP → auto, PCP-renewal-preserves-manual-owner, and manual-re-add-preserves-auto-owner.
- All 32 `tunnel::` unit tests pass (`cargo test -p start-core --lib tunnel::`).

## Housekeeping

- Moved the non-bindings tests in `tunnel/db.rs` to the bottom of the file (the `export_bindings_tunnel_db` test stays inline), per the "tests at the bottom except bindings exports" convention.
- CHANGELOG entry under `[Unreleased] → Fixed`.

Docs already describe the intended Manual/Automatic behavior, so no docs change — this restores documented behavior.

### DIFF
--- a/projects/start-tunnel/CHANGELOG.md
+++ b/projects/start-tunnel/CHANGELOG.md
@@ -21,10 +21,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **Automatic (PCP) mappings now honor their lease.** A forward, pinhole, or SNI route opened automatically by a connected device carries a finite lease that the device renews while it still wants the port. The tunnel now expires and tears down an automatic mapping whose device stops renewing it — because it went offline, rotated its key, or withdrew the exposure — instead of leaving a stale forward in place indefinitely. Manually-added forwards are unaffected and remain persistent.
 - **Admin actions retire a device's forwards immediately.** Deleting a device or demoting it to a client now clears all of its forwards, SNI routes, and IPv6 pinholes (previously a deleted device's v6 pinholes could linger); disabling **automatic port forwarding** for a device clears its automatic forwards while leaving any you added manually. Cleanup no longer waits for the lease to lapse.
 
-### Fixed
-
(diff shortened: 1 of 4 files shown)


MattDHill:
@helix-nine manual sni forwards were not in v1.0, so they can't be fixed in v1.1. This is a new feature for 1.1, remove your fix changelog entry

helix-nine:
Already removed — dropped it in e689b0a5 (pushed just after your comment, same conclusion drbonez reached). The PR no longer touches `CHANGELOG.md` at all; the manual-SNI capability stays documented as a 1.1 feature and there's no "Fixed" bullet. Remaining diff is just the code fix (`tunnel/forward/pcp.rs` + `api.rs`) and the test move in `tunnel/db.rs`.

dr-bonez (approved):

PR #3427, opened 2026-07-07
feat(tunnel): per-IPv4 HTTP→HTTPS redirect on port 80 (default on)

helix-nine:
## What

Adds an **HTTP→HTTPS redirect on port 80** to StartTunnel. A plain `http://` request to an exposed service is answered with a redirect to the same host over `https://` instead of a connection error.

Per the discussion, this is modeled as a **checkbox per public IPv4, default on, port 80 only** (not an arbitrary user-selected SocketAddr):

- **On by default** on every public IPv4 the tunnel holds — on a fresh install *and* after an upgrade. This is a live rule (`HttpRedirects.disabled` is a set of opted-out IPv4s that defaults empty), so no data migration or one-time seed is needed to satisfy "run one on all ipv4 port 80s by default (on bootstrap and migration)".
- **Reuses the existing OS redirect handler** — each address gets a userspace TCP listener that hands every connection to `net::http::handle_http_on_https` (the same 307→`https://<host><uri>` redirect the OS serves on its own TLS ports).
- **Reactive** — the listener set is reconciled from the db on every revision (mirrors the reactive `/webserver` HTTPS listener in `bins/tunnel.rs`), so a user toggle or a port-forward change takes effect with no restart.
- **Mutually exclusive with a port forward** — the redirect *yields* to any forward occupying port 80 on that IP (`PortForwards::occupied`), manual or automatic (PCP/UPnP). Forward the port and the redirect stops; remove the forward and it resumes (unless the user turned it off). Public IPv4 only — matches the web UI's `IpNet.isPublic` (skips loopback + RFC1918).

## Surface

- **UI**: an "HTTP Redirects" card on the Port Forwards page — one checkbox per public IPv4 (greyed while a port-80 forward occupies the address).
- **CLI**: `start-tunnel http-redirect list` and `http-redirect set-enabled <ip> [--enabled]`.
- **db**: new `httpRedirects: { disabled: string[] }` (additive, `#[serde(default)]` — no migration).

## Where

- `shared-libs/crates/start-core/src/tunnel/redirect.rs` — new: `public_ipv4s`, `reconcile`, per-address accept loop.
- `.../tunnel/db.rs` — `HttpRedirects` + `PortForwards::occupied` (+ unit test).
- `.../tunnel/api.rs` — `http-redirect` command group + handlers.
- `.../bins/tunnel.rs` — reactive reconcile thread subscribed to the db root.
- `projects/start-tunnel/web/...` — UI card + API service method + mock.
- Docs (`docs/src/http-redirects.md`, cli-reference, port-forwarding xref), CHANGELOG (rolled into the unreleased **1.1.0**), and regenerated manpages + TS bindings.

## Testing

- `cargo check -p start-core` — clean.
- `cargo test -p start-core --lib tunnel::db::` — 6/6 pass (incl. new `port_forward_occupied_gates_http_redirect`).
- Manpage export test + `make start-core-ts-bindings` regenerated (unrelated pre-existing binding/manpage drift reverted to keep the diff focused).
- `npm run check:tunnel` (UI typecheck) — clean; prettier clean.
- Not exercised locally: live install/boot behavior on a VPS (binding `:80`, actual redirect round-trip) — covered by CI + manual QA.


MattDHill (dismissed):

shared-libs/crates/start-core/src/tunnel/api.rs
```diff
@@ -1675,6 +1717,81 @@ pub async fn set_pinhole_enabled(
...
+                eyre!(
+                    "port 80 on {ip} is forwarded; delete that port forward before enabling the HTTP→HTTPS redirect"
+                ),
+                ErrorKind::InvalidRequest,
+            ));
+        }
+    }
+    ctx.db
+        .mutate(|db| {
+            let mut redirects = db.as_http_redirects().de()?;
```
dr-bonez: Use a .mutate to simplify the de -> ser dance

shared-libs/crates/start-core/src/tunnel/db.rs
dr-bonez: You don't need to iterate the entire set, just look at the port and the one immediately before it ordinally

shared-libs/crates/start-core/src/tunnel/db.rs
```diff
@@ -316,6 +333,17 @@ impl PortForwards {
             (new_lo <= hi && lo <= new_hi).then_some(*src)
         })
     }
+
+    /// Whether any forward on `addr`'s IP has a port span covering `addr.port()`.
+    /// Used to keep the port-80 HTTP redirect mutually exclusive with forwards:
+    /// the redirect yields when a forward already occupies the port.
+    pub fn occupied(&self, addr: SocketAddrV4) -> bool {
+        self.0.iter().any(|(src, pf)| {
```
dr-bonez: Same here

shared-libs/crates/start-core/src/tunnel/api.rs
helix-nine: Done — `set_http_redirect_enabled` now uses `db.as_http_redirects_mut().mutate(|r| …)` instead of the de → modify → ser. (ae7d35542)

shared-libs/crates/start-core/src/tunnel/db.rs
helix-nine: Done. `overlapping` now ranges up to the top of the new span and walks back to the nearest non-`source` entry (`range(..=(ip, new_hi)).rev().take_while(same ip).find(!= source)`), relying on the invariant that entries on one IP never overlap — so only that nearest entry can reach into the range. Did the same to `Pinholes6::overlapping` for parity. Existing overlap/range tests still pass. (ae7d35542)

shared-libs/crates/start-core/src/tunnel/db.rs
helix-nine: Done — `occupied` is now just `range(..=addr).next_back()` and a span check on that one entry. (ae7d35542)

dr-bonez (approved):

PR #3429, opened 2026-07-07
fix(tunnel): rebind device-bound listeners after wg interface reload

dr-bonez:
## What

After any tunnel config change that reloads the WireGuard interface (`add-subnet`, `remove-subnet`, `add-device`, `remove-device`, WAN reassignment — everything that calls `sync_network()`), the PCP server, SSDP responder, and IGD control server went permanently silent until the daemon was restarted.

## Why

`sync_network()` → `WgServer::sync()` runs `wg-quick down` + `wg-quick up`, which **deletes and recreates** `wg-start-tunnel`. The PCP/SSDP/IGD sockets are `SO_BINDTODEVICE`-bound, and that option pins the interface **index** at bind time — after the bounce they reference a dead ifindex and the kernel silently drops every inbound packet. The serve loops' self-restart never fires because a stale device binding doesn't error; it just never receives.

Observed in production on `tunnel-chad` (2026-07-07): an admin device change at 17:09:56 recreated the interface (ifindex 4 → 5); `ss` showed the listeners bound to `%if4` (a nonexistent interface); clients retried PCP MAP + SSDP M-SEARCH unanswered for hours; every auto forward, SNI route, and v6 pinhole lapsed ~1h later when the unanswered renewals hit their lease expiry, taking the exposed services offline.

The same bounce also broke the per-subnet DNS proxy resync with `Address in use`: `wait_for_abort()` on the proxy task only *schedules* the abort of hickory's internal socket tasks (dropping the `Server` drops its `JoinSet`, which aborts without awaiting), so the old `:53` sockets could still be open when `sync()` immediately rebound the same address — leaving the *old* proxy (with stale upstream config) serving.

## How

- `TunnelContextSeed` gains `forward_rebind: Watch<()>`; `sync_network()` bumps it right after `server.sync()`.
- The PCP v4/v6, SSDP, and IGD-control serve loops `select!` between their socket and `forward_rebind.changed()`, and rebind immediately (no 5s error backoff) when signaled. Each loop subscribes (`mark_seen`) **before** binding, so a bounce that lands mid-setup still triggers a rebind rather than being lost.
- `DnsProxyController` now keeps each proxy's hickory `CancellationToken` and shuts down via `shutdown_gracefully` semantics (cancel + await, 5s timeout) so the sockets are provably closed before rebinding.

## Verification

- Root cause and fix mechanism verified live on the affected production server: stale `%if4` bindings confirmed with `ss`/`tcpdump` (requests arriving, zero replies); after a daemon restart rebound the sockets by name, client PCP requests were answered again within seconds.
- `cargo check -p start-core` / `-p start-tunnel` clean; `cargo test -p start-core --features=test tunnel::` — 32 passed.

Changelog entry added under 1.1.0 (unreleased).

## Note (not in this PR)

`cargo test … export_manpage_start_tunnel` regenerates `projects/start-tunnel/man/start-tunnel-device-show-config.1` with the `WAN_ADDR` positional removed — the committed manpage is stale relative to the current CLI on master. Left out of this PR since it's unrelated drift.

### DIFF
--- a/projects/start-tunnel/CHANGELOG.md
+++ b/projects/start-tunnel/CHANGELOG.md
@@ -66,6 +66,7 @@ StartOS provides the client side — see
 
 ### Fixed
 
+- **Automatic port forwarding no longer goes silent after a network config change.** Adding or removing a device or subnet (or any other change that reloads the WireGuard interface) recreated the interface out from under the PCP and UPnP IGD listeners, which are bound to it — leaving them unreachable until the daemon was restarted. Connected devices' renewals then went unanswered, so their automatic forwards, SNI routes, and IPv6 pinholes expired and their exposed services dropped offline. The listeners now rebind whenever the interface is reloaded. The same reload could also leave a subnet's DNS proxy serving a stale upstream configuration ("Address in use" in the logs); proxies now shut down cleanly before rebinding.
 - **Forwarded TCP MSS is clamped to the WireGuard path MTU (#3262, #3261).** A `mangle FORWARD … TCPMSS --clamp-mss-to-pmtu` rule fixes hung TLS handshakes where a large ClientHello (e.g. desktop Firefox/Chromium with the X25519MLKEM768 post-quantum key share) split into a segment that exceeded the tunnel's effective path MTU after WireGuard encapsulation and was silently dropped.
 - **"Default" DNS mode falls back to `/etc/resolv.conf`** when systemd-resolved is absent, unreadable, or yields no usable servers (84e99f1e9).
 - **Gateway type is auto-detected from the StartTunnel marker** when a pasted config declares no type, so outbound-only configs (e.g. Mullvad) are no longer mislabeled as inbound-capable (f76ac4bfe).
--- a/projects/start-tunnel/CHANGELOG.md
+++ b/projects/start-tunnel/CHANGELOG.md
@@ -66,7 +66,6 @@ StartOS provides the client side — see
 
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


shared-libs/crates/start-core/src/tunnel/context.rs
```diff
@@ -174,6 +174,10 @@ pub struct TunnelContextSeed {
     /// Wakes the lease reaper when a stamp may have moved the soonest expiry
     /// earlier, so it can pull its next wake-up forward.
     pub lease_wake: tokio::sync::Notify,
+    /// Bumped after every `wg-quick` bounce: the PCP/IGD listeners are
+    /// `SO_BINDTODEVICE`-bound, which pins the ifindex at bind time, so a
+    /// recreated interface leaves them deaf until they rebind.
+    pub forward_rebind: Watch<()>,
```
dr-bonez: A watch<()> is really just a Notify but with more complexity

projects/start-tunnel/CHANGELOG.md
```diff
@@ -66,6 +66,7 @@ StartOS provides the client side — see
 
 ### Fixed
 
+- **Automatic port forwarding no longer goes silent after a network config change.** Adding or removing a device or subnet (or any other change that reloads the WireGuard interface) recreated the interface out from under the PCP and UPnP IGD listeners, which are bound to it — leaving them unreachable until the daemon was restarted. Connected devices' renewals then went unanswered, so their automatic forwards, SNI routes, and IPv6 pinholes expired and their exposed services dropped offline. The listeners now rebind whenever the interface is reloaded. The same reload could also leave a subnet's DNS proxy serving a stale upstream configuration ("Address in use" in the logs); proxies now shut down cleanly before rebinding.
```
dr-bonez: No need to include a fix for an unreleased feature in the changelog

dr-bonez (commented):
@helix-nine please pick up these fixes

shared-libs/crates/start-core/src/tunnel/context.rs
helix-nine: Done in acca4fad2 — `forward_rebind` is now a `tokio::sync::Notify`. Each listener `enable()`s a `Notified` before binding (keeping the subscribe-before-bind guarantee, so a bounce mid-setup still triggers a rebind), and `sync_network` wakes them all with `notify_waiters()` after the `wg-quick` bounce.

projects/start-tunnel/CHANGELOG.md
helix-nine: Removed in acca4fad2.

PR #3430, opened 2026-07-07
fix(start-cli): allow nested workspaces; refuse init-workspace only inside a package repo

helix-nine:
## Problem

`start-cli s9pk init-workspace` failed for anyone with a `.startos` above the target directory — most commonly a leftover global `~/.startos` from a previous start-cli. The nested-workspace guard walked up, matched it, and refused with *"Cannot create a workspace inside an existing one"* anywhere under it, locking community developers out with a confusing error.

## Approach

Nested workspaces are the **intended model**, not an error. `init-workspace` now just creates the workspace and pays no attention to outer ones.

- **Nesting allowed.** No nested-workspace guard, no home-directory guard, no migration prompts.
- **Walk-up resolution (already how it works).** When building, signing, or reading `host`/`registry` config, start-cli walks up from the cwd and uses the nearest `.startos/` — so an inner workspace transparently overrides an outer one. Conceptually a deep merge of every `.startos/` on the path, innermost first; in practice just a walk, no merge performed. (`build_key()` and `WorkspaceConfig::find()` already did exactly this.)
- **One guard kept:** refuse to initialize **inside a package repo** — a workspace *holds* package repos, it isn't one. Detected via a `package.json` depending on `@start9labs/start-sdk` or the scaffolded `startos/` layout; the walk stops at the nearest workspace marker so an enclosing workspace never trips it.
- **No automatic migration** from an older global `~/.startos`. Users copy `developer.key.pem` → a workspace's `.startos/build-key` themselves if they want to reuse a signing key (documented).

`find_workspace_root` is now marker-aware (a `.startos/` counts as a workspace only once it has a `build-key` or a schema-tagged config), so `init-package` targets the nearest *real* workspace and a bare/legacy `.startos` is ignored.

## Docs / changelog / i18n / tests

- `environment-setup.md`: replaced the migration section with "Nested workspaces and config resolution"; kept the corrected note that (as of SDK 2.0) `make install/publish` resolve host/registry via `start-cli`, not by parsing `~/.startos/config.yaml`.
- `start-cli/CHANGELOG.md` under `[Unreleased]`.
- Removed the now-unused migration/nested i18n keys; `s9pk.init.in-package-repo` present in all 5 locales.
- 4 unit tests: marker detection, package-repo detection, package-repo walk-up stops at a workspace marker, and the scaffolded config parses as a full `WorkspaceConfig`.

## Testing

- `cargo check -p start-core` ✓
- `cargo test -p start-core --features=test s9pk::init::test` ✓ (4 passed)
- `make start-core-format-check` ✓

Not exercised here (needs a built binary): creating a real nested workspace and confirming build/sign resolve to the nearest `.startos/` on a live tree.


shared-libs/crates/start-core/src/s9pk/git_hash.rs
```diff
@@ -12,15 +12,21 @@ use crate::util::Invoke;
...
+    /// package build before its first commit; it populates once one exists.
+    pub async fn from_path(path: impl AsRef<Path>) -> Result<Option<GitHash>, Error> {
+        let Ok(rev) = Command::new("git")
+            .arg("rev-parse")
+            .arg("HEAD")
+            .current_dir(&path)
+            .invoke(ErrorKind::Git)
+            .await
+        else {
+            return Ok(None);
```
dr-bonez, line 27: Maybe still warn? Otherwise you might never notice if git is misconfigured

helix-nine: Good call — done in ca30480b3. `s9pk pack` now warns (to stderr, so it shows during a build) when it can't read a commit hash:

> No git commit found in `<path>` — building without a commit hash in the manifest. Commit your work (or check that git is set up) to include it.

I kept `GitHash::from_path` pure (still returns `None`) and put the warning in the pack command itself, so a misconfigured/missing git is visible, and for a fresh scaffold it doubles as a nudge to make the first commit.

MattDHill (approved):

PR #3432, opened 2026-07-07
fix(install): keep UploadingFileReader wakers registered across polls

dr-bonez:
Follow-up to #3431, whose test suite hangs in CI (master's "Automated Tests" run is currently stuck on it).

## Root cause

`UploadingFileReader::poll_ready{,_for}` created a fresh `watch::Receiver::wait_for` future, polled it **once**, and dropped it. Dropping the future deregisters its waker, so a progress update that lands between polls is lost and the reader's task is never woken — it pends forever.

The bug is latent, not new: production reads gate through `UploadingFile::fetch` (a genuine `.await`, safe), so only `fetch_all` on a still-downloading file reaches the broken poll path. Before #3217 nothing hit it; #3217's inline download guaranteed completion before any read; #3431 restored the background download and its tests read via `fetch_all` mid-download — hanging the suite (locally reproduced: download task completes and records its outcome, reader never wakes).

## Fix

- Store the wait future in the reader across polls (keyed by wait target: `Complete` / `SizeAtLeast` / `DataAt`), so the waker stays registered while pending; drop it only once it resolves.
- Gate `poll_read` on `DataAt(position)` — "data available at my position, or the transfer finished" — instead of `ready_for(position + buf.remaining())`, which mid-download either pended (with the lost waker) or spuriously errored with "file size is less than requested" whenever the destination buffer overshot the file size. Reads now stream as bytes arrive; EOF surfaces naturally once complete. Seeks keep the old `ready_for` semantics.

## Verification

Previously-hanging combinations now pass (were reproducible hangs locally, >30min):
- full `registry::asset` suite serial + parallel: 6 passed in ~5s
- cross-module pair (`net::acme` + asset): passes
- `s9pk::merkle_archive` suite: passes

Full local test-binary sweep and CI will confirm the rest of the crate.

### DIFF
--- a/shared-libs/crates/start-core/src/registry/asset.rs
+++ b/shared-libs/crates/start-core/src/registry/asset.rs
@@ -203,7 +203,7 @@ impl BufferedHttpSource {
         progress: PhaseProgressTrackerHandle,
     ) -> Result<Self, Error> {
         let (handle, file) = UploadingFile::new_for_download(progress).await?;
-        Self::spawn_mirror_download(handle, file, urls, client).await
+        Ok(Self::spawn_mirror_download(handle, file, urls, client))
     }
     async fn from_urls_with_path(
         path: impl AsRef<Path>,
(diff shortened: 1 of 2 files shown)


dr-bonez:
Pushed 41cebf2ca addressing the review's confirmed findings:

- **Truncation race (the CI failure):** a full mirror retry `set_len(0)`s the file, and if that lands between the DataAt gate resolving and the file read, `poll_read` saw a shortened file and returned a false EOF mid-transfer. EOF is now only reported when progress shows a clean `complete` with nothing left past the reader's position; otherwise the read self-wakes and retries (yielding so the writer can catch up). Verified: the two full-retry tests failed ~1/60 runs before, **0/200 after**; full crate suite 464/464.
- **Per-chunk allocation:** `poll_wait` now checks the predicate synchronously via `borrow()` before allocating — the data-already-available case (the common one) no longer boxes a future + clones the receiver per chunk.
- **Gate-logic triplication:** `ready`/`ready_for`/`data_at` collapsed into one `Progress::check` (sync predicate + terminal result) + `Progress::wait` (the awaited form); the fast path and the wait future share the same predicate, so they can't drift.
- **Zero-capacity reads** return `Ready(Ok)` immediately again (matches AsyncRead convention; old pre-PR behavior).

Not addressed here (latent, pre-existing, no in-repo trigger — happy to file follow-ups): negative `SeekFrom::End/Current` wrap producing a misleading error; `SeekFrom::End` resolving against the growing physical file instead of the gated absolute target; stale `to_seek` retained when a gated seek errors; O_DIRECT-write/buffered-read fd mixing on uploads; no test coverage for the seek paths.

PR #3439, opened 2026-07-08
fix(backup): stop enumerating legacy backup folder size on target list

helix-nine:
## What

Listing backup targets walked the **entire** pre-v2 `StartOSBackups` folder (`dir_size`) to report its size. That directory scan is too expensive to run on every target list, so this drops it.

## Backend

- `LegacyBackupInfo` drops `size`; keeps `available` (the target's free space, from a cheap `df`).
- `legacy_backup_info` no longer walks the folder — it reuses the lightweight `has_legacy_backup` existence check and reads free space only. Regenerated the `LegacyBackupInfo` ts-rs binding to match.

## UI

Because the old folder's size is now unknown, the create-backup warning can no longer compute whether a new backup will *fit*. The legacy-backup modal is reworked:

- Removes the "Not enough free space" refusal branch (StartOS can't determine this anymore).
- Reinforces the messaging that the old folder's size is not measured, **shows the target's free space** (`convertBytes(available)`), and lets the user proceed and decide for themselves.
- Drops the now-dead i18n strings and adds the new ones across all five locales (en/de/es/fr/pl).

Docs (`backup-create.md`) and the beta.10 changelog entry are updated to match.

## Verification

- `cargo check -p start-core` ✅
- `make start-core-ts-bindings` → only `LegacyBackupInfo.ts` changes ✅
- `check:i18n` ✅ · `check:ui` ✅ · `check:shared` ✅
- prettier `--check` on all changed frontend/i18n files ✅


projects/start-os/web/ui/src/app/routes/portal/routes/system/routes/backups/legacy.component.ts
```diff
@@ -26,66 +26,42 @@ export interface LegacyBackupData {
...
-              'The old "StartOSBackups" folder is larger than the free space remaining on this drive, so a new backup will not fit. You have two options:'
-                | i18n
-            }}
-          </span>
+    <div tuiNotification appearance="warning">
+      <span tuiTitle>
+        <strong>{{ 'Check your available space' | i18n }}</strong>
+        <span tuiSubtitle>
+          {{
+            'StartOS no longer measures the size of the old "StartOSBackups" folder, so it cannot tell you in advance whether a new backup will fit. Keep an eye on the free space below when choosing what to back up.'
```
MattDHill: In multiple places you reference that we don't or no longer measure the V1 backup size. Beta.10 was never released, and we never measured before beta 10. Telling users that we no longer do something that we never did is silly. These user facing docs and notifications read like notes to yourself rather than proper guidance and info for users. Fix it.

shared-libs/ts-modules/start-core/lib/osBindings/LegacyBackupInfo.ts
```diff
@@ -1,7 +1,8 @@
 // This file was generated by [ts-rs](https://github.com/Aleph-Alpha/ts-rs). Do not edit this file manually.
 
 /**
- * A pre-V2 `StartOSBackups` folder found on a backup target. Surfaced so the
- * UI can warn (or refuse, when it wouldn't fit) before a new V2 backup runs.
+ * A pre-V2 `StartOSBackups` folder on a backup target. Carries the target's
```
MattDHill: Seriously, how many times do you need to remind yourself, both in user-facing copy and code comments, that we don't do something? Shoukd we also remind ourselves and our users that we dont sell hamburgers?

MattDHill (changes requested):

projects/start-os/web/ui/src/app/routes/portal/routes/system/routes/backups/legacy.component.ts
helix-nine: Fixed. Dropped the "no longer / doesn't measure" framing everywhere it appeared — the modal, the docs warning, the changelog, and the code + doc comments. Since beta.10 never shipped there's nothing to say we "used to" do.

The copy is now plain guidance: the modal shows the target's free space and says "Make sure the services you select will fit within the free space remaining on this target." The docs/changelog do the same — show free space, and point at Delete old backup / another drive if it's tight.

shared-libs/ts-modules/start-core/lib/osBindings/LegacyBackupInfo.ts
helix-nine: Removed. The struct/binding doc comment no longer editorializes about what isn't measured — it just states what the type is: a present V1 folder plus the target's free space. Same trim applied to the `has_legacy_backup` comment.

helix-nine:
Reworked the copy per review — pushed. Removed the "no longer / don't measure" self-referential framing from all of it (modal, docs warning, changelog, and the Rust doc + binding comments) and replaced it with plain user-facing guidance built around the free-space figure. Behavior is unchanged; this is a copy/comment pass. Green locally: `cargo check -p start-core`, `check:i18n`, `check:ui`, `check:shared`, prettier.

helix-nine:
Added per @drbonez: free space is now shown for **every** backup target, not only ones with a legacy backup.

- Free space is a target-level field now — `available` on `PartitionInfo` and `CifsBackupTarget` (cheap `df` on the mounted target) — rendered as a new **Free** column in both the Physical Drives and Network Folders tables (shows `—` when the target isn't mountable / space is unknown).
- Since free space no longer lives inside the legacy info, `LegacyBackupInfo` had no data left, so `legacyBackup` is now a plain `boolean` and the type is gone. The legacy modal reads free space from the target entry.

Green locally: `cargo check -p start-core`, bindings regen (PartitionInfo/CifsBackupTarget/BackupTarget updated, LegacyBackupInfo removed), `check:i18n`, `check:ui`, `check:shared`, `check:setup`, prettier.

Note on presentation: I put free space in a new column in both target tables — if you'd prefer it inline (e.g. "X free of Y" in the capacity cell) or elsewhere, easy to change.

dr-bonez (dismissed):

helix-nine:
Per @dr-bonez: legacy detection **and** deletion are now scoped to the current server's ID.

The V1 folder is keyed by server ID (`StartOSBackups/<server_id>`), same as `StartOSBackupsV2` — confirmed by the format-v2 refactor, which renamed `join("StartOSBackups").join(server_id)` → `join(BACKUP_DIR_NAME).join(server_id)`. So a shared target can hold several servers' legacy backups; the old code checked/deleted the whole folder and would flag or destroy another server's backup.

Now:
- `has_legacy_backup(mountpoint, server_id)` checks `StartOSBackups/<server_id>` only.
- `delete_legacy` removes just `StartOSBackups/<server_id>` — other servers' backups and `StartOSBackupsV2` are untouched.
- `disk::util::list` threads the server ID through `part_info`, passing `None` for the install/setup/disk-management callers (no server yet → `legacyBackup: false`); the backup-target list, cifs add/update/list, and the post-backup notification all pass the current server's ID.

Backend-only (the `legacyBackup` bool wire type is unchanged, so no binding/UI/i18n churn). Green locally: `cargo check -p start-core`, `cargo +nightly fmt --check`. Docs + changelog updated to say "this server's" backup.

helix-nine:
Per @dr-bonez: the **Delete old backup** button now shows whenever this server has a V1 backup on the target — dropped the "a current backup already exists" gate that was hiding it in exactly the case cleanup matters most.

And when there's **no** current backup, deleting is guarded by an **extra confirmation**: the normal "Delete old backup?" prompt, then a second "This target has no other backup" prompt warning that this server will be left with no backup there ("Delete anyway").

Two design calls worth flagging:
- **"Current backup" is judged per-server** — `hasThisBackup(entry, serverId)` against this server's `StartOSBackupsV2/<server-id>`, not `hasAnyBackup` (any server on the target). This matches the per-server scoping from the last commit: if another server has a V2 backup on a shared target but *this* server doesn't, you still get the extra confirmation. The service reads the current server id from patch-db.
- **Extra confirmation = a genuine second dialog**, not just escalated wording, since deleting your only backup is irreversible. Easy to make it a single stronger dialog instead if you'd prefer.

Frontend + i18n only (new strings across all 5 locales). Green locally: `check:i18n`, `check:ui`, `check:shared`, prettier.

helix-nine:
Per @dr-bonez: a backup phase whose progress is `false` (running, but not reporting a fraction) now shows **Backing Up** with the loader instead of `0%`. Added an `@else if (leaf === false)` branch in the phase list, reusing the existing lowercase `backing up` key so the host's `text-transform: capitalize` renders it consistently with the sibling `complete`/`waiting` labels.

Scoped to the per-phase rows as asked. The overall-progress header still shows `0%` when its own leaf is `false` — say the word if you want the same treatment there.

Green: `check:i18n`, `check:ui`, prettier.

MattDHill (approved):

PR #3442, opened 2026-07-08
fix(startos): keep lxc-net + mounts up through the graceful teardown on host reboot/poweroff

helix-nine:
## What

On an OS-level `reboot`/`poweroff`, service containers were torn down uncleanly — a burst of `lxc-stop … is not running` / `Connection refused (os error 111)` and a stalled shutdown — even though the graceful-teardown hooks were meant to make it behave like `start-cli server restart`.

## Root cause (corrected after VM testing)

The teardown runs in the `ExecStop` of `startos-restart.service` / `startos-shutdown.service` (`start-cli server {restart,shutdown}` → `startd shutdown_all()` → per-container `Exit` RPC + `lxc-stop`). That hook was ordered only against the power targets, so systemd's shutdown transaction tore down the container's supporting infrastructure **concurrently** with it. The load-bearing culprit is **`lxc.service`**, whose `ExecStop` is Debian's `lxc-containers stop` — it stops the container — plus **`lxc-monitord.service`**. They run at the same time as the graceful hook, so the container is already dead by the time `startd`'s `Exit` RPC reaches it (`"ungracefully dropped"` → `lxc-stop … is not running` → `111`), and the hook then hangs to `TimeoutStopSec`.

`start-cli server restart` on a running system never enters a shutdown transaction, so nothing races the teardown — which is why the manual path was always clean.

## Fix

**Order the hooks after the LXC infrastructure + the data the teardown needs.** `startos-{restart,shutdown}.service` now order:
```
After=startd.service lxc.service lxc-monitord.service lxc-net.service \
      media-startos-data-main.mount media-startos-data-package\x2ddata.mount
```
A unit `After=X` stops *before* X on shutdown, so each of these stays up until the hook's `ExecStop` graceful teardown has finished:
- `startd.service` — startd must be alive to service the RPC
- `lxc.service` / `lxc-monitord.service` — the container-stop infrastructure (this is the fix for the `111`)
- `lxc-net.service` — keep `lxcbr0` up for container networking
- the `/media/startos/data` mounts — hold `embassy.db` and the volume sources

**`Shutdown::execute` no longer force-reboots mid-transaction.** When systemd is already stopping (`systemctl is-system-running` == `stopping`), it leaves the power action to systemd, and now **exits** once the graceful teardown is done — otherwise startd lingered until systemd's SIGTERM and waited out its ~90s stop timeout.

**Raspberry Pi:** the `systemd_is_stopping` skip applies to the *restart* sub-case; *poweroff* still self-drives the standby marker + reboot (rpi has no true power-off).

An earlier revision of this PR used per-container transient "anchor" units to hold the per-guid `.mount` units in the shutdown transaction. VM testing showed those mounts are torn down *by* `lxc-containers stop` (not systemd's `umount.target` wave), so the anchors targeted the wrong vector — **removed** in favor of the `lxc.service`/`lxc-monitord` ordering above.

## VM validation (beta.10, systemd 257)

Deployed this build to a real StartOS beta.10 VM, installed a real service (`start9-pages` from the registry), and rebooted, capturing the shutdown on the serial console:

- **Before the `lxc.service` ordering:** reproduced the bug exactly — `Container … was ungracefully dropped`, `lxc-stop … is not running`, hook hangs 120s → SIGKILL.
- **With this fix:** hook `ExecStop` completes in **<1 s**, `lxc-monitord`/`lxc.service` stop *after* it, and there are **no `111` / lxc-stop errors** — the container is stopped gracefully by startd. With the `Shutdown::execute` exit, startd no longer holds the shutdown for its 90s stop timeout either.


helix-nine:
⚠️ Marking this **draft** — @dr-bonez tested it and it makes things *worse* (the shutdown no longer waits; it force-reboots and SIGTERMs everything). I reproduced the regression on a beta.10 VM with the **real** `ExecStop` (my earlier validation used a `sleep` stub, so it never exercised the self-reboot path — my mistake).

**Confirmed mechanism.** The hook's `ExecStop` is `start-cli server restart`, which makes `startd` run `Shutdown::execute()` → its own `reboot` (the unit comment even notes "startd then re-issues the matching poweroff … idempotent since systemd is already powering off"). That's only idempotent when the hook stops **late**. `After=lxc-net.service` moves the hook to stop **first**, so the self-reboot now fires mid-shutdown:

```
OLD (orderly):
  Stopping startos-restart.service → teardown → Stopping lxc-net.service
  → Stopping systemd-journald → systemd-shutdown: SIGTERM → SIGKILL → Rebooting   (~+3.4s)

NEWPR (this PR, short-circuited):
  Stopping startos-restart.service → TEARDOWN_BEGIN
  → Stopped startos-restart.service (no return mark)
  → Stopping lxc-net.service
  → "Requested transaction contradicts existing jobs: reboot.target has 'start' job
     queued, but 'stop' is included in transaction" → abrupt reboot (~+0.6s)
```

**The tension is fundamental to a unit-file-only fix.** To keep `lxcbr0` up through the teardown, the teardown must stop *before* `lxc-net` (`After=lxc-net.service`) — but that same early stop fires `startd`'s self-`reboot` early, which short-circuits systemd. Ordering is symmetric, so there's no unit-only arrangement that keeps `lxc-net` up *without* moving the teardown early. **The self-reboot has to be decoupled from the teardown.**

Proposed direction (in `shared-libs/crates/start-core/src/shutdown.rs`): when `Shutdown::execute()` is invoked as part of a systemd-initiated shutdown, do the graceful work (stop journald, export the data volume, sound) but **skip the final `reboot`/`shutdown now`** and let systemd complete the transition — detectable via `systemctl is-system-running` == `stopping` (or the presence of the systemd shutdown marker). Then `After=lxc-net.service` + `Before=shutdown.target` on the hooks becomes safe. Alternatives: route the host-reboot teardown through `startd`'s SIGTERM/`None` path (which already tears down without rebooting) by ordering `startd.service After=lxc-net.service`, or add a teardown-only RPC the hook calls.

Since this now lands in `startd`'s shutdown path, I'd rather align before pushing a third attempt.

helix-nine:
Implemented **option 1** (thanks @dr-bonez). New commit: `Shutdown::execute` skips its self-issued `reboot`/`shutdown` when systemd is already driving the transition, detected via `systemctl is-system-running` == `stopping`. The manual `start-cli server restart` path (system `running`) still issues its own reboot; `raspberrypi` keeps its existing branch.

```rust
} else if systemd_is_stopping() {
    // Reached from the startos-{restart,shutdown}.service ExecStop during a
    // systemd-initiated reboot/poweroff: systemd already owns the power action.
    // Re-issuing it here mid-transaction force-reboots and skips the orderly
    // teardown, so leave the final step to systemd.
    tracing::info!("systemd is already shutting down; leaving the final reboot/poweroff to it");
}
```

**Verified on a beta.10 VM (serial console):**
- `is-system-running` returns **`stopping`** at the exact moment the hook's `ExecStop` runs → the skip fires on the systemd path but not the manual path. ✅
- Earlier runs confirmed the ordering half: with `After=lxc-net.service`, `lxc-net` stays up until **after** the teardown completes. ✅
- `cargo check -p start-core` passes.

**One gap I couldn't close here:** a full end-to-end reboot with the *rebuilt* `startd` — `startbox` embeds the built web UI, so it won't compile without a full Angular build in this environment. Each half of the fix is verified independently and the logic joining them is straightforward (no self-reboot command → no transaction collision → systemd's orderly shutdown proceeds), but a smoke test on a real box with running services (`journalctl -b -1`: clean teardown, no `111`/`is not running`, and it *waits*) is worth doing before merge since you have the hardware.

dr-bonez (approved):

Ari4ka (approved):

PR #3446, opened 2026-07-08
fix(sdk): don't destroy a daemon's subcontainer on a dependency pause

helix-nine:
## Problem

Since start-sdk **2.0.0**'s hold/release refactor, `Daemon.term()` unconditionally destroyed the daemon's `SubContainer`, and `HealthDaemon` calls `term()` **every time a dependency goes not-ready** (a *pause*, not a teardown). When the dependency's health recovered, `HealthDaemon.changeRunning(true)` called the daemon's `start()`, whose `hold()` on the already-destroyed subcontainer threw `cannot hold subcontainer …: already destroyed`.

Because that `start()` was **un-awaited**, the throw surfaced as an unhandled promise rejection and the daemon never recovered — the health check reported "not ready" / "daemon crashed" indefinitely, curable only by a full package restart. Any service with a dependency-gated daemon was exposed (e.g. `c-lightning`'s and BTCPay's web UIs).

The 2.0.0 refactor had removed the old `destroySubcontainer` flag that kept the subcontainer alive across a stop, collapsing "pause" and "teardown" into one destroying `term()`.

## Fix

1. **Restore the pause/teardown distinction in `Daemon`.** Split the destroying `term()` into:
   - a non-destroying **`stop()`** — aborts the restart loop, terminates the process, releases the `SubContainer` hold, but leaves the container intact so a later `start()` re-holds it (a restartable pause);
   - **`term()`** — now `stop()` + destroy the `SubContainer` (genuine teardown, one-way).
2. **`HealthDaemon` pauses with `stop()`.** `changeRunning(false)` (a dependency going not-ready) now calls `stop()` instead of `term()`. `HealthDaemon.term()` / `Daemons.term()` still route to the destroying `term()` for real teardown.
3. **Await + catch `start()`.** `changeRunning(true)` now awaits and catches `start()`, so a start failure becomes a health failure instead of a silent unhandled-rejection wedge.
4. **Serialize pause/resume transitions.** `updateStatus()` runs fire-and-forget from dependency watchers, so a fast readiness flap (not-ready → ready before the pause's `stop()` finished draining) could run the resume's `start()` while the previous run loop was still winding down. Because `Daemon.start()` is a no-op while `this.loop` is set, that `start()` would silently do nothing and strand the daemon *stopped-but-believed-running* until the next flap — the same wedge class. Transitions now apply on an in-order promise chain, so a resume always waits for the in-flight pause to finish before re-holding and restarting.

Bumps `@start9labs/start-sdk` to **2.0.3** (+ CHANGELOG entry).

## Adversarial review

Ran a multi-lens adversarial review (concurrency, hold/release invariants, API/call-site consistency, test rigor, docs) with refutation-based verification. Actioned findings:

- Closed the **fast-flap serialization race** above (item 4) — added a regression test proven to fail without the fix (`start()` fires a 2nd time inside the stop window) and pass with it.
- Fixed two stale doc-comments this change touches: `Daemon.start()`'s JSDoc (hold is released by `stop()` too now, not only `term()`) and `CommandController.term()`'s JSDoc (it never destroyed the subcontainer).

## Testing

- New `projects/start-sdk/lib/test/DaemonLifecycle.test.ts` (6 tests): `stop()` vs `term()` subcontainer lifecycle, resume re-holds a live subcontainer (the exact `already destroyed` regression), pause uses `stop()` not `term()`, failed `start()` → health failure (not unhandled rejection), and fast-flap serialization.
- `tsc --noEmit` clean, Prettier clean, **77/77** jest tests pass.

Scope is confined to `projects/start-sdk/lib/mainFn` + its tests; no Rust, OS-bindings, or `dist/` changes.

### DIFF
--- a/projects/start-sdk/CHANGELOG.md
+++ b/projects/start-sdk/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## 2.0.3 — StartOS 0.4.0-beta.10 (2026-07-08)
+
+### Fixed
+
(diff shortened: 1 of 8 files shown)


dr-bonez (dismissed):

dr-bonez (approved):

helix-nine:
Thanks @dr-bonez — merged 🎉 The `style(core)` rustfmt fix landed alongside it, so no need to split it out. `@start9labs/start-sdk` is now **2.0.3**: dependency-gated daemons pause via the non-destroying `stop()` (no more permanent wedge after a dependency's readiness flaps), plus the serialized-transition hardening and the `DaemonLifecycle` regression suite. Give a shout if you'd like me to cut the npm publish from `dist/`.

PR #3454, opened 2026-07-09
feat(packaging): clone the full monorepo, publish the agent context, correct the SDK docs

MattDHill:
## Summary

Three related changes to how a packaging workspace is provisioned and what the guide tells an agent, plus a batch of doc corrections that were each verified against the SDK source rather than the surrounding prose.

### `init-workspace` clones the whole monorepo

The clone was sparse, shallow, and limited to `projects/start-sdk/docs`. It is now the full tree, so a packager has the SDK source (`projects/start-sdk/lib`) and the StartOS source on hand when the guide can't settle a question — and is standing in a repo they can open a fix PR from. `--filter=blob:none` keeps it cheap: measured at ~4s and 25 MB of `.git` (75 MB with the worktree), with full history, so `log` / `blame` / `rebase` all behave normally.

Symlinking an existing `start-technologies` checkout into the workspace before running `init-workspace` skips the clone. That already worked (`Path::exists()` follows symlinks) but was undocumented; `environment-setup.md` now has an **Already have the monorepo?** section with the commands and the caveat that such a checkout is yours to maintain, so skip the session-start sync.

The old `sparse-checkout add` guidance is removed from `workflow.md` and the agent context, since nothing needs fetching anymore.

### The agent context is now a page of the guide

`docs/AGENTS.md` — the file symlinked into every workspace — moves to `docs/src/agent-context.md`. It builds as [Agent Context] in the published book, so it can be read without scaffolding a workspace or browsing the monorepo. `AGENTS_SYMLINK_TARGET` moves with it, and `init-workspace` repoints symlinks left behind by earlier releases (they now dangle rather than silently resolving to the wrong file).

That frees `docs/AGENTS.md` to become the mdbook's *own* scope doc. Along the way: the conventions shared by every book in the repo (mdBook versions, admonitions, tabs, `SUMMARY.md`, the shared `theme/`) live in `projects/start-docs/AGENTS.md`, which is a **sibling**, not an ancestor — so nothing loads it for you. The root `AGENTS.md` now says so explicitly.

The scaffolded `AGENTS.local.md` stub now states what belongs in it, and that anything which would help every packager belongs upstream in the guide instead of drifting in one workspace.

### Doc corrections

Each of these was checked against `projects/start-sdk/lib`, the installed types, or the StartOS docs — not against neighbouring prose.

- **`interfaces.md`** — Tor is opt-in and per-interface (`projects/start-os/docs/src/tor.md`: *"Tor is not included in StartOS by default"*), `type` is a label rather than a control, and a package can never assert it is reachable on Tor or the public internet. The word "tor" previously appeared **zero** times on this page.
- **`file-models.md`** — what an empty `merge()` actually does, per `fileHelper.ts:400-424`: it fills missing keys, replaces invalid ones with their `.catch()` default, re-serializes (dropping comments), and writes only if the output differs. It does not strip. Also: the SDK's `z.object` is *deep*-loose, verified empirically, so `z.looseObject` is never needed.
- **`recipe-web-ui.md`** — prefer the app's own login. `searxng-startos` is the OS-gate reference **only**; it stores plaintext because the reverse proxy needs it, and that does not transfer. Naming it beside the app-native packages is the anchoring path that has produced a plaintext-at-rest bug.
- **`recipe-admin-credentials.md`** — `utils.getDefaultString` takes a required `DefaultString`. The bare `utils.getDefaultString()` written in five recipes **throws** (`TypeError`). Adds the import line and the rule that packages never roll their own password RNG.
- **`actions.md`** — `Value.text` accepts a `RandomString` on `default` and on `generate`, so the OS can mint the secret; neither was documented anywhere. Snippet typechecked against the real SDK inside `hello-world-startos`. Also: don't `as const` what the SDK already types (`setupActions.d.ts:10` declares `version: '1'`).
- **`workflow.md`** — new "A comment is not evidence" section. Four of the corrections above began life as confidently-worded prose that was simply wrong.
- **`package-template/AGENTS.md`** — `recipes.md` promoted to the first pointer, framed as the entry point.

### Also

- Bumps `start-cli` to **1.0.2** with a changelog entry. Fixes the changelog's link definitions, which pointed at `v0.4.0-beta.10` — a tag that has never existed in any form.
- The `cloning-guide` i18n string now names what is actually cloned, in all five locales.

## Verification

- `mdbook build` clean; every new anchor and relative link across the changed pages resolves.
- `npx prettier --check` clean.
- The `Value.text` / `getDefaultString` snippet typechecks (`tsc --noEmit`) inside a real package.
- Clone strategies measured against the real repo; `sparse-checkout disable` confirmed to widen an existing workspace in place (98 → 3078 files, 3s).
- Rust is a `&str` const, a six-line legacy-symlink repoint, and the stub text; the `read_link` / `PathBuf == &Path` idiom was compile-checked standalone. Leaving the full `cargo check -p start-core` to CI.

[Agent Context]: https://docs.start9.com/packaging/agent-context.html

### DIFF
--- a/shared-libs/crates/start-core/src/s9pk/init.rs
+++ b/shared-libs/crates/start-core/src/s9pk/init.rs
@@ -66,26 +66,7 @@ const TEMPLATE_SUBPATH: &str = "projects/start-sdk/docs/package-template";
 const CLAUDE_MD_CONTENTS: &str = "@AGENTS.md\n@AGENTS.local.md\n";
 
 /// Created once and never overwritten by a sync — the user's own context.
-const AGENTS_LOCAL_STUB: &str = r#"# AGENTS.local.md — your workspace preferences
-
-<!--
-Notes and preferences for AI assistants working in this workspace. CLAUDE.md loads this file
-alongside AGENTS.md, and syncing the guide never overwrites it — so it is the right home for
-anything true of *your* setup rather than of packaging in general:
-
-  - the StartOS box you install to, and the registry you publish to
-  - the packages you maintain, and where they live
-  - conventions your team follows that the guide doesn't mandate
-  - any way your workspace departs from the scaffolded layout — for example, a
-    start-technologies/ symlinked to a monorepo checkout you maintain yourself
-
-Rule of thumb: if a note would help *every* packager, it does not belong here. Open a PR
-against start-technologies so it lands in the guide — otherwise it stays trapped in one
-workspace and drifts out of date.
-
-Delete this comment once you have something to say.
--->
-"#;
+const AGENTS_LOCAL_STUB: &str = include_str!("AGENTS.local.md.template");
 
 #[derive(Deserialize, Serialize, Parser)]
 #[group(skip)]
--- a/shared-libs/crates/start-core/src/s9pk/init.rs
+++ b/shared-libs/crates/start-core/src/s9pk/init.rs
@@ -39,20 +39,25 @@ registry:
 /// a package repo.
 const DOCS_URL: &str = "https://docs.start9.com/packaging/environment-setup.html";
 
-/// Default source for the packaging guide (which also carries the package
-/// template). The guide lives in the start-technologies monorepo; the clone is
-/// sparse (see GUIDE_SUBPATH) so packagers don't pull the whole repo. Re-point the
-/// `start-technologies` remote afterward to track a fork; the session-start sync
-/// follows whatever remote is configured.
+/// Default source for the packaging guide (which also carries the package template).
+/// The guide lives in the start-technologies monorepo, and the clone takes the whole
+/// tree: past the guide, packagers get the SDK and StartOS source to answer questions
+/// the guide can't, and a repo they can open a fix PR from. `--filter=blob:none` keeps
+/// it cheap (blobs fetch on demand) while leaving full history, so `log`/`blame`/rebase
+/// all work. Re-point the `start-technologies` remote afterward to track a fork; the
+/// session-start sync follows whatever remote is configured.
 const MONOREPO_URL: &str = "https://github.com/Start9Labs/start-technologies.git";
-/// Workspace-relative path to the sparse monorepo checkout that carries the guide.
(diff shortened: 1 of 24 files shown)


shared-libs/crates/start-core/src/s9pk/init.rs
```diff
@@ -64,9 +69,21 @@ const CLAUDE_MD_CONTENTS: &str = "@AGENTS.md\n@AGENTS.local.md\n";
...
+  - the packages you maintain, and where they live
+  - conventions your team follows that the guide doesn't mandate
+  - any way your workspace departs from the scaffolded layout — for example, a
+    start-technologies/ symlinked to a monorepo checkout you maintain yourself
+
+Rule of thumb: if a note would help *every* packager, it does not belong here. Open a PR
+against start-technologies so it lands in the guide — otherwise it stays trapped in one
+workspace and drifts out of date.
+
+Delete this comment once you have something to say.
```
dr-bonez: extract this to an `include_str!`

MattDHill: Done in 2a85a6db4 — `include_str!("AGENTS.local.md.template")`, matching the `vconsole.conf.template` / `update_details/v0_4_0.md` precedents.

Two notes on the filename:

- **`.template`, not `.md`** — the repo-wide `prettier --write .` would otherwise reformat it, and whatever prettier decides becomes the bytes we write into every packager's workspace. Keeping it off prettier's radar means the emitted file is exactly what's authored.
- **Not literally `AGENTS.local.md`** — an agent walking up the tree from `src/s9pk/` would happily load that as real instructions.

Verified the extracted bytes are identical to the old raw string, and that `include_str!` resolves from `init.rs`'s directory (926 bytes, ends `-->\n`).

Happy to give `WORKSPACE_CONFIG_CONTENTS` the same treatment if you want — it's the other embedded blob in that file, though it's only 8 lines of YAML and a unit test asserts it parses as a `WorkspaceConfig`.

dr-bonez (approved):

PR #3458, opened 2026-07-10
fix(start-wrt): embed the build-start git-hash stamp instead of re-ch…

Dominion5254:
…ecking git

By binary-build time the make run has already touched the tracked package-lock.json (shared-libs/ts-modules/build.mk), which build-rust.sh's fresh `git diff-index` misread as a dirty tree, branding every CI build's version string with -dirty. Read the parse-time GIT_HASH.txt stamp instead (the same source the web config and image basenames use, matching how start-core embeds the root stamp); a checkout that was genuinely dirty at make parse still gets -dirty, and non-make invocations keep the git fallback.


Dominion5254:
Closed in favor of #3460

PR #3460, opened 2026-07-10
fix(build): detect a dirty tree with git status, not stat-only diff-i…

Dominion5254:
…ndex

Builds touch tracked files without changing content — the make recipe for package-lock.json ends in `touch` (shared-libs/ts-modules/build.mk), which runs before the product binaries compile. `git diff-index` trusts cached stat data and reports such files as dirty, so every CI-built startwrt branded its version string -dirty (start-wrt v1.0.0 shipped as `d557594-dirty` in `startwrt-cli verify` while the web UI, stamped at make parse time before the touch, showed clean). `git status --porcelain --untracked-files=no` compares content, keeps diff-index's tracked-only scope, and still flags real modifications.

Converted everywhere the monorepo derives a hash dirty-marker: the GIT_HASH.txt stamp scripts (root + start-wrt), start-wrt's build-rust.sh and ctrl/build.rs fallback, and s9pk packaging's GitHash (start-cli/SDK pack builds can touch tracked files the same way).

### DIFF
--- a/build/env/check-git-hash.sh
+++ b/build/env/check-git-hash.sh
@@ -3,7 +3,10 @@
 cd "$(dirname "${BASH_SOURCE[0]}")"
 
 if [ "$GIT_BRANCH_AS_HASH" != 1 ]; then
-    GIT_HASH="$(git rev-parse HEAD)$(if ! git diff-index --quiet HEAD --; then echo '-modified'; fi)"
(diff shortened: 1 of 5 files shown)


dr-bonez (approved):
As long as this is tested, it looks good

PR #3463, opened 2026-07-10
fix(net): expose the whole host when a public domain is added

helix-nine:
## Problem

Adding a public domain to a host is meant to expose that host as a unit — every binding and range reachable via the domain on its own external port. Instead, `add_public_domain` (`net/host/address.rs`):

- enabled the domain only on the **target** single-port binding,
- **disabled** it on every *other* single-port binding, and
- never touched **port ranges** at all.

Because domains are enabled-by-default (opt-out), a range's synthesized `PublicDomain` address stayed live, so any range on a host that had a domain was silently WAN-forwarded (the `mod.rs` range-forward loop) while the range's own WAN toggle read `disabled` — i.e. the toggle lied. Concretely: attaching `turn.start9.dev` to the coturn `turn` binding invisibly forwarded the 42000–42499 relay range.

## Fix

Replace the target-only enable + sibling-disable cleanup with a single helper, `enable_domain_on_addresses`, applied to **every binding and every range** on the host:

- clear the domain from `disabled` (domains are on-unless-disabled);
- for non-SSL rows — no SNI, so domain packets are indistinguishable from bare-IP packets — force-enable the matching WAN IPv4 and, on bindings, the IPv6 GUA;
- ranges pass `ip_info: None` (they are IPv4-only, non-SSL).

Standalone public-IP defaults are unchanged (still opt-in). The only behavioral change is the sibling/range handling on domain-add.

## Result

Adding a domain now exposes the host as a unit and reflects it honestly. For coturn, `turn.start9.dev` lights up `turn:` 3478, `turns:` 5349, and the relay range 42000–42499 together on a single host, with the range's WAN toggle showing enabled instead of lying.

## Tests

Added unit tests for `enable_domain_on_addresses`:
- non-SSL domain re-enables the domain and force-enables its bare WAN IP;
- SSL domain does **not** expose the bare IP (SNI);
- a domain on one gateway does not touch addresses on another.

`cargo test -p start-core net::host` → 30 passed. `cargo check -p start-core` clean.

## Known follow-up (not in this PR)

`remove_public_domain` does not tear down the force-enabled WAN IPs, so removing a domain leaves the host's bare WAN IPs enabled/forwarding. This was pre-existing (previously only the target binding); this PR extends it to all bindings/ranges. A symmetric teardown on remove — disable the WAN IPs on the removed domain's non-SSL port unless another domain still uses that gateway+port — is worth a follow-up.

### DIFF
--- a/projects/start-os/web/ui/src/app/routes/portal/components/interfaces/addresses/gateway/gateway.component.ts
+++ b/projects/start-os/web/ui/src/app/routes/portal/components/interfaces/addresses/gateway/gateway.component.ts
@@ -274,7 +274,6 @@ export class GatewayComponent {
       fqdn,
       gateway: gatewayId,
       acme: !authority || authority === 'local' ? null : authority,
-      internalPort: iface?.addressInfo.internalPort || 80,
     }
 
     return this.tasks.run(async () => {
--- a/shared-libs/crates/start-core/src/net/host/address.rs
+++ b/shared-libs/crates/start-core/src/net/host/address.rs
@@ -10,6 +10,7 @@ use ts_rs::TS;
 use crate::GatewayId;
 use crate::context::{CliContext, RpcContext};
 use crate::db::model::DatabaseModel;
+use crate::db::model::public::IpInfo;
 use crate::hostname::ServerHostname;
 use crate::net::acme::AcmeProvider;
 use crate::net::dns::QueryDnsRes;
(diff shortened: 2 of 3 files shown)


helix-nine:
Closing in favor of #3464. After team discussion we're going the opposite direction: rather than a public domain exposing the whole host, bindings and port ranges should both stay disabled by default. #3464 keeps the original per-binding scoping and just fixes range parity — a domain is now disabled on sibling ranges too, so a range is WAN-exposed only via its own toggle.

PR #3464, opened 2026-07-10
fix(net): reconcile public-domain exposure on sibling bindings and ranges

helix-nine:
## Problem

A public domain is scoped to its target binding: `add_public_domain` enables it on that binding and, on the same host, is supposed to keep it off *other* bindings. Two gaps:

1. Port **ranges** were never handled. Because a range's synthesized domain row is enabled-by-default (opt-out, like any domain), a domain added to a sibling binding silently WAN-forwarded the **entire range** (the `mod.rs` range-forward loop) while the range's own WAN toggle read "disabled". (This is the coturn relay-range report — attaching a domain to `turn` invisibly forwarded 42000–42499.)
2. The old sibling cleanup blindly wrote the domain into `disabled` regardless of an already-enabled WAN IP. Since a non-SSL domain shares the bare WAN IP's packets (no SNI), a sibling/range whose WAN IPv4 was already enabled would then show the domain "disabled" while it was in fact still reachable — reintroducing the same mismatch on a re-add or when a second domain is added.

## Fix

Replace the blind sibling-disable with a **reconcile** applied identically to sibling bindings and port ranges (`reconcile_domain_on_sibling`):

- **non-SSL** domain row: if the co-located WAN IPv4 (same gateway + port) is already enabled on that sibling, **enable** the domain to match; otherwise **disable** it.
- **SSL** domain row: always isolated (it carries its own SNI).

So a domain and its bare WAN IP stay in lockstep on every binding and every range — in both directions and across re-adds. The direct per-address enable/disable toggle path (`set_address_enabled_on`, shared by bindings and ranges) was already correct and is unchanged; this only fixes the `add_public_domain`-time reconciliation.

## Tests

New unit tests for `reconcile_domain_on_sibling`: non-SSL follows an enabled WAN IP; non-SSL isolated when the IP is off; SSL always isolated; an enabled IP on a *different* gateway is not honored. `cargo test -p start-core net::host` → 31 passed. `cargo check -p start-core` clean.

## Note

Supersedes #3463 (the abandoned "expose the whole host" direction).

### DIFF
--- a/shared-libs/crates/start-core/src/net/host/address.rs
+++ b/shared-libs/crates/start-core/src/net/host/address.rs
@@ -259,14 +259,21 @@ fn reconcile_domain_on_sibling(
             disable.push(port);
             continue;
         }
-        // Non-SSL: honor an already-enabled co-located WAN IPv4 (same gateway +
-        // port) — without SNI it is the same packets as the bare IP.
+        // Non-SSL: honor an already-enabled co-located public WAN address on the
+        // same gateway + port — IPv4 or a WAN IPv6 GUA. Without SNI the domain
+        // and this bare IP are the same packets, so the two must stay in sync.
         let wan_enabled = addresses.available.iter().any(|b| {
             !b.ssl
                 && b.public
-                && matches!(&b.metadata, HostnameMetadata::Ipv4 { gateway: gw2 } if gw2 == gateway)
-                && b.to_socket_addr()
-                    .map_or(false, |sa| sa.port() == port && addresses.enabled.contains(&sa))
+                && matches!(
+                    &b.metadata,
+                    HostnameMetadata::Ipv4 { gateway: gw2 }
+                        | HostnameMetadata::Ipv6 { gateway: gw2, .. }
+                    if gw2 == gateway
+                )
+                && b.to_socket_addr().map_or(false, |sa| {
+                    sa.port() == port && addresses.enabled.contains(&sa)
+                })
         });
         if wan_enabled {
             enable.push(port);
@@ -307,6 +314,17 @@ pub async fn add_public_domain<Kind: HostApiKind>(
                 }
             }
 
+            // Adding a domain that is already present is a no-op for exposure:
+            // we re-affirm the config below, but skip the target force-enable and
+            // the sibling/range reconcile so a re-add can't re-isolate a binding
+            // whose WAN IP the operator has since enabled, or clobber any other
+            // per-address choice. Enabling an existing domain on a binding is done
+            // through set-address-enabled, not by re-adding it.
+            let is_new = !Kind::host_for(&inheritance, db)?
+                .as_public_domains()
+                .keys()?
+                .contains(&fqdn);
+
             Kind::host_for(&inheritance, db)?
                 .as_public_domains_mut()
                 .insert(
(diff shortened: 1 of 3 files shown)


shared-libs/crates/start-core/src/net/host/address.rs
```diff
@@ -232,6 +233,113 @@ pub struct AddPublicDomainRes {
...
+    /// Reconcile a newly-bound single-port binding against the host's existing
+    /// public domains — a domain added earlier must not silently leak onto a
+    /// binding added afterwards. A fresh binding is always a sibling (never a
+    /// domain's target), so this isolates each domain unless the binding's own
+    /// WAN address is already enabled. Call after `update_addresses` has
+    /// synthesized the domain rows, then re-run `update_addresses` so the port
+    /// forwards reflect the isolation. Only run for a genuinely new binding —
+    /// re-running it on a re-bind would clobber a domain's target binding.
+    pub fn reconcile_domains_on_new_binding(&mut self, internal_port: u16) -> Result<(), Error> {
+        let domains = self.domain_gateways()?;
```
dr-bonez: the naming here should probably specify we're talking about public domains. private domains are similarly tied to LAN addresses.

shared-libs/crates/start-core/src/net/net_controller.rs
```diff
@@ -1233,8 +1233,15 @@ impl NetService {
...
                 let host = host_for(db, pkg_id.as_ref().unwrap_or(&PackageId::start_os()), &id)?;
+                let is_new = !host.as_bindings().contains_key(&internal_port)?;
                 host.add_binding(&mut ports, internal_port, options)?;
                 host.update_addresses(&hostname, &gateways, &ports)?;
+                // Isolate a newly-bound binding from any pre-existing public
+                // domain on this host, then re-derive so port forwards match.
+                if is_new {
+                    host.reconcile_domains_on_new_binding(internal_port)?;
+                    host.update_addresses(&hostname, &gateways, &ports)?;
+                }
```
dr-bonez, line 1244: seems inefficient to run host.update_addresses twice. is that necessary? does the reconcile depend on the first update?

shared-libs/crates/start-core/src/net/host/address.rs
helix-nine: Good call — renamed to make the public-domain scope explicit (private domains being the LAN-tied parallel): `domain_gateways`→`public_domain_gateways`, `reconcile_domain_on_sibling`→`reconcile_public_domain_on_sibling`, `reconcile_domains_on_new_binding`/`_range`→`reconcile_public_domains_…`, and `has_nonssl_domain`→`has_nonssl_public_domain`. Done in b0a4e6ab0.

shared-libs/crates/start-core/src/net/net_controller.rs
helix-nine: Yes — the reconcile depends on the first \`update_addresses\`, and the second turns its output into forwards:

1. **First \`update_addresses\`** synthesizes the \`available\` address rows onto every binding/range, including the public-domain rows. The reconcile *reads* \`available\` to find the domain rows and the co-located WAN/LAN IPv4 + GUA it links — without it there is nothing to reconcile against. (This one also runs unconditionally after \`add_binding\`, so it is not new.)
2. **The reconcile** writes the override sets (\`enabled\`/\`disabled\`/\`gua_wan\`).
3. **Second \`update_addresses\`** recomputes \`port_forwards\` from \`enabled_addresses()\` and re-projects \`gua_wan\` → the GUA rows' \`public\` flag, so the forwards reflect what the reconcile decided.

So it is not a redundant double-recompute: (1) is the reconcile's input, (2) is its output. It also only happens on the \`is_new\` path (a genuinely new binding/domain), which is a config-time operation, so the extra pass is cheap in practice. Collapsing to one pass would mean either folding the reconcile into \`update_addresses\` or duplicating the \`port_forwards\` computation into the reconcile — both move logic that lives in one place today.

dr-bonez (approved):

PR #3471, opened 2026-07-12
fix(cli): config precedence + .local (mDNS) resolution — start-cli 1.0.3

### LINKED ISSUE #3469: start-cli: musl-static build cannot resolve .local (mDNS) hostnames — auth login fails after 5s
## Summary

`start-cli` cannot resolve `.local` (mDNS) hostnames on Linux, because it ships as a **statically-linked musl binary** and musl's `getaddrinfo` does not implement NSS. It ignores `/etc/nsswitch.conf` entirely, so `mdns4_minimal` — the only thing that resolves `.local` on a Linux box — is never consulted. musl can read `/etc/hosts` and send unicast DNS to `resolv.conf`, and nothing else.

This surfaces as an opaque, ~5s failure:

```
Unknown Error: auth.login: error sending request for url (https://demo.local/rpc/v1)
```

It is not a TLS, network, or timeout-tuning issue. `auth login` is just where users hit it first — **every** `start-cli` command against a `.local` host is affected, and `.local` is the primary way StartOS users address their server.

## Reproduction

Against a StartOS box advertising `demo.local` (192.168.8.170). No config involved — the host is passed literally, so this reproduces from any directory:

```
$ PASSWORD=wrong start-cli -H https://demo.local auth login
Unknown Error: auth.login: error sending request for url (https://demo.local/rpc/v1)
```

The same box addressed by IP (`-H https://192.168.8.170`) works instantly, from the same shell.

**10/10 runs fail, every one at ~5.04s.** Originally reported as intermittent; see "Why it looks intermittent" below.

`RUST_LOG=trace` shows the request stalling inside connect with no resolver activity at all:

```
TRACE hyper_util::client::legacy::pool: checkout waiting for idle connection: ("https", demo.local)
DEBUG reqwest::connect: starting new connection: https://demo.local/
TRACE hyper_util::client::legacy::connect::http: Http::connect; scheme=Some("https"), host=Some("demo.local"), port=None
   << 5.006s of silence >>
TRACE hyper_util::client::legacy::pool: checkout dropped for ("https", demo.local)
TRACE reqwest::retry: shouldn't retry!
```

Nothing logs in between because `getaddrinfo` is a blocking call inside `spawn_blocking`; hyper never gets far enough to log a resolved IP.

## Evidence that it is purely resolution

| Path | Result |
|---|---|
| `start-cli` → `https://demo.local` | fails at 5.04s |
| `start-cli` → `https://192.168.8.170` | **works, 0.29s** (real `Password Incorrect` from the server) |
| `curl https://demo.local/rpc/v1` | HTTP 200, 126ms |
| `getent ahosts demo.local` (glibc, AF_UNSPEC) | 0.18s |
| Python `getaddrinfo` (glibc, AF_UNSPEC) | 0.04s |

By IP it works and TLS is fine. Every glibc consumer resolves the name in milliseconds; only the musl binary fails. Adding a static `/etc/hosts` entry fixes it immediately — `/etc/hosts` is the one lookup path musl *does* support.

Binary confirmed static musl:

```
$ file $(which start-cli)
... /home/matt/.local/bin/start-cli-release
$ ldd $(which start-cli)
	not a dynamic executable
$ strings $(which start-cli) | grep musl
compiler: .../zigcc-x86_64-unknown-linux-musl-....sh ... --target=x86_64-unknown-linux-musl
```

## Where the 5 seconds comes from

There is no 5s timeout in our code — the reqwest client in `shared-libs/crates/start-core/src/context/cli.rs:186` sets no timeouts at all, and nothing in the CLI request path or `rpc-toolkit` imposes one. The 5s is musl's own.

Skipping NSS, musl sends a plain unicast DNS query for `demo.local` to `127.0.0.53`. systemd-resolved returns **SERVFAIL** in 30ms (it refuses to serve `.local` over unicast, and has no mDNS on the link). musl treats SERVFAIL as *retryable* rather than a final answer, so it retries until its hard-coded 5-second budget is exhausted, then returns `EAI_AGAIN`. Hence exactly 5.0s, every time.

### Why it looks intermittent

Once you're off NSS, the outcome depends entirely on what that machine's `resolv.conf` server happens to do with a `.local` name — SERVFAIL on one box, a cached mDNS answer through a `systemd-resolved` stub with multicast enabled on another, a slow cold multicast round trip elsewhere. The underlying defect is unconditional; only its visible symptom is environment-dependent.

## Scope

Three things worth knowing before scoping a fix:

1. **Every binary is affected, not just the CLI.** All six products build musl-static (`start-cli`, `startbox`, `registrybox`, `tunnelbox`, `startwrt`, `start-container` — see each `projects/*/build.mk`), so none of them have NSS. Corroborating: `shared-libs/crates/start-core/src/net/mdns.rs` already has a `resolve_mdns()` that shells out to `avahi-resolve-host-name` — someone previously worked around the missing NSS server-side, but it was never wired into the CLI's HTTP client.

2. **A reqwest-only fix is incomplete.** `ws_continuation` (`context/cli.rs:321`) hands the URL to `tokio_tungstenite`, which does its own resolution via `TcpStream::connect` and therefore hits musl's `getaddrinfo` too. The WebSocket paths (`logs -f`, install progress) would still fail after the HTTP client is fixed. Both call sites need covering.

3. **macOS is almost certainly unaffected** — darwin's `getaddrinfo` resolves `.local` via mDNSResponder natively. This is a Linux-only defect.

## Secondary issue: the error swallows its cause

`Unknown Error: auth.login: error sending request for url (...)` is reqwest's outer `Display`. The inner source says `dns error: Temporary failure in name resolution`, which would have made this self-diagnosing. Walking `std::error::Error::source()` when converting `reqwest::Error` is a small, independently worthwhile change.

## Proposed fix — @dr-bonez, how do you want to proceed?

**Option A (my recommendation): custom resolver on the client.** reqwest exposes an ungated `ClientBuilder::dns_resolver()`, and start-core already pulls hickory in transitively via `hickory-server 0.26` (`resolver` feature). Implement a `Resolve` that answers `*.local` with a real multicast query and delegates everything else to the default `GaiResolver`, wire it into `CliContext::init`, and use the same resolver to pre-resolve the WebSocket host before handing an address to tungstenite. Pure Rust, works in a static binary, no dependency on avahi being installed on the user's machine.

**Option B: build `start-cli` against glibc.** Restores NSS so `nss-mdns` works the way curl does. But it gives up the portable static binary (glibc version skew across distros) and still requires the user to have `nss-mdns` installed. Likely a non-starter for distribution, but it is the option that actually honors the host's resolver config.

**Option C: shell out to avahi**, reusing the existing `resolve_mdns()`. Rejecting this unless you disagree — it depends on `avahi-utils` being present on the client, which it often isn't (it's not in `PATH` on my Ubuntu box).

Open questions for you:

- Option A vs B, or something else?
- Should the fix live in `start-core` (benefiting every musl binary that resolves a hostname) or be scoped to `CliContext`?
- Is there a reason the musl-static target is load-bearing beyond portability that would rule out B outright?

MattDHill:
> **Now carries two fixes.** #3474 (`.local`/mDNS resolution, closes #3469) was merged into this branch rather than into `master` — that's the stacked-PR mechanic. So **merging this PR is what ships both fixes**, plus the start-cli **1.0.3** release. #3474's own review discussion is [here](https://github.com/Start9Labs/start-technologies/pull/3474); the mDNS change is verified end-to-end against a real box (old: 5.04s failure → new: 0.54s real server response).
>
> The rest of this description covers the config-precedence fix only.

---

## The bug

`start-cli`'s README documents the layering as:

> Configuration is layered: explicit flags override `.startos/config.yaml` (local workspace), which overrides `/etc/startos/config.yaml`.

The code does the opposite for config files. `ClientConfig::load()` merges `~/.startos/config.yaml` and `/etc/startos/config.yaml` into the **same field** as the `-H`/`-r` flag, and `CliContext::init` then feeds that one field to `resolve_target`. A config file therefore resolves with *flag-level* precedence and silently shadows the workspace's `default` profile.

Concretely: a `host:` key in `~/.startos/config.yaml` hijacks every invocation, even deep inside a packaging workspace with its own `default` profile. That's a live trap — a leftover **pre-1.0 flat config** in `$HOME` (exactly what the 1.0 workspace migration leaves behind) quietly overrides the workspace you're standing in, and nothing tells you.

## The fix

A config file is a *fallback*, not a per-invocation override. Only a flag should outrank the workspace.

- `ClientConfig` gains `host_file` / `registry_file`. `merge_with` routes a config file's `host`/`registry` there, leaving `host`/`registry` to mean exactly "as passed on the command line". (`self` starts as the parsed command line; `other` is always a file — see `load_path_rec`.) Both new fields are `#[arg(skip)]` + `#[serde(skip)]`, so they're unreachable from a flag or a file key.
- New `resolve_target_layered(flag, file, profiles)` resolves in precedence order: **flag → workspace `default` → config file → nothing**. It leans on the existing `resolve_target(None, ..)` returning the `default` profile, so the flag and workspace-default cases collapse into one arm.
- `CliContext::init` uses it for both host and registry.

Unqualified behavior is unchanged: with no workspace in scope, a config file still supplies the host exactly as before (how `start-cli` runs outside a workspace, and on the server itself against `/etc/startos/config.yaml`). A flag naming an unknown profile still errors rather than quietly falling through to a file.

## Behavior change worth calling out

`-c <file>` is also demoted below the workspace. It's a flag that selects *a file of defaults*, and I read it as not itself asserting a host — but it is a judgment call, and the one place someone could notice a difference. Happy to special-case it if you'd rather an explicit `-c` keep flag-level precedence.

The shared `ClientConfig` also backs the `tunnelbox` / `registrybox` CLI contexts, so they pick up the same ordering. I left their changelogs alone since those track the server products; say the word if you want entries there.

## Tests

`context/config.rs` had no test module; this adds one covering the merge and all four precedence layers — including `the_workspace_default_outranks_a_config_file`, which fails on `master`.

## Verification

Formatting verified locally (`cargo +nightly fmt --check` on `start-core`, `prettier --check` on the touched markdown). **I did not compile locally** — a `start-core` build is heavyweight on my machine, so I'm leaning on CI for the typecheck and test run.

Found while diagnosing #3469 (unrelated root cause: `.local` mDNS resolution). Independent of it.

## Release

Ships as **start-cli 1.0.3** — crate version bumped, `Cargo.lock` synced, and the changelog entry promoted out of `Unreleased`.

### DIFF
--- a/shared-libs/crates/start-core/src/context/cli.rs
+++ b/shared-libs/crates/start-core/src/context/cli.rs
@@ -26,6 +26,7 @@ use crate::context::config::{ClientConfig, local_config_path, resolve_target};
 use crate::context::{DiagnosticContext, InitContext, RpcContext, SetupContext};
 use crate::developer::{OS_DEVELOPER_KEY_PATH, default_developer_key_path, load_signing_key};
 use crate::middleware::auth::local::LocalAuthContext;
+use crate::net::mdns::pin_mdns_host;
 use crate::prelude::*;
 use crate::rpc_continuations::Guid;
 use crate::s9pk::init::{BUILD_KEY_FILE, STARTOS_DIR};
@@ -130,6 +131,9 @@ impl CliContext {
             Some(url) => url,
             None => "http://localhost".parse()?,
         };
+        // Before anything derives a URL from this one: a `.local` host is unresolvable from a
+        // musl-static binary, so pin it to an address the system resolver found.
+        pin_mdns_host(&mut url)?;
 
         let registry = resolve_target(config.registry.as_ref())?;
 
@@ -164,6 +168,7 @@ impl CliContext {
             },
             registry_url: registry
                 .map(|mut registry| {
+                    pin_mdns_host(&mut registry)?;
                     registry
                         .path_segments_mut()
                         .map_err(|_| eyre!("Url cannot be base"))
--- a/shared-libs/crates/start-core/src/context/cli.rs
+++ b/shared-libs/crates/start-core/src/context/cli.rs
@@ -22,7 +22,7 @@ use tokio_tungstenite::{Connector, MaybeTlsStream, WebSocketStream};
 use tracing::instrument;
 
 use super::setup::CURRENT_SECRET;
-use crate::context::config::{ClientConfig, WorkspaceConfig, local_config_path, resolve_target};
+use crate::context::config::{ClientConfig, local_config_path, resolve_target};
 use crate::context::{DiagnosticContext, InitContext, RpcContext, SetupContext};
 use crate::developer::{OS_DEVELOPER_KEY_PATH, default_developer_key_path, load_signing_key};
 use crate::middleware::auth::local::LocalAuthContext;
@@ -123,19 +123,15 @@ impl CliContext {
     /// BLOCKING
     #[instrument(skip_all)]
     pub fn init(config: ClientConfig) -> Result<Self, Error> {
-        // Resolve -H/-r (profile name or URL) against the workspace config, falling
-        // back to a literal URL, then to localhost / no registry when unset.
-        let workspace = WorkspaceConfig::find()?;
-        let mut url =
-            match resolve_target(config.host.as_deref(), workspace.as_ref().map(|w| &w.host))? {
-                Some(url) => url,
-                None => "http://localhost".parse()?,
-            };
+        // Follow each namespace's `default` profile to a URL (`load` already seeded
+        // -H/-r as `default` and layered every config file in), then localhost / no
+        // registry when unset.
+        let mut url = match resolve_target(config.host.as_ref())? {
+            Some(url) => url,
+            None => "http://localhost".parse()?,
+        };
 
-        let registry = resolve_target(
-            config.registry.as_deref(),
-            workspace.as_ref().map(|w| &w.registry),
-        )?;
+        let registry = resolve_target(config.registry.as_ref())?;
 
         let cookie_path = config.cookie_path.unwrap_or_else(|| {
             local_config_path()
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7709,7 +7709,7 @@ checksum = "6ce2be8dc25455e1f91df71bfa12ad37d7af1092ae736f3a6cd0e37bc7810596"
 
 [[package]]
 name = "start-cli"
-version = "1.0.2"
+version = "1.0.3"
 dependencies = [
  "start-core",
  "tracing",
--- a/projects/start-cli/ARCHITECTURE.md
+++ b/projects/start-cli/ARCHITECTURE.md
@@ -33,7 +33,7 @@ standalone binary it behaves like a dedicated CLI.
 argv ─▶ MultiExecutable ─▶ start_cli::main (in start-core)
                               │
                               ├─ build CliApp from start-core::main_api()
-                              ├─ load ClientConfig (flags + .startos/config.yaml + /etc/startos)
+                              ├─ load ClientConfig (flags + -c + workspace + ~/.startos + /etc/startos)
                               └─ run:
                                    • remote command  ─▶ CliContext ─HTTPS RPC─▶ StartOS server
                                    • local command   ─▶ executed in-process (s9pk, keys, util)
@@ -55,10 +55,18 @@ and `tunnel` groups can target a separate registry/tunnel host via `--registry`/
 
 ### Configuration
 
-`ClientConfig` (in `start-core::context::config`) is parsed from CLI flags and merged with
-config files in priority order: explicit flags → workspace `.startos/config.yaml` →
-`/etc/startos/config.yaml`. It carries the target host/registry/tunnel, proxy, cookie path,
-developer key path, root CAs, and the `--insecure` toggle. The loaded config becomes a
+`ClientConfig` (in `start-core::context::config`) is the single config type — the workspace
+`.startos/config.yaml`, `~/.startos/config.yaml`, `/etc/startos/config.yaml`, and `-c` files all
+deserialize into it (a `schema` key marks a real workspace for the walk-up). `host`/`registry` are
+each a `Profiles` namespace with one config key; `-H`/`-r` parse (via `Profiles`' `FromStr` /
+`ValueParserFactory`) straight into `{ default: <value> }`, so the flag is just the top layer of
+that namespace — no separate selector field. `ClientConfig::load` then layers the config files
+under the flags: a `-c` chain, the discovered workspace (only its `host`/`registry`), `~/.startos`,
+then `/etc/startos`. A bare URL is shorthand for `default`, and the layers combine into one
+namespace — a union of aliases where a name defined at more than one level takes the higher tier's
+value. `resolve_target` follows the `default` profile, chasing a value that names another profile
+until it reaches a URL, which it parses only then. `ClientConfig` also carries tunnel, proxy,
+cookie path, developer key path, root CAs, and the `--insecure` toggle. The loaded config becomes a
 `CliContext` that the RPC handlers use.
 
 ## Man pages
--- a/projects/start-cli/CHANGELOG.md
+++ b/projects/start-cli/CHANGELOG.md
@@ -11,6 +11,35 @@ or the CLI's externally observable behavior.
 
 ## [Unreleased]
 
+## [1.0.3]
+
+### Added
+
+- **`host`/`registry` profiles work in every config file, not just the workspace.** A config
+  file's `host`/`registry` is a namespace of named profiles — `default`, `prod`, whatever you
+  like — and a bare URL is shorthand for the `default` profile, so a legacy flat
+  `host: https://box.local` still targets that URL. Profiles from a `-c` file, the workspace
+  `.startos/config.yaml`, `~/.startos/config.yaml`, and `/etc/startos/config.yaml` all **combine**
+  into one namespace, so you can keep reusable profiles in your home config and reach them from
+  anywhere — inside a packaging workspace or not. A profile's value can be a URL or the name of
+  another profile, and `-H`/`-r` just set the `default` profile for that invocation (`-H prod`
+  points `default` at your `prod` profile); resolution follows the chain to a URL. A value that
+  isn't a URL is reported only when it's the one resolved, so a stale ambient config never fails an
+  unrelated command.
+
+### Fixed
+
+- **An ambient config file no longer shadows the workspace `.startos/config.yaml`.** A `host`
+  set in `~/.startos/config.yaml` or `/etc/startos/config.yaml` was merged into the same field as
+  an explicit `-H`/`-r`, so it silently outranked the workspace's `default` profile — the opposite
+  of the documented layering, and a trap for anyone with a leftover pre-1.0 flat config in their
+  home directory. Profiles now layer in a strict order, highest precedence first: a command-line
+  flag (`-H`/`-r`), a config file named with `-c`, the workspace config, `~/.startos/config.yaml`,
+  then `/etc/startos/config.yaml`. Where the same profile name is defined at more than one level
+  the higher tier wins; what you name on the command line overrides the workspace, what you don't
+  is an ambient default beneath it. A `-H`/`-r` naming an unknown profile still errors rather than
+  quietly falling through to a lower tier.
+
 ## [1.0.2]
 
 ### Changed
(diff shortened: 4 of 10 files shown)


dr-bonez:
-c should override the workspace config. Hiearchy is cli flags, then cli provided config, then workspace config, then home dir config, then global configs

MattDHill:
Done — pushed. That was exactly the call I flagged as uncertain, so thanks for settling it.

Resolution now follows your ladder strictly:

1. command-line flag (`-H`/`-r`)
2. config file named on the command line (`-c`, and any file it chains to)
3. workspace `.startos/config.yaml`
4. `~/.startos/config.yaml`
5. `/etc/startos/config.yaml`

Mechanically: `merge_with` funnels every config file's `host`/`registry` away from the `-H`/`-r` field, and `load` then lifts the `-c` chain into `*_cli_config` before reading the ambient files, which stay in `*_implicit_config`. So the split is "did the user name this file on the command line" — named ones override the workspace, ambient ones sit beneath it. Within tier 4/5, existing merge order already gives `~` priority over `/etc`.

A `-c` file's `host` can name a workspace profile too, same as `-H` (`host: remote` resolves against the workspace's `remote`), and an unknown profile name still errors instead of silently dropping to a lower tier.

Tests cover each rung, including `a_cli_named_config_outranks_the_workspace_default` and `the_workspace_default_outranks_an_implicit_config`. README updated to the 5-tier list.

shared-libs/crates/start-core/src/context/cli.rs
```diff
@@ -123,18 +126,27 @@ impl CliContext {
...
-        // back to a literal URL, then to localhost / no registry when unset.
+        // Resolve -H/-r (profile name or URL) against the workspace config, then a
+        // config file, then localhost / no registry when unset.
         let workspace = WorkspaceConfig::find()?;
-        let mut url =
-            match resolve_target(config.host.as_deref(), workspace.as_ref().map(|w| &w.host))? {
-                Some(url) => url,
-                None => "http://localhost".parse()?,
-            };
+        let mut url = match resolve_target_layered(
```
dr-bonez: Why are we doing this at all? Why not just include the workspace config in the merge_with chain?

MattDHill: Fair — most of this can go, and you're right that I reached for a parallel mechanism where the existing one would do.

The part that genuinely can't join the chain is the **profiles map**, but not for a precedence reason. `merge_with` merges `ClientConfig` into `ClientConfig`, where `host: Option<String>` is *one* value. `WorkspaceConfig.host` is a `BTreeMap<String, Url>` — a namespace, not a value. And it's a namespace the tiers *above* it read: `-H remote`, and a `-c` file's `host: remote`, both resolve against it. So the map has to be discovered and threaded to `resolve_target` regardless of how precedence is computed. It can't be a link in the chain because it isn't a peer of the other layers — it's the table they're interpreted in.

What *can* join the chain — and what I missed — is the workspace's **`default` profile**, which is just a value:

```rust
// in ClientConfig::load, after the -c chain, before ~/.startos:
let workspace = WorkspaceConfig::find()?;
self.merge_with(ClientConfig {
    host: workspace.as_ref().and_then(|w| w.host.get("default")).map(Url::to_string),
    registry: workspace.as_ref().and_then(|w| w.registry.get("default")).map(Url::to_string),
    ..Default::default()
});
self.load_path_rec(local_config_path())?;
self.load_path_rec(Some(CONFIG_PATH))?;
```

Existing `.or()` semantics then produce your hierarchy for free — flags, `-c`, workspace, home, global — purely from *where the layer is injected*. That deletes all four `*_cli_config`/`*_implicit_config` fields, `resolve_target_layered`, and the `merge_with` change, and puts resolution back to a single `resolve_target(config.host, profiles)`. The only thing `ClientConfig` needs to carry forward is the map itself, for name resolution.

I convinced myself the workspace had to sit outside the chain because it's *discovered* differently (walk up from cwd for a schema-tagged file, skipping a legacy flat config) — but that's just a different loader, not a different precedence mechanism, and it doesn't justify a second one.

Behavior should be identical: an ambient file's `host` still can't shadow the workspace `default`, an unknown profile name still errors, and a `-c` file's `host` can still name a profile. Want me to rewrite it that way? It's a strict simplification of the current diff and I'd rather land it than the version you're looking at.

helix-nine:
Took this over to land the config-precedence fix via the unified model @dr-bonez asked for — **namespaces in every config, not just the workspace** — rather than the parallel mechanism. Pushed as 80cb415.

### What changed

The asymmetry @MattDHill diagnosed (workspace = a profile *map*, everything else = a single *value*) is gone. One `Profiles` type now backs `host`/`registry` in **every** config file — the workspace `.startos/config.yaml`, `~/.startos/config.yaml`, `/etc/startos/config.yaml`, and any `-c` file. A bare URL is shorthand for the `default` profile, so a legacy flat `host: https://box.local` still targets that URL.

With the shape unified, resolution collapses into the existing merge chain (answering "why not just include the workspace in `merge_with`?"):

- `ClientConfig::load` layers the tiers in precedence order — `-c` chain → discovered workspace → `~/.startos` → `/etc/startos` — and `merge_with`/`merge_under` **combine** the namespaces per-alias: distinct profile names accumulate into the union, and a name defined at more than one tier takes the higher tier's value. (Per your follow-up: combine, not replace; same hierarchy for matching aliases.)
- `-H`/`-r` are now flag-only *selectors* resolved against the merged namespace by a single `resolve_target`; no selector → the `default` profile.

**Deleted:** the four `*_cli_config`/`*_implicit_config` fields, `resolve_target_layered`, and the separate workspace threading in `cli.rs`. Net simpler than the diff you were looking at.

### Deliberate behavior change

A config file's `host` is now a profile *definition*, not a *selector* — so a `-c`/flat file can no longer name a profile (e.g. `host: remote`); only `-H`/`-r` select. The precedence ladder itself is unchanged.

### Robustness (from an adversarial review pass)

Profile values are kept verbatim and parsed to a URL only at *resolve* time (exactly how a literal `-H <url>` already resolves). So a stale/malformed ambient config errors only if it's the one selected — it never fails a command that doesn't target it (e.g. `s9pk pack`). A null or empty `host:` is an empty namespace, not a hard parse error. Both are covered by new tests.

### Verification

`cargo test` green (19 config tests incl. the null/empty + lazy-parse regressions, plus the workspace-scaffold and man-page tests — no man diff); `cargo fmt`/prettier clean. Didn't compile the full CI matrix locally (darwin/musl); leaning on CI for that.

shared-libs/crates/start-core/src/context/cli.rs
helix-nine: Done — this is exactly the fix. With `host`/`registry` now a shared `Profiles` namespace in every config file, the workspace just folds into the merge chain (`ClientConfig::load` merges it between the `-c` tier and `~/.startos`), so `resolve_target` reads one merged map. The parallel `resolve_target_layered` + `*_cli_config`/`*_implicit_config` fields this thread was on are deleted. Details in the top-level comment.

helix-nine:
Reworked to the full unification @dr-bonez asked for (719568782). This is a strict simplification of the previous push.

### What changed vs the last commit

- **`WorkspaceConfig` is deleted.** There's one config type now: the workspace `.startos/config.yaml`, `~/.startos`, `/etc/startos`, and `-c` files all deserialize into `ClientConfig`. A `schema` key just marks a real workspace for the walk-up.
- **`-H`/`-r` are no longer selectors.** `load` seeds them as the `default` profile at the top of the stack (overriding the files' `default`) — `-H prod` is just `host: { default: prod }`. There's no selector argument anywhere.
- **Resolution is one recursive walk.** `resolve_target(&profiles)` follows `default`; while a profile's value names another profile it chases the reference (cycle-guarded), and parses the first non-reference value as a URL. That single mechanism now covers `-H <url>`, `-H <profile>`, a flat `host: <url>`, and — restored as a nice side effect — a flat `host: <profile>` naming a profile.

Deleted `resolve_target_layered`, the four tier fields, and the selector plumbing. `resolve_target` went from `(selector, map)` to `(map)`.

### Adversarial review caught one that mattered

Folding the discovered workspace in as a whole `ClientConfig` would have let a walked-up `.startos/config.yaml` set `insecure` / `proxy` / `developer-key-path` / `cookie-path` — a real footgun, since that tier is auto-discovered by `cd`-ing into a repo tree. Fixed: the walk-up workspace contributes **only** its `host`/`registry` profiles (its documented contract); the fixed-path `~/.startos` and `/etc` files you own still merge in whole, as before. Also added the missing `pl_PL` locale for the new cycle error and de-duped a README clause.

### Verify

21 tests (recursion, chains, cycle, flag-seed, folded-workspace scaffold, null/empty tolerance), `cargo check`/`fmt`/`prettier` clean, no man-page diff. CI covers the darwin/musl matrix I can't build locally.

helix-nine:
Collapsed the last bit of duality (373783e0b): `host`/`registry` are now a single `Option<Profiles>` each — one key serving both the config file and the `-H`/`-r` flag. `Profiles` grew a `FromStr` + `ValueParserFactory`, so clap parses `-H <value>` straight into `{ default: <value> }`; the flag is just the top layer of the namespace, and the `seed` step in `load` is gone. No `#[serde(skip)]`/`rename`, no separate map field. `resolve_target` takes `Option<&Profiles>`.

Behavior is identical — pure internal cleanup. Verified end-to-end on the built binary: `-H https://host pubkey` parses/resolves then runs the local command; `-H not-a-url pubkey` errors with `'not-a-url' is not a known profile or a valid URL`; `--help` still shows `-H, --host <HOST>`. 21 tests, fmt/prettier clean, no man-page diff.

dr-bonez (approved):

PR #3476, opened 2026-07-13
fix(start-sdk): don't publish a stale dist or leak the file: link into the tarball

MattDHill:
## Symptom

`make publish` from `projects/start-sdk` fails:

```
npm error code E415
npm error 415 Unsupported Media Type - PUT https://registry.npmjs.org/@start9labs%2fstart-sdk
npm error   invalid path: package/../../shared-libs/ts-modules/start-core/dist/node_modules/isomorphic-fetch/.editorconfig
```

## Cause

The `dist` recipe does `rsync -ac node_modules dist/`, which drags along npm's hidden lockfile. That file describes the **source** tree, where the bundled dep is a `file:` symlink:

```json
"node_modules/@start9labs/start-core": {
  "resolved": "../../shared-libs/ts-modules/start-core/dist",
  "link": true
}
```

When `npm publish` runs from `dist/`, Arborist reads that lockfile, decides `@start9labs/start-core` is a link, and packs **through it** — emitting paths that escape the package root — even though the very next line of the recipe (`rsync -aL`) has already put the real dereferenced files in `dist/node_modules/@start9labs/start-core/`. npm strips `.package-lock.json` from the tarball regardless, so it was never doing anything for consumers; deleting it from `dist/` is a no-op for them and removes the link record entirely.

## Why this regressed now, when every previous publish worked

Worth being precise, because the `link: true` record is **not** new — it has always been copied into `dist/`, which is why publishes have always worked. From the npm debug log of the failing run:

```
verbose shrinkwrap failed to load node_modules/.package-lock.json out of date,
        updated: ../../shared-libs/ts-modules/start-core/dist
```

The link record alone is inert. npm only re-resolves through it when it judges the hidden lockfile **stale** — at which point it discards it and re-derives the tree from the `file:` spec in `dist/package.json`. So the bug needs both halves, and the second half is what changed today.

**Confirmed:** the mechanism above (stale hidden lockfile → re-resolve through the `file:` link → escaping paths).

**Hypothesis, not proven:** the trigger was rebuilding `start-core/dist` while working on #3475. `$(START_CORE_DIST)/package.json` is a prerequisite of the SDK's `package-lock.json` and `node_modules` rules, so that rebuild made `make publish` re-run `npm i` + `npm ci` for the first time in a while, regenerating `node_modules` and its hidden lockfile mid-build (`npm i` also rewrote `package.json` — its mtime was the publish time, not the earlier version bump). That plausibly perturbs the mtime relationship npm uses to decide the hidden lockfile is current, so the copy landing in `dist/` was judged stale. Previously `dist/` was reused untouched and the lockfile stayed "fresh."

I could not verify this: npm keeps only the last 10 logs (`logs-max=10`), and both the failing log and the successful 2.0.4 log have since been pruned, so the diff that would settle it is gone. Flagging it as a hypothesis rather than dressing it up as a finding. The fix makes the trigger moot either way — `dist/` no longer carries the lockfile the staleness check operates on.

## Two staleness bugs found alongside it

These are what let a wrong `dist` reach `npm publish` in the first place:

- **`$(START_CORE_DIST)/package.json` had no prerequisites**, so make only ran it when `start-core/dist` was *missing* — never when start-core's sources changed. A rebuilt start-core left the SDK's `dist` stale against it. Now depends on start-core's sources.
- **`dist` is a directory target**, so make's staleness check doesn't track its contents (a directory's mtime only moves when top-level entries change, and the recipe ends in `touch dist`). This is how the failing run packed a **2.0.5** source tree as a **2.0.4** tarball. `publish` now does a clean rebuild rather than trusting incremental `dist`; `bundle` stays incremental so the OS build isn't slowed.

## Verification

Clean rebuild, then packed the real tarball (`npm pack --dry-run` is not usable here — it reports `bundled files: 0`):

- `dist/package.json` → `2.0.5`, matching source
- `2983` files under `package/node_modules/@start9labs/start-core/` — bundled as real files
- **0** paths containing `../`

With this applied, `make publish` succeeded and **`@start9labs/start-sdk@2.0.5` is live on npm**.

Found while publishing 2.0.5 (#3475).

### DIFF
--- a/projects/start-sdk/Makefile
+++ b/projects/start-sdk/Makefile
@@ -3,16 +3,17 @@ START_CORE := ../../shared-libs/ts-modules/start-core
 START_CORE_DIST := $(START_CORE)/dist
 START_CORE_SRC := $(shell git ls-files $(START_CORE)/lib) $(START_CORE)/package.json
 
-.PHONY: all test clean clean-dist bundle fmt check-fmt check publish link node_modules
+.DELETE_ON_ERROR:
+
+.PHONY: all test clean bundle fmt check-fmt check publish link
 
 all: bundle
 
-# start-core is an external dependency now: the SDK imports it and bundles its
-# built lib into the published dist so external authors still install one package.
-# Prerequisites matter here: with none, this only fired when start-core/dist was
-# *missing*, leaving the SDK's dist stale against a rebuilt start-core.
+# Prereqs so a rebuilt start-core restales the SDK dist; touch after because start-core's
+# build regenerates a tracked source (exver.ts) without bumping dist (else it rebuilds every time).
 $(START_CORE_DIST)/package.json: $(START_CORE_SRC)
 	$(MAKE) -C $(START_CORE) dist
+	touch $(START_CORE_DIST)/package.json
 
 test: node_modules
 	npm test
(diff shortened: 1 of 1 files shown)


helix-nine:
Pushed a rework (01c26a4) so `make bundle` on its own always produces a publishable dist — no force clean-rebuild on `publish`.

**Root cause, fixed structurally.** The E415 escaping paths come from the published `dist/package.json` declaring `@start9labs/start-core` as an out-of-tree `file:../../shared-libs/...` link — and that link still shipped verbatim in 2.0.5's metadata. The recipe now rewrites the spec to the in-tarball path `file:./node_modules/@start9labs/start-core`, so npm can never resolve the bundled dep out of tree, at pack time or ever. (Deleting the hidden lockfile removed the pack-time *trigger* but left the link in the shipped package.json.)

**`make bundle` is the reliable sentinel now.** Track `dist/package.json` (a file) instead of the `dist/` directory — make can't see content changes under a directory target, which is how a stale dist packed 2.0.5 as a 2.0.4 tarball. `publish` depends on `bundle` again; `clean-dist` is gone.

**Restored incremental settling** so bundle no-ops when nothing changed: `node_modules` was `.PHONY` (forced `npm ci` every build), and start-core's build regenerates a tracked source (`exver.ts`) without bumping its dist mtime. Added the same `touch` discipline `build.mk` already uses (`touch` start-core/dist, package-lock.json, node_modules).

Verified: from-scratch build packs **0** escaping paths, 2983 start-core files bundled, in-tarball spec, no hidden lockfile; a real consumer `npm install` + `require` works; two consecutive `make bundle` runs are no-ops while a source edit still rebuilds; `tsc`/prettier/jest (80/80) all pass.

Kept your unambiguous fixes (start-core source prereqs). Happy to squash on merge.

projects/start-sdk/Makefile
```diff
@@ -20,17 +24,21 @@ clean:
...
-	rsync -ac node_modules dist/
+	# Don't stage npm's hidden lockfile: it pins start-core to an out-of-tree path.
+	rsync -ac --exclude=.package-lock.json node_modules dist/
 	rm -rf dist/node_modules/@start9labs/start-core
 	mkdir -p dist/node_modules/@start9labs
 	rsync -aL $(START_CORE_DIST)/ dist/node_modules/@start9labs/start-core/
-	cp package.json dist/package.json
+	# Point the bundled dep at its in-tarball copy so npm never resolves start-core
+	# out of tree (the source spec is a file: link out to shared-libs).
+	node -e "const fs=require('fs'),d=JSON.parse(fs.readFileSync('package.json','utf8'));d.dependencies['@start9labs/start-core']='file:./node_modules/@start9labs/start-core';fs.writeFileSync('dist/package.json',JSON.stringify(d,null,2)+'\n')"
```
dr-bonez: Can you use jq instead? Much easier to read and already required by the build

helix-nine: Done — switched to jq in 0b01166. Confirmed jq is already a build dep (build.mk uses it for the firmware roms) and the output is byte-identical to the node version; rebuild still packs 0 escaping paths with start-core bundled, and bundle still settles to a no-op.

dr-bonez (approved):

PR #3480, opened 2026-07-15
backup-fs: kernel idmapped mount (fuser 0.17 + drop internal idmap)

helix-nine:
Switches `startos-backup-fs` from its FUSE-internal `idmapped_root` uid-translation to a real **kernel idmapped mount**, removing the whole `is_root_for` permission machinery — including the `system.posix_acl_*` `EPERM` where an idmapped container-root couldn't set ACLs (`xattr_access_check` branch `C` ignored the idmap and denied any non-literal-uid-0 caller).

Depends on **Start9Labs/fuser#1** (merged; `master` now upstream fuser 0.17 + our `FUSE_SYNCFS`). Pins `4ef196b`.

### Commits
1. **`refactor(backup-fs): migrate to fuser 0.17 API`** — `handler: Mutex<Handler>` for the `&self` methods (`init`/`destroy` stay `&mut self`); newtype/bitflag/`Errno` conversions across the adapter; `Config`-based mount. Also fixes a real deadlock: under 0.17 `spawn()` an `unmount_callable()` taken before spawn is dead (spawn `mem::take`s the mount) and `AutoUnmount` deadlocks an explicit umount — the daemon now exits on external unmount / signal.
2. **`feat(backup-fs): hand permission enforcement to the kernel`** — `default_permissions` + `FUSE_ALLOW_IDMAP` (gated on an `idmapped` flag: requesting it without `default_permissions` aborts the FUSE connection, so unprivileged tests stay on a plain mount). Removes `check_access`/`check_sticky`/`is_root_for`/`xattr_access_check` + the `setattr` uid guards + the `idmapped_root` plumbing. Kept: suid/sgid side-effects (now `uid==0`), `creation_gid`, the fh-truncate write check.
3. **`feat(start-core): idmap the container's backup bind`** — the container's bind into the backup fs (`persistent_container.mount_backup`) becomes an `IdMapped` bind `{from_id:0,to_id:100000,range:65536}` (mirroring volumes) instead of a plain bind + chown-to-100000; drops the `idmapped-root=` mount option + `vec![(100000,65536)]`.

### On-disk id-space (please sanity-check)
Mirroring the volume idmap means backup-fs now stores **container-space (0-based)** ownership instead of host-space (100000+). The container's backup procedure copies volume→backup, both 0-based in its view; StartOS (root=0) writing metadata also stores 0-based. **No migration** — the backup format is an unreleased break, so now is the window.

### Validation
- ✅ fuser: build/clippy/53 tests. backup-fs: build/clippy + **all 72 tests** (incl. FUSE mount/remount round-trips). start-core: `cargo check` clean.
- ⛔ **Merge gate — needs a VM:** unprivileged non-idmapped tests can't exercise the kernel-idmap path. Before merge: backup→restore round-trip on a 6.12+ host with an idmapped LXC + ACLs, confirming chmod/chown/sticky-unlink and the ACL-set case (the original bug) all work under kernel enforcement, and that ownership round-trips.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -74,7 +74,7 @@ dependencies = [
  "cfg-if",
  "once_cell",
  "version_check",
- "zerocopy 0.8.48",
+ "zerocopy",
 ]
 
 [[package]]
@@ -2677,19 +2677,22 @@ checksum = "e6d5a32815ae3f33302d95fdcb2ce17862f8c65363dcfd29360480ba1001fc9c"
 
 [[package]]
 name = "fuser"
-version = "0.14.0"
-source = "git+https://github.com/Start9Labs/fuser.git?rev=b6f91ac#b6f91ac6985310b1f4e3ff8075e9468b120e2fe0"
+version = "0.17.0"
+source = "git+https://github.com/Start9Labs/fuser.git?rev=4ef196b#4ef196b4381010046075d1b70e467f129ccd4493"
 dependencies = [
- "clap 4.6.1",
+ "bitflags 2.11.1",
  "libc",
  "log",
  "memchr",
- "nix 0.28.0",
+ "nix 0.30.1",
+ "num_enum",
  "page_size",
+ "parking_lot 0.12.5",
  "pkg-config",
+ "ref-cast",
  "serde",
  "smallvec",
- "zerocopy 0.7.35",
+ "zerocopy",
 ]
 
 [[package]]
@@ -3042,7 +3045,7 @@ checksum = "6ea2d84b969582b4b1864a92dc5d27cd2b77b622a8d79306834f1be5ba20d84b"
 dependencies = [
  "cfg-if",
  "crunchy",
- "zerocopy 0.8.48",
+ "zerocopy",
 ]
 
 [[package]]
@@ -5769,7 +5772,7 @@ version = "0.2.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "85eae3c4ed2f50dcfe72643da4befc30deadb458a9b590d720cde2f2b1e97da9"
 dependencies = [
- "zerocopy 0.8.48",
+ "zerocopy",
 ]
 
 [[package]]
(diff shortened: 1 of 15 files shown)


helix-nine:
## VM validation — authentic production backup ✅

Built the full StartOS `.deb` from this branch (`f63060f`, x86_64 musl) and ran a **real production backup** on a StartOS VM (kernel **7.0.12**, well past the 6.12 FUSE-idmap floor):

- Deployed this branch's `startbox` (start-core) + `startos-backup-fs` onto the VM and confirmed `startd`'s running exe + `/usr/bin/startos-backup-fs` md5s match the branch build (not stock master).
- Stood up a real **CIFS** backup target (Samba), installed a real package (**filebrowser**), and ran `start-cli backup create` — the actual production path: `persistent_container.mount_backup` (the new **IdMapped bind**) → the container's `/backup/create` procedure → `backup-fs` under **kernel idmap + default_permissions** (internal permission model removed) → encrypted store on the CIFS share.
- **Result: clean success.** `perform_backup{filebrowser}:save_and_unmount` completed with **zero** `EPERM`/`EACCES`/`posix_acl`/panic in the logs — i.e. the exact failure class of the original ACL bug did not occur. Real encrypted output landed on the target (`StartOSBackupsV2/<uuid>/crypt/…`, 56 MB) and is readable back (`backup target info` decrypted the metadata via `backup-fs`).

**Still to exercise:** the full **restore** round-trip reads the same `backup-fs` store back through the same idmapped bind into a container's volume — it requires putting the server into recovery/setup mode (a destructive re-provision), which isn't cleanly triggerable from a running server, so it's a follow-up. A dedicated ACL-setting package would also pin the ACL case directly, though the write-side perm path was exercised clean and the offending `xattr_access_check` is gone entirely.

helix-nine:
## VM validation — RESTORE round-trip ✅ (completes the test)

Restored the exact backup from the previous comment back onto a fresh baseline VM (branch binaries redeployed, md5-confirmed) via `package backup restore cifs-0 <pw> filebrowser`:

- The reverse idmap path ran through the branch code: `install:new:mount_backup` (the new **IdMapped bind**) → container restore procedure (`init … "kind":"restore"`) reading from `backup-fs` → written back into the service volume. **Zero `EPERM`/`EACCES`/`posix_acl`/panic** across the whole restore window.
- filebrowser came back `status: installed`, with its volume fully restored (`.../volumes/filebrowser/data/{config,database,data,main}`).
- **Ownership round-tripped correctly** — restored files are `1000:1000` (service user) and `0:0` (root) on the host, i.e. **container-space (0-based)**, exactly the representation this PR switches backup-fs to store. That's the load-bearing format decision, confirmed end-to-end.

**Both legs of the round-trip are now validated on a real 7.0.12 StartOS VM** running this branch: `backup create` (write path) and `package backup restore` (read path), each clean, with ownership preserved through the kernel idmap. The original ACL-EPERM failure class did not occur in either direction, and the offending `xattr_access_check` is gone.

(Deploy caveat unchanged: binaries were bind-mounted over the read-only squashfs rather than installed via `.deb` overlay, because the 32 GB VM lacks scratch space for `chroot-and-upgrade`; the backup/restore flow itself is byte-identical to production.)

dr-bonez (dismissed):

dr-bonez (approved):

PR #3483, opened 2026-07-15
fix(start-os): fix race during check-port

dr-bonez:
## Summary

Fixes the check-port race and improves the Address Requirements UX around it, plus supporting cleanup and refactor.

When a user adds a public domain or enables a public address, the backend runs DNS / port-forward / firewall reachability tests. Many services auto-restart on such a change, so the service is briefly unreachable and those probes fail for a reason unrelated to the user's network. This PR makes the UI aware of that.

### Probe suppression during a restart
- For a **non-SSL** binding (served directly by the service): if the service left the `running` state at any point during the add/enable request cycle and the external check failed, the probe is shown as **untested** (neutral) rather than failed. The setup modal still opens so the user learns the port needs forwarding and can retest once the service is running again.
- An **SSL** binding is fronted by the always-up OS reverse proxy, so its probe stays valid mid-restart and a failure there is genuine — not suppressed.

### Status-aware Address Requirements modals
- The port-forward and IPv6-firewall Test buttons are disabled while a non-SSL binding's service isn't running, alongside a live status note ("Service must be running to perform this test. Current status: …").
- DNS tests are unaffected — they don't depend on the service running.

### Cleanup / refactor
- Finishes removing `openInternally` from `CheckPortRes` (`utils/gua`, `port-check-icon`, mock API).
- Extracts the shared external-reachability test row into `port-check-test.component`, reused by the domain modal (port-forward + IPv6 firewall) and the standalone public-IP port-forward modal.
- Renames `dns.component` → `domain-validation.component` (the class was already `DomainValidationComponent`).
- Adds `getStatusColor()` to `pkg-status-rendering.service`, shared by the status note and the services-list status.

### Service interfaces: list → detail
- The interfaces page is now a navigable list; each row opens a per-interface detail sub-page (`interfaces/:interfaceId`) with a header (back arrow + name + type badge) and the same per-gateway address content the accordion used to expand.

### Docs & i18n
- Updates `interfaces.md` / `clearnet.md` / `public-ip.md` for the reachability-test behavior (which addresses are testable while stopped vs. only while running).
- Adds/updates i18n keys across all five locales.


helix-nine:
The only failing check was **Formatting & Lockfiles** — `make format-check` → `prettier --check .` flagged `projects/start-os/docs/src/interfaces.md`. The new **Actions** row widened that table, so Prettier wanted to re-pad the other rows to match; the content wasn't reflowed before committing.

Pushed a one-line fix (`541757f48`, prettier `--write` on that file — table padding only, no content change). `prettier --check .` is green locally and CI is re-running. Everything else (all binary builds, Debian packages, automated tests, cargo fmt, SDK/web lockfiles) was already passing.

projects/start-os/web/ui/src/app/routes/portal/components/interfaces/addresses/gateway/domain-health.service.ts
```diff
@@ -62,17 +182,34 @@ export class DomainHealthService {
...
 
+      // A non-SSL binding is served directly by the service, so while the
+      // service auto-restarts on an address change its port/firewall probes fail
+      // for a reason unrelated to the user's network — present those as untested
+      // (null → neutral) rather than failed. An SSL binding is fronted by the
+      // always-up OS reverse proxy, so its probe stays valid even mid-restart and
+      // a failure there is genuine; hence the !addSsl guard. The modal still
+      // opens below (a neutralized probe leaves portOk false) so the user learns
+      // the port needs forwarding and can retest once the service is running.
+      if (!ctx.addSsl && ctx.watch?.restarted) {
```
helix-nine: **#1 (correctness, medium).** `ServiceStatusWatch` latches `restarted` from the watch's synchronous on-subscribe emit, and the watch is started just before the mutating request — so if the service isn't `running` at that instant (stopped, or `getInstalledPrimaryStatus` returning `starting`/`task-required` for a service that's actually up), `sawNonRunning` is set immediately with **no** restart caused by this change. This guard then downgrades a genuinely-closed external port from failed → "untested", masking a real forwarding problem. The intent is a `running → non-running` *transition* — consider only latching after having first observed `running` (or `pairwise`). Flagging for your read since you know the intended semantics.

projects/start-os/web/ui/src/app/routes/portal/components/interfaces/addresses/gateway/domain-validation.component.ts
```diff
@@ -393,19 +303,41 @@ export type DomainValidationData = {
...
+  )
+  protected readonly status = computed(() => {
+    const pkg = this.pkg()
+    return pkg ? renderPkgStatus(pkg).primary : 'running'
+  })
+
+  // A non-SSL binding is served directly by the service, so its port forward /
+  // firewall can't be reached while the service is stopped. An SSL binding is
+  // fronted by the always-up OS reverse proxy, so it stays testable.
+  readonly testDisabled = computed(
```
helix-nine: **#2 (correctness, low).** `testDisabled` keys off `renderPkgStatus(pkg).primary !== 'running'`, but `primary` is `task-required` (active critical task) or `starting` (health check) for a service that's up and serving — so the Test buttons get disabled and the "must be running" note shown incorrectly. It also disagrees with the sibling detail page `interface.component.ts:114`, which calls the same service `running` via `getInstalledBaseStatus`. Suggest sharing one definition (likely `getInstalledBaseStatus(...) === 'running'`).

projects/start-os/web/ui/src/app/utils/gua.ts
```diff
@@ -45,16 +45,16 @@ export function dnsAllPass(
...
  */
 export function portAllPass(
   port: T.CheckPortRes | null | undefined,
   portV6: T.CheckPortV6Res | null | undefined,
   gua: string | null,
 ): boolean {
   if (!port) return false
-  const v4Ok = port.openInternally && port.openExternally && port.hairpinning
-  const v6Ok = !gua || (!!portV6?.openInternally && !!portV6?.openExternally)
+  const v4Ok = port.openExternally && port.hairpinning
```
helix-nine, line 57: **#3 (correctness, low).** Follow-on from dropping `openInternally`: previously `openInternally === false` rendered a neutral "port status cannot be determined while service is not running" cue; now the icon shows a red X whenever `openExternally` is false. Since the Test button gates only on *primary status* (a desired/started state, not a real listener check), a service that reads `running` but has crashed / not yet bound now reports "forwarding broken" — a false diagnosis, and the internal-vs-external distinction is no longer recoverable from the API. Worth a conscious "acceptable" sign-off.

package.json
```diff
@@ -37,8 +37,8 @@
     "build-config": "node shared-libs/ts-modules/build-config.js",
     "format": "prettier --write .",
     "format:check": "prettier --check .",
-    "format:toml": "taplo fmt",
-    "format:toml:check": "taplo fmt --check"
+    "format:toml": "git ls-files -zco --exclude-standard -- '*.toml' | xargs -0 taplo fmt",
```
helix-nine: **#5 (correctness, low/latent).** `… | xargs -0 taplo fmt` with no `--no-run-if-empty`: if the file list is ever empty, `xargs` still runs bare `taplo fmt`, which walks per `taplo.toml` and re-triggers the exact openwrt symlink-loop hang the new `exclude` was added to avoid. Never empty at repo root today, so latent — but cheap to harden with `-r`.

shared-libs/crates/start-core/src/net/gateway.rs
```diff
@@ -628,13 +626,6 @@ pub async fn check_port(
         .map(|a| SocketAddr::new(a, port))
         .collect::<Vec<_>>();
```
helix-nine, line 629: **#6 (cleanup — free fix).** With the `open_internally` probe removed just below, this `internal_ips` binding (621-627) is now dead: an `unused_variables` warning plus a pointless `Vec<SocketAddr>` allocation per `check_port`. Safe to delete.

projects/start-os/web/ui/src/app/routes/portal/components/interfaces/addresses/gateway/port-forward.component.ts
```diff
@@ -124,115 +83,71 @@ export type PortForwardValidationData = {
...
+  )
+  protected readonly status = computed(() => {
+    const pkg = this.pkg()
+    return pkg ? renderPkgStatus(pkg).primary : 'running'
+  })
+
+  // A non-SSL binding is served directly by the service, so its port forward
+  // can't be reached while the service is stopped. An SSL binding is fronted by
+  // the always-up OS reverse proxy, so it stays testable.
+  readonly testDisabled = computed(
```
helix-nine: **#7 (cleanup — reuse).** This `pkg` / `status` / `testDisabled` triple is byte-identical to `domain-validation.component.ts:324-339`. Since #2's status rule lives in both, they'll drift — worth extracting a small shared helper (co-located with `DomainHealthService`).

projects/start-os/web/ui/src/app/routes/portal/components/interfaces/addresses/gateway/test-status-note.component.ts
```diff
@@ -0,0 +1,58 @@
...
+})
+export class TestStatusNoteComponent {
+  readonly pkg = input.required<PackageDataEntry>()
+
+  private readonly status = computed(() => renderPkgStatus(this.pkg()).primary)
+  protected readonly statusText = computed(
+    () => PrimaryRendering[this.status()].display,
+  )
+  protected readonly dots = computed(
+    () => PrimaryRendering[this.status()].showDots,
```
helix-nine, line 54: **#8 (cleanup — reuse).** This re-derives statusText/color/dots from a `PackageDataEntry` that `StatusComponent` (`app-status`) already renders. And because dots here come from `PrimaryRendering[...].showDots` while `app-status` uses a hardcoded list (which omits `restoring`), the two already disagree on dots for the same status. Consider composing `<app-status>` for the status portion and keeping only the icon + label wrapper.

projects/start-os/web/ui/src/app/routes/portal/components/interfaces/addresses/gateway/domain-validation.component.ts
```diff
@@ -393,19 +303,41 @@ export type DomainValidationData = {
...
+  // The package's live status, or undefined for an OS interface (empty id). It
+  // gates the port/firewall tests and drives the status shown while they are
+  // unavailable. DNS resolution does not depend on it, so its Test always stays
+  // on. The OS UI carries its own packageId but stays testable via addSsl.
+  protected readonly pkg = toSignal(
+    this.context.data.packageId
+      ? this.patch.watch$('packageData', this.context.data.packageId)
+      : of(undefined),
+  )
+  protected readonly status = computed(() => {
```
helix-nine: **#9 (cleanup — simplify/efficiency).** `status()` is read only by `testDisabled`, and `testDisabled = status() !== 'running' && !addSsl` evaluates `status()` first — so for the common `addSsl` (SSL) case it eagerly subscribes the packageData watch and reruns `renderPkgStatus` per emission for a result that's always `false`. Reorder to `!addSsl && status() !== 'running'` (ideally gate the watch on `!addSsl`), inlining the single-use `status` while you're at it.

helix-nine (commented):
Ran the code-review skill on this diff (7 finder angles → dedup → adversarial verify). Findings are inline; grouped summary below. Nothing here looks merge-blocking — **#1** is the only one with real behavioral bite (and it's a false-*neutral* that masks a failure, not a false-fail), and **#6** is a free fix.

**Correctness**
1. `ServiceStatusWatch` treats an already-not-running service as "restarted" → masks a genuinely-closed port. *(domain-health.service.ts)*
2. `testDisabled` disables Test for a service that's actually up (`task-required` / health-`starting`), and disagrees with `interface.component.ts`. *(domain-validation / port-forward)*
3. Dropping `openInternally` turns "running but not listening" into a false "forwarding broken". *(gua.ts / gateway.rs)*
4. System › Gateways "Port Forwards" modal wasn't updated — same lost cue, no status gating (detail below).
5. `format:toml` xargs has no `--no-run-if-empty` → latent bare-`taplo fmt` hang. *(package.json)*

**Cleanup**
6. Dead `internal_ips` binding. *(gateway.rs)*
7. Duplicated `pkg`/`status`/`testDisabled` triple across the two modals.
8. `test-status-note` re-implements `app-status` (dots already diverge for `restoring`).
9. `status` computed is single-use + wrong short-circuit order for the SSL fast path.

---

**#4 detail** (no line to anchor — the file is untouched by this PR): `projects/start-os/web/ui/src/app/routes/portal/routes/system/routes/gateways/port-forwards.component.ts` reuses the shared `port-check-icon` / `port-check-warnings`, which lost their `!openInternally` branches — so a row whose backing service is stopped now shows a bare red X with no "service not running" cue, and as a multi-service aggregate (only `gatewayId` in context) it got none of the new `testDisabled` / `test-status-note` gating. Defensible as out of scope; flagging so it's a conscious decision rather than a miss.

projects/start-os/web/ui/src/app/routes/portal/components/interfaces/addresses/gateway/domain-health.service.ts
dr-bonez: No - if the service was not running at any point during the request, even intentionally so with no transition, we cannot assert if the port is open, as nothing was listening

dr-bonez: Perhaps the naming is bad though. `restarted `is the wrong term for that case

helix-nine: Fair — withdrawing #1. You're right: if the service was non-running at any point in the request window, nothing was listening, so the external probe can't distinguish "forwarding broken" from "nothing to forward to" — neutral/untested is the correct result regardless of a transition. I'd wrongly assumed the intent was specifically a `running → non-running` transition.

(The one residual case — a service reporting `starting`/`task-required` while actually bound and listening, where the probe *is* valid but still gets downgraded — is really the "primary status isn't a real listener check" point behind #2/#3, so I'll leave that to your read on those rather than relitigate it here.)

projects/start-os/web/ui/src/app/routes/portal/components/interfaces/addresses/gateway/domain-validation.component.ts
dr-bonez: A failing/starting health check could also prevent the check from working, so this heuristic is better

projects/start-os/web/ui/src/app/utils/gua.ts
dr-bonez: Acceptable

projects/start-os/web/ui/src/app/routes/portal/components/interfaces/addresses/gateway/domain-health.service.ts
helix-nine: Agreed — `restarted` reads as a transition, which is exactly what misled me; the field it returns (`sawNonRunning`) already names the real meaning ("was non-running at any point in the request window"). Happy to push a rename — getter `restarted` → `wasNonRunning`, the two `ctx.watch?.restarted` call sites, and the "restart mid-check" comments so they also cover the intentional-stop case. Want me to, or would you rather fold it into your next push? Pick a name if you'd prefer another.

projects/start-os/web/ui/src/app/utils/gua.ts
helix-nine: Acknowledged — that's exactly the conscious sign-off I was flagging for. Nothing further from me on this one.

projects/start-os/web/ui/src/app/routes/portal/components/interfaces/addresses/gateway/domain-validation.component.ts
helix-nine: Fair — a `starting`/failing health check can mean it isn't actually serving, so gating the Test button on `primary === 'running'` is the safer call. Agreed on the heuristic.

The only loose end is consistency: the sibling detail page (`interface.component.ts:114`) uses the looser `getInstalledBaseStatus(...) === 'running'`, so the two pages disagree for a `starting`/`task-required` service. Worth aligning the detail page to this stricter gate (or sharing one helper, per #7) so they don't drift — your call.

MattDHill:
@dr-bonez — review items are addressed and pushed, plus a merge from `master` to clear the conflict. One item I want your call on before doing anything, because it lands in your backend change.

## The gap

Dropping `open_internally` from `CheckPortRes` left **System › Gateways → port-forwards modal** with no way to say "can't tell". It previously rendered a warning triangle off `!openInternally`; now a stopped non-SSL service's row shows a bare red ✗ asserting "not forwarded" when we genuinely don't know. That is the same untrustworthy-negative this PR exists to kill — we just fixed it in the Address Requirements modals by watching the service status client-side, but that trick doesn't transfer here: this modal is an aggregate over every service on a gateway, so there is no single status to watch.

## Why the UI can't fix it alone

The modal already builds its rows from `patch.watch$('packageData')`, so every contributing package's status is already in hand — gating on it costs no new subscription. What's missing is knowing whether a given forward terminates at the always-up OS reverse proxy (`addSsl`) or directly at the service. `PortForward` carries only `src`/`dst`/`gateway`/`count`, and the src/dst ports are the *address's* port, not the binding's internal port — so there's no key back to the binding. Recovering `addSsl` in TypeScript means re-deriving your `bind.enabled_addresses()` → public → port loop (`net/host/mod.rs:461-493`), override semantics and all. That's duplicated logic that will rot silently, and I'd rather not.

I could gate on package status alone (if any contributing package is running, trust the result). It's sound but incomplete: it under-claims on the shared 443 row when every HTTP service happens to be stopped, since the proxy is still up and 443 really is testable. Fails conservatively — a false "cannot determine" rather than a false ✗ — but it's an approximation.

## Proposed real fix

Have `PortForward` record what it terminates at:

```rust
pub struct PortForward {
    src: SocketAddrV4,
    dst: SocketAddrV4,
    gateway: GatewayId,
    count: u16,
    /// Whether this forward terminates at the OS reverse proxy rather than at
    /// the service itself. A proxied forward stays reachable while the service
    /// is stopped; a direct one does not, so a failed external check against it
    /// is meaningless unless the service was running.
    #[serde(default)]
    ssl: bool,
}
```

Set from `bind.options.add_ssl.is_some()` in the single-port loop (`:487`), `false` in the range loop (`:529` — ranges are non-SSL by construction). `#[serde(default)]` for on-disk compat, same as `count`. The UI then reads `pf.ssl || <any contributing package running>` with no re-derivation, and `addSsl` stays the single predicate this whole PR reasons in.

Two things I want your read on:

1. **Is `ssl` the right predicate, or do you want the more general "terminates at an always-up OS process"?** They coincide today; `ssl` is what the rest of the PR uses.
2. **`port_forwards` is a `BTreeSet`, so a new field widens `Ord`.** Two forwards identical but for `ssl` would stop coalescing. That's a port conflict and shouldn't arise, but it's your call whether to exclude it from the ordering.

Happy to implement it, or leave it to you since it's your change — just say which. If you'd rather not widen the type in this PR, I'll take the status-only approximation with the limitation commented, and we can do this properly as a follow-up.

helix-nine:
Review items all look cleanly addressed in `eda9f09bd` (verified at the code level), not blocking from my side:

- `restarted` → `wasNonRunning` (#1); both Address Requirements modals now share `injectTestStatus`, which only subscribes patch-db for a non-SSL binding (#2/#7/#9).
- app-status dots off `PrimaryRendering`, `restoring` included (#8); `xargs -0r` (#5); dead `internal_ips` dropped (#6). Bonus removal of the inert `disabled`/`isRunning` chain closes the consistency gap behind #2.

Two notes, neither blocking:

- The `master` merge correctly pulled in the husky/pre-commit changes from #3488 **and** keeps this branch's `git ls-files … | xargs -0r taplo fmt`. Heads-up that merging therefore flips `master`'s `format:toml` back off #3437's plain `taplo fmt` — intentional here (bare `taplo fmt` walks into the openwrt symlink loops), just flagging so it's a conscious change rather than a silent merge artifact.
- The `PortForward { ssl }` question is yours and @dr-bonez's call — his backend change, and I've nothing to add beyond that the `ssl`/`addSsl` predicate lines up with how the rest of the PR already reasons.

dr-bonez:
We can trivially re-add openInternally, I only removed it because I thought we weren't using it under the new model

helix-nine:
Re-adding `openInternally` looks like the cleaner fix — it's computed backend-side where the binding is known, so it sidesteps the "no key back to the binding" problem entirely (the port-forwards modal already rendered its triangle off `!openInternally`), and with no `PortForward` field there's no `BTreeSet` `Ord` question either.

One concrete ripple: it reverts #6. `open_internally` is `TcpStream::connect(&*internal_ips)`, and that `internal_ips` binding is exactly what `eda9f09bd` removed — so it comes back with the probe.

It's also a *real* internal-listener probe rather than a status inference, which has two nice properties over the status-only fallback: it's accurate where primary status isn't (the residual #2/#3 point), and it handles the shared 443 row @MattDHill flagged — the proxy is listening, so the probe succeeds even when every HTTP service on it is stopped. Same rendering as before (`openInternally === false` → neutral "can't determine", since it also fails while the service is down).

Approach is your + @MattDHill's call — just flagging the #6 interaction so it isn't missed.

helix-nine:
Re-added `openInternally` — pushed `0cd3c25ca` to this branch.

**Backend:** restored the `open_internally` field on `CheckPortRes`/`CheckPortV6Res` and the `TcpStream::connect` probe in `check_port` + `check_gua_port` (`gateway.rs` is now byte-identical to master; bindings restored to match, so the Generated Artifacts gate stays green). Also restored the mock values and the i18n key (774, all five locales).

**UI:** the shared `port-check-icon` (warning triangle) and `port-check-warnings` ("cannot be determined") read `openInternally` again — that's the fix for the System › Gateways port-forwards modal, which was showing a bare red ✗ for a stopped non-SSL service.

**One design call worth your eyes:** I also re-gated the Address Requirements checks (`portAllPass`, and the `portOk` in domain-health / port-forward) on `openInternally`, i.e. the full pre-#3483 behavior. I initially left those on the client-side status-watch alone, but that produced a visible contradiction — a probe with `openExternally: true, openInternally: false` shows the "cannot be determined" warning **while the Done button stays enabled** (and it's reproducible in mock mode, since the mock returns `openInternally: false`). Re-gating keeps the icon/warning and the Done button coherent. The status-watch still composes on top: it neutralizes the probe to *untested* during a mid-request restart, before `openInternally` is even read.

If you'd rather keep `openInternally` **out** of the Address Requirements gates (status-watch only there), say the word and I'll trim it back and instead suppress the icon/warning in the gate-passing case so it's not self-contradictory.

Verified: `cargo check -p start-core`, `tsc` (check:ui), `check:i18n`, and `prettier --check` all pass.

MattDHill:
@dr-bonez Yes — let's do that, it's better than what I proposed. Withdrawing the `PortForward { ssl }` idea: yours is computed backend-side where the binding is known, so it sidesteps the "no key back to the binding" problem entirely, needs no `BTreeSet`/`Ord` widening, and it's a real listener probe rather than my status inference — which also means it gets the shared 443 row right (the proxy is listening, so it reads open even when every HTTP service on it is stopped). My version was worse on every axis.

One distinction I want to nail down before writing it, because it decides whether this stays additive or quietly undoes the PR.

## `openInternally` doesn't replace the client-side watch

It's a **point sample taken before the external probe**, and the gap isn't small — between the two sit up to 2s per subnet in the `mapped_external_ip` loop, then the echo request at 5s per base URL until one answers. A service that restarts inside that window returns `open_internally: true` next to `open_externally: false`, which renders as a red ✗. That's precisely the spurious failure this PR exists to kill.

The client-side watch covers the *whole* request/response window, which is the claim you made settling #1 — "if the service was not running at any point during the request, even intentionally so with no transition, we cannot assert if the port is open, as nothing was listening." A single sample can't make that claim. So the two measure genuinely different things and I'd keep both:

- **Address Requirements modals** — whole-window watch, unchanged.
- **System › Gateways port-forwards modal** — neutral cue off `!openInternally`, restoring exactly what was there. No watch is possible here (it's an aggregate over every service on the gateway), and a point sample beats nothing by a lot.

Flagging it mainly so "we re-added `openInternally`" doesn't later get read as "the watch is redundant."

## Question on sampling order

As written, the probe is the loosest possible bound on "was anything listening during the external check" — it's sampled before, and answers about a moment that may be seconds stale by the time the external result lands. Do you want it moved after the external check, or bracketed on both sides and AND-ed? Tighter, but it's your call whether the extra round-trip is worth it for a cue that's already only advisory.

## Two smaller things

- @helix-nine is right that this partially reverts #6: `internal_ips` was only dead *because* the field was gone, so it comes back with the probe. Expected, no issue.
- `TcpStream::connect` means a UDP-only binding probes `false` regardless — so every port-range row (the RTP/STUN/TURN case) would sit permanently on "can't determine". Since `check_port` is TCP end-to-end that's arguably the honest answer for a UDP forward, but you'd know better whether those rows should say something more specific.

Happy to implement — it's a `#[ts(export)]` change so it needs a bindings regen, which I'd run rather than hand-edit. Or it's yours if you'd rather; just say which so we don't collide.

helix-nine:
Our comments crossed — I'd already implemented and pushed the re-add (`0cd3c25ca`) ~2 min before yours landed, and my note just above flagged this exact gate question. But your point-sample-vs-whole-window framing is the right root cause (better than the "contradiction" symptom I described), so I've corrected the branch to match your decomposition. Pushed `250dd28a0`:

- **Backend:** `open_internally` probe + field restored (`gateway.rs` byte-identical to master; bindings, mock, and i18n key 774 restored too).
- **Address Requirements modals — unchanged:** reverted the `portAllPass`/`portOk` gates back to the status-watch. You're right that gating on the point sample re-introduces the spurious mid-restart failure — it's sampled before the ~2s/subnet `mapped_external_ip` loop + 5s/URL echo, so it can't cover the window the watch does. Rendering there is functionally identical to before (flag off → same branch).
- **Port-forwards modal only:** the neutral `!openInternally` cue (triangle + "cannot be determined") is scoped via a `checkInternal` flag that only that modal sets — the aggregate-with-no-watch case where the point sample beats nothing.

Left for you/@dr-bonez (didn't want to pre-empt):
- **Sampling order** — still sampled *before* the external check (matching master). Moving it after, or bracketing + AND-ing, is dr-bonez's call on whether the extra round-trip is worth it for an advisory cue.
- **UDP forwards** — `TcpStream::connect` reads `false` for a UDP-only binding, as on master. Unchanged.
- Confirmed the #6 revert (`internal_ips` comes back with the probe).

Green: `cargo check -p start-core`, `tsc`, `check:i18n`, `prettier`. The mechanical re-add + scoping is done — if you'd rather take the sampling-order change yourself, go ahead, just flagging so we don't collide.

MattDHill:
Thanks @helix-nine — took the backend re-add in `0cd3c25ca` as-is and built on it. Two notes on `250dd28a0`, which I've partly reversed in `3f04cca19`.

## Kept: portOk stays on the status-watch

You were right to revert this, and I've left it alone. A point sample taken before the external probe can't make the whole-window claim, so gating the Done button on it would have re-introduced the mid-restart failure. `portAllPass` / `portOk` are untouched.

## Reversed: the cue is no longer scoped to the port-forwards modal

Hiding the cue doesn't follow from that argument — gating a *pass/fail decision* on a stale sample is unsound, but *displaying* what the probe measured is not the same act. The case it left uncovered is a service that is running but isn't listening on the port: `primary === 'running'`, so the status-watch sees nothing wrong, and the row rendered a bare red ✗ asserting "not forwarded" when we don't actually know. That's the same untrustworthy negative in a different costume, and it's exactly the gap you identified yourself ("accurate where primary status isn't"). So `checkInternal` is gone and both modals read the probe.

## The reason it looked unsafe: a precedence bug

Worth calling out, because I think it's what made scoping feel necessary. The icon tested `openInternally` before `openExternally`:

```
@if (checkInternal() && !res.openInternally) {   → triangle
} @else if (!res.openExternally) {               → ✗
} @else {                                        → ✓
```

So a port that genuinely answered the external probe still rendered a warning triangle whenever the internal sample happened to be stale — the icon contradicting its own passing check. Un-gating that would have been noisy, correctly. Reversing the order fixes it: if something answered externally the port is reachable, whatever the earlier sample said. That precedence predates this PR — it's master's order too.

## Also fixed: the IPv6 firewall row explained nothing

`domain-validation` passes `portV6Result()` to `[result]` but never to `[warningResult]`, so once the icon read `openInternally` the V6 row could show a triangle with no message next to it. Both now take one union-typed `result`, which drops the `warningResult` input entirely — hairpinning is an IPv4 NAT artifact and narrows on the field rather than needing a second input.

## Copy

774 went back verbatim as *"Port status cannot be determined while service is not running"*. The probe measures whether anything answered a TCP connect — it doesn't establish why, and that wording is wrong for both a running-but-not-listening service and a UDP-only binding, which never answers the TCP probe at all. Now *"Nothing responded on this port, so its status cannot be determined"*, across all five locales.

Verified: i18n check clean (800 keys), `check:ui` clean, and a full AOT build passes — which is what actually typechecks the templates and the new union.

MattDHill (approved):

PR #3488, opened 2026-07-16
ci(format): fast prettier gate + pre-commit hook to catch formatting before CI

helix-nine:
## Why

PR #3483 went red only on **Formatting & Lockfiles** — an unformatted markdown table in `projects/start-os/docs/src/interfaces.md` — while every build and test passed. The repo has **no local formatting gate** (the husky hooks in `.git/hooks` are a stale, inert v4 artifact from the patch-db subproject), so the only thing enforcing prettier is CI, and a missed `make format` isn't visible until the full matrix has already run. This adds two layers so it's caught earlier.

## What changed

**1. CI — fast prettier gate + fail-fast (`.github/workflows/test.yaml`)**
- New `prettier` job: `npx prettier@<pinned> --check .`, ~20s, no docker/npm ci. Runs on every non-draft PR.
- `test` / `format` / `generated` now `needs: [prettier, changes]`, so a formatting slip **short-circuits** them instead of burning the full matrix.
- Dropped the `pull_request` `**/*.md` paths-ignore so **docs-only PRs are now format-checked** (previously they skipped the format job entirely). A new `changes` job keeps the heavy jobs skipping docs-only PRs, mirroring the existing `startos-iso.yaml` idiom (`gh api .../files` + `grep`, gated via `needs` + `!cancelled() && … && (event != PR || changes.output)`). The `push` paths-ignore is kept — a docs-only push to master was already checked on its PR.

**2. Local — pre-commit hook (`.husky/pre-commit`, `.lintstagedrc.json`, `package.json`)**
- husky v9 + lint-staged auto-run prettier on staged files before they land. No-ops when `node_modules` is absent (fresh worktrees), so it never blocks a commit where deps aren't installed — CI stays the source of truth.

## Notes for review

- **Lockfile diff is large but scoped.** It promotes + upgrades the already-hoisted `husky 4.3.8 → 9.1.7` and `lint-staged 13.3.0 → 15.5.2` (both were in the root lockfile transitively via patch-db-client; the `.husky/` hook convention needs husky v9). No unrelated top-level deps changed; `npm ci --dry-run` confirms the lockfile is in sync. patch-db-client keeps its own v4 via its own lockfile/`npm ci`.
- **Cross-workflow limit:** GitHub Actions can't gate the *separate* product-build workflows (`startos-iso`/`start-cli`) on this preflight; they already self-filter by `paths:` and build Rust/images (prettier-independent), so wasted cost there is bounded. The `needs:` gating covers this workflow's own jobs.
- **Branch protection:** the new **Prettier** check runs on every non-draft PR and is a good required check. On docs-only PRs, `test`/`format`/`generated` report *skipped* via the `changes` filter — no worse than today, where they didn't run at all.
- The hook is bypassable (`--no-verify`) and best-effort by design; it's fast feedback, not enforcement.

## Verification

- `prettier --check .` green (incl. all files in this PR); `npm ci --dry-run` reports the lockfile in sync.
- Pre-commit hook dogfooded: reformatting + re-staging of a deliberately-broken markdown table verified end-to-end — **this very commit was made through the hook**.
- `test.yaml` parses; the `changes`/`needs` logic mirrors the proven `startos-iso.yaml` pattern.
- **Not testable locally:** the live GitHub Actions run (job graph, skip behavior on a real docs-only PR). Worth a glance once checks run here.

Context: follows the #3483 formatting-failure discussion.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -5,14 +5,15 @@ on:
     branches:
       - master
       - next/*
+    # A docs-only push to master needs no re-check — the PR that produced it
+    # already ran the prettier gate below. Pull requests intentionally do NOT
+    # ignore *.md (see the `prettier` job) so docs formatting is always checked.
     paths-ignore:
       - '**/*.md'
   pull_request:
     branches:
       - master
       - next/*
-    paths-ignore:
-      - '**/*.md'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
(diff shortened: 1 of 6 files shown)


dr-bonez (approved):

helix-nine:
Thanks for the review. CI is green across the board — the fast **Prettier** gate (28s) and the new **Detect non-docs changes** job both pass, and `Formatting & Lockfiles` / tests / the build matrix ran behind them, so the `needs:` gating and the lockfile upgrade both check out on a real runner.

One follow-up to make the gate actually *enforced*: add **Prettier** to the required status checks for `master` (and `next/*` if protected) in branch protection — otherwise the job runs but a red result wouldn't block a merge. That's an admin action, so I'll leave it to you. Happy to merge whenever you'd like, or leave it in your hands.

PR #3498, opened 2026-07-17
fix(core): unmount tmp mounts only while the slot has no live guard

helix-nine:
## Bug

Backups to a physical drive target intermittently fail with:

```
[startos_backup_fs] could not load backup: Io(Os { code: 2, kind: NotFound, message: "No such file or directory" })
```

Reproduced on both exFAT and BTRFS (so it's not filesystem-specific). Reported by @dr-bonez against `0.4.0-beta.10` (`e59dd8a6`), which already includes #3442, #3439, and the fuser 0.17 migration from #3480.

## Root cause (this is **not** AutoUnmount)

The `could not load backup` line only fires at `mount.backup-fs` **startup** (`BackupFS::new` → `File::create(data_dir/.lock)` → `ENOENT`) — it's a *fresh* load whose source directory is already gone, not a crashed/OOM'd daemon. The OS logs show the source that vanished is the **physical drive** mount (a plain `BlockDev` mount — no FUSE, so AutoUnmount is irrelevant).

Across a full day of logs the signature is dead consistent: a `backup.target.list` scan RO-mounts the drive at its hashed tmp path and releases it, and shortly after a **detached** `unmount:` of that same path (bare span = a `tokio::spawn` from a `Drop`) fires immediately before each `ENOENT`.

The defect is in the `TMP_MOUNTS` guard layer. `MountGuard` teardown unmounts by **mountpoint path**, and for `Drop` it does so from a detached spawned task. The tmp mountpoint is `hash(source)`, stable across remounts, so:

1. `M1`'s last `Arc<MountGuard>` drops → strong count hits 0 (the slot's `Weak` dies **now**) and `Drop` *spawns* the umount (it can't `.await`, so it defers).
2. A concurrent `mount()` for the same path takes the entry lock first, sees `weak.upgrade() == None`, cleans up the stale mount, and **remounts `M2` fresh at the same path**.
3. The deferred umount from step 1 finally runs and unmounts **`M2`** — its target ("umount this path") was frozen before it ever took the lock.
4. The op that just mounted `M2` (or a backup-fs load off it) then hits `ENOENT`.

A per-entry lock alone doesn't fix this: the refcount→0 (which kills the `Weak` that `mount()` reads to decide "remount") happens *before* the deferred teardown acquires the lock, so serialization can't cancel the stale intent.

## Fix

`MountGuard` now carries an `Arc` handle to its own `TMP_MOUNTS` slot. Teardown (both the explicit `unmount` and the `Drop` spawn) takes **just that per-mountpoint slot lock** and unmounts only while the slot has no live guard:

- `slot.weak.strong_count() == 0` ⟺ no live `TmpMountGuard` owns this path → safe to unmount.
- `> 0` ⟺ a live guard owns it and will tear it down itself → skip.

Because *establishing* a mount (mount/upgrade) requires the same slot lock, the count can't grow while teardown holds it — only shrink via concurrent drops, which only ever moves toward "safe to unmount." So a stale/detached umount can no longer tear out a mount another operation just established, whether the drop came from an error path or a normal `Drop`.

No global lock (that would serialize every mount in the system and re-introduce the `IdMapped` re-entrancy deadlock), and no epoch — `strong_count` answers the only question that matters: *is this path owned right now?* Standalone `MountGuard`s (e.g. container rootfs mounts in `persistent_container`) have `slot: None` and are unaffected.

Design worked through with @dr-bonez.

## Testing

- `cargo check -p start-core` — clean (pre-existing warnings only).
- `cargo fmt -p start-core --check` — clean.
- No unit tests cover the guard (it needs real mounts / root). **Still to do:** a VM repro — install a package, back it up to a physical target while spamming the backup-target listing, and confirm the `ENOENT` is gone. I'll follow up with that result before this should merge.


shared-libs/crates/start-core/src/disk/mount/guard.rs
```diff
@@ -73,29 +81,64 @@ impl MountGuard {
...
+/// `strong_count` means another mount now owns this path and will tear itself
+/// down — the slot lock is held across the umount so no remount slips in
+/// between the check and the umount.
+async fn unmount_slot(
+    slot: &Option<MountSlot>,
+    mountpoint: &Path,
+    lazy: bool,
+    delete_mountpoint: bool,
+) -> Result<(), Error> {
+    let _occupancy = match slot {
```
dr-bonez: I prefer an explicit drop at the end. makes it super clear the scope of the guard

helix-nine: Done — dropped the guard explicitly after the umount/delete (134e4540d), and rebased onto your `SyncMutex` cleanup.

dr-bonez (approved):

PR #3501, opened 2026-07-17
fix: auto-forward public IPv4 on addSsl interfaces + tunnel SNI fallback

helix-nine:
## The bug

Enabling a **public IPv4** address on an **SSL** service interface did not trigger automatic port forwarding — the port stayed closed until forwarded by hand. Adding a public **domain** on the same interface worked fine; only the bare IP:port failed, and only for bindings with `addSsl`. (Reported by @dr-bonez.)

## Root cause (StartOS)

In `net_controller.rs::NetServiceData::update`, the IPv4 auto-map path (`forwards` → `InterfaceForwardEntry::update` → `port_map.ensure`) that opens the router pinhole is built **only** in the non-SSL forwards block. For an `add_ssl` port that block is skipped. The bare public IPv4 got a `*` reverse-proxy listener but no pinhole:

- public **domains** → `hostname_maps` → `sync_hostname_mappings` → `ensure_hostname` ✅
- public **IPv6 GUAs** → `gua_pinholes` → `port_map.ensure` ✅
- bare public **IPv4** on an SSL port → **nothing** ❌

It even affected the StartOS admin UI itself (bound `add_ssl` + `secure: None`): its HTTP port auto-forwarded on a public IPv4 but its HTTPS port did not.

## Fix (StartOS)

For each enabled bare public IPv4 at the SSL port, request the pinhole directly — mirroring the IPv6 `gua_pinholes` mechanism: ask the gateway to forward the SSL port to the box's own **LAN IPv4** (the host TLS `*` vhost is the listener, so it's a pure pinhole, no local DNAT). Tracked in a new `HostBinds.ssl_ip_pinholes` set and withdrawn on the same reconcile as the GUA pinholes. Gated strictly to a public IPv4 (not a GUA-only gateway, not a domain).

## Fix (StartTunnel) — required so the OS fix doesn't regress domains

@dr-bonez flagged that a StartTunnel port was strictly **either** a DNAT **or** an SNI route set — so with the OS change, a service with both a public domain *and* a bare public IPv4 on the same SSL port would have its two PCP mappings collide (whichever landed first won; the other errored), regressing the working domain case.

So a hostname-less forward now acts as the **SNI fallback** (the `PortBindings.fallback` slot that existed but was never wired):

- `PortForward::Sni` gains `fallback: Option<SniRoute>` (`#[serde(default)]` → **no DB migration**).
- A hostname-less MAP on an SNI port sets the fallback; a hostname MAP on a lone **same-owner** DNAT promotes it to the fallback (kernel rule torn down, demux listener registered, lease carried to a new `SniFallback` key).
- Traffic whose TLS SNI matches a named route reaches that device; a bare-IP client or unmatched/absent SNI reaches the fallback device.

The DNAT→SNI promotion is **owner-scoped** (`plan_dnat_conversion`): a DNAT is an exclusive whole-port claim, so only the DNAT's own client (target forced to its tunnel IP) may fold it into the fallback — another client's hostname is rejected, as before. *(This guard was added in response to a self-run adversarial review that flagged the unguarded conversion as a cross-peer boundary issue on shared-WAN-IP subnets.)*

Wired through the data-plane demux, PCP/UPnP backend (lease stamping centralized into `apply_peer_forward_range`), lease seed/reap, startup + WAN-IP reconcile, peer teardown (`clear_for_peer`/`gc_forwards`), and the manual API. The tunnel UI lists the fallback as a hostname-less row.

## Versioning / changelog

- **start-os** `0.4.0-beta.10` is unreleased → CHANGELOG entry added in place, no bump.
- **start-tunnel** `v1.1.1` is released → new `## [1.1.2]` + `Cargo.toml`/`Cargo.lock` bump.
- Regenerated `osBindings/tunnel/PortForward.ts` (only the `fallback` field added).

## Testing

- `cargo test -p start-core` — tunnel (36, incl. new `dnat_conversion_is_owner_scoped`, `fallback_register_ownership_and_coexistence`, `sni_fallback_serde_backward_compat`) and net (180) suites pass.
- Prettier + pinned-nightly (`nightly-2026-05-28`) rustfmt clean.
- **Not** exercised here: end-to-end WAN reachability needs a real PCP/NAT-PMP/UPnP gateway (a libvirt NAT gateway won't grant a pinhole), and the `start-tunnel` binary itself needs its web frontend built (`include_dir!`) which isn't built in this environment — CI covers the full web/tunnel build. Manual confirmation on real hardware (or a StartTunnel VM) that the pinhole is emitted/withdrawn is the one remaining step.

### DIFF
--- a/shared-libs/crates/start-core/src/net/vhost.rs
+++ b/shared-libs/crates/start-core/src/net/vhost.rs
@@ -2,7 +2,7 @@ use std::any::Any;
 use std::collections::{BTreeMap, BTreeSet, HashMap};
 use std::fmt;
 use std::future::Future;
-use std::net::{IpAddr, Ipv4Addr, SocketAddr, SocketAddrV6};
+use std::net::{IpAddr, Ipv6Addr, SocketAddr, SocketAddrV6};
 use std::pin::Pin;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::{Arc, Weak};
@@ -14,8 +14,9 @@ use clap::Parser;
 use color_eyre::eyre::eyre;
 use futures::FutureExt;
 use futures::future::BoxFuture;
-use imbl::OrdMap;
+use imbl::{OrdMap, OrdSet};
 use imbl_value::{InOMap, InternedString};
+use ipnet::IpNet;
 use rpc_toolkit::{Context, HandlerArgs, HandlerExt, ParentHandler, from_fn, from_fn_async};
 use serde::{Deserialize, Serialize};
 use tokio::io::{AsyncRead, AsyncWrite, ReadBuf};
@@ -40,7 +41,7 @@ use crate::net::acme::{
 use crate::net::gateway::{
     GatewayInfo, NetworkInterfaceController, NetworkInterfaceListenerAcceptMetadata,
 };
-use crate::net::port_map::PortMapController;
+use crate::net::port_map::{PortMapController, candidate_gateways};
 use crate::net::ssl::{CertBranding, CertStore, RootCaTlsHandler};
 use crate::net::tls::{ChainedHandler, TlsHandler, TlsHandlerAction, TlsListener, TlsMetadata};
 use crate::net::utils::{bind_mio_listener, ipv6_is_link_local, is_private_ip};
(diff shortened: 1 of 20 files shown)


shared-libs/crates/start-core/src/net/vhost.rs
```diff
@@ -429,21 +434,72 @@ impl VHostController {
...
+        }
+        self.sync_port_maps(owner, desired);
+    }
+
+    fn sync_port_maps(
+        &self,
+        owner: HostMapOwner,
+        desired: BTreeMap<
+            (Ipv4Addr, u16),
+            (u16, Vec<(IpAddr, Option<u32>)>, Vec<Option<InternedString>>),
```
dr-bonez: why vecs? can there be duplicates? do we care about order? if not we should use a BTreeSet

helix-nine: Good call — the hostname list is exactly that: no duplicates (I was hand-deduping with `contains()`) and order-independent, so I switched it to a `BTreeSet<Option<InternedString>>` in 303fe3a9d, which also drops the manual dedup.

The gateway list next to it I kept as a `Vec` on purpose: the port-mapper walks it in order and takes the first gateway that binds (`port_map/client.rs` — the `return` on first success), so it's a preference-ordered list, not a set. It's also the type `candidate_gateways` returns and `port_map.ensure` takes.

dr-bonez (approved):

PR #3502, opened 2026-07-18
fix(os-install): refuse preserve selections that would erase the data drive

dr-bonez:
## Summary

A user updating from 0.3.5.1 (single-drive install, everything on one disk) to 0.4.0-beta.9 chose **Preserve** for their data drive while installing the OS to a *different* drive. The installer's existing-pool lookup only matched a VG guid at the *disk* level or at an exact *partition path* — but the setup wizard always submits a whole-disk path for the data drive, and a 0.3.x single-drive install keeps its pool on a **partition** of that disk. The lookup missed, and the code silently fell through to `setup_data_drive()`, which zeroed the partition table and ran `pvcreate`/`vgcreate`/`lvcreate`/`luksFormat`/`mkfs` over the user's 0.3.5.1 pool — destroying their data with no error, then failing later with the cryptic `Not found \`version\` in db` when the empty pool was attached.

Root cause: with `wipe: false`, a missed pool lookup must never create a fresh pool.

## Fix

`plan_data_drive()` resolves the data-drive plan **before any disk is written** and fails fast with an actionable, localized (5-locale) error:

| Existing layout on data drive | OS drive selection | Before | After |
|---|---|---|---|
| Pool on a partition (0.3.x single-drive) | same drive | attach (works) | attach (unchanged) |
| Pool on a partition (0.3.x single-drive) | different drive | **silently reformatted the data drive** | `partitioned-pool-needs-same-drive` error, nothing written |
| Pool spans whole disk (0.3.x data-only drive) | different drive | attach (works) | attach (unchanged) |
| Pool spans whole disk | same drive | mid-install `Cannot protect the entire disk` | up-front `whole-disk-pool-cannot-share-drive` error with guidance |
| No StartOS pool | any + Preserve | silently created a fresh pool | `no-startos-data-to-preserve` error |

The pre-installed-OS flow (no `os_drive`) attaches whole-disk pools as before and refuses partitioned pools (`partitioned-pool-unsupported-layout`). `wipe: true` behavior is unchanged.

As a side benefit, the post-install attach now uses the plan validated pre-install instead of re-resolving VG guids against the stale pre-install disk list.

## Tests / docs

- 8 new unit tests in `os_install::tests` covering each layout × selection combination, including the data-loss case (`preserve_single_drive_split_selection_errors`).
- `update-040.md` Step 10 now warns that the 0.4.0 drive selection must match the 0.3.5.1 layout.
- CHANGELOG entry under `[0.4.0-beta.10]`.

Verified: `cargo check -p start-core`, `cargo test -p start-core os_install --features=test` (9 passed), `make start-core-format`, prettier.

Not covered (manual): an end-to-end USB-install run exercising the new error paths on real hardware.


helix-nine:
Added the setup-wizard companion to the backend guard (commit 16fb5770d).

`DrivesPage.continue()` now refuses the same "Preserve" selections `plan_data_drive` rejects, before the install is submitted. `preserveBlockedReason()` reproduces all six arms of the backend match (whole-disk pool sharing a drive, partition pool needing the same drive, partition pool on a pre-installed device, no StartOS data to preserve), gated to the Preserve path so the Overwrite/wipe flow is untouched. A blocked selection surfaces an actionable dialog instead of proceeding, so the user is corrected in place rather than hitting the RPC error mid-install.

- `isStartOsPoolGuid` uses the same `EMBASSY_`/`STARTOS_` prefix check; the wizard already receives the same `DiskInfo` (disk + partition guids) from `setup.disk.list`.
- Pre-installed devices send `osDrive: null`, so `sameDrive` is always false — routing partition pools to the unsupported-layout message, matching the backend's `os_drive == None` arm.
- Label + four messages added to all five dictionaries (en/de/es/fr/pl), using each locale's own Preserve/Overwrite wording. No interpolation on the frontend i18n, so the messages say "the selected data drive" rather than the backend's `%{disk}`/`%{partition}`.
- Verified: `check:i18n`, `check:setup`/`check:shared`/`check:ui` (strict + strictTemplates), prettier, and the setup-wizard prod build.

Folded the wizard behavior into the existing changelog entry rather than adding a new line. Happy to split this into a standalone PR if you'd rather keep #3502 backend-only.

MattDHill (approved):

PR #3515, opened 2026-07-21
fix(net): scope SSL vhost public_v4 to SSL-port addresses

helix-nine:
## Problem (A6 from the PCP-delay case file)

In `NetServiceData::update`, the `*` SSL vhost's `server_public_v4` filter spanned **every** enabled public bare IPv4 address on the binding, with no port/ssl scoping — its v6 twin already had `a.ssl` (`net_controller.rs`). Consequence: enabling the bare WAN IP for a binding's **plain** port marked the **SSL** vhost public, so StartOS requested a bare upstream pinhole for the SSL port the operator never enabled. Observed live on hillfam: a StartTunnel SNI-fallback (`—`) row appeared on 5349 (coturn) purely because `204.48.27.17:57551` (the plain port) was enabled — and it kept being re-requested even with `turns://204.48.27.17:5349` disabled.

## Fix

Mirror the v6 arm: `ssl_vhost_public_v4` filters `a.public && a.ssl && Ipv4`, so only addresses on the SSL port make the SSL vhost public. Extracted as a small function with a regression test (`plain_port_bare_ip_does_not_publicize_ssl_vhost`).

No changelog entry: the beta.10 "Automatic port forwarding" entry already *claims* this behavior ("a port's forwards and mappings are driven only by the addresses advertised at that exact port") — this makes the code match it. The buggy scoping shipped only in the unreleased beta.10 line.

## Test plan

- `cargo test -p start-core --features=test --lib` — all pass, incl. the new regression test.

### DIFF
--- a/shared-libs/crates/start-core/src/net/net_controller.rs
+++ b/shared-libs/crates/start-core/src/net/net_controller.rs
@@ -26,7 +26,7 @@ use crate::net::host::binding::{AddSslOptions, BindId, BindOptions, UpstreamCert
 use crate::net::host::{Host, Hosts, host_for};
 use crate::net::port_map::{PortMapController, candidate_gateways};
 use crate::net::service_interface::{
-    AddressInfo, HostnameMetadata, ServiceInterface, ServiceInterfaceType,
+    AddressInfo, HostnameInfo, HostnameMetadata, ServiceInterface, ServiceInterfaceType,
 };
 use crate::net::socks::SocksController;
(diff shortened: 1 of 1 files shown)


dr-bonez (approved):

helix-nine:
Approved and mergeable — the only thing keeping the merge state `UNSTABLE` is the cross-arch + ISO builds still running; unit tests, Prettier, and formatting/lockfile checks are already green. Nothing outstanding on my end, so this is clear to land once those finish. Thanks for the review, @dr-bonez.

PR #3516, opened 2026-07-21
fix(start-tunnel): honor deletes of SNI fallback forwards

helix-nine:
## Problem

Found while diagnosing the PCP forward-delay case on hillfam. Since 1.1.2, a hostname-less forward on an SNI-held port becomes the port's **fallback** — but both delete paths only match `PortForward::Dnat`:

- **PCP lifetime-0 delete** (RFC 6887 §15): `remove_peer_forward` scanned only DNAT entries, so deleting a fallback was a **silent no-op that still returned SUCCESS** (`net/port_map/server/mod.rs:520-531`). The client believed the exposure was withdrawn; the fallback kept serving until its lease lapsed (up to an hour).
- **UPnP `DeletePortMapping`**: `remove_forward_by_source` was likewise DNAT-only → SOAP fault 714.

This is what made the live debugging confusing: the client's port-map actor removed its key and sent a delete, the tunnel acknowledged it, and the `—` fallback row stayed put — indistinguishable from a client-side bug.

## Fix

Both paths now also match an SNI entry whose **fallback target** is the requesting peer's `(peer, internal_port)` and clear it via the existing `remove_sni_fallback` (demux unregister, lease forget, db cleanup — same teardown the reaper uses). Named SNI routes on the port are untouched; another client's fallback is never matched. Matching logic extracted as `peer_forward_matches` with unit tests.

Bumps start-tunnel to **1.1.3** with a changelog entry (1.1.2 is tagged on origin, so this cuts a new heading).

## Test plan

- `cargo test -p start-core tunnel::forward --features=test` — pass, incl. new `peer_delete_matches_dnat_and_own_fallback` (own fallback matches, another client's doesn't, route-only SNI port doesn't, DNAT parity).
- Full lib suite green.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7899,7 +7899,7 @@ dependencies = [
 
 [[package]]
 name = "start-tunnel"
-version = "1.1.2"
+version = "1.1.3"
 dependencies = [
  "include_dir",
  "start-core",
--- a/projects/start-tunnel/CHANGELOG.md
+++ b/projects/start-tunnel/CHANGELOG.md
@@ -5,6 +5,17 @@ All notable changes to StartTunnel are documented here.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.1.3]
+
+### Fixed
+
+- **Deleting a fallback forward now actually deletes it.** A PCP delete
+  (lifetime 0) or UPnP `DeletePortMapping` for a hostname-less forward that had
+  become an SNI port's fallback reported success but silently left the fallback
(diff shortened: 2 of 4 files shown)


dr-bonez (dismissed):

dr-bonez (approved):

helix-nine:
CI is green now, and the red wasn't in the change itself: the `start-tunnel` `1.1.2 → 1.1.3` bump in `Cargo.toml` wasn't mirrored in `Cargo.lock`, so every `--locked` job died at lockfile resolution (`cannot update the lock file … because --locked was passed`) before compiling anything. Synced the lone lockfile entry in a24b4f4e — all checks now pass (tests, manpages, and the full binary/deb matrix). Thanks for the re-review; ready to merge whenever.

PR #3517, opened 2026-07-21
fix(install): pace writeback + preallocate instead of O_DIRECT for s9pk transfers

helix-nine:
## Why

Certain package installs stall — or briefly freeze the whole box — on some devices but not others, consistently across versions of the same package, with a *different* package id per device on identical hardware and the same network. That signature points at **device-local btrfs free-space fragmentation**, not the network.

Two mechanisms compound it, and both are on the `s9pk` transfer path:

1. **No writeback pacing.** Downloading and unpacking a large `s9pk` streams straight through the page cache. Dirty pages accumulate and then flush in one enormous final `fdatasync` — a seconds-to-minutes lock-up whose victim varies by package size and each box's on-disk layout.
2. **No preallocation on CoW btrfs.** The temporary download and the installed archive are written with no `fallocate` into whatever fragmented free space btrfs hands out, scattering a single file across thousands of extents; the concurrent read-while-download re-read then thrashes.

Note the earlier O_DIRECT download writer (`DirectIoFile`) was *not* the culprit here — the stall predates and postdates the switch to buffered writes (#3217), so the common factor is fragmentation + the final flush, which O_DIRECT vs buffered both suffer. O_DIRECT was also incompatible with mirror-resume (it can't seek/truncate).

## What

Replace the download **and** sideload O_DIRECT writer with a buffered `tokio::fs::File` fronted by a new `WritebackPacer` (`util/writeback.rs`):

- **`sync_file_range`** kicks async writeback every 16 MiB and caps un-synced data at a 64 MiB window (`SYNC_FILE_RANGE_WAIT_*`), so dirty RAM stays bounded and the final `fdatasync` is always cheap — this replaces the sole reason O_DIRECT was added (avoiding the giant end-of-stream flush), without O_DIRECT's downsides.
- **`fallocate(FALLOC_FL_KEEP_SIZE)`** reserves the full size up front (when Content-Length is known) so writes land contiguously.
- **`chattr +C`** (nodatacow) on the temp dir (inherited by `download.tmp`/`upload.tmp`) and the installed `s9pk`, so large writes don't fragment via CoW.

The file stays a plain seekable `tokio::fs::File`, so `DownloadHandle`'s mirror-resume (`set_len(0)`/`seek`, range-resume) is untouched, and the concurrent archive reader keeps hitting the page cache. Linux-only mechanics; `pace`/`preallocate` are no-ops on other targets (start-cli on macOS), where the kernel's own writeback throttling bounds dirty memory.

## Testing

- `cargo check -p start-core` clean; pinned-nightly `fmt` clean.
- All 5 `registry::asset::tests::buffered_download_*` pass — including resume-after-truncation, range-ignored restart, mirror fallback, and **serves reads before download completes** (the read-while-write contract).
- Linux-only local check; the `libc` calls (`sync_file_range`, `fallocate`) are cfg-gated with non-Linux no-ops and `AsRawFd`/`RawFd` are available on all Unix targets, so the darwin/musl matrix should be unaffected — worth a matrix confirm.

Changelog updated under `0.4.0-beta.10` (Fixed).


helix-nine:
Reviewed, simplified, and cleaned this up — pushed as `1eb0cc4`. Ran an adversarial multi-agent pass over the writeback/preallocate/nodatacow mechanics; here's what came out.

### Fixed
- **btrfs NOCOW ordering (real bug).** On the install/sideload path in `service_map.rs`, `chattr +C` ran *after* `fallocate` — but btrfs only honors the NOCOW inode flag on an empty file, so preallocating first left those extents copy-on-write and quietly defeated the anti-fragmentation goal on exactly the write this PR targets. Reordered to `create → set_no_cow → preallocate → write`. (The temp-file path was already correct — files inherit `+C` from the `TmpDir` at creation.)
- **`set_no_cow` cross-platform + log level.** It shelled out to `chattr` unconditionally and logged failures at `error`, so start-cli on macOS (no `chattr`) and any non-btrfs host would log an error on every `TmpDir::new()`. Made it linux-only with a no-op stub (matching `preallocate`/`pace`), and dropped the expected off-btrfs miss to `debug`.
- **dead_code on darwin.** cfg-gated `KICK_EVERY`/`DIRTY_WINDOW` to linux (only referenced from the linux `pace`), so the darwin matrix build stays warning-free.
- **DRY.** Folded the two identical "set expected size + preallocate" blocks in `download_from` into one `set_expected_size` helper, and added a one-liner on the 206 branch explaining why resume needs no seek or pacer reset.

### Considered, left as-is
- `Progress.written` now bumps on page-cache write rather than on-sync (old O_DIRECT `bytes_synced`) — that's the *correct* semantics for the buffered read-while-write contract, not a regression.
- The `spawn_blocking(fd)` in `pace` is sound: `pace` awaits the blocking task while holding `&mut self`, so the `File` can't drop underneath it.
- Redundant `preallocate(total)` on repeated 206 resume retries — `fallocate` is idempotent/cheap and only fires on retry; a dedup field wasn't worth the state.
- Changelog entry already describes the shipping (now-correct) behavior; nothing unreleased to amend.

### Verification
`cargo check -p start-core --tests` clean (no new warnings), nightly `make start-core-format-check` clean, all 5 `registry::asset::tests::buffered_download_*` pass (resume-after-truncation, range-ignored restart, mirror fallback, all-mirrors-fail, reads-before-complete). Linux-only local check — the darwin/musl matrix is still worth a CI confirm.

dr-bonez (approved):

PR #3520, opened 2026-07-21
fix(core): tolerate transient failures querying clock-sync status

stupleb:
## Problem

`check_time_is_synchronized()` runs `timedatectl show -p NTPSynchronized`, which reaches systemd-timedated over D-Bus and can fail transiently (e.g. a D-Bus activation timeout under boot load). Both call sites treat a failed query as fatal, with disproportionate consequences:

- **Boot → Diagnostic Mode.** The init sync-clock loop calls the check once per second (~31 calls minimum per boot) and propagates any `Err` out of `init()`; `start_init.rs` catches any init error by dropping the server into Diagnostic Mode. A momentary D-Bus hiccup during boot becomes the scariest screen the OS has, for no durable reason.
- **Stale "Clock sync failure" warning.** The recovery poller spawned in `RpcContext::init` `.unwrap()`s the check every 30 seconds. One failure panics the task — nothing observes its JoinHandle — so `serverInfo.ntpSynced` can never flip to `true`, and the UI warning sticks for the rest of the boot even after time actually synchronizes.

Context: surfaced while triaging a 0.4.0-beta.9 field report of a persistent **Clock Sync Failure** warning (a network where the regional pool.ntp.org zone serves dead servers). The dead-poller mode is one candidate explanation for the warning refusing to clear on a box whose clock is otherwise fine.

## Fix

Treat a failed query as "not synchronized yet": log it (warn + debug) and retry on the existing cadence (1 s during init, 30 s in the poller). No happy-path change; `check_time_is_synchronized()` keeps its `Result` signature — tolerance belongs to the callers. A persistent failure during init now runs out the existing ~31-second no-progress window and boots normally with the warning shown, instead of aborting.

Includes the new i18n keys in all 5 locales and a changelog entry under `0.4.0-beta.10` **Fixed**.

## Verification

- `cargo check -p start-core --locked` passes in a `rust:1-bookworm` (Debian stable rust) container, which also exercises the compile-time i18n key validation. (macOS host; the native toolchain here predates `edition2024`, and the `start9/cargo-zigbuild` image did not fit the local Docker VM.)
- `rustfmt --check` (stable defaults) clean on the two touched files. The repo's pinned-nightly-only options (`group_imports`, `imports_granularity`) affect only `use` statements, which this diff does not touch; CI's containerized format gate remains the authority.
- prettier 3.8.3 `--check` clean on the changelog and locale file.
- Not runtime-tested on a device: this hardens an error path (transient `timedatectl`/D-Bus failure) that is hard to reproduce on demand. The Diagnostic-Mode consequence is established by code trace (`init()` `?` → `setup_or_init` → `DiagnosticContext::init`).

### DIFF
--- a/projects/start-os/CHANGELOG.md
+++ b/projects/start-os/CHANGELOG.md
@@ -243,6 +243,16 @@ file tracks notable changes since the move to the monorepo.
   targets, setup-wizard drive lists, `start-cli disk list`), and a sub-gigabyte
   partition that does remain shows its capacity in MB instead of rounding down
   to "0 GB".
+- **A transient failure querying clock-sync status no longer aborts boot or
+  strands the clock-sync warning.** StartOS asks systemd (`timedatectl`) whether
+  NTP has synchronized — once per second during the boot-time sync wait, then
+  every 30 seconds in the background until the first sync lands. Both callers
+  treated a failed query as fatal: during init it aborted boot into Diagnostic
+  Mode, and in the background poller it silently killed the polling task,
+  leaving the "Clock sync failure" warning up for the rest of the boot even
+  after time synchronized. A failed query (e.g. a D-Bus activation timeout
+  under boot load) is now treated as "not synchronized yet" — logged and
+  retried on the existing cadence.
 
 ### Removed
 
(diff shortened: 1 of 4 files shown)


dr-bonez:
a .log_err() would be simpler and sufficient

dr-bonez (approved):

PR #3526, opened 2026-07-21
feat(auth): replace cookie auth with signature auth for the web UI

### LINKED ISSUE #3511: Web UI lockout: any hostile or malformed cookie on the UI origin collapses into error 34 → forced logout, and re-login silently bounces
## Summary

Session validation treats the browser's entire `Cookie` header as all-or-nothing and collapses **every** failure — unparseable header, foreign cookie, even database errors — into `ErrorKind::Authorization` (34), which the web UI interprets as "log out immediately." Because service UIs share the server hostname's cookie jar (cookies ignore ports), any co-hosted service webapp — or any upstream reached through the `/proxy` route, which passes `Set-Cookie` through — can plant a cookie that makes **every authenticated RPC fail with 34**. Login itself skips cookie validation, so it keeps succeeding: the user gets bounced back to the login screen with **no error message**, in a loop, with no indication of what is wrong. Recovery requires knowing to clear cookies for the origin.

## Observed incident (0.4.0-beta.10)

Real-world report, LAN access via `https://<hostname>.local` (Firefox):

1. User is browsing the StartOS UI and interacting with services. Suddenly the login screen appears; console shows `Unauthenticated, logging out` (the code-34 handler in `embassy-live-api.service.ts:757`).
2. Repeated login attempts with the correct password "do nothing" — no error shown, user ends up back at the login form. (Mechanism: `auth.login` succeeds and sets a fresh cookie; the first authenticated RPC after `setVerified()` — `db.subscribe` — returns 34; `setUnverified()` navigates back to `/login`, superseding the login page's own navigation. Nothing is displayed.)
3. The login page also logs `Cookie with an empty name and an empty value has been rejected` on every attempt — that is `login.page.ts:123` executing `document.cookie = ''`, a browser-rejected no-op (see defect 5).
4. SSH access unaffected. The cookie jar for the origin contained **11 cookies, 10 from co-hosted service UIs** (Nextcloud `nc_*`, `authelia_session`, `__Secure-next-auth.session-token`, JWT `token`/`refresh_token`/`auth`, and a `__stripe_mid` scoped to `Domain=.<hostname>.local` — demonstrating that services do set domain-wide cookies in practice).
5. Clearing all cookies for the origin and logging in once resolved it.

The specific culprit cookie was not captured before the jar was cleared. The defects below are independent of it — each is verifiable in current master, and together they turn "one weird cookie appeared" into "silent total lockout."

## Root-cause chain (all refs to current master)

1. **Whole-header strict parse, any failure → 34.** `ValidSessionToken::from_header` (`shared-libs/crates/start-core/src/middleware/auth/session.rs:202-224`) runs `HeaderValue::to_str()` (fails on any non-visible-ASCII byte anywhere in the header) and `basic_cookies::Cookie::parse` over the **entire** header; both failures map to `ErrorKind::Authorization`. `basic-cookies` 0.1.5's lexer (its `cookie_lexer.rs`, char classes at ~212-236) rejects comma, space, backslash, double-quote, and all non-ASCII in cookie values — characters real-world apps do put in cookies and browsers happily store and resend. One such cookie from any co-hosted app ⇒ every authenticated request fails.
2. **First-match-only on the `session` name.** `session.rs:213` takes the *first* cookie named `session` and never tries later candidates (same pattern at `session.rs:141` on the logout path). Cookie identity is (name, domain, path) — port excluded — so a service app on another port using the very common name `session` with `Path=/` either shadows or outright overwrites the StartOS cookie.
3. **All failures collapse into one error code the UI treats as fatal.** `session.rs:214` (`if let Ok(s) = …`) swallows *any* `from_session` error — including patch-db write errors from the per-request `last_active` bump — and falls through to generic `Authorization` (`session.rs:220-223`). The middleware composite additionally overwrites errors so browsers always see SessionAuth's 34 regardless of which layer failed (`middleware/auth/mod.rs:84-94`). The UI maps 34 → `setUnverified()` → wipe local storage, force logout (`embassy-live-api.service.ts:757`).
4. **Login bypasses the poisoned path.** For `login: true` requests, SessionAuth never touches the cookie header (`session.rs:303-320`), so login succeeds while everything else fails — producing the silent bounce loop instead of a visible error.
5. **The shipped mitigation is a no-op.** The login page runs `this.document.cookie = ''` before each attempt (`web/ui/src/app/routes/login/login.page.ts:123`) — browsers reject an empty-name/empty-value cookie write (it's the exact console warning in the incident), and it could never clear `HttpOnly` service cookies anyway.
6. **The `/proxy` route injects upstream cookies into the UI origin.** `proxy_request` strips `cookie` from the outgoing request (`net/static_server.rs:57`) but returns the upstream response verbatim — `Set-Cookie` included (`net/static_server.rs:253-264`). Any registry/CDN reached through it can plant cookies directly on the StartOS origin, no service tab required.
7. **Weak cookie attributes invite collision.** The session cookie is `session=<tok>; Path=/; SameSite=Strict; Expires=…` (`session.rs:343`) — generic name, no `HttpOnly`, no `Secure`, no `__Host-` prefix.

## Ruled out while investigating

- Code regression: the auth middleware and `auth.rs` are byte-identical from `v0.4.0-beta.9` to master (blob diff across the monorepo path move).
- Disk-full / patch-db write failure masquerading as 34: refuted for this incident (login's own db write kept succeeding — its Set-Cookie kept landing, datable via the browser's 400-day `Expires` cap).
- Frontend auto-logout loops: `auth.logout` is only called from the header menu button.

## Suggested fixes

- **Parse leniently, per cookie.** Split the header on `;`, trim, and extract candidates named `session` without ever failing the whole header on one bad pair. (Or swap `basic-cookies` for a lenient parser.)
- **Try every `session` candidate** against the store instead of only the first.
- **Reserve 34 for "token not in store."** Map header-parse failures and db errors to distinct kinds so the UI doesn't force-logout on non-auth failures; log the underlying error server-side.
- **Harden the cookie:** `__Host-session` (implies `Secure`, `Path=/`, no `Domain` — all compatible with current usage) + `HttpOnly` (the UI never reads the cookie from JS; it receives the token in the login response body). Costs one re-login on upgrade; eliminates the collision/overwrite class entirely.
- **Strip `Set-Cookie` from proxied responses** (`/proxy/{url}` and the s9pk proxy route).
- **Frontend:** remove the no-op `document.cookie = ''`; when a login succeeds but the next RPC returns 34 within a few seconds, surface an explicit error on the login page ("Login succeeded but the session was rejected — try clearing cookies for this site") instead of silently returning to the form.

## Environment

- StartOS 0.4.0-beta.10 (`Cargo.toml` `0.4.0-beta.10`, master line; auth code unchanged since beta.9, `document.cookie=''` predates the monorepo reorg)
- Firefox, LAN `.local` origin; services exposed as ports on the same hostname

helix-nine:
Resolves #3511.

Replaces cookie-based authentication for the web UI (start-os and start-tunnel) with per-request Ed25519 signatures over enrolled device keys.

## How it works

- `auth.login` still takes the master password, but instead of issuing a session cookie it **enrolls a public key** the client generated. The login request itself is signed with the matching secret key, so enrollment proves possession.
- Every RPC and authorized static-asset request is then signed: the signature commits to a fixed protocol tag, timestamp, nonce, body size, and `blake3(body)`, plus the **server identity** (hostname/IP/domain) in the signed bytes — so a signature for one server can't be replayed against another. Pure Ed25519, so it's WebCrypto-compatible.
- Enrolled keys are **managed like sessions were**: each carries its enrolled-at/last-active times and user agent, shows up in System → Active Sessions (and `auth session list`), can be revoked (`kill` / logout), and is reaped after 30 days idle. Rate limiting on login is unchanged.
- All existing sessions are signed out on upgrade; each device logs in once more.

## Notable details

- The browser client uses `@noble/curves` (already a workspace dep) for deterministic cross-browser behavior; byte-level interop with the Rust verifier is locked by an embedded test vector, including a wrong-context negative case.
- **Backwards compatibility:** request signatures in the previous Ed25519ph+context scheme are still accepted on verify, so deployed CLI and tunnel-device clients keep working. Package/registry protocol signatures (which use that scheme with a fixed protocol context) are untouched.
- Signature context now includes the server's own IPs, so signature auth works when the server is addressed by IP rather than hostname.
- The session middleware, token store, and cookie issuance are removed; on-box local auth (the `local` cookie) is unchanged. Replay-protection nonces are now actually recorded (they previously were not).
- The unused `/proxy` relay is removed; s9pk static-asset routes authorize by signature.
- `start-cli` login generates (if needed) and enrolls the developer key; on-box CLI continues to use local auth.
- start-tunnel is bumped to v1.2.0 for the auth change.

## Verification

- `cargo check -p start-core --all-targets` clean; `cargo test -p start-core` 522+52 pass; `cargo test -p startwrt-core` 473 pass
- `cargo fmt --all --check`, `npm run check` (i18n + all web projects), prettier, taplo — clean


Dominion5254 (dismissed):

Dominion5254:
Not really an issue for start-wrt, but this looks like a problem for start-os:

Legacy session cookie survives the upgrade as a permanently valid, JS-readable credential

#3511 renames the session cookie (session → startos-session, now HttpOnly) but does not invalidate what it leaves behind. After upgrading, every previously logged-in browser still holds the old cookie, and the token inside it is still accepted by the server.

▎ The problem: the pre-rename cookie was set with Expires: 9999 and without HttpOnly, and StartOS never expires sessions server-side (they're only removed by explicit logout/kill). Cookies are host-scoped, not port-scoped — so JavaScript in any co-hosted service UI on the same hostname can read the legacy cookie via document.cookie, indefinitely, and its token remains a valid StartOS session forever (replayable as startos-session=<token>). One XSS in one service UI — or one malicious package UI — yields a permanent OS session. The changelog's claim that "the rename logs every existing session out" is only cosmetic: users must re-login, but the old credentials stay live.

Fix (both halves):

1. Revoke server-side: clear all existing sessions in the upgrade migration (this also makes the changelog sentence literally true).
2. Purge client-side: alongside the new cookie, emit Set-Cookie: session=; Path=/; Max-Age=0 so browsers drop the legacy name instead of broadcasting it to every port of the hostname forever.

(start-wrt has the same pattern but it self-heals: the old wrt cookie carried Max-Age=86400 and wrt sessions expire after 1 day of inactivity, so exposure ends ~24 h after upgrade — and it leaks only to loopback uhttpd, which is already inside the trust boundary.)

Dominion5254:
Actually it is just up to 30 days for legacy cookie survival, not permanent; but probably a gap worth closing

helix-nine:
@Dominion5254 good catch on the old approach — but note this was written against the pre-rewrite branch (the `startos-session` rename no longer exists here). In the current version the session middleware, the session token store, and all cookie issuance are deleted outright: no code path accepts `session`/`startos-session` as a credential, so the token inside a legacy cookie authenticates nothing after upgrade. That subsumes both halves of your fix — there's no session store left to revoke, and a purge isn't needed because the string isn't a live credential. (Only the on-box `local` auth cookie remains, unchanged, file-based for CLI-over-SSH.)

dr-bonez (dismissed):

dr-bonez (approved):

PR #3533, opened 2026-07-23
feat: constant-time install backups via btrfs subvolume snapshots

MattDHill:
# feat: constant-time install backups via btrfs subvolume snapshots

## Problem

The install backup taken before every service install/update
(`snapshot_volumes_for_install`) shells out to `cp -a --reflink=always`. A
reflink clone copies no data, but FICLONE is O(extent count) — and a large,
aged database volume on CoW btrfs accumulates enormous extent maps. Observed on
a dev box updating monerod 0.18.4.6 → 0.18.5.1: the 263 GB LMDB volume had
**47,455,315 extents** (avg ~5.8 KB), and the clone pegged the box for **19
minutes** while the update sat frozen at 201/251 ("Updating", no progress
ticks — the finalization phase is indeterminate and contributes 0 to the
overall bar).

The same design also fails in the opposite direction: btrfs refuses FICLONE
between `nodatacow` and datacow inodes, and Bitcoin Core's package chattrs `+C`
its data dir — so **bitcoind's install backup has never been created** (EINVAL
on `blocks/index/LOCK`, WARN in the journal on every update attempt).

Finally, a crash between the Installed db-commit and `remove_install_backup`
orphans the backup forever — nothing enumerates `volumes/`, so it silently
pins extents on disk.

## Change

- **`volumes/<pkg>` becomes a btrfs subvolume.** The backup is then an atomic
  `btrfs subvolume snapshot`: milliseconds and O(1) space at creation,
  regardless of size, fragmentation, or nodatacow flags — and finally a true
  point-in-time snapshot (the file-by-file clone never was). Its position in
  the update flow is unchanged (before the old service exits), so rollback
  semantics are identical.
- **New `util/btrfs.rs`** — `is_btrfs` / `is_subvolume` (inode 256 + statfs
  magic), `create_subvolume` / `snapshot_subvolume`, and a subvolume-aware
  `delete_tree` (recursive `btrfs subvolume delete`, so removal is handed to
  the background cleaner instead of an O(extents) `rm -rf`; plain dirs still
  use `remove_dir_all`). Used by backup removal, restore, and uninstall.
- **New `util/clone.rs`** — a reflink tree walker replacing the `cp` shell-out
  for volumes that are still plain directories. Clones regular files in 1 GiB
  `FICLONERANGE` chunks (real byte progress even inside one huge file),
  matches the source's `+C` flag on the destination before cloning (fixes the
  bitcoind EINVAL), preserves owners/modes/times/xattrs/hardlinks/symlinks/
  fifos, and **never falls back to a byte copy** — on failure the operation
  aborts cleanly to the existing no-backup warn path. Linux-only; darwin gets
  a stub (the CLI never runs this path).
- **One-time boot conversion with real progress.** A new "Optimizing storage"
  phase in `cleanup_and_initialize` runs **before `services.init`** — the only
  point where no container (and no cross-service dependency mount) can hold a
  volume — and converts each installed package's plain-dir volume root into a
  subvolume via reflink clone + rename swap, reporting bytes on the boot
  screen. Interrupted conversions are recovered from filesystem state
  (`.convert-tmp` / `.convert-old` sweep), re-checking on later boots costs a
  stat per package, and per-package failure degrades that package to the
  legacy clone path rather than failing boot.
- **Backup repair & sweep** in the same boot step: an interrupted
  `restore_volumes_from_install_backup` (crash between deleting the live
  volume and renaming the backup in) leaves the backup as the only copy of the
  data — boot now completes the rename. Backups orphaned by a package's
  removal are deleted; backups beside mid-install/update packages are left for
  load recovery; a backup beside a healthy Installed package is left alone
  (indistinguishable from a stranded rollback point) and gets removed as stale
  on that package's next update.
- **Legacy fallback keeps the UI honest.** If a plain-dir volume still needs
  the clone backup at update time (non-btrfs dev dirs, or a dir recreated
  behind our back), the finalization phase now gets byte units/total/progress
  instead of freezing at ~80%.

## Space safety

No path byte-copies volume data. Subvolume snapshots share the b-tree itself
(O(1) space at creation); the clone walker is FICLONE-or-fail. The only real
consumption is CoW divergence while a backup exists — bounded by what the
service writes during the update window, freed when the backup is deleted.
Verified empirically during the monerod incident: 166 G free before, during,
and after a 263 GB "copy".

## Notes for review

- The conversion swap (`rename live → .convert-old`, `rename .convert-tmp →
  live`, delete old) is crash-safe via the boot sweep: mid-swap crash renames
  the original back; post-swap crash deletes the leftover.
- `ensure_volume_root` creates new packages' volume roots as subvolumes at
  install and load; `Bind::pre_mount` can still create a plain dir if a volume
  vanishes at runtime — the boot converter picks those up on the next boot.
- Bitcoin Core's `+C` flag is preserved through backup/conversion (dir flag
  propagated so inheritance matches).

## Verification

Diagnosed and measured on a dev box (0.4.0-beta.10): extent count via
filefrag, clone duration from journal timestamps, disk usage via btrfs fi df.
Reviewed by a multi-agent adversarial pass (4 dimensions, every finding
independently verified); confirmed findings — including a libc const that
doesn't exist (FS_NOCOW_FL is defined locally now), a boot-sweep rule that
could have deleted the only copy of a volume after an interrupted restore,
add-only nocow mirroring, and nested-subvolume snapshot gaps — are all fixed
in this revision. Known residual: the legacy clone (plain-dir volumes only) is
not atomic against a concurrently-writing service, same as the `cp` it
replaces; it degrades to warn + no-backup. CI compiles the full cross-target
matrix (linux gnu/musl, darwin stubs).

### DIFF
--- a/shared-libs/crates/start-core/src/util/btrfs.rs
+++ b/shared-libs/crates/start-core/src/util/btrfs.rs
@@ -1,0 +1,112 @@
+use std::path::{Path, PathBuf};
+
+use tokio::process::Command;
+
+use crate::prelude::*;
+use crate::util::Invoke;
+
+/// Subvolume roots always have this inode number (BTRFS_FIRST_FREE_OBJECTID).
+const SUBVOL_INO: u64 = 256;
+
+#[cfg(target_os = "linux")]
+pub async fn is_btrfs(path: impl AsRef<Path>) -> bool {
+    let path = path.as_ref().to_owned();
+    tokio::task::spawn_blocking(move || {
+        nix::sys::statfs::statfs(&path)
+            .map(|s| s.filesystem_type() == nix::sys::statfs::BTRFS_SUPER_MAGIC)
+            .unwrap_or(false)
+    })
+    .await
+    .unwrap_or(false)
+}
+
+#[cfg(not(target_os = "linux"))]
+pub async fn is_btrfs(_path: impl AsRef<Path>) -> bool {
+    false
+}
+
+pub async fn is_subvolume(path: impl AsRef<Path>) -> bool {
+    use std::os::unix::fs::MetadataExt;
+    match tokio::fs::metadata(path.as_ref()).await {
+        Ok(m) if m.is_dir() && m.ino() == SUBVOL_INO => is_btrfs(path).await,
+        _ => false,
+    }
+}
+
+pub async fn create_subvolume(path: impl AsRef<Path>) -> Result<(), Error> {
+    Command::new("btrfs")
+        .args(["subvolume", "create"])
+        .arg(path.as_ref())
+        .timeout(Some(std::time::Duration::from_secs(60)))
+        .invoke(ErrorKind::Filesystem)
+        .await?;
+    Ok(())
+}
+
+/// Creates a writable point-in-time snapshot of `src` at `dst`. Constant-time
+/// and constant-space regardless of the subvolume's size or fragmentation.
+pub async fn snapshot_subvolume(src: impl AsRef<Path>, dst: impl AsRef<Path>) -> Result<(), Error> {
+    Command::new("btrfs")
+        .args(["subvolume", "snapshot"])
+        .arg(src.as_ref())
+        .arg(dst.as_ref())
+        .timeout(Some(std::time::Duration::from_secs(60)))
+        .invoke(ErrorKind::Filesystem)
+        .await?;
+    Ok(())
+}
+
+/// Lists subvolumes nested anywhere below `path` (exclusive), deepest first.
+pub async fn nested_subvolumes(path: &Path) -> Result<Vec<PathBuf>, Error> {
+    use std::os::unix::fs::MetadataExt;
+    let mut found = Vec::new();
+    let mut stack = vec![path.to_owned()];
+    while let Some(dir) = stack.pop() {
+        let mut entries = tokio::fs::read_dir(&dir)
+            .await
+            .with_ctx(|_| (ErrorKind::Filesystem, lazy_format!("read dir {dir:?}")))?;
+        while let Some(entry) = entries.next_entry().await? {
+            let m = match entry.metadata().await {
+                Ok(m) => m,
+                Err(_) => continue,
+            };
+            if m.is_dir() {
+                if m.ino() == SUBVOL_INO {
+                    found.push(entry.path());
+                }
+                stack.push(entry.path());
+            }
+        }
+    }
+    found.sort_by_key(|p| std::cmp::Reverse(p.components().count()));
+    Ok(found)
+}
+
+async fn delete_subvolume(path: &Path) -> Result<(), Error> {
+    for nested in nested_subvolumes(path).await? {
+        Command::new("btrfs")
+            .args(["subvolume", "delete"])
+            .arg(&nested)
+            .timeout(Some(std::time::Duration::from_secs(60)))
+            .invoke(ErrorKind::Filesystem)
+            .await?;
+    }
+    Command::new("btrfs")
+        .args(["subvolume", "delete"])
+        .arg(path)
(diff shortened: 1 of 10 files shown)


shared-libs/crates/start-core/src/util/btrfs.rs
```diff
@@ -0,0 +1,112 @@
...
+                stack.push(entry.path());
+            }
+        }
+    }
+    found.sort_by_key(|p| std::cmp::Reverse(p.components().count()));
+    Ok(found)
+}
+
+async fn delete_subvolume(path: &Path) -> Result<(), Error> {
+    for nested in nested_subvolumes(path).await? {
```
dr-bonez: Nested subvolumes would cause more problems than just failing deletion. Better to just fallback to `rm -rf` if `subvolume delete fails, and lose the complexity of the recursive walk

shared-libs/crates/start-core/src/util/clone.rs
```diff
@@ -0,0 +1,354 @@
...
+    _ctr: Arc<Counter>,
+) -> Result<(), Error> {
+    Err(Error::new(
+        eyre!("reflink clone is not supported on this platform"),
+        ErrorKind::Filesystem,
+    ))
+}
+
+#[cfg(target_os = "linux")]
+mod linux {
```
dr-bonez, line 56: Needs tests

dr-bonez (commented):
A lot of redundancy / paranoia here. If we're enforcing subvolumes on boot, then we only need to handle the case where they are there. The fallback should be minimal, and it can degrade all the way to no-snapshots.

shared-libs/crates/start-core/src/util/btrfs.rs
helix-nine: Done in 4052c33d0 — the recursive walk is gone. `delete_tree` now just tries `btrfs subvolume delete` and falls back to a plain recursive remove on failure.

shared-libs/crates/start-core/src/util/clone.rs
helix-nine: Added in 4052c33d0: five tests covering contents/modes/xattrs, symlinks and fifos, hardlink preservation, nodatacow mirroring, and multi-chunk clones. Tests that touch regular files probe for reflink support first and skip on filesystems without it (CI), so they run for real on btrfs and harmlessly skip elsewhere; the symlink/fifo tree makes no FICLONERANGE calls and runs everywhere.

helix-nine:
Took over and simplified per review, in 4052c33d0 on top of a rebase onto current master (the branch was conflicting):

- **Fallback is now minimal and degrades to no-snapshots.** Since subvolumes are enforced at boot (conversion) and at install (`ensure_volume_root`), `snapshot_volumes_for_install` no longer handles plain directories at all: subvolume → atomic snapshot, anything else → warn + no backup. This removes `legacy_clone_backup`, the nested-subvolume probe on the snapshot path, and the byte-progress plumbing through install finalization (`service_map.rs` is back to master's version verbatim). `clone_tree` remains, but only as the engine for the one-time boot conversion, which keeps its byte progress.
- **`btrfs::delete_tree` lost the recursive walk.** It tries `btrfs subvolume delete` and falls back to a plain recursive remove on failure, accepting that nested subvolume roots (which package volumes should never contain) may survive.
- **`clone.rs` has tests** — see the inline reply.

Net: −100 lines across `btrfs.rs`/`volume.rs` relative to the previous head, and the CHANGELOG entries now describe the simplified behavior. Verified: `cargo check`, full `cargo test -p start-core` (531 passed, clone tests run for real on btrfs and skip cleanly on tmpfs), `make start-core-format-check`. Darwin gating is unchanged from the commit whose CI passed both darwin targets.

dr-bonez (approved):

PR #3543, opened 2026-07-23
fix(ui): snake invisible-wall deaths and held-key input lag

Dominion5254:
## Summary
Fixes the two bugs behind "snake dies when it didn't hit anything", found by
auditing the game after a report of collisions with nothing visibly there.

## Invisible wall (51a914fcc)

The game's `#111` background sat on a full-width `.game-container` with the
transparent canvas centered inside it, while wall collisions use the canvas
dimensions. The canvas is sized as
`min(floor(containerWidth / 40), floor(0.55 * innerHeight / 26))` grid cells,
so on most window shapes the height cap makes it narrower than the container —
leaving same-colored, out-of-bounds strips left and right. The snake died at
the invisible canvas edge, seemingly mid-field, and the death frame draws the
head fully clipped, so it never appears to touch anything.

Fix: the container now shrink-wraps the canvas (`inline-size: fit-content;
margin-inline: auto`), making the visible box edge the collision boundary.
Grid sizing measures the host instead of `canvas.parentElement`, since the
container's width now derives from the canvas. The overlays also now align
with the playfield instead of the wider dialog.

## Input-queue lag (446dbbb3c)

Every keypress while playing was queued — including OS auto-repeat (~30 ms),
which fires faster than the ~50 ms game tick — and `update()` consumed exactly
one entry per tick, with an inapplicable move (no-op repeat, attempted
reversal) still burning the slot. Holding a key grew the queue without bound
at ~0.4 s of lag per second held; panic-mashing added a tick of delay per
redundant press. Turns then executed late and the snake plowed into walls the
player had steered away from.

Fix: input goes through a single `queueMove()` — arrow keys only, consecutive
duplicates dropped (a held key contributes one entry), buffer capped at 3 —
and `update()` drains until the first applicable move instead of consuming
blindly. Turn semantics are unchanged: reversals still rejected, and the
deliberate two-press corner buffer still works.

Known and deliberately left: a mid-game window resize still desyncs the grid
(food becomes uneatable until the next restart). Rare trigger, no unfair
death, self-heals — not worth the fix for an easter egg.

Changelog: two entries under the unreleased 0.4.0-beta.10 `### Fixed`.

Verified: `npm run check:ui`, prod `npm run build:ui`, prettier clean; both
fixes confirmed by playing the game against the mock backend.

### DIFF
--- a/projects/start-os/CHANGELOG.md
+++ b/projects/start-os/CHANGELOG.md
@@ -351,6 +351,13 @@ file tracks notable changes since the move to the monorepo.
   snake hit the (invisible) boundary well inside the dark box — a seemingly
   causeless game over. The dark box now shrink-wraps the playfield exactly, so
   its edge is the wall.
+- **Snake turns no longer lag behind held or mashed keys.** Every keypress —
+  including keyboard auto-repeat, which fires faster than the game advances —
+  was queued and consumed one per tick, so held or redundant presses delayed
+  the next real turn by a tick each; a key held for a couple of seconds banked
+  a near-second of input lag, reading as the snake ignoring a well-timed turn.
+  Redundant presses are now dropped and stale ones skipped, so a turn lands on
+  the tick after its keypress.
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


MattDHill:
Please remove from the CHANGELOG as not to clutter it up

Dominion5254:
> Please remove from the CHANGELOG as not to clutter it up

done

MattDHill (approved):

PR #3552, opened 2026-07-24
feat(backup): show an "Initializing" progress phase during target mount

helix-nine:
## Problem

During a create backup, the progress display sat at a bare **0% with no phases** from the moment the request was accepted until `perform_backup` added the per-package phases — i.e. the entire window that mounts the target and opens (and, on a large/slow target, replays) its encrypted store. `assure_backing_up` seeded `backup_progress` with an empty `FullProgress`, and the real phases only appeared once the mount finished.

## Change

Add an **Initializing** phase that covers that window:

- `backup_all` creates the `FullProgressTracker` up front, adds an `Initializing` phase (started), and seeds `backup_progress` with its snapshot in the same mutate that atomically claims the "backing up" slot.
- The progress→db stream is started **before** the encrypted mount, so `Initializing` is live during it; its `NonDetachingJoinHandle` is moved into the background backup task so it isn't aborted when the request returns.
- The tracker + phase handle are threaded into `perform_backup`, which `complete()`s `Initializing` once the store is open, then adds the reclaim / per-package / OS-data phases (unchanged) to the same tracker.

Result: instead of a blank 0%, the UI shows an `Initializing` phase while the target mounts / the store opens, then transitions into the per-package phases.

Notes:
- The `ctx.db.mutate` closure runs under `catch_unwind`, so it can't capture the tracker (its `watch` channels aren't unwind-safe) — the mutate seeds a plain `FullProgress` **snapshot** instead.
- `Initializing` is added with no overall-percentage contribution, so it reads as a labeled spinner and doesn't skew the overall bar; the per-package/OS contributions are unchanged.
- New i18n key `backup.bulk.initializing` (all five locales).

## Testing

`cargo check -p start-core` and `cargo fmt` clean. Progress reporting isn't unit-tested here (needs a mounted target + running services), so the phase transition is best confirmed on a device / VM: start a backup and watch the progress show `Initializing` during the mount, then the package phases.

### DIFF
--- a/shared-libs/crates/start-core/src/backup/backup_bulk.rs
+++ b/shared-libs/crates/start-core/src/backup/backup_bulk.rs
@@ -345,11 +345,6 @@ async fn perform_backup(
     let mut package_backups: BTreeMap<PackageId, PackageBackupInfo> =
         backup_guard.metadata.package_backups.clone();
 
-    // The target is mounted and its store is open; close out the Initializing
-    // phase before the reclaim/per-package/OS phases take over. The db-sync task
-    // (spawned in backup_all) keeps streaming as these phases are added.
-    init_phase.complete();
-
     let mut reclaim_phase = if crate::backup::trash::has_trash(disk_guard.path()).await {
         Some(progress.add_phase(
             InternedString::intern(&*t!("backup.bulk.reclaiming-space")),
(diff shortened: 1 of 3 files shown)


shared-libs/crates/start-core/src/backup/backup_bulk.rs
```diff
@@ -308,7 +345,11 @@ async fn perform_backup(
     let mut package_backups: BTreeMap<PackageId, PackageBackupInfo> =
         backup_guard.metadata.package_backups.clone();
 
-    let progress = FullProgressTracker::new();
+    // The target is mounted and its store is open; close out the Initializing
+    // phase before the reclaim/per-package/OS phases take over. The db-sync task
+    // (spawned in backup_all) keeps streaming as these phases are added.
+    init_phase.complete();
```
dr-bonez: after the per-package phase handles are added

helix-nine: Good call — moved `init_phase.complete()` below the reclaim/per-package/OS `add_phase` calls, so the phase list is already populated when Initializing flips to complete (no window where it's the only, completed phase). Pushed.

dr-bonez (approved):

PR #3557, opened 2026-07-24
fix(core): accept plugin-provided addresses as signature auth identities

helix-nine:
Targeted at 0.4.0.1 — the entry goes under the existing untagged heading cut in #3564.

## The bug

`RpcContext::sig_context()` yields the identities a request signature may be bound to: the account hostnames (`<hostname>`, `<hostname>.local`), the admin host's `publicDomains` and `privateDomains`, the gateways' subnet IPs and `wan_ip`, and `localhost`.

Addresses a url plugin exports are not among them. `url.export-url` resolves `host_for(db, start-os, HostId::admin())` — the server's own host — and writes a `HostnameInfo { metadata: Plugin { .. } }` into `serverInfo.network.host.bindings.<port>.addresses.available`, a store nothing feeds back into the identity set.

The browser signs `window.location.hostname` (`AuthKeyService::signHeader`), so a UI loaded at a plugin address signs a string the server has no candidate for: `verify_request_signature` fails against every known identity and the request is rejected with `ErrorKind::Authorization`. Under the cookie auth this worked, cookies being host-scoped.

## The fix

Use the binding enabled addresses as the source of truth for allowed hostnames

## Worth a second opinion

This lets any installed `url-v0` plugin add a string to the server's accepted identity set. The narrow consequence: a signature made for server B is replayable against server A if a plugin on A registered B's hostname *and* the same key is enrolled on both. Per-origin browser keys make that a non-event for the UI, but `start-cli`'s developer key is deliberately shared across servers. Reachable only with a hostile plugin plus interception, and accepting these addresses is the point of the feature — flagging it rather than deciding it.

## Verification

- `cargo check -p start-core` clean
- `cargo test -p start-core --features=test --lib` — 531 passed, incl. a new test pinning the collection
- `make start-core-format-check` and prettier clean

### DIFF
--- a/projects/start-os/CHANGELOG.md
+++ b/projects/start-os/CHANGELOG.md
@@ -21,13 +21,6 @@ file tracks notable changes since the move to the monorepo.
 - **Installing onto a pre-installed Raspberry Pi keeps the data pool you pick.**
   Selecting the data pool by partition path now preserves that choice through
   installation.
-- **The UI now works when reached at an address a plugin provides.** Request
-  signatures are bound to a server identity, and the set the server recognizes
-  as itself covered its hostnames, domains, `localhost`, and interface
-  addresses — but not the addresses plugins add for it, which appear under
-  **Plugin** in the server's address list. Loading the UI at one of those
-  addresses left every request unauthorized. They now count as server
-  identities like the rest.
 
 ## [0.4.0]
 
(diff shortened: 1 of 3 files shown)


helix-nine:
Rebased onto master. #3564 answers the version question I'd left open — root `package.json` is the source of truth and 0.4.0.1 is already cut, so the changelog entry now sits under that existing heading instead of creating one, and no manifest change is needed here. Retested on the new base: 531 pass.

dr-bonez (dismissed):

dr-bonez (approved):

PR #3558, opened 2026-07-24
feat(net)!: assignedSslPort carries every TLS external port

helix-nine:
Per the Matrix thread: `assignedSslPort` should hold the external port for *all* SSL ports, not just `addSsl`.

## What changed

A binding with `secure: {ssl: true}` and no `addSsl` — the container serves its own certificate and StartOS DNATs the port straight through — kept its external port in `assignedPort`, the plaintext field. The old split was "did we allocate a TLS-terminating port"; it's now "does this external port speak TLS":

| binding | before | after |
| --- | --- | --- |
| plaintext | `assignedPort` | `assignedPort` |
| `addSsl`, plaintext upstream | both | both |
| `addSsl` + `secure.ssl` (we rewrap) | `assignedSslPort` | `assignedSslPort` |
| `secure.ssl`, no `addSsl` (own cert) | **`assignedPort`** | **`assignedSslPort`** |

`BindOptions` gains `serves_own_tls` / `preferred_ssl_port` / `wants_plain_port`, so the allocator, `update_addresses` and the forward wiring stop restating the rule in five slightly different forms. The self-TLS domain address (#3130) was `else`-chained on `assignedSslPort.is_none()`, which this makes false — un-chained and re-keyed. The "Non-SSL forwards" block becomes the direct-forward block: every external port with no listener of ours in front, which is what it always meant, so it covers the self-TLS port too and no longer needs `or_not_found`.

Emitted addresses are unchanged for every family — the old code already flagged these `ssl: true` out of the plain branch, and the `map_or` defaults it carried were only reachable when `secure` is null. One deliberate exception: the mDNS SSL leg is now gated on there being an mDNS gateway, matching the plaintext leg, so a self-TLS binding doesn't gain a `.local` address it never had.

## The one deliberate asymmetry

`AvailablePorts`' bool is **not** set for a self-TLS port. Its four readers all use it as "one of our TLS listeners answers here, so a domain can be advertised there and SNI-routed" — a DNAT'd port isn't that. Flagging it would let package A's DNAT capture a domain that package B advertises on the same preferred port. Documented on the struct.

## No package has to change

`sdk.host.getBridgeAddress` (#3560) resolves the binding's *derived* address, not these fields, so for a self-TLS binding it resolves the identical result before and after — same hostname, same port number, still flagged `ssl: true`. A code search across `Start9Labs` and `Start9-Community` finds zero `*-startos` packages reading `net.assignedPort` directly. Its docstrings did describe the old field mapping, so those are corrected here (`GetHostInfo.ts`, `StartSdk.ts`, `service-to-service.md`) — no SDK library code changed, so no version bump; whether to cut 2.0.10 to get the corrected docstrings onto npm is your call. This follows #3567, which changed the same docstrings without one.

## Migration

Folded into the existing (still untagged) `0.4.0.1` node rather than cutting an 0.4.0.2, per the changelog/tags rule in root `AGENTS.md`. It moves the persisted port between fields for `secure.ssl && addSsl == null` bindings across the server host and every package host. **The number is preserved**, so bookmarks, router forwards and the per-address `enabled`/`disabled`/`guaWan` overrides (keyed by port) all keep working. `availablePorts` needs no edit — these ports were already flagged `false` and stay `false`. Idempotent, with a true inverse in `down`.

`update()` now releases both ports before reallocating, so a binding that *changes* how its port is served keeps its number, carrying it between the fields. Without that, dropping `secure.ssl` in a package update handed the binding a fresh random port.

## Rebase note

The first push carried its own 0.4.0.1 version bump; #3564 has since cut 0.4.0.1 properly (root `package.json` as the source of truth, `build/env/version.sh`, the `current_matches_manifest` test) so all of that is dropped — including the `start-core` `Cargo.toml` bump, which `VERSION_BUMP.md` now says to leave alone on a revision. `should_welcome_to_release` landed independently on master with the same predicate. What's left is the net change, the migration, and the doc corrections.

## Not fixed here

`is_ssl`-derived domain advertisements aren't recomputed for other hosts when a port's role changes, so a binding that advertised a domain on someone else's preferred port can go stale. Pre-existing (the old allocator reached the same state one step later, via a third package reclaiming the freed port) and it needs cross-host recomputation — I have a patch for it and can raise it separately.

## Verified

`cargo test -p start-core --features=test` — 586 pass, including new coverage for all four binding shapes, the port-number carry in both directions, an idempotent re-bind not migrating onto a freed preferred port, and the migration. `make start-core-format-check` and prettier 3.8.3 clean. `manage-release.sh pre-check start-os` passes every content check (changelog heading, docs release link, crate label, tag free).

### DIFF
--- a/projects/start-os/CHANGELOG.md
+++ b/projects/start-os/CHANGELOG.md
@@ -10,6 +10,18 @@ file tracks notable changes since the move to the monorepo.
 
 ## [0.4.0.1]
 
+### Changed
+
+- **A service that serves its own TLS certificate now reports its external port
+  as an SSL port.** Every interface whose external port speaks TLS — whether
+  StartOS terminates it or the service presents its own certificate — now
+  carries that port in `assignedSslPort`, and `assignedPort` means a plaintext
+  port. The port number itself is unchanged, so existing addresses, bookmarks
+  and router port-forwards keep working. Packages resolve a dependency's address
+  with `sdk.host.getBridgeAddress`, which is correct under either arrangement;
+  see
+  [Service-to-Service Networking](https://docs.start9.com/packaging/service-to-service.html).
+
 ### Fixed
 
 - **The over-the-air update to 0.4.0 boots on the Server Pure.** The Server
--- a/projects/start-os/CHANGELOG.md
+++ b/projects/start-os/CHANGELOG.md
@@ -28,6 +28,12 @@ file tracks notable changes since the move to the monorepo.
   `sdk.host.getBridgeAddress`, which is correct under either arrangement; see
   [Service-to-Service Networking](https://docs.start9.com/packaging/service-to-service.html).
 
+- **The StartOS web interface holds ports 80 and 443.** StartOS runs as root, so
+  its own interface is the one binding that may claim the privileged range, and
+  it now does so through the same port allocator every service uses. HTTPS was
+  already served on 443; the plaintext address — offered only over loopback and
+  the service bridge — moves from a random high port to 80.
+
 ### Fixed
 
 - **The over-the-air update to 0.4.0 boots on the Server Pure.** The Server
--- a/projects/start-os/CHANGELOG.md
+++ b/projects/start-os/CHANGELOG.md
@@ -13,13 +13,19 @@ file tracks notable changes since the move to the monorepo.
 ### Changed
 
 - **A service that serves its own TLS certificate now reports its external port
-  as an SSL port.** Every interface whose external port speaks TLS — whether
-  StartOS terminates it or the service presents its own certificate — now
-  carries that port in `assignedSslPort`, and `assignedPort` means a plaintext
-  port. The port number itself is unchanged, so existing addresses, bookmarks
-  and router port-forwards keep working. Packages resolve a dependency's address
-  with `sdk.host.getBridgeAddress`, which is correct under either arrangement;
-  see
+  as an SSL port, and StartOS serves it like one.** Every interface whose
+  external port speaks TLS — whether StartOS terminates it or the service
+  presents its own certificate — now carries that port in `assignedSslPort`,
+  and `assignedPort` means a plaintext port. A self-TLS port is now answered by
+  the StartOS SNI router, which pipes the raw TLS stream to the service with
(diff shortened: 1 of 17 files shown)


helix-nine:
Rebased onto master (`e777da755`). Two of the merged PRs did require changes:

**#3564 (cut 0.4.0.1)** supersedes the whole version-bump half of this PR, and did it better — root `package.json` as the source of truth, `build/env/version.sh`, the `current_matches_manifest` test. Dropped all of it: both `Cargo.toml`s, `Cargo.lock`, `package.json`/lock, the three man pages, `manage-release.sh`, `installing-startos.md`, and `version/mod.rs`. Notably `shared-libs/crates/start-core/Cargo.toml` must stay `0.4.0` — the rewritten `VERSION_BUMP.md` says to leave it alone on a revision bump since it feeds all four products' man pages, and my bump had been dragging start-registry's and start-tunnel's along with it. `should_welcome_to_release` landed on master with the same predicate I'd written, so that's dropped too.

Since `start-os/v0.4.0.1` isn't tagged yet, the migration folds into the existing `v0_4_0_1` node and the changelog entry goes under the existing heading, rather than cutting an 0.4.0.2.

**#3560 (`sdk.host.getBridgeAddress`)** turns out to need no behavioral change: it resolves the binding's derived address rather than these fields, so for a self-TLS binding it resolves the identical address before and after — same hostname, same port, still flagged `ssl: true`. Its docstrings did encode the old field mapping though ("a passthrough binding only \`assignedPort\`"), so those are corrected in `GetHostInfo.ts`, `StartSdk.ts` and `service-to-service.md`. While there I fixed the adjacent "only one of those is ever populated" — that's wrong on master today, since an `addSsl` binding with `secure: null` carries both; #3567 corrected that sentence in `GetHostInfo.ts` and missed the two copies.

No SDK library code changed, so no version bump — same call #3567 made. If you'd rather the corrected docstrings reach npm, that's a 2.0.10 cut and I'm happy to do it.

Fleet check: zero `*-startos` packages in either org read `net.assignedPort` in code (only prose in three AGENTS/README files, all describing shapes this doesn't change).

dr-bonez (approved):

PR #3573, opened 2026-07-25
fix(tunnel): publish ports on a VPS whose public address is NAT'd at the edge

### LINKED ISSUE #3565: [bug]: StartTunnel 1.2.1 — port forwarding is impossible on a NAT'd VPS: subnet set-wan is uninvokable and upgraded installs are left with wanIp: null
### Prerequisites

### Server Hardware

Lenovo M920q 32gb RAM 2 TB M.2

### StartOS Version

0.4.0

### Client OS

MacOS

### Client OS Version

15.7.7

### Browser

Firefox

### Browser Version

Version 152.0.5 (64-bit)

## Current Behavior

On a VPS behind 1:1 NAT (where the public IPv4 is assigned at the provider edge and not bound to any local interface—the typical AWS, Oracle, GCP, or Azure model), no new port forwards can be created via the CLI or web UI after upgrading. Both setup paths fail identically:

* **CLI:**
  ```bash
  $ sudo start-tunnel port-forward add --label "LND" 9735 10.59.203.3:9735
  Network Error: no WAN IP available for device 10.59.203.3
  ```
* **Web UI (Add published port, IPv4):**
  ```text
  RPC ERROR: Network Error no WAN IP available for device 10.59.203.3
  ```

Two independent defects combine to make this state entirely unrecoverable for the end user:

### Defect A: The subnet `wanIp` is null after migration
A `db dump` reveals that `wanIp` is explicitly `null` on the gateway, the subnet, and every downstream device. This occurs even though `get-available-ips` actively returns a perfectly usable local interface IP (`172.16.0.12`). 

While all relevant port-forward schema migrations successfully executed (`PortForwardAuto`, `PortForwardEntry`, `PortForwardKind`, `ReclaimPort80Redirect`, `WgClientKind`), none of them backfilled the new `wanIp` field. Port forwards that existed prior to the upgrade continue to function normally because they carry the external IP embedded directly in their entry keys, allowing their DNAT rules to remain present and pass traffic. As a result, only new additions fail.

### Defect B: `subnet … set-wan` is uninvokable
The `subnet ... set-wan` CLI administrative command—which is intended to manually correct a missing `wanIp`—is broken at the argument-parsing level. Its own `--help` page emits a duplicate `<SUBNET>` positional argument:

```text
Usage: start-tunnel subnet <SUBNET> set-wan [OPTIONS] <SUBNET>
```

Because both the parent `subnet <SUBNET>` argument and the nested `set-wan <SUBNET>` argument attempt to deserialize into the exact same database key structure, any physical command invocation fails:
* **Providing the subnet once:** 
  ```text
  error: the following required arguments were not provided: <SUBNET>
  ```
* **Providing the subnet twice:** 
  ```text
  General error: error: Serialization Error: duplicate key: subnet
  ```

Consequently, there is no valid structural syntax to invoke this command.

---

## Absence of Workaround

Direct manual patching via `db apply` is entirely blocked. The utility expects an undocumented, `jq`-like expression format (where the parser enforces tokens like `def` / `if` / `(` / `[` / `..` / `{` / `-` / `.`). 

Neither a broad full-document replacement nor a hyper-targeted JSON-pointer operation will validate against the internal structural model, consistently resulting in schema rejection messages:
```text
missing field 'webserver': result does not match database model
```

As it stands, end users hosting on a NAT'd cloud VPS have zero supported paths to manually or programmatically set `wanIp`.

### 1. Fix CLI Argument Inheritance for `set-wan`
The `set-wan` command must correctly inherit the `<SUBNET>` positional argument from its parent context without requiring a second, duplicate positional argument at the tail.

* **Target Syntax:** `subnet <SUBNET> set-wan --wan-ip <IP>`
* **Expected Action:** Inherit the `<SUBNET>` parameter and successfully update the subnet's WAN IP.

### 2. Backfill `wanIp` via Migration
The database upgrade migration routine must be updated to prevent existing installations from entering an unusable state.

* **Detection Logic:** Automatically scan and extract the `wanIp` from the active network interface.
* **Fallback Logic:** If interface detection is blocked, extract and backfill the `wanIp` using the embedded IP addresses found inside legacy `port-forward` entries.

### 3. Implement Fallback Resolution for Port Forwarding
Both the CLI `port-forward add` command and the Web UI **Add published port** dialog must support NAT'd VPS setups.

* **Dynamic Lookup:** If the `wanIp` state field evaluates to `null`, the application must query `get-available-ips` to dynamically resolve it.
* **Validation Rule:** Prevent hard-failing with a network error when the necessary IP configuration data is already discoverable on the system.

### 1. Confirm Environment and Available Interfaces
Run StartTunnel 1.2.1 on a Debian 13 VPS whose public IPv4 is NAT'd and not present on the NIC (only a private IP on the interface).

```bash
$ ip -4 addr show
2: ens3: ... inet 172.16.0.12/20 ... scope global dynamic ens3

$ sudo start-tunnel web get-available-ips
[ "172.16.0.12", "<global-IPv6>", "<link-local-IPv6>", "127.0.0.1", "::1" ]
```

### 2. Attempt to Add Port Forward via CLI
Try to add a port forward using the command-line interface.

```bash
$ sudo start-tunnel port-forward add --label "LND" 9735 10.59.203.3:9735
```
**Verbatim Error:**
```text
Network Error: no WAN IP available for device 10.59.203.3
```

### 3. Attempt to Add Port Forward via Web UI
Navigate to **Port Forwards** → **Add published port** and enter the details:
* **Protocol:** IPv4
* **External Port:** 9735
* **Internal Port:** 9735
* **Target:** Server device

**Verbatim Error:**
```text
RPC ERROR: Network Error no WAN IP available for device 10.59.203.3
```

### 4. Attempt to Set Subnet WAN IP (Both Forms Fail)
Try to manually set the subnet WAN IP so the forward can resolve. Both syntax forms fail to invoke properly.

**Form A (Missing argument error):**
```bash
$ sudo start-tunnel subnet 10.59.203.1/24 set-wan --wan-ip 172.16.0.12
```
**Verbatim Error:**
```text
error: the following required arguments were not provided:
  <SUBNET>
```

**Form B (Duplicate key error):**
```bash
$ sudo start-tunnel subnet 10.59.203.1/24 set-wan --wan-ip 172.16.0.12 10.59.203.1/24
```
**Verbatim Error:**
```text
General error: error: Serialization Error: duplicate key: subnet
```

### 📋 Full Environment Profile
* **Component:** StartTunnel 1.2.1
* **Update Verification:** 
  ```json
  {"status":"up-to-date","installed":"1.2.1","candidate":"1.2.1"}
  ```
* **Installation Protocol:** Upgraded from legacy 0.x/1.0.x line via:
  ```bash
  curl -sSL https://start9.com/start-tunnel/install.sh | sh
  ```
* **Network Topology:** 1:1 NAT Cloud VPS (Public IPv4 at provider edge; local interface `ens3` bound to private IP `172.16.0.12/20`). Matches standard documentation models for AWS EC2, Oracle Cloud, GCP, and Azure.
* **Subnet Layout:** One active subnet (`10.59.203.1/24`) containing one StartOS server device (`10.59.203.3`) and three downstream clients.

---

### 🗄️ Database Evidence (`db dump` Analysis)
A redacted look at the post-migration database state shows exactly where state validation collapses. Existing entries remain functional due to inline legacy keys, while the new `wanIp` parent tracking keys remain explicitly empty:

```jsonc
{
  "gateways": { 
    "ens3": { 
      "ipInfo": { "wanIp": null } 
    } 
  },
  "wg": { 
    "subnets": { 
      "10.59.203.1/24": {
        "clients": {
          "10.59.203.3": { "kind": "server", "allowAutoPortForward": true, "wanIp": null }
          // All other sub-connected clients are similarly set to "wanIp": null
        },
        "wanIp": null  // <-- CRITICAL BLOCK: Gating condition for all new port-forward adds
      } 
    } 
  },
  "portForwards": {
    // Grandfathered entries carry the external IP inline; DNAT mechanics process normally
    "172.16.0.12:443":  { "target": "10.59.203.2:443",  "kind": "dnat", "enabled": true  },  
    "172.16.0.12:8333": { "target": "10.59.203.3:8333", "kind": "dnat", "enabled": false }
  }
}
```

---

### 💥 Severity & Scope of Impact
Port forwarding configuration is **fundamentally broken** across all virtual private servers operating behind an edge NAT layout (including AWS, Azure, GCP, and Oracle Cloud). 

Because pre-existing DNAT rules continue to pass traffic perfectly, the underlying translation layer remains healthy. The failure is strictly administrative: the software blocks new additions when `wanIp` is null, yet provides no valid command sequence to populate it.

---

### 🛠️ Recommended Code Fixes

1. **Fix CLI Argument Inheritance:** Remove the redundant trailing `<SUBNET>` positional argument requirement from the `set-wan` command structure. Allow it to natively inherit context from the parent `subnet` command space.
2. **CLI Sub-Command Audit:** Audit sister sub-commands (specifically `set-dns` and `set-ipv6`) to ensure this duplicate positional parameter defect is not replicated across other network configuration modules.
3. **Enhance Migration Routine:** Update database migration logic to automatically backfill the `wanIp` data property using either the active interface IP or values parsed from legacy pre-upgrade port-forward records.
4. **Implement Port Forwarding Fallbacks:** Modify the validation check within the `port-forward add` sequence. If `wanIp` resolves to null, fallback to query `get-available-ips` dynamically instead of executing an immediate hard failure.

helix-nine:
Fixes #3565.

## What breaks, and why

The reporter runs on a 1:1-NAT VPS: their dedicated public IPv4 lives at the provider's edge and `ens3` holds only `172.16.0.12/20`. `port-forward add` fails with `no WAN IP available for device 10.59.203.3`, from the CLI and the UI alike.

`add_forward` derives the external IP from `external_ipv4` → `default_wan`, which filtered the host's addresses through `is_wan_candidate` — a predicate that rejects RFC1918. No public address on the NIC, so `None`, so a hard error. Ports published *before* the upgrade keep working because their external IP is embedded in the forward key and `resync_forward_keys` falls back to it (`.unwrap_or(*src.ip())`), which is why only new adds fail.

Three things the issue reports are worth correcting:

- **`wanIp: null` is not a migration defect.** StartTunnel never populates `gateways.<iface>.ipInfo.wanIp` on *any* install: `TunnelContext::init` fills `gateways` once from `load_ip_info()` (`ip -o addr show`), which sets name/scope_id/subnets/device_type and nothing else. The UPnP + echoip WAN probe lives in `net/gateway.rs`'s watcher, a StartOS-server component `start-tunneld` never starts. So `default_wan`'s `ip_info.wan_ip` branch is dead code in this product, and a backfill migration would be clobbered on the next boot — `gateways` is derived from the kernel at every start.
- **The blast radius is larger than manual publishing.** Every `external_ipv4` caller is affected, so automatic publishing (PCP `MAP` → `CANNOT_PROVIDE_EXTERNAL`, UPnP `GetExternalIPAddress` → SOAP 501) from a StartOS server on such a tunnel is dead too.
- **This is a 1.1.0 regression, not 1.2.x.** 49ea23e3b (#3306) replaced `port-forward add`'s caller-supplied `source: SocketAddrV4` with a server-derived IP, on the rationale that a forward's inbound IP must equal the device's egress WAN.

## The fix

`default_wan` keeps today's behaviour exactly wherever the host holds a public address, and otherwise falls back to any address it does hold. That private address is what inbound packets carry once the provider has translated them, so it is what the DNAT rule has to match — `build/lib/scripts/forward-port` emits `ip daddr $sip`, a destination match that needs no address ownership beyond the packet actually carrying it. Extracted as `default_wan_of(&OrdMap<GatewayId, NetworkInterfaceInfo>)` so it is unit-testable; there was no coverage of this path at all before.

**Why not just document the WAN IP override.** Pinning `wan_ip` reaches the dataplane correctly (`assigned_wan_for` applies no filter), but it is also read at precedence #2 by `show_config` to build the WireGuard `Endpoint`. Telling these users to pin their private address would stamp an endpoint their devices cannot dial into every config generated afterwards. Resolving it in `default_wan` leaves that chain untouched. Two related footguns are deliberately left alone here, noted below.

## Second defect: the escape hatch was uninvokable

`subnet set-wan` and `subnet set-ipv6` each declared their own `<SUBNET>` positional instead of inheriting the parent's, so `ParentHandler::cli_command` appended a second one and `cli_parse` → `util::combine` rejected it as `duplicate key: subnet`. Nothing catches this at compile time — `impl<A, B> OrEmpty<Flat<A, B>> for Empty` lets a 2-arg handler register under a params-carrying parent by silently discarding the inherited params. Both now take `SubnetParams` as inherited params like their `add`/`remove`/`set-dns` siblings. A repo-wide audit found these two were the only instances; `host_api`, the `s9pk` parents, `logs` and the `address`/`binding` families all inherit correctly.

`set-ipv6` has been equally unusable from the CLI since 1.1.0 — the CLI reference documents the inherited form the code never implemented.

The RPC path never calls `combine`, so the web UI was unaffected and its payload is unchanged: `{subnet, wanIp}`. Only the generated binding and six type annotations move (mirroring the already-correct `setSubnetDns`); the two real call sites already pass `subnet`.

## UI

`wanOptions()` mirrors `is_wan_candidate` in TypeScript, so on these hosts the WAN IP menu was empty — the setting was unreachable from the UI even once the CLI worked. It now lists what the backend accepts, with `defaultWanIp` mirroring the same two passes so the `System default (…)` parenthetical stays truthful.

## Verification

- `cargo test -p start-core --features test tunnel::` — 45 passed, including 4 new `default_wan_of` cases (public wins over private; private-only host; wg interface excluded; loopback/link-local yield `None`)
- `npm run check:tunnel`, `npm run check:i18n:tunnel` — clean
- `make manpages` — all four affected pages lose the duplicate `<SUBNET>`
- `make start-core-ts-bindings` — `SetSubnetWanParams` becomes `{ wanIp }`
- `make start-core-format-check`, `make start-tunnel-format-check`, prettier — clean
- Behaviour was reproduced against a built `tunnelbox` before the change (`Usage: start-tunnel subnet <SUBNET> set-wan [OPTIONS] <SUBNET>`, then both reported errors verbatim) and after (help loses the positional, invocation reaches transport)

Not verified on real hardware: the live PCP/UPnP path on a provider-NAT'd VPS. #3306's own body flagged that path as needing a router + VPS pair to exercise end to end.

## Deliberately out of scope

- **`set-wan` still accepts an address the host does not hold.** Now that the CLI works, a user who pins their real public IP loses every working forward in one command: `resync_forward_keys` re-keys them onto an address no packet carries, and the SNI listener still comes up because `build_listen_socket` sets `IP_FREEBIND`, so the bind succeeds and silently blackholes. A validation belongs here, but rejecting outright forecloses pinning an address that is about to be assigned — worth a decision rather than a guess.
- **The WireGuard `Endpoint` is already wrong on this topology**, independent of any pin: `show_config`'s chain falls through to the webserver listen address because its `local_addr` branch is dead (the `ConnectInfo` middleware is registered nowhere). Registering it would make `devices.md`'s existing promise — "StartTunnel uses the address you are accessing it over" — true, and is the cheaper of the two ways to separate the advertised endpoint from the dataplane address.
- **Migration `m_04` deletes any port-80 forward** with no publicness check, on the assumption the HTTP→HTTPS redirect replaces it — but the redirect binds only public addresses, so on this host nothing takes over. Worth asking the reporter whether they had a `172.16.0.12:80` forward before upgrading; if so it is a separate 1.2.1 data-loss bug.


helix-nine (commented):
Two things from the analysis on #3565, one of which I think is a real defect in the second pass.

**The private fallback can select a bridge address.** `default_wan_of` filters out only `wg0` and `Loopback`, and `net_iface` holds every interface `ip -o addr show` reports (`tunnel/context.rs:242-252`), so on a host running Docker or libvirt the second pass sees `docker0` = 172.17.0.1 and `virbr0` = 192.168.122.1 alongside the uplink. `OrdMap<GatewayId, _>` iterates by interface name, and `docker0` sorts before `ens3`/`eth0` (as does `virbr0` before `wlan0`), so the fallback picks the container bridge rather than the address inbound packets actually carry — and the resulting DNAT rule never matches.

That only bites when the first pass finds nothing, which is exactly the NAT'd-VPS case this PR targets, and Docker on such a VPS is common.

`net/gateway.rs:2314-2318` already draws this line for WAN probing:

```rust
let forwardable = !subnets.is_empty()
    && !matches!(
        device_type,
        Some(NetworkInterfaceType::Bridge | NetworkInterfaceType::Loopback)
    );
```

Adding `Bridge` to `egress()`'s filter would match it. Worth considering whether the *first* pass should stay bridge-inclusive — a public address on a bridge is plausible on a virtualization host, and it was reachable before this PR — in which case the exclusion belongs on the fallback closure only.

**`external_ipv4` also answers downstream StartOS boxes, and one of those paths won't filter the private address out.** It backs `GatewayBackend::external_ipv4`, so it feeds UPnP `GetExternalIPAddress` (`net/port_map/server/igd.rs:199`) and the PCP MAP external address (`net/port_map/server/mod.rs:338`). A downstream UPnP client discards a private answer (`net/port_map/upnp.rs:104-105`), but the PCP client returns it unfiltered (`net/port_map/client.rs:328-337`, `:361`), and `net/gateway.rs:652-675` short-circuits the reachability probe once auto-forwarding has succeeded — so the attached server would report `open_externally: true` for an RFC1918 address. Today those paths fail closed on this topology (`CANNOT_PROVIDE_EXTERNAL`), which is less useful but honest. If the intent is dataplane-only, the relaxation could be confined to the callers that key rules (`add_forward`, `resync_forward_keys`) and left out of the `GatewayBackend` impl.

Unrelated to the above: this needs a rebase — GitHub now reports it conflicting, because #3570 landed the same CLI fix (`api.rs`, the four man pages, both bindings, the three api-service files, CHANGELOG/Cargo.toml at 1.2.2) on master about ten minutes before this went up. Apologies for the collision; two sessions picked up #3565 in parallel and I didn't see this PR until after mine was merged. The `forward/igd.rs`, `wan.ts`, help-content and docs work here is all additional to what landed.

PR #3574, opened 2026-07-25
docs(start-os): converge the 0.4.0 update paths on start.local

MattDHill:
## Summary

The **Over the Air** tab of the 0.4.0 update guide described the wrong address for the migration. It told users that once 0.4.0 boots, an initialization screen appears at their server's own address (`https://adjective-noun.local`), and built an entire step around repeatedly opening private windows there to defeat the browser cache.

That isn't what happens. After the restart, the server does **not** come back at its old address — it comes back at `start.local`, the same address the USB installer serves, showing the same migration progress screen a flash update reaches after the password step. A Pi reaches it too.

## Changes

- **OTA:** the restart step now says plainly that the server does not return at its old address, and sends the user to `http://start.local` after a few minutes.
- **The tabs now end where the paths actually diverge.** Everything from the progress screen onward is shared prose in a new **Step 8: Wait for the Migration** — the migration wait, the reboot, and the return to `https://adjective-noun.local`.
- **The 85% tip is now shared.** It was flash-only, but it always applied to both paths — it describes the migration, not the installer.
- **Deduplicated the cache guidance.** Both tabs were carrying independent copies before this PR; there is now one, placed only where it's actually needed — the return to the server's own address, which is the sole point where the browser has a stale copy. Nothing is cached at `start.local`, so the long part of the migration no longer needs any of it.
- **Raspberry Pi rejoins at Step 8** instead of skipping to *After the Update*, which had it bypassing the migration wait entirely.
- Steps 8/9/10 renumbered to 9/10/11; the reboot step's USB clause now excludes a Pi's microSD.

Net: `update-040.md` gains guidance while losing 7 lines, since two duplicated blocks collapse into one.

### DIFF
--- a/projects/start-os/docs/src/update-040.md
+++ b/projects/start-os/docs/src/update-040.md
@@ -85,7 +85,7 @@ With all services stopped, create a [full system backup](/0.3.5.x/user-manual/ba
 
 ### Step 7: Install 0.4.0
 
-Pick your method below. Everything above applies to both.
+Pick your method below. Everything above applies to both, and the two paths rejoin at [Step 8](#step-8-wait-for-the-migration).
 
 {{#tabs}}
 {{#tab name="Flash Update (Recommended)"}}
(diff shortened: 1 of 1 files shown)


stupleb (approved):
Seems to cover it.

dr-bonez (approved):

PR #3578, opened 2026-07-25
test(bins): check every CLI tree for arguments a subcommand shares with an ancestor

helix-nine:
Follow-up to #3570, which landed while this was being written.

#3570 fixed `subnet set-wan`/`set-ipv6` re-declaring the parent's `<SUBNET>` positional (#3565). Nothing would have caught that: there is no test over the CLI surface, and the one artifact that showed the doubled positional — the generated man page — is regenerated from the same code, so a future refactor would regenerate the bug along with its evidence.

Per review, this checks the whole class rather than that one command. `assert_no_shadowed_args` walks a product CLI's clap tree and rejects any subcommand whose argument ids intersect an ancestor's, and every product CLI runs it: `start-tunnel`, `start-cli`, `start-registry`, `start-container`.

## Why an intersection is always a bug

`ParentHandler<_, P>::cli_command` uses `P::command()` as the base and attaches each child's own command beneath it (`rpc-toolkit/src/handler/parent.rs:352-366`), so the two argument sets live on separate clap `Command`s and clap never objects. At parse time `cli_parse` merges the two serialized param objects with `combine` (`parent.rs:389` → `util.rs:31-48`), which errors on any duplicate key. So a shared id is uninvokable in both directions — pass the value once and the ancestor consumes it, leaving the child's copy missing; pass it twice and the merge fails — and the failure only appears when the command is run.

The walk starts fresh at each top-level subcommand: the root's own arguments are the CLI's config (`--host`, `--config`, …), which `CliApp` parses into `Config` separately and never merges into params (`rpc-toolkit/src/cli.rs:57-78`). Global args and clap's own `help`/`version` are excluded for the same reason.

## Result

All four trees pass as-is, so `subnet set-wan` and `set-ipv6` were the only instances in the repo.

Verified the check actually catches the original bug rather than just being green — reverting `set-ipv6`'s `.with_inherited`:

```
these subcommands redeclare an ancestor's argument and cannot be invoked:
  tunnel subnet set-ipv6 <subnet>

failures:
    bins::start_cli::no_shadowed_args_start_cli
    bins::tunnel::no_shadowed_args_start_tunnel
```

Note it fires in the `start-cli` tree too, which the first (subnet-specific) version of this test missed entirely.

## Known limit

The check reads clap arguments, so it cannot see a field that is `#[arg(skip)]` but still serialized — `ShowConfigParams::local_addr` is the one such field today. Catching those would mean comparing serialized param keys, which no trait currently exposes.

Test-only, so no changelog entry and no version bump — 1.2.2 is already open for the fix itself.

### DIFF
--- a/shared-libs/crates/start-core/src/bins/container_cli.rs
+++ b/shared-libs/crates/start-core/src/bins/container_cli.rs
@@ -43,6 +43,11 @@ pub fn main(args: impl IntoIterator<Item = OsString>) {
 }
 
 #[test]
+fn no_shadowed_args_start_container() {
+    super::assert_no_shadowed_args(app().into_command());
+}
+
+#[test]
 fn export_manpage_start_container() {
     // start-container is part of the start-os product; anchored to start-core's crate dir.
     let dir = concat!(
--- a/shared-libs/crates/start-core/src/bins/mod.rs
+++ b/shared-libs/crates/start-core/src/bins/mod.rs
@@ -235,3 +235,60 @@ impl MultiExecutable {
         std::process::exit(1);
     }
 }
+
+/// Assert that no subcommand of `root` redeclares an argument an ancestor owns.
+///
+/// `ParentHandler<_, P>` renders `P` as the parent command's arguments and each
+/// subcommand's own `Params` beneath it, then merges the two serialized objects
+/// with `combine`, which rejects a duplicate key. A subcommand that redeclares
+/// one is therefore uninvokable both ways: pass the value once and the ancestor
+/// consumes it, leaving the subcommand's copy missing; pass it twice and the
+/// merge fails. Clap sees two separate commands, so nothing catches it until the
+/// command is run. Inherit with `.with_inherited(...)` and take the ancestor's
+/// params as the handler's last argument instead.
+#[cfg(test)]
+pub(crate) fn assert_no_shadowed_args(root: clap::Command) {
+    use std::collections::BTreeSet;
+
+    fn declared(cmd: &clap::Command) -> BTreeSet<String> {
+        cmd.get_arguments()
+            .filter(|a| !a.is_global_set())
+            .map(|a| a.get_id().to_string())
+            .filter(|id| id != "help" && id != "version")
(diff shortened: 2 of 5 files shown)


dr-bonez:
can you find a test that can more generally test this kind of shadowing issue?

helix-nine:
Done — pushed as 7562eb2.

`assert_no_shadowed_args` (`bins/mod.rs`) walks a CLI's clap tree and rejects any subcommand whose argument ids intersect an ancestor's; every product CLI now runs it — `start-tunnel`, `start-cli`, `start-registry`, `start-container`.

An intersection is always the bug rather than a heuristic: `ParentHandler<_, P>::cli_command` uses `P::command()` as the base and attaches each child beneath it (`rpc-toolkit/src/handler/parent.rs:352-366`), so the two argument sets sit on separate clap `Command`s and clap never objects; `cli_parse` then merges the serialized param objects with `combine` (`parent.rs:389` → `util.rs:31-48`), which rejects a duplicate key. The walk starts fresh at each top-level subcommand because the root's arguments are the CLI config, parsed into `Config` and never merged into params (`cli.rs:57-78`); globals and clap's `help`/`version` are excluded for the same reason.

All four trees pass unchanged, so `subnet set-wan`/`set-ipv6` were the only instances. Confirmed it catches the original rather than being vacuously green — reverting `set-ipv6`'s `.with_inherited`:

```
these subcommands redeclare an ancestor's argument and cannot be invoked:
  tunnel subnet set-ipv6 <subnet>

failures:
    bins::start_cli::no_shadowed_args_start_cli
    bins::tunnel::no_shadowed_args_start_tunnel
```

It fires in the `start-cli` tree as well, which the subnet-specific version missed.

One limit worth naming: it reads clap arguments, so a field that is `#[arg(skip)]` but still serialized is invisible to it — `ShowConfigParams::local_addr` is the only one today. Catching that class would mean comparing serialized param keys, and nothing exposes those generically right now. Say the word if you want me to try it from the handler side instead.

dr-bonez (approved):

PR #3586, opened 2026-07-26
fix(migration): report why a service failed to convert from 0.3.5.1

helix-nine:
The 0.3.5.1 → 0.4.0 migration collected every per-package failure into one
aggregate notification listing service names, and threw the error itself away to
the logs (`tracing::error!("Error reinstalling {id}: {e}")`).

## Why the install machinery didn't already cover it

`ServiceMap::install` does emit a per-package failure notification, but only
through `ServiceRefReloadCancelGuard`, which is not constructed until
`service_map.rs:209`. `notify(` appears exactly once in that file (`:584`, inside
`ServiceRefReloadInfo::reload`), reachable only via `handle`/`handle_last`/`Drop`
— so the whole prologue is silent by construction.

Of the 25 error-return points in `install` (lines 150–418), 5 are unguarded, all
in that prologue: `:167` `s9pk().await?`, `:169` `validate_and_filter`, `:171`
`commitment()`, `:180-188` the reserved-id return, `:189` `icon_data_url()`. For
the migration, `:167` **is** the docker-driven v1→v2 s9pk conversion — the step
most likely to fail — so the most common migration failure reported nothing.

Two more silent steps sit outside `install` entirely: the `File::open` /
`MultiCursorFile::open` ahead of it, and the post-install `db.mutate` after it
(`handle_last` disarms the guard on success, so nothing covers that one).

## What this changes

Hoist the s9pk load out of the closure passed to `install` and notify against the
package when the conversion fails, matching the shape of an install failure:
`package_id: Some(id)`, localized title, the error as the body, subtype `()`. The
post-install bookkeeping mutate reports itself the same way.

Failures *inside* `install` deliberately do **not** get a notification from here —
they already raise their own "Installing Failed" — so nothing is duplicated. The
`failures` map carries the error only when nothing else has reported it; the
aggregate summary still lists every failure either way.

Deliberately **out of scope** (agreed on-thread): `install`'s remaining prologue
holes (`validate_and_filter` / `commitment()` / `icon_data_url()`). Covering those
means wrapping them in `reload_guard.handle(...)` — merely arming the guard
earlier is not enough, because `Drop` passes `None` and `reload` only notifies on
`Some(error)`. It also changes failed-*update* behaviour: with a packageData entry
present, `load(…, Undo)` takes the `Installed` arm and bounces the running
service. That belongs in its own change.

## Renamed packages were always reported as failed

`s9pk::v2::compat` rewrites five ids during the conversion (`nostr` →
`nostr-rs-relay`, `ghost`/`synapse`/`monerod` → `*-legacy`, `fedimintd` →
`fedimint-guardian`), and `install` inserts into `packageData` under
`manifest.id`. The post-install lookup used the old archive-directory id, and
`up()` resets `packageData` to `{}` — so `or_not_found` always missed and those
five landed in `failures` even when they migrated fine, telling users to
re-install them under ids the marketplace no longer has. The rename map already
existed inline a few lines above for the `.version` write; it is now a shared
helper used by both.

Also: clear a package from `failures` once it succeeds. The archive is walked per
version directory with no dedup by id, so a package with a stale version dir
alongside the current one can be installed twice and would otherwise stay marked
failed after a later attempt worked.

## Verification

`cargo check -p start-core`, `make start-core-format-check`, and prettier over the
changed markdown all pass. The notification path itself is not exercised by any
existing test and I have not run a 0.3.5.1 → 0.4.0 migration on a VM, so the
runtime behaviour is reasoned from the source, not observed.

### DIFF
--- a/projects/start-os/CHANGELOG.md
+++ b/projects/start-os/CHANGELOG.md
@@ -40,6 +40,20 @@ file tracks notable changes since the move to the monorepo.
   onion addresses come across with it and answer as soon as the update
   finishes, with nothing to start by hand.
 
+- **A service that fails to convert while migrating from 0.3.5.1 reports the
+  reason.** The v1→v2 package conversion raises a notification against that
+  service carrying the error that stopped it, in the same form as an install
+  failure — alongside the summary notification listing every service to
+  re-install.
+
+- **A service that is renamed during migration is recorded as migrated.** Ghost,
+  Synapse, Monero, Nostr and Fedimint are installed under new ids on 0.4.0
+  (`ghost-legacy`, `synapse-legacy`, `monerod-legacy`, `nostr-rs-relay` and
+  `fedimint-guardian`), and the migration looks each one up under the id it was
+  installed as. These services complete their migration without appearing among
+  the failures, and the failure list names services by ids that exist in the
+  marketplace.
+
 - **The over-the-air update to 0.4.0 boots on the Server Pure.** The Server
   Pure's PureBoot firmware reads the boot configuration itself rather than
   running GRUB, and it takes the kernel and initramfs paths literally. The
--- a/projects/start-os/docs/src/update-040.md
+++ b/projects/start-os/docs/src/update-040.md
@@ -180,6 +180,12 @@ Depending on the speed of your drive, plan on 3-5 minutes per GB of backup data.
 
 ## Post-Migration Notes
 
+### If a Service Fails to Migrate
+
+Check your notifications. A service that fails to migrate raises a notification naming the service and the reason it failed, and a summary notification lists everything that needs re-installing.
+
+Your data is safe — it stays on disk where the service left it. Install the service again from the marketplace and it will pick that data back up.
+
 ### Tor Cleanup
 
 During migration, the **Tor** service is automatically installed and started, with all your existing onion addresses intact and reachable as soon as the update finishes. However, Tor is rarely needed in StartOS 0.4.0 — most users will be better served by other networking options.
--- a/shared-libs/crates/start-core/locales/i18n.yaml
+++ b/shared-libs/crates/start-core/locales/i18n.yaml
@@ -1996,6 +1996,13 @@ migration.migrating-package:
   fr_FR: "Migration du paquet %{package}..."
   pl_PL: "Migracja pakietu %{package}..."
 
+migration.service-failed-title:
+  en_US: "Migration Failed"
+  de_DE: "Migration fehlgeschlagen"
+  es_ES: "Migración fallida"
+  fr_FR: "Échec de la migration"
+  pl_PL: "Migracja nie powiodła się"
+
 migration.services-failed-title:
   en_US: "Services Failed To Migrate"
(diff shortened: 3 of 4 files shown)


dr-bonez (approved):

helix-nine:
Tested on a VM.

Built an ISO from this branch (`make start-os-iso`, x86_64 devmode), installed it to a fresh libvirt VM, and confirmed the server is running this branch:

```
$ start-cli --host https://<vm> git-info
@fix/per-service-migration-failure-notifications
```

To exercise `post_up` without a full 0.3.5.1 install, I planted two deliberately-broken v1 s9pks in the archive (v1 magic `3b 3b 01` + garbage) and re-armed the migration todo by appending a CBOR JSON-patch to `/media/startos/data/main/embassy.db` (patch-db stores `revision`, `persistent`, then a stream of patches), then restarted `startd`:

```
archive/ghost/3.0.0/ghost.s9pk
archive/bitcoind/27.0.0/bitcoind.s9pk
postInitMigrationTodos = {"0.3.6-alpha.0": null}
```

`post_up` ran with `input=Null`, walked the archive and failed conversion on both:

```
INFO  post_up{input=Null}: v0_3_6_alpha_0:502: Migrating package ghost...
INFO  post_up{input=Null}: s9pk:45: Converting package to v2 s9pk
ERROR post_up{input=Null}: v0_3_6_alpha_0:576: Error converting ghost: S9PK Parsing Error: signature error: Cannot decompress Edwards point
...same for bitcoind...
INFO  init: Running migrations: complete
```

Resulting notifications:

```json
[
  { "id": 2, "packageId": null,
    "title": "Services Failed To Migrate",
    "message": "Failed to migrate the following services from 0.3.5.1: bitcoind, ghost-legacy\nYOUR DATA IS SAFE. ..." },
  { "id": 1, "packageId": "ghost-legacy",
    "title": "Migration Failed",
    "message": "S9PK Parsing Error: signature error: Cannot decompress Edwards point" },
  { "id": 0, "packageId": "bitcoind",
    "title": "Migration Failed",
    "message": "S9PK Parsing Error: signature error: Cannot decompress Edwards point" }
]
```

That covers the three things the change is for:

- **Per-service notification carrying the reason** — one per failed package, attributed via `packageId`, error as the body. Before this change the same scenario produced only the aggregate, with the error going to the logs alone.
- **The rename fix** — `ghost` is reported as `ghost-legacy` (an id that exists in the marketplace), while `bitcoind` is unchanged, so the map applies only where it should. The aggregate's list follows suit: `bitcoind, ghost-legacy`.
- **No duplication** — exactly 3 notifications for 2 failures, `unreadNotificationCount: 3`, and the todo was retired (`postInitMigrationTodos: {}`) so the pass doesn't re-run.

Not covered by this run, to be explicit: a failure *inside* `ServiceMap::install` (i.e. after its reload guard is armed) still relies on the existing "Installing Failed" notification — I did not fabricate an s9pk that converts cleanly but then fails to install, so the no-duplicate claim for that path is reasoned from the guard placement rather than observed. The `.version`-write rename path also wasn't exercised, since a null migration input skips it.

PR #3596, opened 2026-07-27
fix(start-sdk): report which daemon failed, and why, when runUntilSuccess times out

MattDHill:
## Problem

A `runUntilSuccess` timeout reported only a list of ids:

```
Timed out waiting for postgres,upgrade,__RUN_UNTIL_SUCCESS
```

Three things wrong with it:

1. **It leaks the internal sentinel.** `__RUN_UNTIL_SUCCESS` is the oneshot `runUntilSuccess` appends to detect that every user daemon is ready. It depends on all of them, so it is never ready when anything else isn't — it can only ever add noise to a message a user reads.
2. **It cannot distinguish slow from dead.** "postgres never became ready" is the same sentence whether the process is still starting or has exited 47 times.
3. **The error that says which is discarded.** `Daemon.runLoop` catches the exit error, logs it, and retries forever on a backoff. Nothing retains it, so it never reaches the caller.

Together these make a crash-looping daemon indistinguishable from a slow one for the entire timeout. A user hit this updating Nextcloud: the Postgres entrypoint exited 1 on every start, the chain retried for the full 30 minutes, and the only surfaced output was the line above. The real cause — `Database is uninitialized and superuser password is not specified` — was in the service logs the whole time, but nothing in the failure pointed there.

## Change

- `Daemon` retains `lastExitError` and a `failedExits` count.
- `HealthDaemon` appends the cause to its `<id> daemon crashed` health message.
- The timeout message reports, per un-ready daemon, its health result and message plus the exit count and last exit error, and states the budget that elapsed. The sentinel is filtered out, and its id is now a named constant rather than a literal in two places.
- `clearTimeout` on success, which otherwise left a pending long-duration timer alive after the chain finished.

Same failure now reads:

```
Timed out after 1800000ms waiting for postgres (loading; 47 failed exit(s), last: docker-entrypoint.sh exited with code 1), upgrade (waiting; postgres)
```

The exit count is reported *alongside* current health rather than instead of it, because the two disagree in exactly the case that matters: a `ready` check that returns `loading` on a failed probe — the common shape, and what Nextcloud does — overwrites the crash back to `loading` between restarts, so health alone always reads "still starting".

## Not changed

Timeout semantics. A crash-looping daemon still runs out the full budget rather than failing fast. Failing early on N consecutive crashes would be a behavior change for packages whose daemons legitimately crash a few times before coming up, so it is left as a separate decision. This PR is scoped to making the existing failure legible.

## Tests

Four new cases in `lib/test/Daemons.test.ts`, driving `runUntilSuccess` against JS-fn daemons so no container is needed:

- names the un-ready daemon with its health result and message
- omits the sentinel from the message
- reports the elapsed budget
- surfaces the exit cause and count of a crash-looping daemon (the Nextcloud shape: process dies every start, ready check reports `loading`)

`make test` 84 passed, `make check` clean, root `prettier --check` clean on all changed files.

## Docs

`docs/src/init.md` gains a "When it times out" section under the `runUntilSuccess` pattern: how to read the new message, and the distinction between a daemon with failed exits (process is dying — read the service logs) and one without (running, but not passing its ready check).

### DIFF
--- a/projects/start-sdk/lib/mainFn/Daemon.ts
+++ b/projects/start-sdk/lib/mainFn/Daemon.ts
@@ -33,7 +33,9 @@ export class Daemon<
   private onExitFns: ((success: boolean) => void)[] = []
   private loop: { abort: AbortController; done: Promise<void> } | null = null
   private releaseSubcontainerHold: (() => Promise<void>) | null = null
-  private _detached = false
+  private detached = false
+  private _lastExitError: Error | null = null
+  private _failedExits = 0
   protected constructor(
     readonly subcontainer: C,
     private startCommand: () => Promise<CommandController<Manifest, C>>,
@@ -49,6 +51,24 @@ export class Daemon<
   /** Returns true if this daemon is a one-shot process (exits after success) */
   isOneshot(): this is Oneshot<Manifest> {
     return this.oneshot
+  }
+  /**
+   * The error from the most recent abnormal exit, or `null` if none has been
+   * recorded. An exit driven by {@link stop} or {@link term} is deliberate and
+   * does not set this; the restart loop otherwise logs the error and retries,
+   * retaining it here rather than reporting it to anyone.
+   */
+  get lastExitError(): Error | null {
+    return this._lastExitError
+  }
+  /**
+   * How many times the process has exited abnormally over this daemon's
+   * lifetime. Nothing resets it — neither a restart nor a
+   * {@link stop}/{@link start} cycle — so a non-zero count means the process
+   * has died at least once, not that it is dying now.
+   */
+  get failedExits(): number {
+    return this._failedExits
   }
   /**
    * Factory method to create a new Daemon.
@@ -69,7 +89,7 @@ export class Daemon<
         CommandController.of<Manifest, C>()(effects, subcontainer, exec)
       const res = new Daemon<Manifest, C>(subcontainer, startCommand)
       effects.onLeaveContext(() => {
-        if (res._detached) return
+        if (res.detached) return
         res.term().catch(e => logErrorOnce(asError(e)))
       })
       return res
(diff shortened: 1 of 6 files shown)


projects/start-sdk/lib/mainFn/Daemon.ts
```diff
@@ -34,6 +34,8 @@ export class Daemon<
   private loop: { abort: AbortController; done: Promise<void> } | null = null
   private releaseSubcontainerHold: (() => Promise<void>) | null = null
   private _detached = false
+  private _lastExitError: Error | null = null
```
dr-bonez, line 37: What's the convention for underscores vs not here? It seems inconsistent

MattDHill: In this module the underscore only marks a name a public accessor has already taken — TS won't let a field and a getter share one. `HealthDaemon`'s `_health`/`get health()` is the existing instance; `_lastExitError` and `_failedExits` are the same shape. Where the getter reads a field under a *different* name there's no collision and no underscore: `HealthDaemon.resolvedReady`/`get isReady()`, `Daemons.builtHealthDaemons`/`get healthDaemons()`.

The real inconsistency is the line right above: `_detached` predates this PR and has no accessor at all, so nothing forced its underscore — and `Daemons`/`DaemonsReconciler` already spell the same flag plain. Renamed it to `detached`, so the rule holds uniformly now.

Also rebased onto master (changelog conflict), and corrected the new doc comments where they overstated: `failedExits` survives a `stop`/`start` cycle rather than counting "since start", `lastExitError` skips exits driven by `stop`/`term`, and the docs' sample timeout output showed a `waiting` message that the example chain above it — `display: null` — would not actually produce.

dr-bonez (approved):

PR #3600, opened 2026-07-28
fix(cli): resolve a .local host lazily, not at context build

MattDHill:
## Problem

`CliContext::init` resolved the configured host to an address *before* the subcommand was dispatched. Resolution goes through the system resolver (`getent ahosts`), so it needs the named box to be answering mDNS — and when it isn't, **every** `start-cli` invocation dies at context construction:

```
$ start-cli --registry=https://alpha-registry-x.start9.com registry package get lnd full --format=json
Network Error: Failed to resolve hostname: demo.local     (exit 9)
```

That command never contacts a StartOS host. Neither do the rest of the registry commands, nor `s9pk`, nor the tunnel commands — but a `host` profile pointing at an offline `.local` box takes all of them down. It surfaced while promoting packages between registries from a workspace whose `.startos/config.yaml` names a dev box that happened to be powered off.

The eager pinning itself is correct and stays (#3471): a musl-static binary can't resolve `.local` in-process. Only its *timing* was wrong.

## Fix

`LazyPinnedUrl` (in `net::mdns`) holds the URL as configured and resolves it on first use, memoizing via `OnceCell::get_or_try_init`. `CliContextSeed` stores the host and registry URLs that way, and the `rpc/v1` and `rpc/v0` endpoints are derived from them on demand instead of being baked at init.

Consequences:

- Commands that talk to the host fail exactly as they did, with the same message, at the request that needs the address.
- Commands that don't, run regardless of whether the box is up.
- Failure isn't memoized, so a box that comes up mid-session resolves on the next attempt rather than needing a fresh invocation.
- `host_identity` still comes from the *unresolved* URL, so request signatures keep committing to the `.local` name rather than the pinned IP — unchanged from 1.1.0.

`registry_url` is now a method returning `Result<Option<Url>, Error>`; its three call sites (`install`, `registry::context`, `registry::package::promote`) are updated.

## Notes

- start-cli `1.1.0` is a cut tag, so this opens a new prospective `1.1.1` heading and bumps the manifest + lockfile with it.
- Not compiled locally — a cold `start-core` build is heavy on my machine, so I'm leaning on the CI matrix here. Worth a look at the darwin leg in particular: `pin_mdns_host` is a no-op there, so `LazyPinnedUrl::get` just clones, but that path isn't exercised locally.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7729,7 +7729,7 @@ checksum = "6ce2be8dc25455e1f91df71bfa12ad37d7af1092ae736f3a6cd0e37bc7810596"
 
 [[package]]
 name = "start-cli"
-version = "1.1.0"
+version = "1.1.1"
 dependencies = [
  "start-core",
  "tracing",
--- a/projects/start-cli/CHANGELOG.md
+++ b/projects/start-cli/CHANGELOG.md
@@ -9,6 +9,19 @@ Because `start-cli` is a thin client over `start-core`, most user-visible CLI ch
 in `start-core`; record here anything that changes this crate's entrypoint, features, packaging,
 or the CLI's externally observable behavior.
 
+## [1.1.1]
+
+### Fixed
+
+- **A configured `.local` host no longer breaks commands that never use it.** Resolving that
+  host to an address happened while the client context was built, before the subcommand was
+  dispatched, so whenever the named box was off the LAN _every_ invocation died with
+  `Network Error: Failed to resolve hostname: <name>` — including the registry, s9pk, and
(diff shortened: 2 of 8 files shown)


MattDHill:
Closing in favour of #3601, which lays out this approach alongside the connect-time-resolver alternative that would delete the `.local` special-casing rather than defer it.

Branch `fix/cli-lazy-mdns-host-resolution` is left in place as the reference implementation of option A — compiles green on the full matrix; the one CI failure was a stale `man/start-cli.1` after the 1.1.1 version bump, not the logic.

PR #3604, opened 2026-07-29
fix(start-os): map ports over TCP and UDP

remcoros:
## Summary

StartOS automatically maps only requested TCP ports, preventing UDP services such as WireGuard from being exposed through PCP, NAT-PMP, or UPnP.

This change:

- Requests both TCP and UDP for standard port mappings
- Tracks each protocol independently for renewal and removal
- Keeps SNI hostname mappings TCP-only

## Validation

- `cargo check -p start-core --lib`
- `cargo build -p start-core --lib`
- 15 targeted port-mapping tests passed

## Manual testing

On a fresh 0.4.0.1 box + this patch:
- installed wireguard-startos (https://github.com/remcoros/wireguard-startos)
- enabled public address <WAN-IP>:51820
- observed in my router that both a TCP and UDP rule were added, whereas before, only TCP was added.

### DIFF
--- a/projects/start-os/CHANGELOG.md
+++ b/projects/start-os/CHANGELOG.md
@@ -19,6 +19,15 @@ file tracks notable changes since the move to the monorepo.
   `WAN: N/A` and `NTP: Unknown`. It now uses a private temporary file and
   removes it on every exit path.
 
+- **A service that uses UDP is reachable from the Internet on a public IP
+  address.** StartOS already passed UDP through to the service, but the mapping
+  it asked your gateway for over PCP, NAT-PMP or UPnP covered TCP only, and a
+  router maps each protocol separately — so a VPN, a video-call relay or a game
+  server needed a UDP rule added to the router by hand. StartOS now asks for
+  both protocols: on a public IP address, over IPv4 and IPv6 alike, and on a
+  published port range. It withdraws both when the address is disabled or
+  deleted.
+
 ## [0.4.0.1]
 
 ### Changed
--- a/shared-libs/crates/start-core/src/net/port_map/client.rs
+++ b/shared-libs/crates/start-core/src/net/port_map/client.rs
@@ -32,6 +32,7 @@ use chrono::{DateTime, Utc};
 use crab_nat::{
     InternetProtocol, MappingFailure, PortMapping, PortMappingOptions, TimeoutConfig, pcp,
 };
+use igd_next::PortMappingProtocol;
 use igd_next::aio::Gateway;
 use igd_next::aio::tokio::Tokio;
 use imbl::OrdMap;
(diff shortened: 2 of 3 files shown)


dr-bonez (dismissed):

helix-nine:
Added the changelog entry, and rebased the branch to make room for it.

**Rebase.** `master` grew the `## [0.4.0.2]` heading (#3615) after this branch was cut, so the entry couldn't be added on the old base without colliding with it on merge. Rebased onto `07a14b15c`; the code diff is identical line-for-line to the one you approved, and no commit on master since the branch point touches `net/port_map/`.

**Entry** — `projects/start-os/CHANGELOG.md`, under the existing unreleased `## [0.4.0.2]` → `### Fixed`:

> **A service that uses UDP is reachable from the Internet on a public IP address.** StartOS already passed UDP through to the service, but the mapping it asked your gateway for over PCP, NAT-PMP or UPnP covered TCP only, and a router maps each protocol separately — so a VPN, a video-call relay or a game server needed a UDP rule added to the router by hand. StartOS now asks for both protocols: on a public IP address, over IPv4 and IPv6 alike, and on a published port range. It withdraws both when the address is disabled or deleted.

Placement: the newest cut tag is `start-os/v0.4.0.1`, so `0.4.0.2` is unreleased — entry goes under it, no new heading and no manifest bump. Automatic port forwarding shipped in the released `0.4.0`, so this is a genuine `### Fixed` rather than an edit to that feature's `## [0.4.0]` entry.

Scope of the claim: UDP is asserted only for public IP addresses and port ranges — `PortMapController::ensure` / `ensure_range` (`net/forward.rs:756,764`) plus the IPv6 GUA pinhole, which goes through the same `ensure` (`net/forward.rs:999`). Public *domains* map via `ensure_hostname` (`net/vhost.rs:582`), which this PR pins to `Tcp`, so neither the entry nor the docs claim domains gained UDP. No docs change ships with this: every page that mentions automatic opening is protocol-agnostic and stays true, and adding "TCP and UDP" to `docs/src/clearnet.md` would be wrong for the SNI path it describes.

**Verified:** `cargo test -p start-core --features=test port_map` → 54 passed, including the new `tcp_and_udp_share_a_port_without_clobbering`; `prettier --check` clean on the changelog.

The push dismissed your approval (the `master` ruleset dismisses stale reviews on push), so it needs a re-approve to merge.

---

One observation from reading the surrounding machinery, not a blocker and not changed here: `GatewayPortMapCapabilities` has no protocol dimension, so the UDP attempt's outcome overwrites the TCP one on the same verdict. A PCP success writes `pcp=true, nat_pmp=false` (`client.rs:892-896`); if the UDP MAP that follows it times out, `report_crab_nat_failure` writes `pcp=false` (`client.rs:517`), both are then fresh-false, and `pcp_fresh_dead` (`client.rs:996`) short-circuits every later crab_nat attempt on that gateway — including TCP re-maps — straight to the UPnP fallback until the 5-minute no-verdict TTL lapses (`db/model/public.rs:317-324`). A protocol-level refusal is handled correctly (`Pcp(_)` → `pcp=true`), so this needs silence rather than a "no" — but at two attempts per port instead of one it is twice as likely to be rolled. Pre-existing and self-healing; worth a follow-up if you want the verdict to carry a protocol, or `report_crab_nat_failure` to ignore UDP-derived timeouts.

dr-bonez (approved):

PR #3616, opened 2026-08-01
fix(ui): align mobile notification checkboxes

BeeJoe:
## Summary

- replace notification icons with selection checkboxes on mobile without shifting or overlapping text
- bottom-align checkboxes with the notification title line
- add the StartOS 0.4.0.2 release metadata and no-op version migration (not sure it's worth a version bump but followed instructions)

## Testing

- npm run check:ui
- npm run build:ui
- cargo test -p start-core --features test version::
- Prettier and Rust formatting checks
- deployed to test server and verified matching startbox hash, active startd, and running system state


MattDHill:
Thanks for tracking this down — the diagnosis and the two files were right, and the version plumbing is exactly per `VERSION_BUMP.md`.

I've pushed a commit that changes how the checkbox is placed. Measuring the offset against the icon it replaces, `0.75rem` lands 1.7px high on notifications without a service avatar and 4px high on those with one — because 840019c60 also pinned `.service` to grid row 1, so the avatar (2rem) now sets the title row's height. There's no single constant that's correct for both, and a constant here is what broke in the first place.

So the checkbox is now projected into `.title` ahead of the icon. On mobile it's an in-flow flex item and `.title`'s existing `align-items: center` places it — the same rule that places the icon — and the two swap via `display` (both 1rem behind the same gap, so nothing moves). Desktop keeps absolute positioning, anchored to the row rather than the date cell. Nothing left to go stale on the next layout change.

Your version bump and changelog entry are unchanged and still describe it correctly. Note that #3617 also adds `version/v0_4_0_2.rs` and bumps the manifest, so whichever of these lands second will need that dropped.

helix-nine:
Rebased onto `master` (`33a532c39`) and pushed — `2d2d2ae`.

**The conflict.** `projects/start-os/CHANGELOG.md` only. #3615 landed in the meantime and cut the `## [0.4.0.2]` heading itself, so the two sides were each adding that heading with a different bullet under it. Resolved by keeping master's heading and `### Fixed` and adding the notification entry as the second bullet, after the login-banner one — per the root `AGENTS.md` rule, `start-os/v0.4.0.2` has no origin tag, so the right move is to add under the existing heading rather than cut a new one. No wording changed on either entry.

**The version bump stays.** #3615 cut the heading without moving the manifest, so `master` today has the changelog at `0.4.0.2` and `package.json`/`Cargo.toml` label/`version::Current` still at `0.4.0.1` — `manage-release.sh pre-check start-os` fails on master for that reason. This PR completes it, and pre-check now passes every content gate (heading, the two `installing-startos.md` release links, crate label `0.4.0-rev.2`). @MattDHill's note still holds: #3617 carries an identical `v0_4_0_2.rs` + manifest bump, so whichever lands second drops it.

**One thing I had to fix beyond the conflict.** The bump left `projects/start-os/man/start-container.1` at `0.4.0.1`. That page is generated, not hand-written — `start-container` reports `bins::startos_version()`, which `start-core/build.rs` bakes from root `package.json`, so bumping `package.json` changes it. `make manpages-check` (the "Generated Artifacts" job, which runs because this PR touches non-`.md` files) regenerates and fails on any diff, so CI would have gone red. Regenerated with `cargo test -p start-core --lib --features test export_manpage_` and folded the two-line result into the version-bump commit; the other three products' pages regenerate byte-identical. #3617 already carries the same hunk, which is what confirmed it belongs to the bump.

Worth noting `VERSION_BUMP.md`'s checklist doesn't mention the man page — that's the gap that made this easy to miss. Happy to send a one-line follow-up to that checklist if you want it.

**Verified on the rebased tree:** `npm run check:ui`, `npm run build:ui`, `cargo test -p start-core --features test version::` (14 passed, incl. `current_matches_manifest`), `prettier --check` on every changed file, `make manpages-check` equivalent clean, `manage-release.sh pre-check start-os` clean on all repo-content gates.

Aside from the CHANGELOG hunk and the man page, the diff is byte-identical to the pre-rebase one — both UI components are untouched.

dr-bonez (approved):

helix-nine:
CI has settled on your merge commit (`707e3bdb`). Everything passes except one job, which isn't actionable:

**Create Docker Image** — `failed to push ghcr.io/start9labs/startos-registry:3616-merge: denied: installation not allowed to Write organization package`. The image builds fine for all three arches; only the GHCR push fails, because the `pull_request` event's token is read-only when the head repo is a fork. Both runs on `BeeJoe/start-technologies` fail this way and every same-repo-headed run of `start-registry.yaml` succeeds. `master` has no required status checks, so the PR is `MERGEABLE`.

**Generated Artifacts** passed, which is the one I was watching — `make manpages-check` is the gate the regenerated `projects/start-os/man/start-container.1` was for.

Also confirmed your master merge came through clean: the effective diff against `f56502a9` is still the same 11 files, with the changelog carrying all three 0.4.0.2 entries (login banner, notifications, UDP) and the version bump and man page intact.

Issue #3618, opened 2026-08-03
[feat]: surface internal (on-box) service addresses in the UI as a first-class "internal" address type


MattDHill:
### Describe the Feature Request

Every binding already gets plain-HTTP addresses on the container bridge and loopback (`10.0.3.1:<assignedPort>`, `127.0.0.1:<assignedPort>`) alongside the TLS addresses advertised on real gateways (`<assignedSslPort>`). The backend already models these as a category — `HostnameInfo::is_internal()` in `shared-libs/crates/start-core/src/net/service_interface.rs`, whose doc comment calls them "the always-on internal interfaces" — and treats them as never operator-disablable (`net/host/binding.rs:848`, `:891`).

The UI never shows them. This asks for a first-class **`internal`** address type, rendered on the service Interfaces page like any other address and copyable.

Where they get dropped today:

- `projects/start-os/web/ui/src/app/services/gateway.service.ts:33-37` — the gateway list filters out `deviceType === 'bridge'` and `'loopback'`.
- `projects/start-os/web/ui/src/app/routes/portal/components/interfaces/interface.service.ts:264-282` — `buildGatewayGroups` seeds its group map from that gateway list and drops any address whose gateway isn't in it (`if (!list) continue`); `utils.filterNonLocal` then strips localhost/link-local/bridge again unless the browser itself is on localhost.
- `getAddressType()` (`interface.service.ts:90`) has no case for it — a bridge address arrives as `HostnameMetadata::Ipv4 { gateway: lxcbr0 }`, indistinguishable from a LAN IPv4 by kind alone.

There is no `isInternal` equivalent anywhere in `shared-libs/ts-modules` or `projects/start-os/web`.

### Describe the Use Case

A user running two services on one server who wants one to talk to the other. The live example is Uptime Kuma monitoring and sending ntfy notifications to another package on the same box (Start9Labs/uptime-kuma-startos#12).

Every address the UI offers them is a TLS address whose chain terminates at that box's own root CA (`<hostname> Local Root CA` → `StartOS Local Intermediate CA` → leaf). Nothing inside a service container trusts that CA — there is no CA plumbing anywhere in start-core's service/LXC path or in container-runtime — so the request dies with `SELF_SIGNED_CERT_IN_CHAIN`.

The internal address would just work, with no TLS involved: the OS installs a direct DNAT for it, e.g.

```
[...]

PR #3619, opened 2026-08-03
feat(sdk): add sdk.getRootCa for this server's root CA

MattDHill:
Adds `sdk.getRootCa(effects)`, returning this server's root CA certificate as PEM.

## Why

A service that dials an address the **user** supplies — a monitor target, a notification endpoint, a webhook — can't resolve a bridge address for it, and users can't see bridge addresses anyway ([#3618](https://github.com/Start9Labs/start-technologies/issues/3618)). So what they paste is whatever StartOS showed them, which on the LAN is always HTTPS with a certificate chaining to this server's root CA. No container trusts that root, so the dial fails verification with `SELF_SIGNED_CERT_IN_CHAIN`.

Until now the only way to get that root was to mint a certificate you didn't want and take the last link of the chain — and packages doing it by hand have taken the wrong link:

| Package | Code | Result |
| --- | --- | --- |
| `hermes-agent` | `certs.slice(-1)` | correct |
| `openclaw` | `certs.slice(-1)` | correct |
| `dev-env` | `certs[0]` | **the leaf** — installed as a trust anchor, trusts nothing, raises no error |

One in three wrong, on a call whose failure is invisible. That is the whole argument for this change.

## Implementation note

Built over `getSslCertificate` rather than as a new OS effect, so it works on shipping StartOS instead of gating every caller on the next release. The pretext hostname is the OS bridge IP — the one hostname admitted by the allowlist's compiled-in `HOST_IP` subnet check (`net/ssl.rs`) rather than by whatever the gateway table currently holds, which is what admits `127.0.0.1`.

If the OS later exposes the root directly, only `GetRootCa.fetch` changes; the SDK surface is already right and no package needs touching.

It returns a `Watchable` to match the other readers, not to fix a reactivity problem — the root is minted once at setup (`AccountInfo::new`, `setup.rs`) with a ten-year life and there is no rotation flow, so in practice it never changes.

## Changes

- `shared-libs/ts-modules/start-core/lib/util/GetRootCa.ts` — new `Watchable<string>`
- `projects/start-sdk/lib/StartSdk.ts` — `sdk.getRootCa`
- `docs/src/service-to-service.md` — new "Trusting this server's certificates" section, with the "don't index `getSslCertificate`" warning and the rootfs-not-volume placement rule
- `docs/src/recipes.md`, `CHANGELOG.md` (2.0.10, unreleased)

## Verified

- `tsc` clean in `shared-libs/ts-modules/start-core` and `projects/start-sdk` (the latter built against this branch's `start-core` dist, not the installed one)
- start-core TS suite: 86 passed, 9 skipped
- `prettier --check` clean on every changed file

Not exercised against a running service — the SDK surface is a thin wrapper over an effect that already works, but the trust install itself is verified end-to-end in the consumer PR below.

## Follow-on

- [Start9Labs/uptime-kuma-startos#12](https://github.com/Start9Labs/uptime-kuma-startos/pull/12) is the first consumer; it currently hand-rolls this and can collapse to `sdk.getRootCa`.
- [#3618](https://github.com/Start9Labs/start-technologies/issues/3618) is the underlying gap — if internal addresses become visible in the UI, most callers won't need this at all.


shared-libs/ts-modules/start-core/lib/util/GetRootCa.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Reactive reader for this server's root CA certificate, PEM encoded.
+ *
+ * Every LAN-facing address StartOS advertises — mDNS, LAN IP, private domain —
+ * is served with a certificate chaining to this root, and nothing inside a
+ * service container trusts it by default. A container that has to reach another
+ * service on this server over HTTPS needs this certificate installed, either
+ * into its own trust store or via a runtime that takes one (e.g. Node's
+ * `NODE_EXTRA_CA_CERTS`).
+ */
+export class GetRootCa extends Watchable<string> {
```
dr-bonez: Expires in 10 years... Do we really need to make it watchable?

shared-libs/ts-modules/start-core/lib/util/GetRootCa.ts
```diff
@@ -0,0 +1,29 @@
...
+  protected readonly label = 'GetRootCa'
+
+  // The OS exposes the root only as the last link of a fullchain, so obtaining
+  // it means asking for a leaf we discard. The OS's own bridge address is the
+  // one hostname guaranteed to be issuable: it is the sole entry admitted by a
+  // compiled-in subnet check rather than by whatever the gateway table happens
+  // to hold. Collapses to a single call if the OS ever exposes the root direct.
+  protected async fetch(callback?: () => void) {
+    const [, , rootCa] = await this.effects.getSslCertificate({
+      hostnames: [await this.effects.getOsIp()],
```
dr-bonez: Can this be an empty array? Might be nice to not have to make 2 separate rpc calls

shared-libs/ts-modules/start-core/lib/util/GetRootCa.ts
MattDHill: No — dropped it. It's a plain `Promise<string>` now, same as `getOsIp`: minted once at setup, ten-year life, no rotation flow, so there's nothing to subscribe to.

shared-libs/ts-modules/start-core/lib/util/GetRootCa.ts
MattDHill: Yes, and it now is — one RPC instead of two. The validation loop has nothing to check, and `cert_for` mints the discarded leaf with `SANInfo::new` over an empty set; `make_leaf_cert` already falls back to `CN=localhost` with no DNS name, and an empty `SubjectAlternativeName` builds fine on openssl 0.10.80 (I compiled a throwaway crate against the pinned version to be sure — the modern stack-based `build` encodes an empty `GENERAL_NAMES` rather than choking on an empty config string, which is what would have failed on older rust-openssl).

Nice side effect: the leaf it leaves behind is now inert rather than a plausible-looking certificate for the OS's own bridge address.

helix-nine:
Done in 1694b09 — `getRootCa` now takes the last link instead of index 2.

```diff
-  const [, , rootCa] = await effects.getSslCertificate({ hostnames: [] })
-  return rootCa
+  const chain = await effects.getSslCertificate({ hostnames: [] })
+  return chain[chain.length - 1]
```

To be clear about what this does and doesn't fix: index 2 **is** the root today, on every reachable path. `fullchain_ed25519`/`fullchain_nistp256` are straight-line `vec![leaf, int, root]` literals with no branch (`shared-libs/crates/start-core/src/net/ssl.rs:193-198`), the effect maps them to PEM in order (`service/effects/net/ssl.rs:119-129`), and `EffectCreator.ts:261-266` is a verbatim passthrough. So this is not a live defect and no test can distinguish the two forms against the current OS.

What makes it worth changing is where the `3` is actually pinned. Rust's signature is `Result<Vec<String>, Error>` (`service/effects/net/ssl.rs:41`) — unbounded. There is no ts-rs binding for the response; `Effects.ts:237` hand-asserts `Promise<[string, string, string]>`, and the container runtime reaches it via a bare `as ReturnType<...>` cast with no length validation. So a chain that ever grew a link would not fail a build — it would silently return a cross-signing intermediate and install it as a trust anchor, which is the same silent-failure class the doc's WARNING is about. Indexing from the end names the invariant the OS holds to (root last) rather than the depth it happens to have. `start-tunnel` already does exactly this — `tunnel/web.rs:521-529` pops the chain to print the Root CA.

The comment also lost its now-redundant first clause; the load-bearing half is why we ask for a cert we don't want.

**Verified**

- `tsc --noEmit` clean in `shared-libs/ts-modules/start-core` and in `projects/start-sdk` (built against this branch's `start-core` dist, not the installed one). `chain[chain.length - 1]` types as `string` with no assertion — `noUncheckedIndexedAccess` is not set in this package. (`.at(-1)` would not have compiled at all: `target` is es2021.)
- start-core suite: 86 passed, 9 skipped — unchanged.
- `prettier@3.8.3 --check` clean.
- Ran the emitted `getRootCa.js` against fake chains to confirm the property rather than assert it: `[leaf,int,root]`, `[leaf,int,x-signed,root]`, `[leaf,root]` and `[root]` all return `root`. The old form returns `root`, `x-signed`, `undefined`, `undefined`.

**One thing I did not do:** widen `Effects.ts:237` off the 3-tuple to match the Rust `Vec<String>`. That is where the arity assumption actually lives, and leaving it means `chain.length` still has literal type `3` — the change buys runtime robustness, not type safety. Widening ripples to `GetSslCertificate.ts:4` (`Watchable<[string, string, string]>`) and `interfaces.md:344`, and it is a breaking type change for any package destructuring the result, so it did not seem like something to fold into a review fix. Say the word if you want it here rather than as a follow-up.

dr-bonez (approved):

PR #3622, opened 2026-08-03
docs(wrt): Settings fixes, ChromeOS CA trust, DIY Wi-Fi via web UI

Dominion5254:
Three user-guide improvements, each verified against the code/UI before writing:

Settings page mirrors the screen + Remote Access explained (tester feedback)

- Reordered the Settings doc to the actual tab order (General, Password, SSH Keys, Activity, Logs, Backup, Advanced) and added the previously undocumented SSH Keys and Backup sections, each linking to its dedicated page. The General tab's subsections now follow their on-screen order too.
- Rewrote Remote Access around the question the old text never answered — how to manage the router when away from home. It now leads with the Inbound VPN (works regardless of the setting, no port forwards or whitelists needed), then describes each mode by who can reach the admin UI from the WAN side; notably, "When behind NAT" only admits private sources while the WAN address is itself private, so the public Internet can never reach the UI in that mode (semantics verified in system.rs).
- Initial Setup's "Explore the Web Interface" bullets now link each feature to its documentation page.

ChromeOS root-CA trust instructions

- Added a ChromeOS tab to both "Trusting Your Root CA" pages (StartWRT and StartOS) — Chromebooks have no keychain/MMC/profile/terminal trust store, so none of the existing five platform tabs applied. Leads with the current chrome://certificate-manager UI, keeps the older Authorities-tab dialog as a fallback, and notes the rename-to-.pem escape hatch (our .crt is already PEM).
- Extended the canonical global="platform" label set in the start-docs conventions with ChromeOS; a page may omit a platform that genuinely doesn't apply (tabs.js degrades gracefully). The start-os inbound-vpn and start-tunnel installing tab groups deliberately don't get a ChromeOS tab — they'd need real ChromeOS VPN content.
- ⚠️ The ChromeOS steps are sourced from current documentation, not bench-tested — worth verifying on a real Chromebook before release.

DIY Wi-Fi password setup routed through the web interface

- The DIY/unprogrammed-board instructions no longer send users to startwrt-cli set-wifi-password over Ethernet/serial. The web UI handles this end to end: after initial setup, adding the first Admin-profile password on Points of Entry → Wi-Fi → Passwords switches the radios on (auto_enable_radios_on_first_admin_password in wifi.rs) — the docs call out selecting Admin explicitly, since a first password on any other profile leaves the AP down. Cross-references in wifi.md, factory-reset.md, and faq.md aligned; the CLI remains documented in the developer-facing init-reflash.md for headless provisioning.


projects/start-wrt/CHANGELOG.md
```diff
@@ -5,7 +5,24 @@ All notable changes to StartWRT are documented here.
...
+  existing platforms' steps applied to Chromebooks.
+
+### Changed
+
+- **The DIY Wi-Fi password instructions now use the web interface.** Bringing
+  up Wi-Fi on a board with no EEPROM password no longer routes through the
+  command line: connect over Ethernet, complete initial setup, and add an
+  Admin-profile password on the Wi-Fi Passwords page — the first Admin
+  password switches the radios on. The `startwrt-cli set-wifi-password`
+  utility still exists for headless provisioning.
```
dr-bonez: Documentation changes do not belong in the changelog.

Dominion5254: removed

dr-bonez (dismissed):

dr-bonez (approved):

PR #3632, opened 2026-08-05
fix(start-os): route v6 replies out the interface that owns the source address

dr-bonez:
An inbound IPv6 connection to a tunnel-delegated GUA hangs until it times out.

## Root cause

The reply path is at fault, not the inbound one. The SYN arrives fine and conntrack shows the interface mark set correctly, but the entry stays `[UNREPLIED]`:

```
tcp 6 104 SYN_SENT src=2001:470:... dst=...:f002 dport=443 [UNREPLIED] ... mark=1005
```

StartOS sends a reply back out the interface its connection arrived on by restoring a CONNMARK in a `type route` output chain. That works for **forwarded** traffic. It cannot catch a **locally generated** reply: the kernel routes a SYN-ACK *before* the output hook runs (`ir_mark` = `sk_mark` = 0, since `tcp_fwmark_accept` is 0). So it falls to the priority-75 catch-all, reaches the default outbound's table, and hits the `blackhole default` leak guard installed when that gateway carries no v6 of its own.

A blackhole lookup returns `EINVAL`, so `tcp_v6_send_synack` drops the packet **before it is ever built** — leaving the reroute chain no packet to act on.

IPv4 escapes this by accident: its catch-all table holds a real default, so the packet *is* built, hits the output hook, gets its mark restored, and is rerouted correctly. That is the exact v4/v6 asymmetry observed — over v4 the box replies to a source outside the tunnel subnet; the identical v6 test fails.

Tell-tale on an affected box:

```
$ ip -6 route get <external>              # blackhole
RTNETLINK answers: Invalid argument
$ ip -6 route get <server tunnel addr>    # fine
... dev wg1 table 1002 ...
```

## Fix

Install a priority-60 `from <global v6 addr> lookup <iface table>` rule per global v6 address on the interface, sitting between the priority-50 CONNMARK rule and the priority-75 catch-all. The source address *is* known at SYN-ACK route-lookup time, so this routes the reply out the interface that owns it.

Rules reconcile as addresses change, and are GC'd for removed interfaces alongside the existing priority-50 rules.

## Verification

Reproduced and fixed against a live tunnel client. The rule alone takes HTTPS over IPv6 from a hard timeout to a normal response:

| | result |
|---|---|
| before | `code=000`, 10.0s timeout |
| after | `code=200`, **0.25s** |

Confirmed the server side was never at fault — proxy NDP, routing, the nft ruleset and forwarding into the tunnel all behaved correctly; the SYN was observed arriving on `eth0` and leaving on the wg interface.

Also verified the printed rule form (`from 2604:… lookup 1005` — bare compressed address) matches what the reconciler compares against, so it does not churn on every poll.

`cargo check -p start-core` and pinned rustfmt are clean.

## Scope

This does **not** change unbound outbound v6, which still blackholes by design — a `from` rule cannot match a socket whose source address is unselected at route-lookup time, and the blackhole is the deliberate leak guard against v6 escaping via an arbitrary interface. A socket **bound** to its tunnel GUA does now get working outbound v6 (verified: works for an address with the rule, fails for one without).

No docs change: `gateways.md` and `clearnet.md` already describe delegated-GUA IPv6 as working, so this restores documented behavior.

## Left out deliberately

The changelog entry is filed under a new `## [0.4.0.2]` heading, but the version mechanics are **not** included — no manifest bump, no `version/v0_4_0_2.rs` migration node. Cutting the number is a release decision. Outstanding when it is cut: root `package.json` + `package-lock.json`, the `projects/start-os/Cargo.toml` label + `Cargo.lock`, the migration node with its 5 `mod.rs` updates, and the docs release links.


helix-nine:
### Code review

Rebased onto `master` (`33a532c`) and force-pushed. The only conflict was that both sides opened a `## [0.4.0.2]` heading in `projects/start-os/CHANGELOG.md` — resolved by keeping #3615's merged login-banner entry and appending this one under the same `### Fixed`, in landing order. `cargo check -p start-core` and `make start-core-format-check` are both clean on the rebased head.

No blocking issues. Two notes.

**The no-churn claim holds — verified independently.** I round-tripped Rust's `Ipv6Addr::to_string()` output through `ip -6 rule add`/`show` in a netns across the compression edge cases (`2001:db8:0:0:0:0:0:5`, `2600:1f18:0:0:1:0:0:1`, `2607:f8b0:4004:0c07::64`): the printed selector is byte-identical to what `global_v6` holds, bare with no `/128`, so the add/delete sets can't oscillate. Running both new parsers verbatim over real `ip -6 rule show` output also behaves — `existing_src` takes only its own table's addresses and ignores the priority-50 fwmark line and a priority-`600` rule, and the GC skips active tables and only emits stale `(table, addr)` pairs.

**One precision point, on lines this PR adds.** The new comment says the CONNMARK rule "cannot catch a locally generated reply", and the rustdoc says "a host-terminated one is routed before the output hook can restore that mark":

https://github.com/Start9Labs/start-technologies/blob/4eebea3592f8e8d08cc7d3ced0c73f22ed7ad91a/shared-libs/crates/start-core/src/net/gateway.rs#L2224-L2232

https://github.com/Start9Labs/start-technologies/blob/4eebea3592f8e8d08cc7d3ced0c73f22ed7ad91a/shared-libs/crates/start-core/src/net/gateway.rs#L2063-L2069

Both are broader than the defect. `mangle_output` is deliberately a `type route` chain precisely so restore-mark *does* reroute locally-generated packets:

https://github.com/Start9Labs/start-technologies/blob/4eebea3592f8e8d08cc7d3ced0c73f22ed7ad91a/shared-libs/crates/start-core/src/net/startos-base-v6.nft#L14-L17

What the mark can't save is specifically the reply that *opens* the connection, because the route lookup happens inside `tcp_v6_send_synack` before there is an skb for the output hook to act on — which is how the changelog entry words it ("the kernel routes the reply that _opens_ a connection"):

https://github.com/Start9Labs/start-technologies/blob/4eebea3592f8e8d08cc7d3ced0c73f22ed7ad91a/projects/start-os/CHANGELOG.md#L22-L26

As written the new comments also read as contradicting the untouched one 30 lines above, which credits the priority-50 rule with host-terminated replies:

https://github.com/Start9Labs/start-technologies/blob/4eebea3592f8e8d08cc7d3ced0c73f22ed7ad91a/shared-libs/crates/start-core/src/net/gateway.rs#L2192-L2198

Narrowing the two new comments to the connection-opening reply resolves that without having to touch the older one. Comment-only; nothing about the routing behaviour changes.

helix-nine:
Applied in 36e5d2012 — the one review item, comment-only.

Both new comments now say "the reply that opens a connection" instead of "a locally generated reply" / "a host-terminated one". That is the accurate scope: `mangle_output` is a `type route` chain specifically so restore-mark reroutes locally generated packets, so an established connection's replies do get rerouted; what the mark can't save is the SYN-ACK, routed inside `tcp_v6_send_synack` before there's an skb for the output hook. It also removes the apparent contradiction with the priority-50 block's comment 30 lines above, so that one stays untouched.

No behavioural change. `cargo check -p start-core` and `make start-core-format-check` clean at the new head.

Dominion5254 (dismissed):

helix-nine:
@Dominion5254 heads up — this push dismissed your approval. Sorry for the round trip; the PR had gone `CONFLICTING` again, so it couldn't have merged as approved either way.

Master picked up 9 commits since ada9a6133 and `projects/start-os/CHANGELOG.md` conflicted in the `## [0.4.0.2]` section again. Resolved in 309506392 by merging master in, matching the strategy already used on this branch rather than re-rebasing.

The delta is CHANGELOG-only:

- keeps all four `### Fixed` entries — login banner, this PR's IPv6 one, notification checkboxes (#3616), UDP port mapping (#3604)
- preserves master's relative ordering of its three, and this PR's entry where it already sat
- restores the blank line between entries that master's section uses and the previous merge had dropped

Net diff against master is still exactly the two intended files: `gateway.rs` (+131/-17) and the one CHANGELOG entry (+11). `cargo check -p start-core`, `make start-core-format-check` and `prettier --check` are clean at the new head, and it is now `MERGEABLE`.

One note since the PR body defers it: master has since cut 0.4.0.2 for real — `version/v0_4_0_2.rs` wired into `mod.rs` as `Current`, root `package.json` at `0.4.0.2`, `projects/start-os/Cargo.toml` at `0.4.0-rev.2`, `man/start-container.1` regenerated. So the "Left out deliberately" section is moot; nothing version-related is outstanding on this branch.

Dominion5254 (approved):

PR #3634, opened 2026-08-06
feat(start-wrt): automatic port forwarding (PCP + UPnP IGD server)

### LINKED ISSUE #3451: StartWRT — Published Ports: message the port 443/80/22 ↔ web UI / Remote Access collision
_Migrated from Start9Labs/start-wrt#74._

### Context

Forwarding a port in **Published Ports** creates an fw4 `redirect` (`src=wan`, `target=DNAT`) for that `dest_port`. The router's own web UI listens on `[::]:80` / `[::]:443`. Because the DNAT runs in `nat prerouting` — *before* the routing decision that would deliver a packet to the router's local socket — forwarding **443** (or **80** / **22**) rewrites WAN-side traffic for that port to the target device before it can ever reach the router's own listener.

### Impact

- **Default config: no problem.** The `wan` zone is `input REJECT`, and Remote Access defaults to `default` mode, which only opens 80/443/22 on the WAN when the WAN IP is *itself* private (double-NAT), scoped to RFC1918/ULA sources. On a normal public-IP WAN it opens nothing. And LAN access to the UI (via the router's LAN IP) is never affected — the redirect is `src=wan`-scoped, so LAN→LAN-IP traffic never traverses it.
- **Remote Access = `always` + forwarding 443: silent collision.** `always` unscopes 80/443/22 on the WAN input to expose the UI over the internet (help text: *"manage the router's settings and configurations from outside the local network"*). But a 443 port-forward's prerouting DNAT hijacks that traffic to the forwarded device — an input-chain ACCEPT can't win against a prerouting rewrite. The user loses remote UI access with no indication why.

### Ask (messaging)

When a user forwards a **router-reserved port (80 / 443 / 22)** in the Published Ports flow, surface a message explaining that it can override the router's own web UI / SSH / Remote Access on the WAN side. Scope here is **communication, not a hard block** — just make the tradeoff visible instead of silent.

### Out of scope

The default posture is safe; this is specifically about clarity for the `always` / remote-admin case.

Dominion5254:
What this does

Implements the server half of gateway autoconfiguration (start-os #3306) for StartWRT: devices on the LAN can request WAN port forwards to themselves via PCP (UDP 5351) or UPnP IGD (SSDP + HTTP control on 49001), built on the shared start-core protocol cores. A new GatewayBackend impl maps requests onto auto-tagged UCI redirects (_apf_*), with a per-minute lease-expiry sweep.

Authorization is per-device and off by default. An "Allow automatic port forwarding" toggle on the device detail page sets an _allow_pcp flag on the device's DHCP host entry; only flagged devices can create forwards, and only to themselves. The Published Ports page gains a read-only "Automatic" section showing what's currently held.

No forward outlives the user's intent:
- A "permanent" mapping request is granted a week, not forever — an unattended router never carries a WAN port with nothing scheduled to close it. Clients that still want the port renew (in memory, no flash writes).
- Turning the toggle off — or forgetting the device — closes that device's forwards immediately.
- A forward is bound to the DHCP assignment behind it: if the owner loses the address it points at, the sweep drops it (RFC 6887 §15, RFC 6970 §5.10).
- Conflicts are enforced both ways: a device request overlapping any enabled WAN DNAT redirect is refused; a manual publish over an auto-held port wins and removes the auto forward.

Security hardening that rode along

- Router self-collision, both halves of #3451. A prerouting DNAT beats an input ACCEPT, so a forward on a port the router answers on from the WAN (Remote Access 80/443/22, an inbound VPN's listen port) silently diverts the router's own service. The automatic path refuses these outright, reading the reservation from the live WAN-input rules rather than a hardcoded list. The manual path warns: published-ports.set detects the collision (transport-aware — a UDP forward on 443 collides with nothing), applies nothing, and the UI raises a one-time-per-rule confirmation.
- GetExternalIPAddress now requires an authorized client, matching AddPortMapping, so guest-VLAN hosts can't read the WAN IP through the well-known control path.
- Firewall changes now apply via fw4 reload (one atomic nft -f transaction) instead of restart (stop then start — a window with no firewall and no NAT). A failed reload leaves the running ruleset in place; restart remains only as the fallback for the not-yet-loaded case.

Shared start-core changes (affect StartTunnel)

- The IGD server is now recognizable to real third-party clients — discovered on hardware when upnpc refused the router. Added WANCommonInterfaceConfig (the service clients use to identify an IGD at all), GetStatusInfo/GetConnectionTypeInfo/GetNATRSIPStatus/GetCommonLinkProperties, and mapping read-back (GetSpecificPortMappingEntry/GetGenericPortMappingEntry, scoped to the requesting peer's own forwards). Additive throughout; StartOS's client is unaffected.
- root_desc.xml no longer hardcodes "StartTunnel" — the gateway identifies as the product serving it.
- The IGD layer honors NewLeaseDuration, so a lease-tracking backend expires mappings on the schedule reported to the client (StartTunnel's permanent DB-persisted forwards unchanged).
- GatewayBackend::sni() replaces supports_hostname(): a gateway without an SNI dataplane can't advertise the HOSTNAME capability by accident.

UI and docs

- The permission toggle lives inside the device settings form (Save applies, Cancel discards; revocation never fires from merely renaming a device), styled per house patterns (fieldset/legend, visible secondary copy).
- Layout fixes so pages stay fully visible on small screens: the help aside no longer forces the shell past the viewport, the nav rail collapses in step with the mobile pinning, and a hand-expanded sidebar yields a real horizontal scrollbar instead of clipping Save.
- New user-guide page for automatic forwards under Published Ports; changelog entries filed for both StartWRT and StartTunnel (the StartTunnel entries under its unreleased version).
- All five locale dictionaries updated.

Verification

- 492 cargo tests, check:wrt, build:wrt, check:i18n:wrt green.
- IGD surface validated on hardware against upnpc (miniupnpc); small-screen layout verified headless via CDP.


projects/start-wrt/web/src/app/app.ts
```diff
@@ -133,10 +139,25 @@ import { Language } from 'src/app/utils/languages'
...
+  // re-evaluating on navigation re-collapses a nav expanded by hand on mobile.
+  protected readonly open = linkedSignal<string[], boolean>({
+    source: () => [this.breakpoint(), String(this.navigation())],
+    computation: (source, previous) =>
+      previous?.value !== false && source[0] !== 'mobile',
+  })
   protected readonly scrollbar = viewChild(TuiScrollbar, { read: ElementRef })
-  protected readonly _ = inject(Router)
-    .events.pipe(takeUntilDestroyed())
+  protected readonly _ = this.router.events
```
waterplea, line 163: I wonder if we can just use "top" with this:
https://angular.dev/api/router/InMemoryScrollingOptions#scrollPositionRestoration

projects/start-wrt/web/src/app/services/router-port-override.ts
```diff
@@ -0,0 +1,37 @@
...
+import { RouterPortCollision } from 'src/app/services/api/api.service'
+
+/**
+ * If `pending` is non-empty, prompt the user to confirm publishing port(s) the
+ * router itself answers on from the WAN (remote access to its web interface,
+ * SSH, or a VPN server) — the forward would capture that traffic. Returns true
+ * when there is nothing to confirm or the user confirmed, false when they
+ * cancelled.
+ */
+export async function confirmRouterPortOverride(
```
waterplea: It is used just once, do you plan on reusing it or maybe we can just leave it at the place where it's used?

projects/start-wrt/web/src/app/routes/published-ports/service.ts
```diff
@@ -80,25 +132,38 @@ export class PublishedPortsService extends FormService<PublishedPortDisplay[]> {
...
+    label: a.label,
+    deviceMac: a.device_mac,
+    deviceName: a.device_name ?? undefined,
+    ports: a.ports,
+    publicPorts: a.public_ports,
+    expiresSecs: a.expires_secs ?? undefined,
+  }
+}
+
+/** Map backend snake_case response to frontend camelCase types */
```
waterplea, line 180: Do we have to do this everywhere? Maybe we can agree on a single format or come up with a universal util to convert naming?

projects/start-wrt/web/src/app/services/router-port-override.ts
Dominion5254: No plans of reuse. Moved into the service.

projects/start-wrt/web/src/app/routes/published-ports/service.ts
Dominion5254: This is the result of a BE inconsistency present in a couple places. I'll standardize around camelCase in a separate PR.

projects/start-wrt/web/src/app/app.ts
Dominion5254: scrollPositionRestoration goes through ViewportScroller, which scrolls the window — but the window never scrolls in this shell; the scroll container is the tui-scrollbar inside main. Using 'top' would mean providing a custom ViewportScroller pointed at that element, which is more moving parts than the subscription.

If there's a Taiga-friendly way to route the router option into tui-scrollbar, happy to switch.

waterplea: Let's keep it as is for now

waterplea (dismissed):

projects/start-wrt/backend/ctrl/src/bins/daemon.rs
```diff
@@ -334,6 +334,16 @@ async fn inner_main() -> Result<(), Error> {
     // Initialize SSL: ensure Root CA and server cert exist
     let tls_ready = init_ssl().await;
 
+    if !setup_mode {
+        // Port-control servers (PCP + UPnP IGD): automatic port forwarding for
+        // per-device-authorized LAN clients. After init_ssl so the IGD device
+        // UUID (derived from the root CA) is stable.
```
dr-bonez, line 352: what guided the decision to base the UUID on the root CA?

projects/start-tunnel/CHANGELOG.md
```diff
@@ -26,6 +26,33 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  found", or created a mapping and then reported it as failed because it could
+  not read it back. StartOS was unaffected; its client requires none of these.
+
+### Security
+
+- **Reading the tunnel's public IP address over UPnP now requires an authorized
+  device.** The `GetExternalIPAddress` action answered any device that could
+  reach the tunnel's UPnP control endpoint, including one whose **Auto-publish**
+  permission was off. It now answers only devices allowed to create automatic
+  forwards — the same permission `AddPortMapping` already required.
```
dr-bonez: don't see why this is necessary. GetExternalIPAddress should work for whatever WAN the client's profile has access to.

projects/start-wrt/backend/ctrl/src/port_control.rs
```diff
@@ -0,0 +1,2064 @@
+//! PCP + UPnP IGD port-control server ("automatic port forwarding").
```
dr-bonez, line 1: I thought we were doing UPnP only for now, due to the UDP src spoofing issues?

projects/start-wrt/backend/ctrl/src/bins/daemon.rs
Dominion5254: Deriving the UDN from the root CA allows it to be stable across reboots without persisting new state. It also has the benefit of regenerating the CA and UDN on factory reset.

projects/start-wrt/backend/ctrl/src/port_control.rs
Dominion5254: We had discussed that initially but the added protections with default-off (a device with automatic port forwarding enabled can't forward ports of devices with automatic port forwarding disabled) as well as the protections against cross-profile spoofing PCP is no less secure than the UPnP implementation. The only gap is that a device that has opted into automatic port forwarding can forward the ports of other devices that have also opted into automatic port forwarding.

The same gap can be achieved in UPnP with ARP-Poisioning to expose a victim device's ports to the WAN (or delete their mappings) - this also requires both the attacker and victim to have opted into automatic port forwarding. Both devices also need to be on the same profile.

As implemented both PCP and UPnP har hardened beyond their standards, if we are going to remove PCP for the spoofing gap then UPnP shouldn't be supported either.

projects/start-tunnel/CHANGELOG.md
Dominion5254: This was meant to close a gap against DNS rebinding (VPN/Tor deanonymization) on start-wrt in the shared code. This had no security value in start-tunnel because the WAN IP isn't a secret from the WG peer dialing it; this is also functionally free on start-tunnel as a client only reaches this after it passed the same gate at AddPortMapping.

However I did realize this left an opening on start-wrt for a device that had opted in to automatic port forwarding to the DNS rebinding. I added some additional hardening on this point.

dr-bonez: It's not really a secret from start-wrt either? If the device can access the wan, it can figure out its wan ip. This just makes it easier and more reliable.

projects/start-wrt/backend/ctrl/src/port_control.rs
dr-bonez: I'm not sure I agree that an ARP-poisoning attack is the same tier as spoofing a UDP packet src address, but I'll concede that it's not a big deal to support both

projects/start-tunnel/CHANGELOG.md
Dominion5254: Good point. I was conflating the deanonymization issue fixed in 722fc7b03. Removed this gate.

dr-bonez (dismissed):

dr-bonez (approved):

PR #3643, opened 2026-08-07
fix(sdk): don't SIGKILL a database dump copy after 30 seconds

### LINKED ISSUE #3636: [bug]: Backup/restore SIGKILLs the database dump copy after exactly 30 s — `cp` in `Backups.ts` inherits the default `exec` timeout
## Environment

- **StartOS:** 0.4.0.1
- **Affected SDK:** `@start9labs/start-sdk` 2.0.9 (bug present from 1.5.2 through current `master`)
- **Observed on:** Immich v3.1.0 (`immich-startos` 3.1.0:0), backup to a network target

## Current Behavior

For any package using `Backups.withPgDump` / `withMysqlDump`, backups fail once the database dump is large enough that copying it to the backup target takes more than 30 seconds:

```
Immich   Failed: Unknown Error: Error: cp terminated with signal SIGKILL:   1m 55s
```

The dump itself succeeds. The failure is the `cp` that moves the completed dump onto the backup target, and it is killed at **exactly 30.0 s**, every time.

From one server, two consecutive daily backups of the same service (Immich v3.1.0, SDK 2.0.9):

| | 2026-08-05 (succeeded) | 2026-08-06 (failed) |
| --- | --- | --- |
| `[pg-dump] dumping database` | 14:17:20 | 16:14:55 |
| `[pg-dump] copying dump to backup target` | 14:18:45 | 16:16:18 |
| next event | 14:19:09 `stopping postgres` (**cp took 24 s**) | 16:16:48 `cp terminated with signal SIGKILL` (**cp killed at 30.0 s**) |

`pg_dump` took 85 s and 83 s respectively — i.e. the database barely changed. The successful copy completed with only a **6-second margin** under the limit. Once a copy crosses 30 s it fails permanently, and the margin shrinks as the database grows.

There is no OOM or disk-space event in the server logs — the process is killed by the SDK's own timer.

## Expected Behavior

Copying the dump to (or from) the backup target should not be subject to a fixed 30-second timeout. The duration of that copy is proportional to database size and backup-target write speed, neither of which the SDK can bound, exactly like the `pg_dump` / `pg_restore` calls that already opt out of the timeout.

## Steps to Reproduce

1. Install a package whose backup uses `Backups.withPgDump` or `Backups.withMysqlDump` (e.g. Immich, Nextcloud, BTCPay Server, Invidious, Ghost, Mempool).
2. Use it until its database dump is large enough that copying the dump file to the backup target takes longer than 30 seconds (a few GB, or less on a slow/network target).
3. Run a backup.
4. It fails with `Error: cp terminated with signal SIGKILL:` after ~30 s in the copy phase.

### Root cause

`SubContainer.exec()` defaults to a 30-second timeout and enforces it with `SIGKILL`:

```ts
// projects/start-sdk/lib/util/SubContainer.ts:685 (and :1122)
timeoutMs: number | null = 30000,

// projects/start-sdk/lib/util/SubContainer.ts:768-770
if (timeoutMs !== null && child.pid) {
  killTimeout = setTimeout(() => child.kill('SIGKILL'), timeoutMs)
}
```

`execFail()` declares `timeoutMs?: number | null` with **no** default and passes it straight through, so omitting the argument lands on `exec`'s 30 000 ms default.

In `projects/start-sdk/lib/backup/Backups.ts` the long-running database commands correctly opt out with an explicit `null`, but the four `cp` calls that move the dump to/from the backup target do not:

| Line | Call | Timeout |
| --- | --- | --- |
| 293–297 | `pg_dump` | `null` — no timeout ✅ |
| **298** | `cp tmpDumpFile dumpFile` (pg backup) | **30 s** ❌ |
| **318** | `cp dumpFile tmpDumpFile` (pg restore) | **30 s** ❌ |
| **536** | `cp tmpDumpFile dumpFile` (mysql backup) | **30 s** ❌ |
| **592** | `cp dumpFile tmpDumpFile` (mysql restore) | **30 s** ❌ |

### How it was introduced (1.5.2)

`git log -L 293,298:projects/start-sdk/lib/backup/Backups.ts` shows this is a regression rather than an oversight from the start:

- `7ffb4623` added `pg_dump` writing **directly** to `dumpFile` on the backup mount, with an explicit `null` timeout — correct, because that single command did the slow write to the target.
- `336eda75` — *"fix(sdk): stage db dump through /tmp before writing to backup-fs (1.5.2)"* — split that into two steps: `pg_dump` to a local `tmpDumpFile`, then `cp tmpDumpFile dumpFile`. The `null` stayed on `pg_dump`, which is now the *fast local* step, while the newly added `cp` — now the step that actually writes to the slow backup target — was added with no timeout argument and silently picked up the 30 s default.

So the opt-out is still present, but attached to the wrong half of the operation. Affects every SDK release from **1.5.2** through current `master`.

### Secondary defect: a timeout kill is indistinguishable from an OOM kill

When the timer fires, the resulting error is built purely from the signal:

```ts
// projects/start-sdk/lib/util/SubContainer.ts:1279-1281
} else if (result.exitSignal) {
  message = `${command} terminated with signal ${result.exitSignal}: ${result.stderr}`
}
```

It never states that a timeout occurred, or what the limit was — and `cp` writes nothing to stderr when SIGKILLed, so the user-facing notification is:

```
Failed: Unknown Error: Error: cp terminated with signal SIGKILL:
```

A SIGKILL raised by the SDK's *own* timer is therefore reported identically to an OOM kill, a cgroup kill, or an operator `kill -9`. Diagnosing this bug required ruling out memory pressure from server logs and then matching the elapsed time against the default to infer the cause; the notification alone points nowhere.

Suggested improvement, independent of the fix above: when the kill originates from `killTimeout`, say so — e.g. `cp timed out after 30000 ms (SIGKILL)`. That single change would make this class of failure self-diagnosing.

### Impact

- **Backups fail permanently** for any affected package once its dump crosses the threshold. Six packages in a single workspace use these helpers: Immich, Nextcloud, BTCPay Server, Invidious (`withPgDump`), Ghost, Mempool (`withMysqlDump`).
- **Restore is affected too, and is the more serious half.** Line 318 (`cp dumpFile tmpDumpFile`, staging the dump off the FUSE before `pg_restore` reads it) has the same 30 s cap. A database whose dump takes more than 30 s to copy **cannot be restored** — and that is discovered during disaster recovery, when the backup is all the user has.
- The threshold is invisible until crossed: a backup that has succeeded for months starts failing with no configuration change.

### Suggested fix

Pass `null` at the four `cp` sites, matching the convention already used for `pg_dump` / `pg_restore` in the same functions:

```ts
await sub.execFail(['cp', tmpDumpFile, dumpFile], { user: 'root' }, null)
```

If an unbounded copy is unwanted, a generous configurable timeout in the style of the `readyTimeout` option added in #3376 would also work — but it must scale with dump size, not be a fixed 30 s.

### Related

- #3376 made the pg/mysql *readiness* timeout configurable in this same file, but did not touch the `exec` timeouts on the copies.
- #3315 (`cp --reflink` hanging on fragmented volumes) is a different `cp`, in the OS-side pre-update volume snapshot path — noting it only to avoid confusion.

helix-nine:
Fixes #3636.

## The bug

`SubContainer.exec` defaults `timeoutMs` to 30 000 and enforces it with `SIGKILL`. `execFail` declares `timeoutMs` with **no** default and forwards it, so any call site that omits the argument silently inherits 30 s.

`336eda75` (1.5.2) split "`pg_dump` writes onto the backup target" into "`pg_dump` writes to `/tmp`, then `cp` through the FUSE". The explicit `null` stayed on `pg_dump`; the new `cp` — the step whose duration is set by the size of the dump and the speed of the target — got none. Once a dump grew past a thirty-second copy, every backup failed:

```
Failed: Unknown Error: Error: cp terminated with signal SIGKILL:
```

Restore stages the dump off the target through the same kind of copy under the same cap, so a database whose dump took longer than 30 s to copy **could not be restored at all** — discovered during recovery, when the backup is all the user has.

## What changed

**Every step of a dump or restore whose duration follows the data now opts out**, through a named `NO_TIMEOUT` (which also replaces the file's existing bare `null`s, so the opt-out is greppable): both copies, `pg_ctl` start/stop, the recursive `chown`s over the data directory, `initdb`, `mysql_install_db` / `mysqld --initialize-insecure`, and the foreground `mysqld` MariaDB runs for the length of the dump. The steps left at 30 s touch a fixed, tiny number of inodes (`touch`, `mkdir -p`, `rm -f`, single-file `chown`, `createdb`, `psql ALTER USER`) or are already bounded by their own poll loop (`pg_isready`, `mysqladmin ping`).

**`readyTimeout` now reaches `pg_ctl`.** `pg_ctl` bounds itself — `-w` is its default and it gives up after `-t` seconds, 60 by default — so lifting the SDK cap alone would only have moved the ceiling 30 s → 60 s, out of reach of the one knob packages are told to raise for slow clusters. `readyTimeout` now supplies `-t`. Its 60 000 ms default is `pg_ctl`'s own default, so behaviour is unchanged until someone raises it.

**A timeout kill now says it timed out.** The error was built from the signal alone, so the SDK's own timer was indistinguishable from an OOM kill, a cgroup kill, or an operator's `kill -9` — and `cp` writes nothing to stderr when killed, leaving the bare `cp terminated with signal SIGKILL:` above as the entire notification. `exec`'s result now carries `timedOutAfterMs` and the message names the limit:

```
cp timed out after 30000ms and was killed with SIGKILL:
```

The flag is decided in the **exit handler**, not the timer: the deadline can elapse after something else has already killed the process (the `'exit'` event is delivered in the poll phase, after timers), and setting it from the timer reported an external `SIGTERM` as an SDK timeout — the same misattribution inverted. `timedOutAfterMs` is set only when our timer fired *and* the process died of `SIGKILL`.

**Docs.** The packaging guide documented `exec`/`execFail` without ever mentioning the 30 s default — the omission this issue is made of, and its example was a `git clone`. It now documents the third argument. Two TSDoc claims that 1.5.2 falsified are corrected: the dump is staged in `/tmp` rather than "written directly to the backup target — no data duplication on disk", and `withMysqlDump` referred to a `dumpVolume` option that does not exist in `MysqlDumpConfig`.

While in the file, the four hand-rolled structural `sub:` parameter types in `Backups.ts` are replaced with the real `SubContainer<M>` — one of them had to be widened by hand just to accept the new third argument.

## Verification

- `make check`, `make test` (87), `make check-fmt`, repo-root `prettier --check`, `make bundle` — all clean. New `exitError.test.ts` covers the three message branches.
- `pg_ctl`'s self-bounding was measured, not assumed: in `postgres:16-alpine`, a `smart` stop held open by a live client and `PGCTLTIMEOUT=5` returned **exit 1 at exactly 5 s**. So dropping the SDK cap on `pg_ctl` cannot hang a backup — it converts a `SIGKILL` into a real `pg_ctl` error.
- Worth knowing, and not addressed here: the SDK's `SIGKILL` lands on the host-side `start-container subcontainer exec` wrapper. `SIGKILL` cannot be in that wrapper's forwarded-signal set (`FWD_SIGNALS`), and it `setns`es into the subcontainer's PID namespace before spawning the real command, so the `cp` itself is not signalled — it is orphaned until the subcontainer is torn down. That is why the timeout note in the guide describes a timeout as "the SDK stopped waiting", not "the work stopped".

Not verified end-to-end on a server with a multi-gigabyte dump; the failing path is a wall-clock timer, which is impractical to exercise here.

### DIFF
--- a/projects/start-sdk/lib/backup/Backups.ts
+++ b/projects/start-sdk/lib/backup/Backups.ts
@@ -12,6 +12,8 @@ import {
 
 const BACKUP_HOST_PATH = '/media/startos/backup'
 const BACKUP_CONTAINER_MOUNT = '/backup-target'
+// these steps can outlast exec's 30 s default
+const NO_TIMEOUT = null
 
 /** A password value, or a function that returns one. Functions are resolved lazily (only during restore). */
 export type LazyPassword = string | (() => string | Promise<string>) | null
@@ -41,7 +43,7 @@ export type PgDumpConfig<M extends T.SDKManifest> = {
   initdbArgs?: string[]
   /** Additional options passed to `pg_ctl start -o` (e.g. '-c shared_preload_libraries=vectorchord'). Appended after `-c listen_addresses=`. */
   pgOptions?: string
-  /** Milliseconds to wait for PostgreSQL to accept connections before failing (default 60000). Raise for large clusters that need longer to start or run crash recovery. */
+  /** Milliseconds to wait for PostgreSQL to accept connections before failing (default 60000). Also passed to `pg_ctl` as `-t`, bounding its own wait for startup and for the shutdown checkpoint. Raise for large clusters that need longer to start, run crash recovery, or shut down. */
   readyTimeout?: number
 }
 
@@ -193,7 +195,8 @@ export class Backups<M extends T.SDKManifest> implements InitScript {
    * this uses `pg_dump` to create a logical dump before backup and `pg_restore` to rebuild
    * the database after restore.
    *
-   * The dump file is written directly to the backup target — no data duplication on disk.
+   * The dump is staged in the subcontainer's `/tmp` and copied to the backup
+   * target, so it needs transient room for one copy of the dump.
    *
    * @returns A configured Backups instance with pre/post hooks. Chain `.addVolume()` or
    * `.addSync()` to include additional volumes/paths in the backup.
@@ -214,6 +217,7 @@ export class Backups<M extends T.SDKManifest> implements InitScript {
       readyTimeout = 60_000,
     } = config
     const pgdata = `${mountpoint}${pgdataPath}`
+    const pgCtlTimeout = String(Math.ceil(readyTimeout / 1000))
     const dumpFile = `${BACKUP_CONTAINER_MOUNT}/${database}-db.dump`
     // pg_dump's writes are silently dropped on the backup-fs FUSE mount —
     // pg_dump exits 0 but the file stays 0 bytes. `cp` writes through the
@@ -231,17 +235,7 @@ export class Backups<M extends T.SDKManifest> implements InitScript {
       })
     }
 
-    async function startPg(
-      sub: {
-        exec(cmd: string[], opts?: any): Promise<{ exitCode: number | null }>
-        execFail(
-          cmd: string[],
-          opts?: any,
-          timeout?: number | null,
-        ): Promise<any>
-      },
-      label: string,
-    ) {
+    async function startPg(sub: SubContainer<M>, label: string) {
       await sub.exec(['rm', '-f', `${pgdata}/postmaster.pid`], {
         user: 'postgres',
       })
@@ -255,9 +249,20 @@ export class Backups<M extends T.SDKManifest> implements InitScript {
       const pgStartOpts = pgOptions
         ? `-c listen_addresses= ${pgOptions}`
         : '-c listen_addresses='
-      await sub.execFail(['pg_ctl', 'start', '-D', pgdata, '-o', pgStartOpts], {
-        user: 'postgres',
-      })
+      await sub.execFail(
+        [
+          'pg_ctl',
+          'start',
+          '-D',
+          pgdata,
+          '-t',
+          pgCtlTimeout,
+          '-o',
+          pgStartOpts,
+        ],
+        { user: 'postgres' },
+        NO_TIMEOUT,
+      )
       for (let elapsed = 0; elapsed < readyTimeout; elapsed += 1000) {
         const { exitCode } = await sub.exec(['pg_isready', '-U', user], {
           user: 'postgres',
@@ -292,14 +297,20 @@ export class Backups<M extends T.SDKManifest> implements InitScript {
             await sub.execFail(
               ['pg_dump', '-U', user, '-Fc', '-f', tmpDumpFile, database],
               { user: 'postgres' },
-              null,
+              NO_TIMEOUT,
             )
             console.log('[pg-dump] copying dump to backup target')
-            await sub.execFail(['cp', tmpDumpFile, dumpFile], { user: 'root' })
+            await sub.execFail(
+              ['cp', tmpDumpFile, dumpFile],
+              { user: 'root' },
+              NO_TIMEOUT,
+            )
             console.log('[pg-dump] stopping postgres')
-            await sub.execFail(['pg_ctl', 'stop', '-D', pgdata, '-w'], {
-              user: 'postgres',
-            })
+            await sub.execFail(
+              ['pg_ctl', 'stop', '-D', pgdata, '-w', '-t', pgCtlTimeout],
+              { user: 'postgres' },
+              NO_TIMEOUT,
+            )
             console.log('[pg-dump] complete')
           },
         )
@@ -315,17 +326,23 @@ export class Backups<M extends T.SDKManifest> implements InitScript {
             await mountBackupTarget(sub.rootfs)
             // Stage the dump off the FUSE before pg_restore reads it — see
             // the comment on `tmpDumpFile` above.
-            await sub.execFail(['cp', dumpFile, tmpDumpFile], { user: 'root' })
+            await sub.execFail(
+              ['cp', dumpFile, tmpDumpFile],
+              { user: 'root' },
+              NO_TIMEOUT,
+            )
             await sub.execFail(['chown', 'postgres:postgres', tmpDumpFile], {
               user: 'root',
             })
             await sub.execFail(
               ['chown', '-R', 'postgres:postgres', mountpoint],
               { user: 'root' },
+              NO_TIMEOUT,
             )
             await sub.execFail(
               ['initdb', '-D', pgdata, '-U', user, ...initdbArgs],
               { user: 'postgres' },
+              NO_TIMEOUT,
             )
             await startPg(sub, 'pg-restore')
             await sub.execFail(['createdb', '-U', user, database], {
@@ -343,7 +360,7 @@ export class Backups<M extends T.SDKManifest> implements InitScript {
                 tmpDumpFile,
               ],
               { user: 'postgres' },
-              null,
+              NO_TIMEOUT,
             )
             if (resolvedPassword !== null) {
               await sub.execFail(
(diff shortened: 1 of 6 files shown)


projects/start-sdk/lib/backup/Backups.ts
```diff
@@ -255,9 +249,20 @@ export class Backups<M extends T.SDKManifest> implements InitScript {
...
+          'start',
+          '-D',
+          pgdata,
+          '-t',
+          pgCtlTimeout,
+          '-o',
+          pgStartOpts,
+        ],
+        { user: 'postgres' },
+        NO_TIMEOUT,
```
dr-bonez: can we make this part of the named opts instead?

helix-nine: Done — `64127f292`.

`timeoutMs` and `abort` both move into the options object, so this site is now:

```ts
await sub.execFail(
  ['pg_ctl', 'start', '-D', pgdata, '-t', pgCtlTimeout, '-o', pgStartOpts],
  { user: 'postgres', timeoutMs: null },
)
```

That also retires the `NO_TIMEOUT` constant I'd added — it only existed to make the bare positional legible, and `timeoutMs: null` says it directly.

I moved `abort` along with it because dropping only `timeoutMs` would have left `abort` sliding into third position, where a caller passing the old third argument would hit a type mismatch rather than a clear error. Passing either positionally is now a compile error, so any straggler says so at build time. In-repo the only positional callers were `DockerProcedureContainer`'s two `subcontainer.exec` calls, updated here; its own wrapper signatures are unchanged.

One thing that fell out while wiring it up: `exec` forwards the remainder of its options straight to `cp.spawn`, so the two fields have to come back off. The existing code does that by `delete`-ing `user`/`cwd` from the caller's own object — with `timeoutMs` on it, a reused options object would have silently reverted to the 30 s default on its second use, which is exactly the failure this PR is about. It's a destructure now, so the deletes hit a copy.

Verified: `make check`, `make test` (87), `make check-fmt`, `make bundle`, root `prettier --check`, and container-runtime's `tsc --noEmit` against the rebuilt `dist/`.

Flagging one judgement call for you: this is a breaking signature change on a public SDK API landing in a patch release. It's undocumented today — the third argument was never mentioned in the packaging guide, which is half of why #3636 happened — so I'd expect approximately zero external callers, and any that exist fail loudly at compile. Say the word if you'd rather I keep the positional form working as a deprecated fallback instead.

dr-bonez: ms is assumed in js. just `timeout`

helix-nine: Renamed — `445131880`.

```ts
await sub.execFail(['cp', tmpDumpFile, dumpFile], { user: 'root', timeout: null })
```

You're consistent with the surrounding surface here — `readyTimeout`, `sigtermTimeout`, `gracePeriod` and the six existing `timeout` fields all go unsuffixed already, so `timeoutMs` was the odd one out.

I applied the same rule to the result field this PR adds: `timedOutAfterMs` → `timedOutAfter`. Shout if you only meant the option and want that one left long. The unit now lives in the option's doc comment, and the error message still spells it out for whoever has to read it (`cp timed out after 30000ms and was killed with SIGKILL:`).

Scope note: container-runtime's `System` adapters have carried `timeoutMs` parameters for a long time and I left them alone — that's a separate API layer and none of it is what this PR touches. Only the one boundary where `DockerProcedureContainer` hands its `timeoutMs` to `subcontainer.exec` is mapped across.

Verified again after the rename: `make check`, `make test` (87), `make check-fmt`, `make bundle`, root `prettier --check`, container-runtime `tsc --noEmit` against the rebuilt `dist/`, and no `timeoutMs`/`timedOutAfterMs` left anywhere under `projects/start-sdk`.

dr-bonez (approved):

PR #3652, opened 2026-08-10
chore(ci): continuously deploy master to alpha, add integration/ and live-docs branches

dr-bonez:
Implements the new branch strategy: `master` is always shippable and continuously deploys to alpha, big work stages on `integration/<topic>`, and `docs.start9.com` is served from a `live-docs` branch.

## Branches

- `next/*` → `integration/*` across every workflow trigger. `next/*` is retired.
- `cancel-in-progress` now fires only on pull requests — a master push publishes, so it must not be aborted halfway by the next push.
- A PR from `integration/*` into `master` bypasses the path-gated image filter and builds the **full** platform matrix, so the batch is flashable before it ships. Same gate on start-wrt's image job.

## Alpha channels

Every push to master publishes:

| Product | Destination |
| --- | --- |
| StartOS | images to S3, indexed into the alpha registry |
| start-cli, StartTunnel, start-registry | `.deb`s into a new `alpha` apt suite, published by a single writer (`apt-publish-alpha.yml`) |
| StartWRT | alpha registry — inert until the `STARTWRT_ALPHA_REGISTRY` repo variable is set, since its image build is multi-hour and there is no alpha registry yet |
| Start SDK | nothing — an npm version can never be republished, so burning one per commit would make `pre-check` refuse the real release |

`debian/publish.sh` needed fixing before this was safe. It synced the **whole bucket** down and back up with `--delete-removed`, evicted old versions from a **shared** pool, and ran `dpkg-scanpackages` over `pool/` entirely — so publishing an alpha deb would have dropped the stable `.deb` out of the pool and broken `stable` for real users. Now each suite owns its own pool, the scan and the sync are both suite-scoped, and the upload is `--delete-after`. `stable` keeps its historical pool path, so nothing moves under the live repo.

The `alpha` suite is signed with a **CI-only key** and carries its own keyring (`apt/start9-alpha.list`). Trust is per-suite — the production release key never goes near CI.

## live-docs

- `docs-deploy.yml` now watches `live-docs`.
- `docs-sync-on-tag.yml` — on `<project>/v<version>`, copies that tag's `docs/` tree plus the shared `projects/start-docs/` infra onto `live-docs`. `versions.conf` rides along, so set it in master **before** tagging.
- `docs-backport.yml` — a push to `live-docs` lands the same change on `master` unattended. It was already reviewed on the way into live-docs, so it does not ask for a second approval. Only a genuine conflict falls back to a branch + PR.

Both use a GitHub App that is a ruleset bypass actor. Granting that bypass to `github-actions[bot]` instead would have been a repo-wide escalation — any workflow on any branch could then push to master, unreviewed. The App's private key is an **environment** secret whose deployment policy admits only the ref its workflow triggers on, so a run on any other ref is refused the secret and cannot mint a bypassing token.

The backport commit is created via the GraphQL `createCommitOnBranch` mutation rather than `git commit`, so GitHub signs it — master's `required_signatures` is satisfied honestly instead of bypassed. `expectedHeadOid` makes the write an atomic compare-and-swap, so a race with a concurrent master push loses cleanly and retries.

Note that App tokens are **not** exempt from the workflow-recursion rule, so the sync's push raises a real push event. `docs-deploy` fires on its own (the explicit dispatch is gone, it would double-deploy), and the `[skip-backport]` marker is now the only thing preventing a loop.

## Deploying a CI build

```sh
make start-os-update-from-gha REMOTE=start9@<ip>              # latest master build
make start-os-update-from-gha REMOTE=start9@<ip> RUN_ID=<id>  # a specific run
```

Reads the platform off the device, downloads that run's squashfs, and runs the same checksummed upgrade as `start-os-update-squashfs`.

## Before merging

1. **Bootstrap `live-docs` with the new workflows — do this first, or the docs automation is inert.** Actions reads push-triggered workflow definitions from the *pushed branch's* tree, and `live-docs` (`5a64e43c1`) still carries the old set: `docs-deploy.yml` listening to `master`, and no `docs-backport.yml` or `docs-sync-on-tag.yml` at all. So until it is bootstrapped, a sync commits to live-docs without deploying, and a live-docs PR does not backport.

   Open a PR into `live-docs` carrying **only** `.github/workflows/docs-*.yml` from merged master. Do **not** fast-forward live-docs to master — that would publish master's unreleased books. (Tag-triggered runs read from the tag's tree, so `docs-sync-on-tag.yml` itself works from day one; it is the two branch-triggered halves that need this.)
2. Environments `docs-backport` (ref policy: `live-docs`) and `docs-publish` (ref policy: `*/v*` tags), each holding `DOCS_BOT_APP_ID` + `DOCS_BOT_PRIVATE_KEY`.
3. The docs App added as a bypass actor on **Critical branchs** and **Docs Branch**.
4. Secrets `APT_GPG_PRIVATE_KEY` / `APT_GPG_KEY_ID`, and `apt/start9-alpha.gpg` committed for testers.

## Known behavior

Alpha re-publishes the **same** version number with a fresh squashfs on each master push, so both deploy jobs run `registry os version remove` before re-registering. `registry os asset add` upserts the platform slot but rejects a write whose blake3 differs from the commitment already there (`registry/os/asset/add.rs` → `commitment-mismatch`), so without the removal the second master push onto a version would fail. Removing the version drops *every* platform indexed under it, so the step is gated to full-matrix deploys — which every master push is. A dispatch naming a single platform skips it rather than silently wiping the other eight; if that version already holds a build, the index step fails on the commitment mismatch.

Because the version number is unchanged, an alpha box comparing versions will not see "update available" between version bumps. `start-cli update` and `update-from-gha` both work regardless.

Pushes *to* an `integration/*` branch build the full 9-platform matrix (inherited from how `next/*` behaved). Useful — every merge into the branch yields a flashable image — but it is real CI time on a branch that may see frequent pushes. Easy to path-gate if it proves too much.


debian/publish.sh
```diff
@@ -28,6 +29,17 @@ SUITE="${SUITE:-stable}"
...
 
+# Every suite owns its own pool, so publishing to one can never evict another's
+# .deb (the eviction below and the Packages scan are both pool-scoped). `stable`
+# keeps the historical pool/<component>/ path because the published repo is
+# already laid out that way — moving it would break `stable` for the window
+# between the delete and the re-upload.
+if [ "$SUITE" = stable ]; then
+    POOL_ROOT="pool"
+else
+    POOL_ROOT="pool/${SUITE}"
```
helix-nine: Blocking: `pool/alpha` is still inside stable's `pool` root. A later stable publish syncs `pool/` and scans it at line 126, so `dists/stable/.../Packages` gains entries like `Filename: pool/alpha/...`. I reproduced this with two dummy debs: scanning `pool/` emitted both the stable package and the nested alpha package. Either make non-stable pools siblings of `pool/` (for example `alpha-pool/`) or make the stable scan/upload scope exclude suite subdirectories (scanning `pool/$COMPONENT/` is the minimal fix).

debian/publish.sh
```diff
@@ -60,9 +72,16 @@ else
...
-s3 sync --no-mime-magic "s3://${BUCKET}/" "$REPO_DIR/" 2>/dev/null || true
+# Sync down only what this suite owns — its pool and its dists/ directory.
+# Everything below (the eviction, the Packages scan, the --delete-removed upload)
+# then operates on a tree that physically cannot contain another suite's files,
+# so publishing `alpha` from CI can neither index nor delete anything in
+# `stable`. Trust is already separated by the per-suite Release signature; this
+# separates write access too.
+echo "Syncing the ${SUITE} suite from s3://${BUCKET}/ ..."
+mkdir -p "$REPO_DIR/$POOL_ROOT" "$REPO_DIR/dists/${SUITE}"
+s3 sync --no-mime-magic "s3://${BUCKET}/${POOL_ROOT}/" "$REPO_DIR/${POOL_ROOT}/" 2>/dev/null || true
```
helix-nine: This `|| true` is dangerous in combination with the later `--delete-removed` uploads. If the download sync fails or is partial, the script proceeds with an incomplete mirror and can prune every object missing from that mirror. This predates the PR, but this script is becoming the continuous publisher; the download needs to fail loudly except for an explicitly detected first-publish/empty-prefix case.

.github/workflows/start-cli.yaml
```diff
@@ -184,3 +186,31 @@ jobs:
           name: start-cli_${{ matrix.arch }}.deb
           path: results/start-cli-*_${{ matrix.arch }}.deb
           retention-days: 14
+
+  # master is only ever merged to once a change is tested and shippable, so every
+  # push publishes into the `alpha` apt suite for testers. The `stable` suite is
+  # untouched — each suite owns its own pool (debian/publish.sh) — and real
+  # releases still go out only through scripts/manage-release.sh.
+  publish-alpha:
```
helix-nine: The alpha apt publish needs a cross-workflow concurrency group. The workflow-level groups include `github.workflow`, so master pushes to start-cli, start-tunnel, and start-registry can publish concurrently into the same `pool/alpha/` + `dists/alpha/`. Each job syncs down, adds only its own debs, then uploads with `--delete-removed`; the last writer can silently remove another product's just-published debs. A fixed job-level group such as `apt-publish-alpha` with `cancel-in-progress: false` in all three workflows would serialize the writers.

.github/workflows/startos-iso.yaml
```diff
@@ -549,6 +575,25 @@ jobs:
...
+      # already holds a build fails. Alpha deliberately rolls the same version
+      # number on every master push, so the version entry has to be dropped
+      # first. Beta never republishes in place and is left alone.
+      #
+      # Removing the version drops *every* platform indexed under it, so this is
+      # gated to full-matrix deploys — which every master push is. A dispatch
+      # naming a single platform deliberately skips it rather than silently
+      # wiping the other eight; if that version already holds a build, the index
+      # step below fails on the commitment mismatch, which is the honest outcome.
+      - name: Drop the previous alpha build of this version
```
helix-nine, line 642: This protects only the alpha registry index; the bytes have already been overwritten in the shared release namespace at line 574. Alpha uploads use `s3://startos-images/v$VERSION/...`, the same bucket and prefix as `manage-release.sh` uses for beta/production. After a beta/prod build of that version has been indexed, the next master push uploads different bytes over the same URLs, leaving the beta/prod registry commitments pointing at content with a different hash. Please give alpha images a separate bucket or prefix (and CDN URL), and mirror that for StartWRT's `startwrt-images/v$VERSION/` path.

.github/workflows/docs-deploy.yml
```diff
@@ -1,14 +1,20 @@
...
-    branches: ['master']
+    branches: ['live-docs']
     paths:
       - 'projects/start-docs/**'
       - 'projects/start-os/docs/**'
       - 'projects/start-tunnel/docs/**'
       - 'projects/start-sdk/docs/**'
       - 'projects/start-wrt/docs/**'
+  # For redeploying the current live-docs without a new commit.
   workflow_dispatch:
```
helix-nine, line 21: A manual dispatch is not pinned to `live-docs`: GitHub's Run workflow UI defaults to the default branch, and the checkout below has no `ref:`. A default dispatch would build and deploy master's unreleased books and `versions.conf` to docs.start9.com. Add a job guard for `github.ref == 'refs/heads/live-docs'` or check out `ref: live-docs` explicitly.

.github/workflows/start-wrt.yaml
```diff
@@ -103,14 +105,23 @@ jobs:
...
+  # variable — unset, master pushes stay on the fast `compile` path only, since
+  # there is no alpha StartWRT registry to publish to yet.
   image:
     name: Build OpenWrt image
-    if: ${{ github.event_name == 'workflow_dispatch' && (github.event.inputs.build-image == 'true' || github.event.inputs.deploy == 'release') }}
+    if: >-
+      ${{
+        (github.event_name == 'workflow_dispatch' && (github.event.inputs.build-image == 'true' || github.event.inputs.deploy == 'release'))
+        || (github.event_name == 'push' && github.ref == 'refs/heads/master' && vars.STARTWRT_ALPHA_REGISTRY != '')
+        || (github.event_name == 'pull_request' && github.base_ref == 'master' && startsWith(github.head_ref, 'integration/'))
```
helix-nine: `github.head_ref` is just the source branch name, including for fork PRs. A fork branch named `integration/anything` targeting master can therefore trigger the multi-hour image build. The same pattern exists in `startos-iso.yaml`'s `changes` step. If this gate is meant for maintainer integration branches, also require `github.event.pull_request.head.repo.full_name == github.repository` (and consider the draft gate used by the other expensive jobs).

scripts/update-from-gha.sh
```diff
@@ -0,0 +1,185 @@
...
+remote_sh 'sudo /usr/lib/startos/scripts/prune-boot'
+
+echo "Uploading the image..."
+# Streamed through tee rather than scp'd: /media/startos/images is root-owned,
+# and this needs no writable staging area on a server that was just pruned for
+# exactly this file's worth of space.
+remote_sh 'sudo tee /media/startos/images/next.rootfs > /dev/null' < "$SQUASHFS"
+
+echo "Upgrading..."
+remote_sh "sudo CHECKSUM=$SQFS_SUM /usr/lib/startos/scripts/upgrade /media/startos/images/next.rootfs"
```
helix-nine: This does not actually run a checksummed upgrade: `projects/start-os/build/lib/scripts/upgrade` only compares when `$2` is non-empty, and every caller passes only the squashfs path. A corrupt-but-mountable CI download would be installed despite the script/docs advertising checksum verification. Pass an explicit second argument here (and preferably at the other callers), or change `upgrade` to validate whenever `CHECKSUM` is set.

.github/workflows/docs-backport.yml
```diff
@@ -0,0 +1,186 @@
...
+# Serialized: two live-docs merges in quick succession would otherwise race each
+# other onto master.
+concurrency:
+  group: docs-backport
+  cancel-in-progress: false
+
+jobs:
+  backport:
+    name: Backport to master
+    if: ${{ !contains(github.event.head_commit.message, '[skip-backport]') }}
```
helix-nine: The loop-prevention marker is controlled by anyone landing the head commit on `live-docs`, so a normal docs commit containing `[skip-backport]` (even accidentally) silently never reaches master. Scope the skip to the docs bot's own sync commits, for example by also checking the head commit author, and emit a warning if a non-bot commit is skipped.

.github/workflows/docs-sync-on-tag.yml
```diff
@@ -0,0 +1,84 @@
...
+            echo "::notice::$PROJECT ships no docs/ book — nothing to publish."
+            exit 0
+          fi
+
+          git config user.name "start9-docs-bot[bot]"
+          git config user.email "start9-docs-bot[bot]@users.noreply.github.com"
+
+          # Delete then restore from the tag, so files the release removed are
+          # actually gone rather than left behind by a copy-over.
+          rm -rf "projects/$PROJECT/docs" projects/start-docs
```
helix-nine: Any matching tag replaces the entire shared `projects/start-docs` tree, including `versions.conf`, build scripts, theme, and deploy infrastructure. A backfilled or re-cut tag from an old commit would roll the whole published site back to that old infrastructure, and `[skip-backport]` keeps master from reconciling it. Add an ancestry guard (for example requiring the tagged commit to be an ancestor of current master) or separate the per-book sync from the shared infrastructure update.

helix-nine (changes requested):
The branch strategy and most of the workflow mechanics look sound, but I think this needs changes before merge. The new continuous-publish paths do not yet preserve the separation the PR is aiming for: alpha debs land under stable's scan root, the three apt publishers can race and delete each other's packages, and alpha OS/WRT images overwrite the release CDN namespace. I also found a docs dispatch path that can publish master's unreleased docs, a fork-triggerable expensive image path, and an advertised checksum path that never checks the checksum.

Validation run locally: `git diff --check`, `bash -n` on both changed shell scripts, `shellcheck` (only style/info diagnostics), `make -n start-os-update-from-gha` for the `RUN_ID` and `BRANCH` forms, and a dummy-deb reproduction of the stable/alpha Packages-index contamination below. GitHub CI is still running at review time; the completed jobs are green.

debian/publish.sh
dr-bonez: Fixed, and thanks — reproduced it exactly as you describe. Suite pool roots are now siblings rather than nested: `stable` keeps `pool/`, everything else gets `pool-<suite>/`. I took that over scoping the scan to `pool/$COMPONENT/` because it makes the isolation structural — the scan root, both sync directions, and `--delete-removed` all operate on disjoint prefixes, so there is no path-nesting subtlety left to get wrong later.

debian/publish.sh
dr-bonez: Fixed. The download is now a `sync_down` helper that lists the prefix first: an empty listing is the explicit first-publish case, and anything else must succeed or the script aborts before it can reach a `--delete-removed` upload. You are right that this predates the PR but is only now load-bearing.

.github/workflows/start-cli.yaml
dr-bonez: Fixed. All three `publish-alpha` jobs now share a fixed job-level `concurrency: {group: apt-publish-alpha, cancel-in-progress: false}`, with a comment saying explicitly not to scope it by `github.workflow`, since that is exactly what makes the workflow-level group insufficient here.

.github/workflows/startos-iso.yaml
dr-bonez: Half right, and I have made the change — but the stated mechanism does not hold, so recording why for the archive.

Basenames carry the commit hash and the environment suffix. A master push produces `startos-0.4.0.2-6a1f75d~dev_x86_64.squashfs` (I ran `build/env/basename.sh` with the push-path `ENVIRONMENT=dev` to confirm); a release build has `ENVIRONMENT` empty and produces `startos-0.4.0.2_x86_64.squashfs`. Two commits never share a URL, so alpha cannot overwrite bytes a beta/prod commitment points at.

The namespace concern is still fair, so alpha now uploads under an `alpha/` key prefix in both `startos-iso.yaml` and `start-wrt.yaml`, which keeps per-commit builds out of the signed release namespace and makes them separately expirable.

Your comment did surface something else: master pushes build with `ENVIRONMENT=dev` (password SSH before setup, no frontend compression), because the `ENVIRONMENT` expression defaults to `dev` when no dispatch input is present. That is pre-existing, but this PR is the first thing to *ship* those builds anywhere. Raised with the maintainer separately — it is a deliberate decision, not a mechanical fix.

.github/workflows/docs-deploy.yml
dr-bonez: Fixed, and this was the most dangerous one — a default dispatch would have published master unreleased. The checkout now pins `ref: live-docs` explicitly, so the ref the run was triggered from cannot influence what gets deployed. I pinned the checkout rather than guarding on `github.ref` so that a dispatch from any ref still does the right thing instead of failing.

.github/workflows/start-wrt.yaml
dr-bonez: Fixed in both places. The gate now also requires `github.event.pull_request.head.repo.full_name == github.repository`, and start-wrt`s image job picks up the `draft != true` gate the other expensive jobs carry.

scripts/update-from-gha.sh
dr-bonez: Confirmed and fixed at the root. `upgrade` line 21 was `if [ -n "$2" ] && [ "$hash" != "$CHECKSUM" ]`, and no caller has ever passed `$2` — so `start-os-update-squashfs` and `wormhole-squashfs` never verified either.

`upgrade` now keys off `CHECKSUM` and reports both values on mismatch. Callers also pass the checksum positionally, because a server running an older image still has the old script, where `CHECKSUM` alone is silently ignored — belt and braces until fleets turn over. Added a `### Fixed` entry under the unreleased `0.4.0.2` in the StartOS changelog, since updates silently skipping verification is user-visible.

.github/workflows/docs-backport.yml
dr-bonez: Fixed. The skip is now conditional on the head commit author being `start9-docs-bot[bot]` as well as the marker, so an ordinary docs commit containing the string is backported normally. A non-bot commit carrying the marker also emits a warning saying the marker is reserved for the tag sync.

.github/workflows/docs-sync-on-tag.yml
dr-bonez: Fixed, though not with the ancestry check as written — a backfilled tag from an old commit *is* an ancestor of master, so that guard alone would not have caught the rollback you describe.

Two guards instead. The tagged commit must be an ancestor of master, which rejects tags cut off unmerged branches (AGENTS.md notes start-sdk has shipped that way). Separately, each sync records `Source-Commit: <sha>` in its commit message, and a tag whose commit is an ancestor of the currently published source syncs only its own book, leaving `projects/start-docs` alone with a warning. Ancestry rather than dates, since tag order is not chronology.

.github/workflows/docs-sync-on-tag.yml
```diff
@@ -0,0 +1,118 @@
...
+          fi
+
+          # [skip-backport] is load-bearing, not a nicety: this push DOES raise a
+          # push event (App tokens are exempt from the GITHUB_TOKEN no-recursion
+          # rule), so docs-backport.yml will see it. The content came from master
+          # already, so it must not be bounced back. Source-Commit is what the
+          # next run's rollback guard reads.
+          git commit -m "docs: publish $PROJECT docs from $TAG [skip-backport]" \
+                     -m "Source-Commit: $SRC"
+          git push origin HEAD:live-docs
```
helix-nine, line 129: Blocking rollout issue: for a push event, Actions reads the workflow definitions from the pushed branch's tree. Current `origin/live-docs` (`5a64e43c1`) still has the old `docs-deploy.yml` listening to `master`, and it has no `docs-sync-on-tag.yml` or `docs-backport.yml` at all. So after this PR merges, this push to `live-docs` will not trigger a deploy, and a human docs PR merged to `live-docs` will not trigger the backport. The rollout needs a one-time bootstrap PR to `live-docs` that adds only the new `.github/workflows/docs-*.yml` files from the merged master (not a fast-forward of master, which would publish unreleased docs). The PR body's current `git push origin master:live-docs` instruction is also stale now that the branch exists.

.github/workflows/docs-sync-on-tag.yml
```diff
@@ -0,0 +1,118 @@
...
+            exit 1
+          fi
+
+          # The shared projects/start-docs tree (theme, build.sh, versions.conf,
+          # deploy infra) is not versioned per product, so a backfilled or re-cut
+          # tag from an old commit would roll the whole site back to that commit's
+          # infrastructure. Each sync records the commit it published from; if
+          # this tag predates that, publish the book alone and leave the shared
+          # tree where it is. Ancestry, not dates — tag order is not chronology.
+          PREV=$(git log -1 --format=%B -- projects/start-docs | sed -n 's/^Source-Commit: *//p' | head -n1)
```
helix-nine: This reads the trailer from only the newest commit touching `projects/start-docs`. A hand PR to `live-docs` that fixes the shared docs infrastructure has no `Source-Commit:` trailer, so `PREV` becomes empty, `SYNC_INFRA` stays enabled, and the next old/backfilled tag that passes the ancestry check can roll that infra back — the case this guard is meant to prevent. Walk back to the newest trailer-bearing commit instead (for example, drop `-1` from the log and take the first parsed trailer), or otherwise preserve the watermark across human infra commits. Until the first sync adds a trailer, the guard is also necessarily inert.

.github/workflows/startos-iso.yaml
```diff
@@ -441,24 +462,40 @@ jobs:
     needs: [image]
     # Explicit status check so the skipped PR-only `changes` job can't poison
     # the implicit success() through the needs chain (actions/runner#2205).
+    # Two ways in: a push to master, which continuously deploys to alpha (master
+    # is only ever merged to when a change is already tested and shippable), or a
+    # dispatch naming a channel.
     # Allowlist the channels rather than excluding NONE: a `!= 'NONE'` gate also admits
     # an empty deploy input, which uploaded a branch build to the public bucket and then
     # died on the empty REGISTRY below. Keep in sync with the REGISTRY map.
     if: >-
```
helix-nine, line 488: The workflow-level concurrency group is scoped by `github.event_name`, so a manual `deploy=alpha` dispatch can run concurrently with a master-push alpha deploy. Both now operate on the same alpha registry version and `alpha/v$VERSION/` prefix, with remove → add → asset-add sequences that can interleave. Please give the deploy job a shared alpha deploy concurrency group (or another lock) as was done for the apt publishers.

.github/workflows/docs-backport.yml
```diff
@@ -0,0 +1,202 @@
...
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+          AFTER: ${{ github.event.after }}
+          BEFORE: ${{ github.event.before }}
+        run: |
+          set -euo pipefail
+          BRANCH="backport/live-docs-$(git rev-parse --short "$AFTER")"
+          git reset --hard origin/master
+          git checkout -b "$BRANCH"
+          git cherry-pick -x $(git rev-list --reverse "${BEFORE}..${AFTER}") || true
```
helix-nine: If a multi-commit range has conflicts in more than one commit, this stops at the first conflict; the single `--continue` resumes only until the next conflict, then `|| true` lets the branch be force-pushed without the remaining commits while the PR body says it backports the whole range. Loop the continue until the sequencer finishes, or cherry-pick one commit at a time and fail/skip explicitly, so the fallback PR cannot silently drop tail commits.

helix-nine (changes requested):
Most of the follow-up fixes check out. I re-ran the dummy-deb reproduction against the new sibling pools and stable no longer sees `pool-alpha`; `bash -n` and `git diff --check` are clean; the checksum gate now matches the production caller; and the fork/draft/S3-prefix changes are internally consistent.

I found one launch blocker in the live-docs rollout and a few residual edge cases below. The important one is that push-triggered workflows are read from the pushed branch, and the existing `live-docs` branch still has the old workflow set. Unless the workflow files are bootstrapped onto `live-docs` out of band, the new sync can commit there without deploying, and live-docs PRs will not backport at all.

Non-blocking residual: the fixed `apt-publish-alpha` concurrency group prevents corruption, but GitHub retains only one pending job per group; a third publisher can supersede a pending one even with `cancel-in-progress: false`, leaving that product's alpha package stale until its next push. CI is still running at the time of this review.

.github/workflows/startos-iso.yaml
dr-bonez: Correction to my reply above — the maintainer confirmed alpha is the registry beta and production are promoted *from*, which changes the answer in both directions.

`registry/os/promote.rs` copies each asset's `urls[0]` and signed commitment through verbatim, so the bytes a master push uploads are the bytes production eventually serves, from the same S3 key. That means:

1. **The `alpha/` prefix I added is reverted.** Prefixing alpha uploads would leave promoted production assets pointing inside an alpha namespace — wrong, and actively dangerous if those objects were ever lifecycle-expired. Alpha and release artifacts share the namespace by design.
2. **`ENVIRONMENT` is now empty on a push to master**, so CD builds are release-equivalent. PRs and `integration/*` branches still default to `dev`. Applied to startos-iso, start-cli, start-tunnel and start-registry.

So your instinct that alpha and release artifacts should not be casually mixed was right, but the correct resolution was the opposite of a separate namespace: make the alpha artifact the real one. Both in 3db0c5bdd.

helix-nine: Confirmed — I checked `registry/os/promote.rs` and the image basename generation. Promotion does carry `urls[0]` and the signed commitment through unchanged, and the CI basename includes the commit hash, so the shared `v$VERSION/` namespace is not an overwrite hazard once master builds are release-equivalent. The new `ENVIRONMENT` expression also resolves as intended: empty on a master push, `dev` for PRs and `integration/*`, and `NONE` still yields empty on dispatch. Reverting the prefix makes sense; this resolves my concern.

.github/workflows/docs-sync-on-tag.yml
dr-bonez: Confirmed and it is exactly as you describe. `origin/live-docs` is `5a64e43c1`, its `docs-deploy.yml` still has `branches: ['master']`, and neither `docs-backport.yml` nor `docs-sync-on-tag.yml` exists there.

This one cannot be fixed from inside the PR — the branch has to be bootstrapped out of band, and doing it from a workflow would mean granting the docs bot `Workflows: write`, which is exactly the permission worth withholding from an automation that writes to master.

I have rewritten the PR body's rollout step accordingly: open a PR into `live-docs` carrying only `.github/workflows/docs-*.yml` from merged master, explicitly **not** a fast-forward (the old instruction said `git push origin master:live-docs`, which would have published master's unreleased books — thanks for catching that too).

Worth noting for whoever does it: tag-triggered runs read from the tag's tree, so `docs-sync-on-tag.yml` works from day one. It is only the two branch-triggered halves that are inert until the bootstrap lands.

.github/workflows/docs-sync-on-tag.yml
dr-bonez: Fixed. Dropped the `-1` so the watermark walks back to the newest commit that actually carries a trailer, with a comment explaining that a trailer-less human infra commit at the tip is precisely what would otherwise blank it.

Agreed the guard is inert until the first sync writes a trailer; that is unavoidable for a watermark that lives in the history it protects, and the ancestry-of-master check still applies in the meantime.

.github/workflows/startos-iso.yaml
dr-bonez: Fixed, and applied to StartWRT's deploy job as well for the same reason. Both now carry a per-channel job-level group shared across event types (`startos-deploy-<channel>` / `startwrt-deploy-<channel>`) with `cancel-in-progress: false`.

One correction: the `alpha/v$VERSION/` prefix you mention was reverted in 3db0c5bdd — promotion copies `urls[0]` verbatim, so prefixing alpha uploads would have left promoted production assets pointing inside an alpha namespace. Alpha and release share the namespace deliberately now. The registry-version race you identified is real regardless, which is what the group fixes.

.github/workflows/docs-backport.yml
dr-bonez: Fixed. The fallback now cherry-picks one commit at a time with `--allow-empty --keep-redundant-commits`, resolving each conflict by keeping the markers, so a range that conflicts in several places carries all of it. Commits that cannot be applied at all are collected and named explicitly in the PR body under a warning callout, rather than being absent from a branch whose description claims the whole range.

Reproduced the original failure first: a two-commit range conflicting in two files carried only the first commit under the old code, and carries both (markers intact, nothing dropped) under the new.

.github/workflows/startos-iso.yaml
dr-bonez: Thanks for re-deriving it independently — agreed on all three points.

dr-bonez (commented):
All four actionable findings from the last round are fixed in 2eb8f8f44, plus the rollout blocker documented in the PR body.

The `live-docs` bootstrap is the one that cannot be closed from inside this PR. Confirmed the branch is at `5a64e43c1` with `docs-deploy.yml` still listening to `master` and neither new workflow present, so both branch-triggered halves are inert until someone opens a workflows-only PR into `live-docs`. Automating it would mean granting the docs bot `Workflows: write`, which is the last permission that automation should have given it can write to master — so this stays a deliberate human step, now spelled out in the PR body with the warning not to fast-forward.

On the non-blocking apt concurrency point: agreed, and I documented it in all three workflows rather than engineering around it. A superseded run never writes, so the suite stays internally consistent; the cost is that one product's alpha deb can lag until its next push to master, which seems an acceptable trade on a testing channel. If it turns out to bite in practice, the real fix is one publisher that collects all three products' artifacts, not a smarter lock.

Remaining setup before merge is unchanged and all on the maintainer side: the two Actions environments with their ref policies, the App added as bypass actor on both rulesets, the apt signing secrets, and `apt/start9-alpha.gpg`.

dr-bonez (commented):
Follow-up on the non-blocking apt concurrency point — the maintainer agreed a single publisher is the right model, so I have built it rather than documenting the limitation (d0bbcefe4).

The three per-product `publish-alpha` jobs are replaced by one `apt-publish-alpha.yml`, triggered by `workflow_run` on the completion of any deb product's master build. The important design detail is that it collects **each product's latest successful master build**, not the artifacts of the run that triggered it. That is what makes supersession harmless rather than merely rarer: whichever run survives the pending-slot contention republishes the current state of all three, so a dropped run costs nothing at all instead of leaving one product stale.

A product whose last successful build has aged past the 14-day artifact retention is skipped with a warning rather than failing the publish, so its existing package stays put instead of the whole suite going unwritten.

One new hand-maintained mirror this introduces: `workflow_run.workflows` matches on workflow **name**, not filename, so renaming a product workflow would silently stop alpha receiving that product. Recorded in the root AGENTS.md coupled-changes list.

.github/workflows/apt-publish-alpha.yml
```diff
@@ -0,0 +1,82 @@
...
+      - name: Collect the latest master debs from every product
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          mkdir -p debs
+          collected=0
+          for wf in start-cli.yaml start-tunnel.yaml start-registry.yaml; do
+            product="${wf%.yaml}"
+            run_id=$(gh run list -R "$GITHUB_REPOSITORY" -w "$wf" -b master \
```
helix-nine, line 72: Blocking security issue: `-b master` is only a head-branch filter. A fork PR opened from a branch named `master` satisfies it, its successful `pull_request` build uploads attacker-built `.deb` artifacts, and its completion also matches this workflow's `workflow_run.branches: [master]` trigger. Since this query takes the latest successful run without checking the event or source repository, the privileged publisher can collect and sign those fork-built debs into alpha. The same query also admits maintainer dispatches, which can be dev-flavored or partial-arch builds. Gate the job on `github.event.workflow_run.event == 'push'` and add `--event push` to this query (or equivalently verify the source repository as well), so only same-repo master pushes are collectible.

.github/workflows/docs-sync-on-tag.yml
```diff
@@ -0,0 +1,123 @@
...
+          # tag from an old commit would roll the whole site back to that commit's
+          # infrastructure. Each sync records the commit it published from; if
+          # this tag predates that, publish the book alone and leave the shared
+          # tree where it is. Ancestry, not dates — tag order is not chronology.
+          # Walk back to the newest commit that actually carries a trailer, not
+          # just the newest commit touching the tree: a hand PR to live-docs
+          # fixing the shared infrastructure has no trailer, and reading only the
+          # tip would blank the watermark and re-open the rollback it guards
+          # against. (Necessarily inert until the first sync writes one.)
+          PREV=$(git log --format=%B -- projects/start-docs | sed -n 's/^Source-Commit: *//p' | head -n1)
```
helix-nine: This fixes the watermark logic, but introduces a deterministic future `SIGPIPE` failure under `pipefail`. `head -n1` closes the pipe after the first trailer; once there are enough later `Source-Commit:` lines for `sed` to flush again, the pipeline exits 141 and the whole tag sync aborts. I reproduced this with 120 trailer-carrying commits (`status=141`). Keep the walk but avoid the early-closing pipe — for example capture the filtered lines and take the first in shell, or use a git-side filter that still limits the result.

.github/workflows/startos-iso.yaml
```diff
@@ -441,21 +479,38 @@ jobs:
...
+          && contains(fromJson('["alpha", "beta"]'), github.event.inputs.deploy)
+        )
+      )
     runs-on: ubuntu-latest
+    # The workflow-level group is scoped by event, so a dispatched alpha deploy
+    # could otherwise run alongside a master-push one and interleave their
+    # remove -> add -> asset-add sequences against the same registry version.
+    # This group is per channel and shared across event types.
+    concurrency:
+      group: startos-deploy-${{ github.event_name == 'push' && 'alpha' || github.event.inputs.deploy }}
```
helix-nine: The per-channel groups serialize alpha-vs-alpha, but alpha and beta still share the same S3 keys for the same commit. A beta dispatch of a commit already alpha-published now rebuilds for hours and then fails the overwrite guard; if it runs concurrently with the alpha deploy, both can pass the `ls` check and race the same key. If alpha→beta promotion is now the supported path, the beta dispatch should probably be removed or changed to promote/reuse the existing artifact; otherwise this needs a cross-channel lock plus a hash-verified reuse path. Same issue applies to StartWRT's release dispatch.

helix-nine (changes requested):
The single-publisher direction is right, but the new collection query loses the trust boundary that the removed per-product jobs had. That needs fixing before merge. I also reproduced a latent `pipefail` failure in the new docs watermark walk, and there is one remaining beta-dispatch/S3 interaction to resolve or document.

Validation: `git diff --check` and `bash -n` are clean; the watermark SIGPIPE reproduces deterministically with 120 trailer-carrying commits. The current start-cli CI failure is an SSL/network failure fetching `https://github.com/dr-bonez/cbor.git`, not an obvious PR-code failure, but it will need a rerun before merge.

.github/workflows/apt-publish-alpha.yml
dr-bonez: Confirmed and fixed — this was a real supply-chain hole and I should have caught it when I moved the collection off the triggering run.

Verified the premise against the live API: `head_branch` on a `pull_request` run is the PR's own head branch, and this repo has `allow_forking: true`, so a fork PR from a branch named `master` would have matched both the `workflow_run.branches` trigger and the `-b master` query. The publisher holds the apt signing key, so it would have signed fork-built debs into alpha.

Two gates now. The job requires `github.event.workflow_run.event == 'push'`, and the query passes `--event push` and re-asserts `select(.event == "push")` on the returned row so it fails closed. A push run can only originate from a ref in this repository, which also excludes the dispatch builds you noted — `gh run list --json` has no `headRepository` field, so the event filter is the check that carries this.

.github/workflows/docs-sync-on-tag.yml
dr-bonez: Confirmed and fixed. My first repro attempt did not trigger it and I nearly wrote back saying so — the fake shas were short enough that the matched output stayed inside sed's 4KB flush buffer. With realistic 40-char shas it reproduces at exactly your 120 (`rc=141`), and at every larger size I tried.

Replaced with a full capture sliced in the shell (`TRAILERS=$(... | sed ...)`, then `${TRAILERS%%$'\n'*}`), which has no early-closing consumer. Verified rc=0 at 120/1000/5000/20000 trailers, still returning the newest one. I checked the other `set -euo pipefail` steps in this PR for the same shape; this was the only one.

.github/workflows/startos-iso.yaml
dr-bonez: Half fixed, half a question for the maintainer.

Fixed: both deploy jobs now take a single `startos-deploy` / `startwrt-deploy` lock instead of one per channel. You are right that the channels are not independent — they write the same `v$VERSION/` keys for a commit — so per-channel groups did not close the race.

Not fixed, deliberately: the beta dispatch rebuilding for hours only to fail the overwrite guard. Making that path reuse the existing artifact is not free — builds are not reproducible, so "reuse" means indexing a commitment computed from bytes already on S3, which the deploy job does not have without downloading them back. And the alternative you float — dropping the beta dispatch in favour of promotion — is the right shape but a process decision, not a mechanical one: `manage-release.sh` already treats alpha -> beta as a promote, which would make the beta deploy option vestigial.

I have put that to the maintainer rather than removing a deploy path unilaterally. If it stays, the proportionate fix is a cheap preflight job that checks the keys before the image matrix runs, so the dispatch fails in seconds instead of hours.

helix-nine: Confirmed in `ba1511262`: the apt publisher is now limited to same-repo master pushes, the watermark rewrite survives the 150-trailer SIGPIPE repro (`rc=0`, newest trailer selected), and the single `startos-deploy`/`startwrt-deploy` locks close the cross-channel S3 race. The remaining beta-dispatch ergonomics are reasonable to settle with the maintainer.

Still open from the supplemental comment above: the GraphQL backport path needs to divert symlink/mode changes to the git fallback, and `projects/start-docs/ARCHITECTURE.md:95` still documents deployment from `master` instead of `live-docs`.

.github/workflows/docs-backport.yml
```diff
@@ -0,0 +1,269 @@
...
+
+            # createCommitOnBranch can only express regular files: additions
+            # carry base64 content and land as mode 100644. A symlink (every
+            # book's docs/theme is one, mode 120000) would be read through to
+            # its target — and where that target is a directory, base64 fails,
+            # the command substitution swallows it, and the mutation happily
+            # writes contents:"" *over the symlink*. Divert anything that is not
+            # a plain 100644-to-100644 change to the git-based fallback, which
+            # represents modes natively.
+            UNSUPPORTED=$(git diff --cached --raw "$BASE" | awk -F'\t' '{
```
helix-nine: One adjacent case still slips through: renames. For an `R100` diff, the additions query below (`--diff-filter=d`) emits only the new path, while the deletions query (`--diff-filter=D`) emits nothing, so the GraphQL commit adds the renamed page but leaves the old page behind. I reproduced that locally. Either pass `--no-renames` to all three diff commands (so a rename is represented as add + delete, which the API can express) or detect `R`/`C` here and divert to the git fallback alongside mode changes.

helix-nine (changes requested):
The symlink/mode diversion and architecture-doc update both check out. I found one adjacent rename case in the same GraphQL payload construction; it is small and should be fixed with `--no-renames` or another divert.

.github/workflows/docs-backport.yml
dr-bonez: Confirmed and fixed in 0e57e7559, and it was worse than the missing deletion.

Reproduced the row: `:100644 100644 8cd52a7 8cd52a7 R100<TAB>d/old.md<TAB>d/new.md`. So besides `--diff-filter=D` emitting nothing while `--diff-filter=d` emitted the new path — leaving the old page behind exactly as you say — that row also breaks the mode check I added in the previous commit: with `-F'\t'`, `$2` is the **old** path on a rename, so a rename that also changed mode would have diverted while naming the wrong file.

Went with `--no-renames` on all three diffs rather than diverting. Renaming a docs page is ordinary enough that pushing it to a manual PR would be a poor trade, and delete + add is precisely what the API expresses; it also guarantees one path per raw row, which is what makes the awk correct. Verified against a rename plus an unrelated edit: additions `["d/keep.md","d/renamed.md"]`, deletions `["d/old.md"]`, nothing diverted.

The flag looks removable at a glance, so the comment above it now says why it is load-bearing.

helix-nine (dismissed):
LGTM at 0e57e7559.

I re-validated the final backport payload behavior locally: a rename now becomes add + delete, and symlink/mode changes are diverted to the git fallback. The apt publisher's same-repo master-push gate, the docs watermark rewrite, the shared deploy locks, and the S3 overwrite guard all match the reviewed behavior. Static checks (`git diff --check`, `bash -n`) are clean.

Remaining gates are operational/CI rather than code: the workflows-only `live-docs` bootstrap, the documented environments/ruleset bypass/secrets setup, and the currently running checks.

scripts/manage-release.sh
```diff
@@ -324,6 +329,64 @@ changelog_section() {
...
+promote_alpha_debs() {
+    local arch darch idx entry hash filename want
+    want=$(tag_commit_sha)
+    echo "Promoting ${PROJECT} debs from the ${APT_ALPHA_SUITE} suite (commit ${want})..."
+    for arch in $DEB_ARCHES; do
+        darch=$(deb_arch "$arch")
+        idx="${APT_BASE_URL}/dists/${APT_ALPHA_SUITE}/${APT_COMPONENT}/binary-${darch}/Packages"
+        # Emitted at the stanza boundary, not on a particular field: the index
+        # does not guarantee an order, and Git-Hash follows Filename in practice.
+        entry=$(curl -fsSL "$idx" | awk -v pkg="$PROJECT" -v ver="$VERSION" '
```
helix-nine: Blocking: this verifies only a self-asserted field in the unsigned `Packages` file, then downloads the pool object without checking it. Because every build of a version reuses the same `<pkg>_<version>_<arch>.deb` pool key, an alpha publish between this index read and the `curl` below can serve a different build after the `Git-Hash` check has passed. An S3-only writer could also forge a stanza naming the public target commit and get those bytes stable-signed by the releaser. Please use the trust chain we already publish: download and verify alpha's `InRelease` against `apt/start9-alpha.gpg`, verify `Packages` against the Release hash, capture the stanza's `SHA256`, and verify the downloaded `.deb` before staging it.

scripts/manage-release.sh
```diff
@@ -324,6 +329,64 @@ changelog_section() {
...
+            >&2 echo "    master, or use 'pull-gha' with the run that built ${want}."
+            return 1
+        fi
+        echo "  ${arch}: ${filename}"
+        curl -fsSL "${APT_BASE_URL}/${filename}" -o "$(basename "$filename")"
+    done
+    # Same marker pull-gha writes, so cmd_tag still refuses to tag a different
+    # commit than the staged assets were built from — the check above only holds
+    # for the COMMIT in effect at promotion time.
+    echo "$want" > "$(gha_commit_file)"
```
helix-nine: The marker is written only after every arch succeeds. If promotion fails partway (hash mismatch or a download error), earlier downloads and any previous files remain in the release directory with no `.gha-commit`, and `cmd_tag` only enforces the commit check when that marker exists. A manual `tag`/`push` sequence can then ship a mixed or stale staging dir. Write the marker before the download loop, as `resolve_gha_run` does, and clear staged debs on failure (or require a clean release dir for this path).

scripts/manage-release.sh
```diff
@@ -324,6 +329,64 @@ changelog_section() {
...
+        entry=${entry%%$'\n'*}
+        if [ -z "$entry" ]; then
+            >&2 echo "  ✗ ${darch}: no ${PROJECT} ${VERSION} deb in the ${APT_ALPHA_SUITE} suite"
+            return 1
+        fi
+        hash=${entry%%$'\t'*}
+        filename=${entry#*$'\t'}
+        if [ "$hash" != "$want" ]; then
+            >&2 echo "  ✗ ${darch}: ${APT_ALPHA_SUITE} holds a build of ${hash:-an unrecorded commit}, not ${want}"
+            >&2 echo "    Alpha carries the newest master build. Cut the release from an up-to-date"
```
helix-nine: This remedy is often impossible with the path-filtered product workflows. If master advances on a docs-only or StartOS-only commit, the product workflow does not run, so alpha legitimately holds the latest product-affecting commit while no build of current master exists. Cutting from an up-to-date master cannot produce one until an allowlisted path changes. The error and the new AGENTS.md guidance should name the usable path (`COMMIT=<alpha hash>` where appropriate), or the workflows need to build every master commit.

helix-nine (changes requested):
The promotion model is good, and the `Git-Hash` control field works: I built a dummy deb and confirmed `dpkg-scanpackages` carries both `Git-Hash` and `SHA256` into the Packages stanza. `bash -n`, `git diff --check`, and `shellcheck -S warning` are clean.

I think this commit needs one more pass before merge. The promotion currently trusts an unsigned index and never verifies the downloaded payload, so an S3-side change or an alpha republish between the index read and the download can swap the bytes after the commit check passes. There is also a partial-staging hole around the commit marker, and the path-filtered workflows make the documented remedy misleading in a common case.

scripts/manage-release.sh
dr-bonez: Fixed in 1725f7a26. You are right, and the irony is that alpha is signed with a CI key precisely so its consumers can verify it — a releaser promoting into stable is one, and I read straight past that.

The chain is now anchored end to end: `InRelease` verified against `apt/start9-alpha.gpg`, `Packages` verified against the SHA256 the signed Release commits to, the stanza taken from that verified index, and the downloaded `.deb` verified against the SHA256 that index commits to. Nothing in the path trusts plain HTTPS.

Tested against a locally built, really-signed fixture suite (`file://` base URL), covering all three cases you raised:

| scenario | result |
| --- | --- |
| valid suite, matching commit | promotes, marker written |
| pool object mutated after signing | rejected at the payload hash |
| `Packages` stanza forged to name the target commit | rejected at the index hash — the attacker cannot re-sign `Release` |
| suite signed by a different key | rejected at `InRelease` |

That last row is the S3-only-writer case you described.

scripts/manage-release.sh
dr-bonez: Fixed in 1725f7a26, and I took the stronger of the two options you offered rather than only writing the marker earlier.

The marker is written and stale `*.deb` cleared before anything is fetched, and downloads now land in a temp dir that is moved into the release directory only after every arch has verified. So an interrupted promotion leaves a marker and an empty dir — `push` fails on "No .deb files" — instead of a mixed set that `cmd_tag` would wave through. Confirmed in the fixture runs above that a failure at any stage stages zero debs.

scripts/manage-release.sh
dr-bonez: You were right that the remedy was often impossible, and the maintainer confirmed the underlying rule: the tag has to match the hash of the built asset. So rather than printing a better error, 724ddde37 makes it the default behaviour.

`pull-alpha` now **adopts** alpha's commit and tags there, reporting how far HEAD has moved on and printing the `git checkout` for that tree. `alpha-commit <project>` prints the same hash for scripting. Two guards: adoption is confined to a commit already in the current branch's history, and an explicit `COMMIT=` is never overridden — that is someone asserting a different intent, so it fails loudly instead.

I also added a check you did not raise but which the same path-filtering makes reachable: every architecture must report the same `Git-Hash`, so a half-published suite cannot yield a release assembled from two different builds.

scripts/manage-release.sh
```diff
@@ -709,6 +942,29 @@ cmd_pull_gha() {
...
+cmd_alpha_commit() {
+    require_kind cli deb
+    alpha_build_commit
+}
+
+cmd_pull_alpha() {
+    require_kind cli deb
+    ensure_release_dir
+    if [ "$KIND" = cli ]; then
+        resolve_gha_run
```
helix-nine, line 1009: Blocking: for `start-cli`, this resolves and downloads the GHA binaries before `promote_alpha_debs` adopts alpha's commit. If HEAD has moved past start-cli's last path-filtered build and the operator supplies a run for HEAD, `resolve_gha_run` passes against HEAD; `promote_alpha_debs` then sets `COMMIT` to alpha's older commit and overwrites `.gha-commit` with that value. `cmd_tag` tags the older commit and the marker check passes, while the staged per-triple binaries are from HEAD. Conversely, supplying the run for alpha's commit fails before adoption is ever reached. Determine/adopt the alpha commit before `resolve_gha_run`, and validate/download the CLI binaries against that adopted commit, so both halves really are pinned as the comment says.

scripts/manage-release.sh
```diff
@@ -1180,8 +1436,13 @@ cmd_release() {
             cmd_sign
             ;;
         cli | deb)
+            # Promote the debs CI published to alpha rather than rebuilding
+            # trust from a CI run: what testers have been running is what ships.
+            # (start-cli's per-triple binaries still come from the run — they are
+            # published only as release assets, so there is no channel to promote
+            # them from. Both halves are pinned to the tagged commit.)
             cmd_pre_check
```
helix-nine, line 1494: Adoption currently happens after `cmd_pre_check`, and the working tree remains at HEAD. `VERSION`, the changelog validation, and `release_notes` are therefore all read from HEAD while `cmd_tag` may tag an older adopted commit. If the changelog body for the same version changed after alpha's build, the GitHub release ships notes that are not in the tagged tree. Either require/check out the adopted commit before running the release pipeline, or re-derive and validate the release metadata from the adopted commit after adoption.

scripts/manage-release.sh
```diff
@@ -324,6 +337,213 @@ changelog_section() {
...
+        /^SHA256:/   { s = $2 }
+        /^Filename:/ { f = $2 }
+        END { flush() }
+    ' "$1")
+    # First line taken in the shell rather than piped through `head -1`, which
+    # would close the pipe early and take SIGPIPE under `pipefail`.
+    echo "${entry%%$'\n'*}"
+}
+
+# The commit alpha's current build of this project was produced from.
```
helix-nine: `alpha-commit` is weaker than the promotion path it feeds: it checks only the first architecture and accepts an empty `Git-Hash`, so it can print a commit (or an empty string with exit 0) that `pull-alpha` then rejects for mixed/missing hashes. Reuse the same all-arch common-commit validation here, or have this helper call the shared logic, so the suggested `git checkout "$(...)"` cannot point somewhere unreleasable.

helix-nine (changes requested):
The signed `InRelease` → `Packages` → `.deb` chain in `1725f7a26` addresses the earlier S3 trust/race finding, and the origin/no-cache clarification in `66f53f3f` is sound. `bash -n`, `git diff --check`, and shellcheck are clean locally; CI is still running.

The new commit-adoption behavior introduces a start-cli ordering bug that can still mix builds, and the adopted tag's metadata/notes still come from HEAD's tree. Those need one more pass. Lower-priority follow-ups: the new direct `sha256sum` calls should reuse the macOS `shasum -a 256` fallback, and the new `\u2717` strings print literally under bash `echo`.

scripts/manage-release.sh
dr-bonez: Confirmed and fixed in 47c1239bb. You traced it exactly: the run validated against HEAD, adoption then moved `COMMIT` to alpha's older commit and rewrote `.gha-commit` to match, so `cmd_tag` tagged the older commit with HEAD's binaries staged beside it and every check passing. The comment claiming both halves were pinned was simply wrong.

`cmd_pull_alpha` now collects and adopts the alpha commit **before** `resolve_gha_run`, so the run is validated against the commit actually being tagged. The inverse case you noted — supplying the run for alpha's commit and failing before adoption — is gone with it.

scripts/manage-release.sh
dr-bonez: Fixed in the same commit, by removing the second implementation rather than teaching it the same rules. `alpha_collect` is now the one path: it verifies the signed Release, verifies every architecture's Packages against it, requires a non-empty `Git-Hash`, and requires all architectures to agree. `alpha-commit` and `pull-alpha` both call it, so the hash the former prints is by construction one the latter accepts.

scripts/manage-release.sh
dr-bonez: Fixed in 47c1239bb, taking the cheaper half of your either/or.

Re-deriving all release metadata from the adopted commit is the thorough fix but a large change to a script that already reads the working tree in several places. Instead the promotion compares the changelog between the adopted commit and HEAD: identical means the notes describe the tagged tree regardless, so the common case stays frictionless; different means it stops and prints the `git checkout <adopted>` + re-run, which is the path the maintainer asked to optimise for anyway.

That leaves a narrower residual than before — metadata other than the changelog could in principle differ — but the version is already pinned by the deb's own `Version`, and the changelog is what `release_notes` actually publishes.

debian/publish.sh
```diff
@@ -26,8 +27,27 @@ ENDPOINT="${S3_ENDPOINT:-https://nyc3.digitaloceanspaces.com}"
 GPG_KEY_ID="${GPG_KEY_ID:-5259ADFC2D63C217}"
 SUITE="${SUITE:-stable}"
 COMPONENT="${COMPONENT:-main}"
+# How long the signed Release stays valid. A suite that goes this long without a
+# publish expires and apt refuses it — which is the point: for a rolling channel
+# like `alpha`, silence is information.
+VALID_DAYS="${VALID_DAYS:-30}"
```
helix-nine: Blocking: this applies the 30-day `Valid-Until` to every suite, including the default `stable` suite used by `publish_debs`. Stable can legitimately go more than 30 days without a start-cli/start-tunnel/start-registry release; once it does, every user's `apt update` fails on an expired Release file. The replay rationale is specific to the rolling alpha channel, so please emit/enforce the field only for `alpha` (or choose a deliberately documented stable cadence/window), not unconditionally.

scripts/manage-release.sh
```diff
@@ -324,6 +337,246 @@ changelog_section() {
...
+    # Trimmed by substitution, not split on ": " — the timestamp's own colons
+    # would otherwise truncate it to the hour and shift the comparison.
+    valid_until=$(awk '/^Valid-Until:/ { sub(/^Valid-Until:[[:space:]]*/, ""); print; exit }' "$tmp/Release")
+    if [ -z "$valid_until" ]; then
+        >&2 echo "  ✗ the ${APT_ALPHA_SUITE} Release carries no Valid-Until, so its age cannot be bounded."
+        >&2 echo "    Suites published before that field was added need one republish (any master"
+        >&2 echo "    push to a deb product); until then use 'pull-gha'."
+        return 1
+    fi
+    expires=$(date -u -d "$valid_until" +%s 2> /dev/null || echo 0)
```
helix-nine: This is GNU-only (`date -d`), and the new hash checks below call `sha256sum` directly. On macOS — which this script explicitly supports with the bash/`shasum` notes — every alpha promotion now fails closed: BSD date has no `-d`, and `sha256sum` is absent. Please use a portable date parse and route the hash checks through the same `sha256sum`/`shasum -a 256` fallback used by `checksum_block`.

scripts/manage-release.sh
```diff
@@ -709,6 +975,44 @@ cmd_pull_gha() {
...
+    # while the debs (and the tag) ended up on alpha's older commit — binaries
+    # from one commit, debs and tag from another, with the marker agreeing.
+    alpha_collect || { rm -rf "${ALPHA_TMP:-}"; exit 1; }
+    resolve_alpha_commit "$ALPHA_COMMIT" || { rm -rf "$ALPHA_TMP"; exit 1; }
+    assert_metadata_matches_adopted || { rm -rf "$ALPHA_TMP"; exit 1; }
+
+    if [ "$KIND" = cli ]; then
+        # Now validated against the adopted commit, so both halves really are
+        # pinned to the commit being tagged.
+        resolve_gha_run
```
helix-nine, line 1009: There is still a partial-staging hole around the CLI half. `resolve_gha_run` writes the adopted marker before `pull_gha_cli_binaries`; if that download fails, the marker is fresh but debs from an earlier staging attempt are still present because `promote_alpha_debs` has not cleared them yet. If promotion later fails after the new binaries download, the reverse partial set remains. Clear both `*.deb` and stale `start-cli_*` payloads before staging either half, or stage both halves into a temp area and swap only after all succeed.

scripts/manage-release.sh
```diff
@@ -1234,6 +1543,17 @@ crate's — or package.json for start-sdk); the git tag / GitHub release is
 Subcommands:
   pre-check          Verify the changelog documents this version and that the
                      version is not already tagged/released.
+  alpha-commit       Print the commit alpha's current build of this project came
+                     from. `git checkout "$(... alpha-commit <project>)"` puts a
+                     tree on it. (cli/deb.)
+  pull-alpha         Stage a deb release by promoting the `alpha` apt suite's
```
helix-nine, line 1551: This usage text still describes the pre-adoption behavior: it says a mismatch stops because master moved past the tag. The implementation now adopts alpha's commit and tags there when `COMMIT` is unset. Please update the help text so operators do not expect the old failure mode.

helix-nine (changes requested):
The start-cli ordering fix, shared all-arch `alpha_collect`, changelog guard, and alpha freshness check all line up with the prior findings. I re-ran `bash -n`, `git diff --check`, and shellcheck; all are clean, and CI is still running.

The freshness bound needs one more iteration: it currently expires `stable` as well as `alpha`, and the new verification path is not portable to the macOS releaser environment the script otherwise supports. I also found one remaining partial-staging path in the start-cli flow and a stale usage string.

debian/publish.sh
dr-bonez: You are right, and the maintainer independently flagged it — reverted entirely in 44453c042 rather than scoped to alpha.

Scoping would have fixed the stable breakage you found, but the idea was wrong for alpha too: it expires a rolling channel on a timer unrelated to release cadence, to close a replay of a build that was legitimate when published. What actually guards that is showing the operator which commit is being tagged, where a replay shows up as an unexpectedly old hash. I have recorded that reasoning in AGENTS.md so it does not get reintroduced.

I should have argued that when you first proposed it instead of implementing it.

scripts/manage-release.sh
dr-bonez: Fixed in 44453c042. `date -d` is gone with the Valid-Until revert, and hashing now goes through a shared `sha256_of` that reuses the sha256sum/shasum fallback `checksum_block` has carried all along — which I should have noticed was already there.

Root cause on my side: I was testing by extracting these functions and eval-ing them on a GNU box, which could never have surfaced this. `APT_BASE_URL` now takes an override like the other endpoints, so the promotion runs end to end against a locally built signed suite through the real script. Doing that immediately caught a second thing the harness had hidden — my fixture only had amd64, while the real path walks all three arches.

scripts/manage-release.sh
dr-bonez: Fixed in 44453c042. Both halves are now cleared, and the marker written, before either is fetched — `clear_alpha_staging` removes `*.deb` and, for cli, `start-cli_*`. So a failure at any point leaves an empty staging dir and `push` fails on it, rather than half a release beside a marker vouching for both.

This is the same class of bug I had just fixed for the debs and did not carry across to the cli half.

scripts/manage-release.sh
dr-bonez: Fixed in 44453c042. The help now describes adoption: with COMMIT unset it tags the commit alpha built, says which, and prints the matching `git checkout`; setting COMMIT to assert a different one fails rather than overriding. Stale help after a behaviour change was my omission.

helix-nine (approved):
LGTM at 44453c042.

I verified the latest changes: `Valid-Until` is fully removed from apt publishing, the signed alpha verification path now uses the shared SHA-256 fallback, alpha commit adoption happens before the start-cli GHA run is resolved, both artifact halves are cleared before staging, and the `pull-alpha` help matches the adoption behavior. `bash -n`, `git diff --check`, and `shellcheck -S warning` are clean.

One non-blocking residual: if a start-cli binary download is interrupted and someone later runs the subcommands manually, `push` can upload partial CLI binaries before failing on the empty deb set. The normal `release` flow aborts before that, so I do not think it needs to hold this PR.

Remaining gates are CI and the documented operational/bootstrap steps.

PR #3658, opened 2026-08-11
feat(start-os): add net gateway set-secure/unset-secure

### LINKED ISSUE #3649: `NetworkInterfaceInfo.secure` has no writer, so a gateway can never be marked secure and `secure: null` plaintext ports stay bridge-only (db/model/public.rs:259)
### Environment

- StartOS `master` @ `6a1f75d6c`. Found by reading source, not observed on a running box — no version/arch/hardware applies.
- Area: networking.

### What happens

`NetworkInterfaceInfo::secure()` is `self.secure.unwrap_or_else(|| self.is_intrinsically_secure())`, and `is_intrinsically_secure()` matches only `Loopback | Bridge`. The `secure` override that would let it return `true` for any other interface is `Option<bool>`, persisted in the DB and TS-exported — but **nothing anywhere writes it**. It is constructed `None` and carried forward as `None` for the life of the server.

So `secure()` is effectively the constant `is_intrinsically_secure()`, and every branch keyed on an operator having marked a gateway secure is unreachable. Concretely, for a binding declared `secure: null` — which is what `protocol: 'http'` expands to, and what electrs, fulcrum and others spell out directly — the plaintext external port is published and forwarded on `lo` and `lxcbr0` only. No configuration exists by which an operator can make it reachable over a LAN interface or a WireGuard/StartTunnel gateway.

Nothing is broken for users today: bridge-only is the correct default, and it is what lets a package name its interface "Electrum (SSL)" honestly. The defect is that the model carries persisted, exported state that no code path can set, and three call sites branch on it as though it were live.

This is also undocumented — no page under `projects/start-sdk/docs/src/` or `projects/start-os/docs/src/` describes marking a gateway secure — so nothing external depends on the capability either way.

One of:

1. **Make it settable** — an RPC + `start-cli` subcommand + a UI toggle on the gateway, which is what the field's `Option<bool>` shape and the `unwrap_or_else` fallback both imply was intended. This is the more useful outcome: it is how an operator would expose a plaintext service over a trusted tunnel without the package having to declare `secure: { ssl: false }`.
2. **Remove it** — drop the field, make `secure()` just `is_intrinsically_secure()`, and collapse the dead branches at the three consumers below.

Either is fine. Persisted, TS-exported state that nothing can write is not.

No reproduction steps: this was found by reading the source at `6a1f75d6c`, and its effect is the *absence* of a state transition.

Definition and fallback:

- `shared-libs/crates/start-core/src/db/model/public.rs:259` — `pub secure: Option<bool>`
- `shared-libs/crates/start-core/src/db/model/public.rs:336` — `secure()` = `self.secure.unwrap_or_else(|| self.is_intrinsically_secure())`
- `shared-libs/crates/start-core/src/db/model/public.rs:351` — `is_intrinsically_secure()` matches `Some(NetworkInterfaceType::Loopback | NetworkInterfaceType::Bridge)`

Every construction site, none of which ever sets `Some(true)`:

- `shared-libs/crates/start-core/src/db/model/public.rs:84` — `secure: None` at init
- `shared-libs/crates/start-core/src/net/tunnel.rs:88` — `secure: None` for a tunnel gateway
- `shared-libs/crates/start-core/src/net/gateway.rs:2258-2290` — rescan carries `i.secure` forward from the previous entry; preserve-only

There is no `as_secure_mut()` in the tree, no RPC or CLI subcommand that sets it, no generic DB write path, and the web references it only in `mock-patch.ts` and `api.fixures.ts`.

Consumers whose non-default branch is therefore dead:

- `shared-libs/crates/start-core/src/net/host/mod.rs:148-151` — `opt.secure.map_or(gateway_secure, |_| opt.wants_plain_port())` gates whether a plaintext IP address is published for the binding
- `shared-libs/crates/start-core/src/net/host/mod.rs:217-225` and `:295-303` — the mDNS and private-domain equivalents
- `shared-libs/crates/start-core/src/net/forward.rs:705` — `if !reqs.secure && !info.secure() { continue }` skips installing the plaintext forward on every non-bridge gateway
- `shared-libs/crates/start-core/src/net/net_controller.rs:613-618` — the IPv6 mirror, which skips a LAN-only GUA unless some gateway reports `info.secure()`

### Logs & evidence

```
$ grep -rn "as_secure_mut\|as_secure()" --include=*.rs shared-libs projects
(no matches)

$ grep -rn "setSecure\|set-secure\|set_secure" --include=*.rs --include=*.ts shared-libs projects
(no matches)

$ grep -rn "\.secure\b\|secure:" --include=*.ts projects/start-os/web/ui/src/app | grep -v insecure
projects/start-os/web/ui/src/app/services/api/mock-patch.ts:132:  secure: null,
...  (mock-patch.ts and api.fixures.ts only)
```

Surfaced while tracing why a `secure: null` Electrum binding is TLS-only to every wallet (Start9-Community/electrs-startos#83). The answer turned out to rest on this field never being settable, which is load-bearing for that package's interface name — so if it does become settable, packages that named an interface "(SSL)" on the strength of it need revisiting.

helix-nine:
Closes #3649 by taking option 1 — give the field a writer.

```
start-cli net gateway set-secure <GATEWAY> [SECURE]
start-cli net gateway unset-secure <GATEWAY>
```

Modelled on `set-public`/`unset-public` (removed in #3117), which is the command the field's `Option<bool>` + `unwrap_or_else` shape was clearly cut from. `net gateway list` gains a `SECURE` column that distinguishes an operator override from an inferred value (`YES` vs `YES (auto)`), since otherwise the writer has no reader.

Three commits: the i18n orphans #3117 left behind, the feature, and a fix for a hole I found reviewing my own writer before opening this.

### The WAN invariant is untouched

Marking a gateway secure buys LAN reachability and nothing else. Both forward paths gate public exposure on the **binding's** security option, never the gateway's:

- `forward.rs:704-713` — `public = reqs.public_gateways.contains(gw_id) && reqs.secure`, where `reqs.secure` is `bind.options.secure.is_some()` (`net_controller.rs:585`). The gateway's `secure()` appears only on the next line, and only to *admit* a forward. A plaintext binding can never be `public`, so `src_filter` is `Some(subnet)` or the iteration `continue`s, and the upstream port-map request lives entirely inside `if public`.
- `net_controller.rs:604-625` (IPv6 GUA) — an unrestricted `src_filter` needs `a.public && secure_exposure`; a plaintext binding falls to the `else`, which ends in `Some(subnet)` or `continue`. The v6 pinhole is opened only `if src_filter.is_none()`.

So I added no guard rail there — the invariant is enforced a level up and this change can't reach it.

### What the review caught

**The guard failed open.** I'd written it as `secure == Some(false) && info.is_intrinsically_secure()`, and `is_intrinsically_secure()` reads `ip_info.device_type` — which the controller itself nulls on every boot seed (`gateway.rs:2962-2966`) and on any NM rescan that drops the device (`:1411-1416`). So `set-secure lxcbr0 false` inside a `lxc-net` restart window was accepted and persisted, and would then suppress the plaintext bridge address (`host/mod.rs:162`) and its DNATs — which is what every `secure: null` dependency binding resolves through. An unknown device type now fails closed with its own message.

### What an operator is actually turning on

The first draft of the docs undersold this and I've rewritten them. It's one switch for the whole server: every installed service's non-SSL addresses appear on that network at once, enabled immediately (private addresses are enabled-by-default — `binding.rs:74-93`), with no per-service confirmation. The docs now say that, and replace "the public internet is never secure" with the mechanism that actually holds — traffic is restricted to the gateway's own network segment, which on a VPS or a bridged modem is the provider's network.

It does **not** expose the StartOS UI. An earlier draft of these docs claimed it did; that was wrong, and @dr-bonez caught it. `WildcardListener::new(80)` (`startd.rs:162`) is unconditional and independent of any gateway setting, so the server UI has always answered plain HTTP on every interface — and in an insecure context the SPA serves the certificate wizard rather than a login form (`login.page.ts:23` branches on `isSecureContext()`). What this setting changes for the OS host is only whether its plaintext address is *listed* among the server's addresses.

### Verification

`cargo check -p start-core` clean (no new warnings) · `make start-core-test` 294 passed · `make manpages-check` and `make start-core-ts-bindings-check` both exit 0 · `make start-core-format-check` clean · prettier clean on the markdown · mdbook builds and the `#secure-gateways` anchor resolves · no raw i18n keys in the regenerated man pages.

Tests: `secure()`'s `Some(_)` branches (never reachable before this PR), the disconnected-gateway precondition the guard turns on, and `secure_gateways`. I mutation-checked the last one — an implementation ignoring its `candidates` argument passed the first version, so it now includes a secure gateway absent from the candidate list.

**Not run:** anything on a box. The propagation argument is source-level — writes go through the `ip_info` watch rather than the db, because `NetworkInterfaceController::sync` re-serializes the whole map on every gateway event, and the same `db.mutate` sweeps `update_addresses` over all hosts (`gateway.rs:2725-2726`), which is what carries the change to the address readers. Worth confirming on a VM before merge.

### Open questions

1. **Is `SECURE` worth keeping as an argument?** `Some(false)` is a no-op for every gateway that isn't refused — it equals the inferred default everywhere except `lo`/`lxcbr0`, where it's now rejected. Its only real use is pinning against a future change in the inference. Happy to drop it and make `set-secure` a bare flag, which would also delete both guards.
2. **No UI.** You asked for the RPC, so `System > Gateways` doesn't show or set this. The field is already in `GatewayInfo` and the TS bindings if you want a toggle.

Rebased onto current master. `a473c02c9` landed the same loopback-DNAT guard I had written for the OS host, so that half of my fix is gone — master's version is narrower (it guards only the `forwards.insert`, where mine also skipped the IPv6 GUA block) and I dropped mine in favour of it.

### Adjacent, deliberately not fixed here

- `host/mod.rs:322` inserts a private-domain address even when its gateway set is empty, unlike its mDNS sibling at `:243` — a pre-existing dead UI row, in the block this PR revives.
- `start-core/AGENTS.md:39` ends with "Compile-time validation catches missing keys." It doesn't — `bins/mod.rs:68` resolves them at runtime through `t!(s.to_string())` with a silent `{locale}.{key}` fallback, and nothing in CI checks coverage. The only detector is a raw key surfacing in a regenerated man page, which is why I grepped for it above.
- `check-port-v6` has a committed man page but no `cli-reference.md` entry.


shared-libs/crates/start-core/src/net/gateway.rs
```diff
@@ -3102,6 +3164,61 @@ impl NetworkInterfaceController {
...
+                }
+            }
+            std::mem::replace(&mut info.secure, secure) != secure
+        });
+        if let Some(e) = err {
+            return Err(e);
+        }
+        if changed {
+            sub.recv().await;
+        }
```
dr-bonez: use a TypedDbWatch + wait_for, which is the new recommended pattern

helix-nine: Done in 690aa6f19.

```rust
let mut watch = self
    .db
    .watch(
        "/public/serverInfo/network/gateways"
            .parse::<JsonPointer>()
            .with_kind(ErrorKind::Database)?
            .join_end(interface.as_str())
            .join_end("secure"),
    )
    .await
    .typed::<Option<bool>>();
// ...
watch.wait_for(|persisted| *persisted == secure).await?;
```

Matches what `net/tunnel.rs:107-121` already does for a gateway write, so the two paths now read the same.

It's also stricter than what it replaces, which I hadn't clocked: `sub.recv()` returned on the first revision touching that scope whether or not it carried this write, and it needed the `if changed` guard to avoid waiting forever when the value was already what was asked for. `wait_for` peeks first, so the already-satisfied case returns immediately and the guard is gone.

`cargo check` clean, 302 tests pass, fmt + `manpages-check` + `start-core-ts-bindings-check` all green.

dr-bonez (approved):

Issue #3601, opened 2026-07-28
start-cli: host resolution is eager, and .local is special-cased at three connect points


MattDHill:
## Bug

A `host` profile naming a `.local` box that is off the LAN breaks **every** `start-cli` command, including those that contact no StartOS host:

```
$ start-cli -H https://bogus.local --registry=https://alpha-registry-x.start9.com registry package get lnd full
Network Error: Failed to resolve hostname: bogus.local        exit 9

$ start-cli -H https://bogus.invalid --registry=https://alpha-registry-x.start9.com registry package get lnd full
{"categories":[],"best":{"0.21.1-beta:9": ...                  exit 0
```

`.invalid` is reserved by RFC 2606 to never resolve, and that one succeeds. The failure is not about reachability — nothing connects at init.

Hit while promoting packages between registries from a workspace whose `.startos/config.yaml` names a dev box that was powered off.

## Cause

`CliContext::init` calls `pin_mdns_host` on the host URL before the subcommand is dispatched (`context/cli.rs:81`). It resolves via `resolve_system` → `getent ahosts`; an `Err` aborts context construction, so no subcommand runs.

`pin_mdns_host` filters `host.ends_with(".local")` and returns `Ok(())` for everything else, so only `.local` names reach the fallible path.

The filter is load-bearing for the current approach: `pin_mdns_host` **rewrites the URL host to an IP**, which is only safe where the cert carries IP SANs (`net/ssl.rs:468`). Applied generally it would pin a CDN edge, break SNI and Host routing, and fail validation against a normal DNS-SAN cert.

## Fragmentation

Three connect paths, three treatments, two resolvers:

| Path | `.local` | Resolver |
|---|---|---|
| reqwest HTTP client | inherits the rewritten URL | `resolve_system` (`getent`) |
| tokio-tungstenite ws | inherits the same rewritten URL | rides on the rewrite |
| `net/socks.rs:49` `connect_target` | own match arm | `resolve_mdns` (`avahi-resolve-host-name`) |

`mdns.rs` notes `avahi-resolve-host-name` ships in `avahi-utils`, "which mDNS on Linux does not depend on and which most desktops don't have." `socks.rs` also carries a `.onion` arm the other two lack.

## Options

[...]

PR #3660, opened 2026-08-11
fix(sdk): fail the build when list-ingredients fails, instead of repacking stale

MattDHill:
`s9pk.mk` could print `✅ Build Complete!`, exit 0, and leave the previous s9pk in place — omitting the changes just made to the package, with nothing to indicate anything was wrong.

## Cause

```make
INGREDIENTS := $(shell start-cli s9pk list-ingredients 2>/dev/null)
$(BASE_NAME)_%.s9pk: $(INGREDIENTS) $(GIT_DEPS) | check-deps
```

`list-ingredients` produces the s9pk's **entire** source-dependency list, `./javascript/index.js` included. Nothing else in the file ties the package to its TypeScript.

`$(shell)` discards exit status. When the command fails it exits 9, its stderr is swallowed by `2>/dev/null`, and `INGREDIENTS` is simply empty — so:

- every source file detaches from the s9pk's prerequisites;
- `javascript/index.js` is orphaned from the graph entirely, so its own rule (which *does* correctly depend on `$(shell find startos -type f)`) is never visited;
- the target's only surviving prerequisites are `.git/HEAD` and `.git/index`. If the existing s9pk is newer than both, make declares it up to date;
- `arch/%` has no file behind it, so its recipe runs regardless and prints the success banner.

Reproduced with source at `09:36:01`, s9pk at `09:36:00`, `.git/index` at `09:28:42`:

```
$ make x86
Network Error: Failed to resolve hostname: demo.local
✅ Build Complete!
$ echo $?
0                       # s9pk mtime unchanged — not rebuilt
```

The window is the ordinary inner loop: edit a source file, rebuild, no git activity in between. Run `git add` first and `.git/index` is newer, so the target *is* out of date, reaches `pack`, and fails loudly (exit 2) — which is why this presents intermittently and reads as a network problem rather than a stale build.

## Fix

Append a sentinel on failure and `$(error)` on it. Deliberately not `$(.SHELLSTATUS)`: it is undefined on the GNU Make 3.81 that macOS still ships, where `ifneq ($(.SHELLSTATUS),0)` would fire the guard on **every** build.

## Verified

Against a real package (`tor-startos`) with the configured host unresolvable:

- **failure path** — errors with the message below, exit 2, no s9pk produced;
- **success path** — with `list-ingredients` returning normally, `INGREDIENTS` holds the clean list with no sentinel, the guard stays silent, and a touched source file correctly triggers an `ncc` rebuild (exit 0) — the very rebuild the bug was suppressing.

```
s9pk.mk:14: *** `start-cli s9pk list-ingredients` failed, so make cannot tell
which files the s9pk depends on and would silently repack the previous build.
Run it directly to see the error — a host in .startos/config.yaml that no longer
resolves is a common cause, since start-cli resolves it even for commands that
contact no server.  Stop.
```

## Relationship to #3601

#3601 is today's trigger: start-cli resolves the configured host even for commands that contact no server, so a workspace naming a `.local` box that is off the LAN fails every invocation. That remains a separate fix. This one is independent — it keeps *any* future failure of `list-ingredients` from silently producing a stale package, which is the part that can ship wrong bits.

## Also here

`projects/start-os/AGENTS.md` listed `projects/start-sdk/s9pk.mk` under "don't edit generated binding files", beside the ts-rs `osBindings`. Nothing generates it, and this repo is where it is meant to be edited — its `DO NOT EDIT` header addresses package authors reading the copy npm put in their `node_modules`. Followed as written, that line forbids the fix above.

### DIFF
--- a/projects/start-sdk/s9pk.mk
+++ b/projects/start-sdk/s9pk.mk
@@ -2,7 +2,17 @@
 # This file is imported by ./Makefile. Make edits there
 
 PACKAGE_ID := $(shell awk -F"'" '/id:/ {print $$2}' startos/manifest/index.ts)
-INGREDIENTS := $(shell start-cli s9pk list-ingredients 2>/dev/null)
+# INGREDIENTS is the whole source-dependency list, javascript/index.js included —
+# nothing else ties the s9pk to your TypeScript. $(shell) discards exit status,
+# so a failure here would leave it empty, detaching every source file from the
+# s9pk's prerequisites; make would then call an existing s9pk up to date and
+# print "Build Complete" over the *previous* build. The sentinel turns that into
+# an error. Not $(.SHELLSTATUS): it is undefined on the GNU Make 3.81 that
+# macOS still ships, where the guard would fire on every build instead.
+INGREDIENTS := $(shell start-cli s9pk list-ingredients 2>/dev/null || echo __LIST_INGREDIENTS_FAILED__)
+ifneq ($(filter __LIST_INGREDIENTS_FAILED__,$(INGREDIENTS)),)
+$(error `start-cli s9pk list-ingredients` failed, so make cannot tell which files the s9pk depends on and would silently repack the previous build. Run it directly to see the error — a host in .startos/config.yaml that no longer resolves is a common cause, since start-cli resolves it even for commands that contact no server.)
+endif
 # Resolve the actual git dir so this works inside git worktrees, where .git
 # is a file pointing at <main>/.git/worktrees/<name> rather than a directory.
 GIT_DIR := $(shell git rev-parse --git-dir 2>/dev/null)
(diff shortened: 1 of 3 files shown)


projects/start-sdk/s9pk.mk
```diff
@@ -2,7 +2,17 @@
...
+# INGREDIENTS is the whole source-dependency list, javascript/index.js included —
+# nothing else ties the s9pk to your TypeScript. $(shell) discards exit status,
+# so a failure here would leave it empty, detaching every source file from the
+# s9pk's prerequisites; make would then call an existing s9pk up to date and
+# print "Build Complete" over the *previous* build. The sentinel turns that into
+# an error. Not $(.SHELLSTATUS): it is undefined on the GNU Make 3.81 that
+# macOS still ships, where the guard would fire on every build instead.
+INGREDIENTS := $(shell start-cli s9pk list-ingredients 2>/dev/null || echo __LIST_INGREDIENTS_FAILED__)
+ifneq ($(filter __LIST_INGREDIENTS_FAILED__,$(INGREDIENTS)),)
+$(error `start-cli s9pk list-ingredients` failed, so make cannot tell which files the s9pk depends on and would silently repack the previous build. Run it directly to see the error — a host in .startos/config.yaml that no longer resolves is a common cause, since start-cli resolves it even for commands that contact no server.)
```
dr-bonez: mentioning the specific defect that caused us to find this is *not* appropriate for the error message

dr-bonez (approved):

PR #3693, opened 2026-08-12
fix(start-os): unbreak the NVIDIA image on GB10 — drop unusable GSP firmware, stage MOK enrollment

### LINKED ISSUE #3702: MOK enrollment is never staged on a fresh Secure Boot install — enroll_mok runs before the password it requires is written (setup.rs:185 vs :226)
### Environment

- StartOS 0.4.0.2 (`fix/nvidia-image-drop-nouveau-firmware` CI build), `aarch64-nvidia`
- NVIDIA DGX Spark (GB10), AMI BIOS `5.36_0ACUM018`, **Secure Boot enabled**
- Area: install/setup flow, Secure Boot module signing

### What happens

On a fresh install to a machine with Secure Boot enabled, the DKMS MOK is generated and the NVIDIA modules are signed, but the enrollment is **never staged**, so the user is never shown the MokManager prompt and the key is never trusted.

The result is silent: `mokutil --sb-state` reports `SecureBoot enabled`, `mokutil --test-key /var/lib/dkms/mok.pub` reports not enrolled, `lsmod | grep nvidia` is empty, and the kernel logs only `Kernel is locked down from EFI Secure Boot`. Every GPU service then fails in a way that points somewhere else entirely — `vllm` dies with `Failed to infer device type` and `No CUDA runtime is found`, and `nvidia-smi` reports "couldn't communicate with the NVIDIA driver".

Once the user sets their password, the MOK import should be staged, so the very next boot prompts for enrollment — which is what the setup wizard's MOK dialog leads the user to expect.

### Steps to reproduce

1. Install StartOS on a UEFI machine with Secure Boot enabled and an NVIDIA GPU (`aarch64-nvidia` or `x86_64-nvidia`).
2. Complete the setup wizard, setting a password.
3. Observe that no MokManager prompt appears on the next boot, `mokutil --test-key /var/lib/dkms/mok.pub` reports the key is not enrolled, and no `nvidia` module is loaded.

`enroll_mok` (`shared-libs/crates/start-core/src/util/mok.rs:73`) refuses to stage the import unless the `start9` user already has a password hash, via `start9_shadow_hash` (`mok.rs:52`):

```rust
let Some(hash) = start9_shadow_hash().await? else {
    tracing::info!("enroll_mok: start9 user password not set, skipping");
    return Ok(false);
};
```

It is called from exactly two places, and on a fresh install neither can satisfy that precondition:

1. **`shared-libs/crates/start-core/src/os_install/mod.rs:481`**, during install. `mok_pub` is correctly resolved inside the target overlay, but `start9_shadow_hash` reads the hardcoded host path `/etc/shadow` (`mok.rs:53`) — the **installer's** shadow file, not the target's. The target has no password at this point either, so this call cannot succeed regardless of the path bug.

2. **`shared-libs/crates/start-core/src/init.rs:177`**, early in `init()`, once per boot. On the boot where setup runs, `setup_init` (`shared-libs/crates/start-core/src/setup.rs:178`) calls `init(...)` on **line 185** and `write_shadow(&password)` on **line 226** — so the enrollment attempt always runs before the password it depends on is written, in the same function.

The import is therefore first staged on the *next* boot, and MokManager only prompts on the boot after that — two reboots after setup completes, with no indication to the user. On a server, which is not routinely rebooted, this means it effectively never happens.

Every skip path is `tracing::info!`, and the caller in `os_install` logs only on `Err`, so a skip is invisible in the UI and easy to miss in logs.

Suggested fix: stage the enrollment where the password becomes known — after `write_shadow` in `setup_init`, and in `auth.rs`'s password-change path (`auth.rs:574`) — rather than only at the start of `init()`. The `os_install` call is dead code in practice and should either read `<overlay>/etc/shadow` or be dropped. Surfacing a skip to the user would also help, since the failure otherwise presents as a broken GPU driver.

### Logs & evidence

```
root@sparky:~# lsmod | grep nvidia
root@sparky:~# dmesg | grep -iE 'nvrm|nvidia|module verification|lockdown|signature'
[    0.000000] Kernel is locked down from EFI Secure Boot; see man kernel_lockdown.7
[    0.106957] DMI: NVIDIA NVIDIA_DGX_Spark/P4242, BIOS 5.36_0ACUM018 08/06/2025
[    1.797089] Lockdown: swapper/0: hibernation is restricted
[    9.654081] Lockdown: systemd: /dev/mem,kmem,port is restricted
root@sparky:~# mokutil --sb-state
SecureBoot enabled
```

`mokutil --test-key /var/lib/dkms/mok.pub` reports the key is not enrolled, and no enrollment prompt was ever presented.

helix-nine:
Four fixes for the `aarch64-nvidia` image on an NVIDIA GB10 (DGX Spark). The first is confirmed on hardware; the second is the fix for #3702, found while chasing the fallout from the first.

## 1. Drop nouveau GSP firmware from the NVIDIA images — the boot fix

The `-nvidia` images blacklist nouveau and drive the GPU with NVIDIA's proprietary driver, yet ship graphics firmware they cannot use. `NON_FREE=1` adds `non-free-firmware` to `ARCHIVE_AREAS`; live-build's `--firmware-chroot` (default true, only `raspberrypi` disables it) then installs every `firmware-*` package, including `firmware-nvidia-graphics` (nouveau's chip-named GSP) and `firmware-nvidia-tesla-535-gsp` (the 535-series driver's). Neither is in `dpkg-deps/`. `initramfs-tools` then copies the firmware every *included* module declares, blacklist or not, and `nouveau.ko` is included under `MODULES=most` — so 159,225,325 B across 381 entries, 57% of the 266 MB uncompressed initramfs, was nouveau firmware.

Measured on this PR's CI artifact vs. the published 0.4.0.1 ISO:

| | 0.4.0.1 | this PR | delta |
|---|---|---|---|
| initrd | 163,574,628 B | 54,048,930 B | **−67.0%** |
| squashfs | 2,680,631,296 B | 2,398,883,840 B | −10.5% |
| ISO | 2,904,543,232 B | 2,513,270,784 B | −373 MB |

### Hardware result

| build | kernel | kernel + initrd | result |
|---|---|---|---|
| 0.4.0 | 7.0.13 | 208,739,061 B | boots |
| 0.4.0.1 | 7.1.3 | 210,494,244 B | hangs |
| 0.4.0.1 + `arm64.nompam` | 7.1.3 | same | hangs |
| 0.4.0.1 + `initcall_blacklist=tegra234_cbb_init` | 7.1.3 | same | hangs |
| this branch | 7.1.3 | 100,968,546 B | **boots** |

**The mechanism is not fully identified.** 0.4.0 booted a payload of essentially the same size under 7.0.13, so this is an interaction between the 7.1.3 kernel and a payload of that size, not a fixed limit — 0.4.0 sat at 99.2% of the size that later failed. This buys margin (101 MB against a ~209 MB failure point) but does not remove the class. The kernel arrived via an **unpinned** `linux-image-*` float from trixie-backports, which #3589 explicitly listed as not addressed; pinning it is the durable fix and is deliberately not in this PR, since it also freezes CVE fixes across seven platforms.

Nothing was lost by the purge — verified at the ELF level. `nvidia.ko`'s only firmware references are `nvidia/580.173.02/gsp_{ga10x,tu10x}.bin`, both present in the built image at the `.run`'s exact byte sizes; neither purged package owns any path under `580.*`, and the NVIDIA userspace is byte-identical to 0.4.0.1.

## 2. Stage MOK enrollment when the password is set — fixes #3702

On the machine above, the GPU then did not work: `nvidia-smi` could not talk to the driver, `lsmod | grep nvidia` was empty, and the only relevant kernel line was `Kernel is locked down from EFI Secure Boot`.

`enroll_mok` will not stage the import unless the `start9` user already has a password hash, and on a fresh Secure Boot install neither caller could satisfy that. `os_install/mod.rs` resolved `mok_pub` inside the target overlay but read the hardcoded **host** `/etc/shadow`; and `init.rs` runs once per boot, while `setup_init` calls `init()` at `setup.rs:185` and `write_shadow()` at `setup.rs:226` — so the attempt always preceded the password it depends on, in the same function. The import was first staged on the boot *after* setup, and MokManager only prompts on the boot after that: two restarts, with every skip logged at `info`. On a server, the modules stay unloadable indefinitely.

Staging now happens in `write_shadow`, the single point every password-setting path goes through (setup, restore, password change), so the prompt appears on the next restart. `enroll_mok` takes the root holding both the certificate and `etc/shadow`, which also corrects the install-time read.

## 3. Name the second firmware package correctly

`firmware-nvidia-tesla-535-gsp` is NVIDIA's proprietary GSP firmware for the 535 driver series (`nvidia/535.274.02/gsp_{ga10x,tu10x}.bin`), not nouveau's. Only `firmware-nvidia-graphics` carries nouveau's chip-named blobs, and only those reach the initramfs. Both are still correct to purge.

## 4. Retry the container-runtime base image download

The riscv64 job here failed on a truncated transfer from `images.linuxcontainers.org` (`curl: (18)`, ~79 MB short) *after* the Rust compile succeeded. There was no retry, under `set -e`. The rootfs download also moves from `> file` to `-o file`, which is load-bearing: curl rewinds a file it owns before each retry but cannot rewind stdout, so `--retry` with a shell redirect concatenates partial attempts — measured, 1200 B from three 400 B attempts vs 400 B with `-o`.

## Ruled out along the way

MPAM (`arm64.nompam`) and `SOC_TEGRA_CBB` (`initcall_blacklist=tegra234_cbb_init`), both tested on the hardware and refuted — the blacklist commit was dropped rather than kept as an unexplained fence. The 580.159.03 → 580.173.02 driver bump and the `--kernel-module-type=proprietary` pin from #3589 are inert before `switch_root` (no `nvidia*.ko` is in either initramfs), and 0.4.0 already shipped the proprietary flavour, so it is not a regression vector.

### DIFF
--- a/projects/start-os/CHANGELOG.md
+++ b/projects/start-os/CHANGELOG.md
@@ -52,9 +52,10 @@ file tracks notable changes since the move to the monorepo.
 - **The 64-bit ARM NVIDIA image boots on NVIDIA GB10 hardware again, such as the
   DGX Spark, and is considerably smaller.** These images drive the GPU with
   NVIDIA's own driver and turn nouveau off, but they were still built with
-  nouveau's firmware for every supported chip — 152 MB of it, most of it inside
-  the initramfs that the bootloader must read in full before the kernel starts.
-  On GB10 that left the machine sitting on the last line the bootloader printed.
+  graphics firmware they cannot use — nouveau's, for every chip it supports,
+  along with an older driver series' — and 152 MB of it sat inside the initramfs
+  that the bootloader must read in full before the kernel starts. On GB10 that
+  left the machine sitting on the last line the bootloader printed.
   The images now carry only the graphics firmware they can use, which more than
   halves what is read at startup and takes a substantial amount off the
   download.
--- a/projects/start-os/CHANGELOG.md
+++ b/projects/start-os/CHANGELOG.md
@@ -49,6 +49,15 @@ file tracks notable changes since the move to the monorepo.
   produced, and only the container's boot-time device pass — which races with
   the grant — widened them; on the losing side of that race a CI runner's jobs
   failed to start with `Failed to open() /dev/net/tun: Permission denied`.
+- **The 64-bit ARM NVIDIA image boots on NVIDIA GB10 hardware again, such as the
+  DGX Spark, and is considerably smaller.** These images drive the GPU with
+  NVIDIA's own driver and turn nouveau off, but they were still built with
+  nouveau's firmware for every supported chip — 152 MB of it, most of it inside
+  the initramfs that the bootloader must read in full before the kernel starts.
+  On GB10 that left the machine sitting on the last line the bootloader printed.
+  The images now carry only the graphics firmware they can use, which more than
+  halves what is read at startup and takes a substantial amount off the
+  download.
 
 - **Your server answers to its own addresses and no others.** A name that
   resolved to your server but was never configured on it — a domain you pointed
--- a/projects/start-os/CHANGELOG.md
+++ b/projects/start-os/CHANGELOG.md
@@ -36,6 +36,17 @@ file tracks notable changes since the move to the monorepo.
   plaintext address, including the passwords typed into it. See
   [Gateways](https://docs.start9.com/start-os/gateways.html).
 
+### Changed
+
+- **The NVIDIA images now use NVIDIA's open kernel modules, which support GeForce
+  RTX 20-series, Quadro RTX and newer.** This is what makes current cards work at
+  all — an RTX 50-series, an RTX PRO 6000 or an NVIDIA GB10 can only be driven by
+  these modules. The trade is at the other end of the range: **a GeForce GTX
+  900-series or 10-series card, a Titan X or Xp, or a Tesla M40, P40, P100 or
+  V100 will no longer be driven by the NVIDIA images.** If you rely on one of
+  those, stay on 0.4.0.1 or use the Standard image, whose open-source `nouveau`
+  driver still provides display output without GPU compute.
+
 ### Fixed
 
 - Trim whitespace on form inputs.
@@ -83,8 +94,8 @@ file tracks notable changes since the move to the monorepo.
   the image built NVIDIA's driver in the flavour that does not support this
   generation of GPU, so the driver found the card but could never bring it up.
   The images now carry only the graphics firmware they can use, switch that
-  driver off, and build the flavour NVIDIA supports here — which is what NVIDIA's
-  own operating system does on the same hardware.
+  driver off, and build the kernel modules NVIDIA supports here — which is what
+  NVIDIA's own operating system does on the same hardware.
 
 - **Your server answers to its own addresses and no others.** A name that
   resolved to your server but was never configured on it — a domain you pointed
--- a/projects/start-os/CHANGELOG.md
+++ b/projects/start-os/CHANGELOG.md
@@ -49,6 +49,15 @@ file tracks notable changes since the move to the monorepo.
   produced, and only the container's boot-time device pass — which races with
   the grant — widened them; on the losing side of that race a CI runner's jobs
   failed to start with `Failed to open() /dev/net/tun: Permission denied`.
+- **Reinstalling with "Preserve" copies your server's configuration before it
+  rewrites the drive, so the configuration survives.** The copy used to be taken
+  afterwards, by which point there was nothing left to read — so the server came
+  back with its configuration reset, and said nothing about it. Among the things
+  lost was the key your server uses to sign its add-on drivers, which on a
+  machine with Secure Boot left hardware such as an NVIDIA GPU unavailable
+  afterwards. A reinstall that cannot read the configuration it was asked to keep
+  now stops and says so, rather than continuing and discarding it.
+
 - **On a server with Secure Boot enabled, hardware that needs an add-on driver —
   an NVIDIA GPU, most commonly — works after setup.** Secure Boot only loads such
   a driver once you approve the key your server signs it with, and approving that
--- a/projects/start-os/CHANGELOG.md
+++ b/projects/start-os/CHANGELOG.md
@@ -81,21 +81,26 @@ file tracks notable changes since the move to the monorepo.
   [Initial Setup](https://docs.start9.com/start-os/initial-setup.html).
 
 - **The 64-bit ARM NVIDIA image boots on NVIDIA GB10 hardware again, such as the
-  DGX Spark, its GPU works, and the image is considerably smaller.** Three
-  separate faults were involved, two of which each stopped the machine on the
-  last line the bootloader printed — so fixing either one alone changed nothing.
-  These images drive the GPU with NVIDIA's own driver and turn nouveau off, yet
-  were still built with graphics firmware they cannot use — nouveau's, for every
-  chip it supports, along with an older driver series' — and 152 MB of it sat
-  inside the initramfs that the bootloader must read in full before the kernel
-  starts. Separately, a Tegra fabric driver that recent kernels build in claims
-  one of the GB10's internal buses and reads a protected register on it, stopping
-  the machine during hardware detection before it can display anything. Finally,
-  the image built NVIDIA's driver in the flavour that does not support this
-  generation of GPU, so the driver found the card but could never bring it up.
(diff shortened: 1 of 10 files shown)


helix-nine (commented):
Reviewed f6c3374e across all four areas (firmware/open-modules/ATS/initcall shell work, `preserve_config` rework, MOK staging, download retry + docs). CI green on the full matrix incl. `aarch64-nvidia`. Overall this looks solid — one corner-case finding below, one question, and the verified-OK list.

## Finding (corner case): restore is gated on `CONFIG_BAK` existing, not on this attempt having preserved

`install_os_to`'s restore runs on `tokio::fs::metadata(CONFIG_BAK).await.is_ok()`, and `/tmp` persists across install retries within one live-boot. Two reachable shapes, both need a failed first attempt:

1. **Cross-attempt, cross-disk leak.** Attempt A (protect, real config staged, then fails late). User retries on a *different* disk that has a pool partition (`protect.is_some()`) but no OS root → `previous_os_root` returns `None` → `preserve_config` returns `Ok` **without clearing `CONFIG_BAK`** (mod.rs:289-294) → attempt A's config — including `overlay/etc/shadow` and `overlay/var/lib/dkms` — is installed on disk B. That is the Secure Boot key/cert desync this PR exists to fix.
2. **Same-disk retry can overwrite the good backup with defaults.** If attempt A failed *after* the new root's default `config/` was created, attempt B's `previous_os_root` finds that default config (it only checks for a `config` dir) and `preserve_config` deletes the real staged bak and re-stages the defaults over it. If attempt A failed *before* that, the leftover bak is the *good* one and restoring it is exactly right — so the `None` branch can't be fixed by unconditionally deleting `CONFIG_BAK` either.

The `protect.is_none()` branch already clears `CONFIG_BAK` (good); it's the `protect.is_some()` + `None`-root path that's ambiguous. Suggested direction: gate the restore on an explicit signal that *this attempt* preserved (return a `bool` from `preserve_config`, or a marker file written after the rename), and note the same-disk-retry trade-off explicitly — with a flag, a retry whose preserve finds nothing restores nothing, even when the leftover bak is the user's real config. Fine as a follow-up if you'd rather not grow this PR, but worth tracking.

## Question

`--no-runlevel-check` was dropped from the NVIDIA installer invocation (it ran on the old `sh $RUN_PATH --silent ...` line). The extracted `./nvidia-installer` clearly passes in the live-build chroot (CI built `aarch64-nvidia` and `x86_64-nvidia`), so this is curiosity more than concern — was the drop deliberate, or incidental to the extract-only restructure?

## Verified OK

- All three `enroll_mok` call sites consistent with the new `root` signature (`init.rs` → `/`, `auth.rs` → `/` after the shadow copy, `os_install` → `overlay.path()`); no other callers. `mok_enrolled`'s wizard consumer (`success.page.ts`) is unaffected — the flag was effectively always false pre-PR (fresh install has no password at install time; that was #3702), and the preserve-reinstall case now gets the success-page MOK step, which is a strict improvement.
- Preserve prune list is byte-identical to the old five entries; pruning the staged copy instead of the read-write-mounted original means a failed preserve leaves the previous install intact — better than before. `delete_dir`/`delete_file` are ENOENT-tolerant; staged-rename is same-tmpfs atomic.
- Firmware purge runs before `update-initramfs`; `apt-get purge` of installed-but-unwanted firmware packages is safe under `set -e`. ATS-guard sed has the `grep -qF` fail-loud guard. The `0 &&` sed replacement escapes its `&`s correctly.
- `postinst` platform guard reads `/usr/lib/startos/PLATFORM.txt` (shipped in the deb), and both consumers of `/etc/default/grub` — image build and the `upgrade` script — run `update-grub` afterwards, so upgrades pick up the blacklist too.
- `download-base-image.sh` is `#!/bin/bash`, so the `RETRY` array is fine; `-o` vs `>` reasoning is correct (curl can't rewind stdout across retries).
- Changelog: top heading `[0.4.0.2]`, no matching origin tag (latest is `start-os/v0.4.0.1`) → correctly accumulated in place; sections right (`Changed` for the open-modules move, `Fixed` for the rest). The Turing+/Maxwell-Pascal-Volta support matrix is accurate. `[!IMPORTANT]` admonition has precedent in the book.
- `mokutil` restaging on password change is a quiet win: a user who changes their password before the enrollment boot gets the pending import re-staged with the new hash.

dr-bonez (approved):

PR #3694, opened 2026-08-12
feat(sdk): add uses to daemon/oneshot entries for Daemons.dynamic

helix-nine:
Surfaced by the lnd-startos#178 review: an import daemon's `exec.fn` captured the origin host/password at builder time, and because `configHash` reduces function-form execs to `{__fn: true}`, a rebuilt chain with corrected credentials diffed to "leave alone" — the running daemon kept retrying with the old password.

`configHash` can't see closures (JS offers no introspection), so this adds the explicit escape hatch the `Daemons.dynamic` docstring already alludes to: an optional `uses` on `addDaemon()`/`addOneshot()` (including the pre-built `{ daemon }` form, which is otherwise entirely opaque to the hash). The field declares the values a daemon's closures use; any change to them restarts the entry on the next reconcile.

Only JSON-serializable values are useful, and the API says so at every touchpoint (field docstring, `configHash` docstring, changelog, packaging recipe). Normalization never throws — `configHash` runs inside the reconciler, so throwing would turn a packager's mistake into a failed reconcile:

- functions, symbols, cycles and `undefined` hash as distinct `UNSERIALIZABLE:*` sentinels — recognizable in the hash, and different kinds don't collide (two values of the same unserializable kind still hash alike: no restart, matching the JSON-only contract)
- BigInts hash as their decimal string, so they still signal a change

- `lib/mainFn/Daemons.ts`: `uses` on `NewDaemonParams` + the `addDaemon` intersection, recorded on the entry, canonicalized into `configHash`. The `exec` options object is now canonicalized in full (the old whitelist dropped a fn-form's `sigtermTimeout` and all `onStdout`/`onStderr` callbacks); function values hash as constant sentinels, so closure contents stay invisible and `uses` remains the escape hatch
- `lib/test/Daemons.test.ts`: hash changes on value change, key-order-independent, participates for oneshots, unserializable kinds get distinct sentinels, circular values hash stably while sibling fields still signal
- CHANGELOG entry under the unreleased 2.0.10; `recipe-dynamic-daemons.md` updated in the same change

`tsc --noEmit` + full jest suite (96 tests) green.

### DIFF
--- a/projects/start-sdk/lib/test/Daemons.test.ts
+++ b/projects/start-sdk/lib/test/Daemons.test.ts
@@ -322,6 +322,44 @@ describe('configHash', () => {
     expect(configHash(a.entries[1])).not.toEqual(configHash(b.entries[1]))
   })
 
+  it("changes when a fn-form exec's sigtermTimeout changes", () => {
+    const e = fakeEffects()
+    const make = (sigtermTimeout?: number) =>
+      Daemons.of<Manifest>({ effects: e }).addDaemon('a', {
+        subcontainer: lazy(e),
+        exec: { fn: async () => null, sigtermTimeout },
+        ready: baseReady,
+        requires: [],
+      }).entries[0]
+    expect(configHash(make(5000))).not.toEqual(configHash(make(6000)))
+    // different fn closures still hash alike — their contents are invisible
+    expect(configHash(make(5000))).toEqual(
+      configHash(
+        Daemons.of<Manifest>({ effects: e }).addDaemon('a', {
+          subcontainer: lazy(e),
+          exec: { fn: async () => null, sigtermTimeout: 5000 },
+          ready: baseReady,
+          requires: [],
+        }).entries[0],
+      ),
+    )
+  })
+
+  it('changes when an output callback is added to a command exec', () => {
+    const e = fakeEffects()
+    const make = (withCallback: boolean) =>
+      Daemons.of<Manifest>({ effects: e }).addDaemon('a', {
+        subcontainer: lazy(e),
+        exec: {
+          command: ['cmd'],
+          ...(withCallback ? { onStdout: () => {} } : {}),
+        },
+        ready: baseReady,
+        requires: [],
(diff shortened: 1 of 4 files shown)


projects/start-sdk/lib/test/Daemons.test.ts
```diff
@@ -134,6 +134,70 @@ describe('configHash', () => {
...
+    const circular: any = {}
+    circular.self = circular
+    const entry = Daemons.of<Manifest>({ effects: e }).addDaemon('a', {
+      subcontainer: lazy(e),
+      exec: { command: ['cmd'] },
+      ready: baseReady,
+      requires: [],
+      hashExtra: circular,
+    }).entries[0]
+    expect(() => configHash(entry)).toThrow(
```
dr-bonez: this shouldn't throw, should just normalize

helix-nine: Agreed — throwing put a packager's mistake in the one place it can stop the service: `configHash` runs inside the reconciler, so an unserializable `hashExtra` failed the reconcile rather than just failing to signal a change. Fixed in 0c8b59b, and the throw is gone entirely — `canonicalizeExtra` is deleted and `hashExtra` goes through the same `canonicalize` as every other hashed field.

Normalization now:

- **cycles** — `canonicalize` tracks the ancestor chain and collapses a back-reference to `null`; the rest of the value still hashes, so `{ port, self }` keeps reacting to `port`
- **symbols** — `null`, alongside functions (previously they fell through and `JSON.stringify` silently dropped the key, which made a symbol and `undefined` hash differently for no good reason)
- **BigInt** — its decimal string, so it still signals a change instead of going quiet at `null`

Tests replaced the throw case with cycle and BigInt cases (hash is stable, and still changes when a sibling field changes); docstrings, changelog and the packaging recipe updated to match. `tsc --noEmit` + 94 jest tests green.

dr-bonez (approved):

PR #3697, opened 2026-08-12
docs: add SECURITY.md

helix-nine:
The repo had no security policy, so GitHub showed no "Report a vulnerability" entry in the issue chooser and the only pointer anywhere was a bare `security@start9.com` line in README and CONTRIBUTING.

Start9 already publishes a full policy at <https://start9.com/security> plus an RFC 9116 `security.txt`, so this file does not invent a second one. It applies that policy to this repository:

- **Two private channels** — `security@start9.com`, or GitHub's advisory form (private vulnerability reporting is enabled here, verified `{"enabled":true}`).
- **The OpenPGP key** for `security@start9.com`, fingerprint `5456 DBFF 1B9D F905 041F A776 5259 ADFC 2D63 C217` — the same key as `apt/start9.gpg`, which I confirmed is byte-identical to <https://start9.com/start9.gpg> and is served by keys.openpgp.org under both UIDs.
- **Scope** — the six released products in this repo, the shared libs, the two website projects, and the reusable CI; versus the separate `*-startos` package repos, where an issue is a public disclosure.
- **What is not a vulnerability report** — support questions.

Legal text (safe harbour) is linked, not restated, so the repo and the website cannot drift. The operational commitments (5 business days, credit) are quoted and attributed.

`CONTRIBUTING.md` says the root's docs are exactly four files and must be updated in the same change, so that list goes to five; README's Documentation list and the two `projects/start-os/` security lines are wired up the same way.

## Two things worth your attention

**1. `security.txt` advertises a key that cannot encrypt.** The file has `Encryption: https://start9.com/start9.gpg`, which under RFC 9116 §2.5 means "encrypt your report to this key". But the key is `scaSCA` — sign/certify/authenticate, no encryption subkey:

```
$ gpg --encrypt -r security@start9.com
gpg: security@start9.com: skipped: Unusable public key
```

So a researcher who does the responsible thing and encrypts the exploit hits an unexplained failure. Worse, `apt/start9-alpha.gpg` *does* carry a cv25519 `[E]` subkey and encrypts fine, so someone troubleshooting can find a Start9 key that "works" and encrypt to the CI key.

I wrote the key section around what the key provably does (checking a Start9 signature, and giving an independent copy of the fingerprint to compare against start9.com and keys.openpgp.org) and route confidentiality to the advisory form. The real fix is on the website — drop the `Encryption:` field, or add an encryption subkey. Say which and I will match the wording.

**2. `start-cli -H <host> git-info` reports the client's build, not the server's.** I had drafted that as the way to identify an alpha build; it is wrong. `git-info` is registered at `shared-libs/crates/start-core/src/lib.rs:142` as `from_fn(|_: C| version::git_info())` with no `.with_call_remote::<CliContext>()` — unlike `echo` and `state` directly below it — and `git_info()` returns the `include_str!`-baked `COMMIT_HASH` of the running binary. `-H` is parsed and ignored, so you get the hash of whatever start-cli the reporter has installed.

SECURITY.md now points at the About dialog instead. But **the same wrong invocation is in `AGENTS.md:98` and `.github/ISSUE_TEMPLATE/1-bug-start-os.yml`** ("From a client, `start-cli -H https://<your-server> git-info` prints the exact build") — so bug reports are being filed with the wrong hash today. I left those alone to keep this PR docs-only. Want a follow-up fixing both, or a `.with_call_remote::<CliContext>()` on line 143 to make the documented behaviour real?

## Open questions for you

- **Supported versions.** I deliberately omitted a version table — nothing published states whether StartOS 0.3.5.x still gets security fixes. Worth stating if there is an answer.
- **Org-wide default.** No repo in either org has a SECURITY.md. A `Start9Labs/.github/SECURITY.md` would cover all ~130 repos including the `*-startos` packages, which is where package reports actually land.

Verification: prettier clean; every relative link resolves; every external URL fetched (start9.com 404s properly on bogus paths, so `/security` and `/.well-known/security.txt` are real pages, not soft-404s).

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -16,21 +16,26 @@ Include:
 - The git hash, if you tested a build from `master`. Every `master` build of a product publishes under the same version number, so the hash is the only thing that identifies it. StartOS shows it in the web UI under the server's About dialog, next to the version.
 - What an attacker gains, and what access they need to start.
 - Steps to reproduce, ideally from a clean install.
-- A proof of concept, if you have one. It saves everyone time. If you would rather keep an exploit off email, the GitHub advisory form holds the report privately between you and the maintainers.
+- A proof of concept, if you have one. It saves everyone time. To keep an exploit out of plaintext mail, encrypt it to the key below, or use the GitHub advisory form.
 - Whether you want credit in the release notes, and the name to use.
 
 Machine-readable contact details are at <https://start9.com/.well-known/security.txt>, per RFC 9116.
 
 ### Our OpenPGP key
 
-`security@start9.com` shares an OpenPGP key with the rest of Start9:
+Encrypt your report to this key, or use it to check a Start9 signature:
 
 ```
 Start9 <security@start9.com>
 5456 DBFF 1B9D F905 041F  A776 5259 ADFC 2D63 C217
 ```
 
-It is in this repository as [`apt/start9.gpg`](apt/start9.gpg), and published at <https://start9.com/start9.gpg> and on <https://keys.openpgp.org>. It is the key that signs the `stable` suite of Start9's Debian repository ([`apt/start9.list`](apt/start9.list)), so a machine that already installs Start9 packages carries a copy you can compare the fingerprint against. Use it to check a Start9 signature. To send a report confidentially, use the GitHub advisory form above.
+```
+gpg --fetch-keys https://start9.com/start9.gpg
+gpg --encrypt --armor --recipient security@start9.com report.txt
+```
+
+The key is also in this repository as [`apt/start9.gpg`](apt/start9.gpg) and on <https://keys.openpgp.org>. The same key signs the `stable` suite of Start9's Debian repository ([`apt/start9.list`](apt/start9.list)), so a machine that already installs Start9 packages carries a copy you can compare the fingerprint against.
 
 ## What happens next
 
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,83 @@
+# Security Policy
+
+Start9's vulnerability disclosure policy is published at <https://start9.com/security>. That page is authoritative. This file says how it applies to this repository.
+
+## Reporting a vulnerability
+
+Email [security@start9.com](mailto:security@start9.com), or file privately on GitHub through the [draft advisory form](https://github.com/Start9Labs/start-technologies/security/advisories/new) — private vulnerability reporting is enabled on this repository. Use one of those two channels rather than a public issue, the forum, or a Matrix room.
+
+Please give us a chance to fix the problem before you publish. We will not ask you to stay quiet indefinitely, and we will not ask you to sign anything. If you have a date in mind, say so in your report and we will work to it.
+
+If you are unsure whether something is in bounds, ask at [security@start9.com](mailto:security@start9.com) first. We would rather answer the question than argue about it afterwards.
+
+Include:
+
+- The product and version you tested, and how it was installed (Start9 hardware, your own hardware, or a virtual machine). Quote the version exactly as the server reports it — StartOS versions may carry a fourth revision segment, as in `0.4.0.1`.
+- The git hash, if you tested a build from `master`. Every `master` build of a product publishes under the same version number, so the hash is the only thing that identifies it. StartOS shows it in the web UI under the server's About dialog, next to the version.
+- What an attacker gains, and what access they need to start.
+- Steps to reproduce, ideally from a clean install.
+- A proof of concept, if you have one. It saves everyone time. If you would rather keep an exploit off email, the GitHub advisory form holds the report privately between you and the maintainers.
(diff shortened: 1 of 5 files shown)


SECURITY.md
```diff
@@ -0,0 +1,83 @@
...
+### Our OpenPGP key
+
+`security@start9.com` shares an OpenPGP key with the rest of Start9:
+
+```
+Start9 <security@start9.com>
+5456 DBFF 1B9D F905 041F  A776 5259 ADFC 2D63 C217
+```
+
+It is in this repository as [`apt/start9.gpg`](apt/start9.gpg), and published at <https://start9.com/start9.gpg> and on <https://keys.openpgp.org>. It is the key that signs the `stable` suite of Start9's Debian repository ([`apt/start9.list`](apt/start9.list)), so a machine that already installs Start9 packages carries a copy you can compare the fingerprint against. Use it to check a Start9 signature. To send a report confidentially, use the GitHub advisory form above.
```
dr-bonez: Note that security@start9.com also accepts pgp encrypted emails

dr-bonez (dismissed):

SECURITY.md
helix-nine: Happy to add that, but I don't think a reporter can act on it yet — there's no key published that they can encrypt to.

The key at `start9.com/start9.gpg` (and the identical one on keys.openpgp.org, under both `security@` and `dev@`) is an Ed25519 primary with **zero subkeys**, capabilities `scaSCA`:

```
$ gpg --encrypt -r security@start9.com
gpg: security@start9.com: skipped: Unusable public key
```

That isn't a missing flag — Ed25519 is a signature algorithm, so there's no ECDH to encrypt against. Encryption needs a cv25519/ECDH or RSA subkey, and there's no WKD record for start9.com either.

The part that worries me: the only Start9 key anywhere with an `[E]` subkey is the **alpha CI key** (`apt/start9-alpha.gpg`, `Start9 Alpha Repository <ops@start9.com>`, `sub:…:e`), and it encrypts fine. So a researcher who hits the error above and goes looking will find a Start9 key that works — and encrypt an exploit to the key CI holds. That's why I'd rather not print "we accept PGP mail" next to a key that can't receive it.

Two ways to make it true, and they're not equivalent:

1. **Add an ECDH encryption subkey to `5456DBFF…`.** Cheapest. Existing `apt` verification is unaffected — the signing material doesn't change, so already-installed keyrings keep validating `InRelease`. But it attaches inbound-mail decryption material to what AGENTS.md:130 calls the production release key, so whoever reads `security@` needs that private subkey. Different threat model than an offline release key.
2. **Publish a separate `security@start9.com` encryption key.** Keeps the release key doing one job. Costs a second key to manage and publish.

Either way `security.txt`'s `Encryption:` field should point at whichever key can actually receive — right now it points at this one, and RFC 9116 §2.5 says that field means "encrypt to this".

Tell me which you want and I'll push the line plus the key details in this PR. If you'd rather ship the file as-is and sort the key separately, that works too — say the word and I'll leave it.

helix-nine: Done in 3bf2fa32 — the key now has an ECDH encryption subkey (`b7075ac28`), so this is accurate.

The key section now leads with encryption and carries the commands:

```
gpg --fetch-keys https://start9.com/start9.gpg
gpg --encrypt --armor --recipient security@start9.com report.txt
```

I ran both against the key as committed, served over HTTP from a throwaway keyring: import succeeds, `--encrypt -r security@start9.com` produces a 293-byte `BEGIN PGP MESSAGE` with `algo 18`. The PoC bullet now offers encrypted mail alongside the advisory form rather than routing everything confidential to GitHub.

**Retracting something I said above:** I claimed `security.txt`'s `Encryption:` field should be repointed. That was wrong. RFC 9116 §2.5.4 requires that field to be a *URI* where the key can be retrieved, and `https://start9.com/start9.gpg` is exactly that — the subkey is what makes the existing field truthful. It needs no change. (The RFC also explicitly allows this key to double as the apt signing key: "Researchers must not assume that this key is used to generate the digital signature.")

Two notes on the key itself, both verified against `origin/master`:

- The subkey is encryption-only (`algo 18`, `key flags: 0C`), which is what keeps apt safe. The primary still issues signatures, and since the fingerprint is a hash over the primary key packet and is unchanged, the 400-byte keyring already deployed at `/usr/share/keyrings/start9.gpg` still verifies new `stable` signatures. No keyring rollout needed.
- Propagation is still in flight: `start9.com/start9.gpg` and keys.openpgp.org were both serving the pre-subkey key when I last checked. The website copy is the one this file and `security.txt` point reporters at, so the commands above start working there once it deploys. Happy to re-verify all three when it lands.

Sorry — this push dismissed your approval. The delta from the sha you approved is `SECURITY.md` only: the PoC bullet, and the key section rewritten to lead with encryption. No other file changed, and prettier is clean.

dr-bonez (approved):

PR #3709, opened 2026-08-14
docs(start-sdk): give service READMEs a fixed section contract

MattDHill:
Targets `live-docs` because this corrects instructions the site is serving **today** and depends on nothing unreleased. On merge it deploys, and `docs-backport.yml` carries the same commit to master.

## Why it can't wait for the SDK tag

The **currently published** `writing-readmes.md` tells packagers to:

- put an `**Upstream docs:**` link at the top of the README
- write a `## What Is Unchanged from Upstream` section
- write a `## Contributing` section

The pre-publish checklist *requires* the first two. All three are now wrong, and the page also names `## Configuration Management` and `## Actions (StartOS UI)` while having no File Models, Tasks, or Troubleshooting.

The readers most likely to reach the website rather than a monorepo checkout are new and community packagers — community-registry is 64 of ~125 packages — and they are exactly the people writing a README from scratch against whatever the site says. A per-package sweep to the new contract is starting, so a stale published guide also invites a contributor to "correct" a swept README back.

## The change

A package README is the only technical file an AI support agent is given, and the file an assistant administering a server reads to operate a service. The four readers are **nested** — each reads a prefix of the same list — and every document has at least two readers except `AGENTS.md`, which is why "repeat nothing" is safe there and unsafe anywhere else.

**Headings are now a fixed, ordered set** in four groups: what the package is made of, how it behaves, what to expect when it doesn't, then the machine-readable summary. A renamed section doesn't fail loudly — it silently degrades retrieval to loading the whole file.

**Added:** `File Models` (replacing `Configuration Management`, which named an abstraction rather than the config files it covered and overlapped `Actions`), `Tasks` (what raises each one, its severity, what clears it — StartOS reports only which are *currently* raised), and `Troubleshooting`.

**Removed:** `Contributing` (identical boilerplate in every package, already in `AGENTS.md`) and `What Is Unchanged from Upstream` (an unbounded list that inverts the scoping note above it: once it exists, a feature missing from it reads as changed).

**Two content rules:** don't restate `instructions.md` — every reader of the README also has it, and upstream links belong only in its `## Documentation` section, which is parsed; and don't re-encode what StartOS exposes — action ids, descriptions, warnings, `allowedStatuses`, input schemas and health-check ids are live, so document instead when to run a thing, what it costs, whether it is safe to repeat, and what it changes.

The quick reference mirrors the sections in order. It deliberately has **no** `backups` key: `withPgDump` means a volume is dumped and replayed rather than copied, and an `included`/`excluded` list would state that backwards in the format that reads most authoritative.

**`AGENTS.md`** is scoped to repo mechanics, prohibitions, extension points, naming traps and package-specific build invocations. The template stops shipping "Inspecting a running install", inherited by all 57 start9-registry packages and package-independent apart from the subcontainer names, which now belong in the README's operable surface; the `attach` `-n`-vs-`-s` trap moves into `workflow.md`.

## Safe under the live-docs rule

`AGENTS.md`: *never merge a docs change into `live-docs` that isn't already true of the published software.* Verified against origin tags:

| Referenced by these pages | Landed in |
| --- | --- |
| `Backups.withPgDump` / `withMysqlDump` | `0.4.0-beta.65` |
| `createTask` / `createOwnTask` | ≤ `0.4.0-beta.63` |
| `ActionMetadata.access` | `2.0.0` |
| `start-cli package attach -n` | released `start-cli/v1.1.0` and `start-os/v0.4.0.1` |

## Scope notes

- **No changelog entry.** The SDK changelog documents the `@start9labs/start-sdk` npm package; neither the guide nor `package-template/` ships in it, and no version is bumped.
- `package-template/` is included because it is live code — `init-package` copies it out of the checkout — so new packages scaffold conformant. Its quick-ref `package_id` is quoted (`'{{id}}'`) because Prettier rewrites a bare `{{id}}` inside a YAML fence into `{ { id } }`, which would not interpolate.
- Built by applying only this change's patch onto `live-docs`, so the unrelated pending edits master carries on these pages — and on `interfaces.md`, `main.md`, `versions.md`, `tasks.md` and others documenting unreleased 2.0.10 features — are **not** published as a side effect.

## Follow-on

Existing packages don't conform yet. A per-package sweep follows and must be source-verified rather than mechanical: File Models, Tasks, subcontainers and the quick reference all have to be read out of `startos/`. The first package through the shape turned up a quick reference listing environment variables that no longer exist, omitting an action entirely, and giving a health check's display string where its id was needed.


projects/start-sdk/docs/package-template/README.md
```diff
@@ -5,49 +5,100 @@
...
+The headings below are a fixed set, in this order: they are how an agent fetches part
+of this file without loading all of it. Don't rename or reorder them. They run in four
+groups — what the package is made of, how it behaves, what to expect when it doesn't,
+then meta. Delete a section only where the guide says you may (Tasks and
+Troubleshooting, if the package has neither). Open every section with a sentence or two
+of prose before any table — that text becomes the section's summary in the generated
+index.
+
+Don't restate what StartOS already exposes (action ids, descriptions, allowed
+statuses, input schemas): document when to run a thing, what it costs, whether it is
```
dr-bonez: Is this the only source of this guidance? I'd worry there guidance would be lost when the TODO is deleted. Maybe best to not frame it as a TODO?

projects/start-sdk/docs/src/writing-readmes.md
```diff
@@ -1,17 +1,46 @@
...
-2. **AI assistants** -- structured data that can be parsed programmatically
+Two consequences:
 
-## Recommended Structure
+- **The README is the only technical file a support agent gets.** It reads this and `instructions.md` from your repository and nothing else — not `AGENTS.md`, not your source. If a fact is needed to diagnose a user's problem, it belongs here.
+- **Where readers conflict, the operable surface wins.** An agent wants predictable structure it can address; a developer wants prose. Both are served by fixed headings with real writing underneath — not by prose that wanders across section boundaries.
+
+`instructions.md` is for the user (see [Writing Service Instructions](writing-instructions.md)); `AGENTS.md` is for whoever changes the package, and repeats nothing this file says.
+
+## Guiding principles
```
dr-bonez, line 23: I see, looks like a lot of the same guidance is here. Why not just link to this?

dr-bonez (dismissed):

projects/start-sdk/docs/src/writing-readmes.md
MattDHill: Agreed — done in 087c6644. The header comment was a compressed copy of the Guiding principles, which is the exact duplication this PR exists to stop, so it is now a pointer to the guide.

The per-section prompts stay. They label the slot rather than restate the rationale (`## Actions` → *when to run it, what it changes, cost, repeat safety*), so the rules are still in front of whoever fills the template in, and nothing is lost when the comments are deleted.

projects/start-sdk/docs/package-template/README.md
MattDHill: This turned out to be the more serious of the two, and in a place I had not looked — thanks.

For the template README the answer is your other comment: link, don't copy, and the TODO framing then becomes fine, because what gets deleted is a pointer.

For the template **AGENTS.md** you were literally right: that comment was the *only* statement of what belongs in a package `AGENTS.md`, so deleting it destroyed the rule. Worse, when I first moved it I put it in `writing-readmes.md` — where it doesn't belong, and where it became a *third* copy, because `project-structure.md` § **AGENTS.md and CLAUDE.md** already owned this and already said "do not restate generic guide content".

It now lives there and only there, extended with what to put in instead and why the file is short: it is the one package document with a single reader, so it must not restate `README.md` or `instructions.md` either.

One deliberate exception, called out in that section: **prohibitions get inlined in the package's `AGENTS.md` rather than linked** — an imperative behind a pointer is a suggestion, which is how start-sdk 2.0.4 and 2.0.5 shipped untagged.

PR #3733, opened 2026-08-17
fix(core/net): close the client leg when the upstream leg closes

### LINKED ISSUE #3731: HTTP-aware vhost proxy pins each client connection to one upstream connection and keeps the client side open after it dies — backend keep-alive expiry surfaces as client-visible aborts (net/http.rs run_http1_proxy)
### Environment

- StartOS 0.4.0.1 (current stable; the `10.0.3.1` bridge-gateway signature in the reporter's service log matches the 0.4.x container architecture). The mechanism is verified on `master` @ e1d036db and is present unchanged in the released `start-os/v0.4.0.1` (checked at the tag)
- Area: networking / reverse proxy (vhost HTTP-aware path)
- Observed via the `nextcloud` package (33.x, stock `nextcloud:*-apache` image) with Nextcloud Desktop on Windows and Linux, connecting over TLS addresses (`.local` / router domain / StartTunnel) — i.e. through the OS reverse proxy

### What happens

Nextcloud Desktop periodically flips **Connected → "Network error"** and auto-recovers a few seconds later. Sync itself works; credentials stay valid; no user action is needed. Client debug logs show the `ConnectionValidator` receiving `QNetworkReply::RemoteHostClosedError` ("Connection closed") on its periodic `status.php` poll. At the same moments, the service's Apache access log gains paired entries of the form:

```
10.0.3.1 - - "-" 408 0 "-" "-"
10.0.3.1 - - "-" 408 0 "-" "-"
```

followed ~10 s later by a fresh-connection burst (`/index.php/204`, `/status.php`, `PROPFIND /remote.php/dav/...`) as the client reconnects. `10.0.3.1` is `HOST_IP` — the host's address on the container bridge, i.e. the OS proxy's dial source.

Any HTTP client that holds persistent connections open between requests (sync clients, API pollers) is affected behind any `http`/`https`-preset binding. Browsers experience the same aborts but mask them by silently retrying idempotent requests, which is why the web UI never shows it.

A backend closing an idle keep-alive connection is routine (Apache's default `KeepAliveTimeout` is 5 s; Node's default is also ~5 s) and must be invisible to clients. Standard reverse-proxy semantics: the upstream leg is per-request (dial lazily, re-dial when a pooled upstream connection has died), or — at minimum — when the upstream leg ends, the client leg must be closed promptly so the client's connection pool sees the EOF *before* it tries to reuse the connection.

### Steps to reproduce

Found by reading code from a user report; the chain below is verified at the source level plus empirically against the backend image. Deterministic repros:

**1. Half-dead reuse abort (the flap).** Against any service bound with the SDK `http` preset and an Apache/short-keep-alive backend (e.g. nextcloud), from a LAN machine:

```python
import socket, ssl, time
ctx = ssl._create_unverified_context()
s = ctx.wrap_socket(socket.create_connection(("nextcloud.local", 443)), server_hostname="nextcloud.local")
s.sendall(b"GET /status.php HTTP/1.1\r\nHost: nextcloud.local\r\n\r\n"); s.recv(8192)  # 200 OK
time.sleep(30)          # backend closed its leg at ~5s; through the proxy this socket still looks open
s.sendall(b"GET /status.php HTTP/1.1\r\nHost: nextcloud.local\r\n\r\n")
print(s.recv(8192))     # b'' — abrupt close, no response: the RemoteHostClosedError the desktop client reports
```

Control: run the same two-request sequence against the container directly (`start-cli package attach nextcloud -n nextcloud-sub -- curl` or from the host to the container IP) — the socket turns readable-EOF ~5 s after the first response, so a client pool discards it before reuse. The proxy hides that EOF for up to 60 s.

**2. The 408 pairs.** Complete a TLS handshake and send nothing (`openssl s_client -connect nextcloud.local:443`): the proxy has already dialed the backend at ClientHello, so ~50 s later the service log gains a `"-" 408 0 "-" "-"` line (~30 s of `TCP_DEFER_ACCEPT` before Apache sees the request-less connection, then mod_reqtimeout's 20 s). The desktop client's speculative spare sockets produce exactly this, in pairs.

All references on `master` @ e1d036db.

1. **Every `http`/`https`/`ws`/`wss` binding takes the HTTP-aware proxy path.** The SDK presets set `addXForwardedHeaders: true` (`shared-libs/ts-modules/start-core/lib/interfaces/Host.ts:23-45`), and `handle_stream` selects `run_http_proxy` whenever forwarded headers or an auth gate are configured (`shared-libs/crates/start-core/src/net/vhost.rs:1206`). The SDK also pins ALPN to `http/1.1`, so this is always `run_http1_proxy`.

2. **One upstream connection per client connection, dialed eagerly, never replaced.** `preprocess` dials the backend during the TLS ClientHello — before any HTTP bytes exist (`vhost.rs:1088-1133`). `run_http1_proxy` performs a single hyper client handshake on that one stream and shares the `SendRequest` behind `Arc<Mutex>` (`shared-libs/crates/start-core/src/net/http.rs:327-332`). Nothing anywhere re-dials.

3. **The client leg outlives the upstream leg.** `futures::future::try_join(from, to)` (`http.rs:412`): when Apache closes its idle connection (silent FIN — verified: zero bytes on the wire), the upstream `Connection` future resolves `Ok`, and `try_join` keeps driving the client-side connection for up to `header_read_timeout` = 60 s (`http.rs:335-337`). The pair is now half-dead.

4. **Reuse aborts with no response.** The next client request hits `client.lock().await.send_request(req)` on the dead connection; the error is returned out of the `service_fn` (`http.rs:366-369`), which makes hyper abort the client connection without writing a response. A client that has just written a request onto a connection it had every reason to believe healthy gets exactly `RemoteHostClosedError`.

5. **The timing turns a rare race into a near-certainty.** Backend closes at 5 s idle (verified: `Keep-Alive: timeout=5, max=100`, reuse at 8 s fails); the proxy hides the close for up to 60 s; the Nextcloud validator polls every 30 s — squarely inside the dead window, every idle cycle. On the raw-splice path (non-HTTP-aware bindings) `copy_bidirectional` (`vhost.rs:1226`) propagates the FIN immediately and client pools recover invisibly — which is why only HTTP-aware bindings exhibit this.

6. **The 408 noise is the same eager dial.** Request-less backend connections (client speculative/spare sockets held silent) get reaped by the image's `mod_reqtimeout` (`header=20-40`): verified close at 20.0 s after first byte, ~52 s for a fully silent connection (`TCP_DEFER_ACCEPT` ≈ 30 s first), logged as `"-" 408 0 "-" "-"` with nothing sent on the wire.

Related: #3216 added the client-side idle bound (`header_read_timeout`, per-target cap) — it bounds resource usage but is also what fixes the half-dead window at 60 s; before it the window was unbounded. Distinct from #3577 (stale `ProxyTarget` after container recreate). `run_http2_proxy` shares the one-upstream structure but its 25 s keep-alive PINGs mask idle expiry for h2 backends; h1 is the affected path in practice.

Fix directions, in order of preference:

1. **Per-request upstream dialing** (lazy dial on first request, re-dial on a dead pooled connection) — standard reverse-proxy behavior; also eliminates the 408 churn, since a client connection that never sends a request never touches the backend.
2. **Minimal:** when the upstream `Connection` future completes, call `graceful_shutdown()` on the server connection (replace the bare `try_join`) so the client sees the close as promptly as the splice path shows it.

A package can work around it by raising its backend's timeouts above every hop in front (nextcloud-startos does this now — see the cross-referenced PR), but every affected package shouldn't need to know proxy internals; the OS should make backend keep-alive expiry unobservable.

### Logs & evidence

From the reporting user (Nextcloud service log):

```
10.0.3.1 - - " -" 408 0 "-" "-"
10.0.3.1 - - " -" 408 0 "-" "-"
# ~10s later:
GET /index.php/204            -> 204
GET /status.php               -> 200
PROPFIND /remote.php/dav/files/embassy/ -> 207
```

Client (Nextcloud Desktop debug log): `ConnectionValidator` receives `QNetworkReply::RemoteHostClosedError "Connection closed"`, then `AccountState state change: "Connected" -> "Network error"`, then auto-reconnect. Same client symptom over the service's `.onion` address — that leg bypasses the vhost proxy (tor dials the container directly), where the raw 5 s Apache close over a high-latency transport hits the same client race; the package-side timeout raise covers it.

Empirical timings, stock `nextcloud:33.0.8-apache` (Debian Apache defaults, mpm_prefork, mod_reqtimeout enabled):

| Probe | Behavior |
| --- | --- |
| Keep-alive advertisement | `Keep-Alive: timeout=5, max=100` |
| Reuse after 8 s idle | connection already closed |
| Stalled partial request (`GET ` + silence) | closed at 20.0 s, 0 bytes sent, access log `"-" 408 0 "-" "-"` |
| Fully silent connection | EOF at ~52 s (≈30 s `TCP_DEFER_ACCEPT` + 20 s), same 408 log line |

helix-nine:
Fixes #3731.

## The bug

The HTTP-aware vhost proxy pins one client connection to one upstream connection. `run_http1_proxy` ended in `futures::future::try_join(from, to)`, which short-circuits only on `Err`. A backend closing its idle keep-alive leg is a clean FIN, which hyper reports as **`Ok(())`** — so `try_join` parked the upstream side and went on driving the client side for up to `header_read_timeout` (60s).

The client was then holding a connection it had every reason to believe was healthy. The next request on it reached `send_request` against a dead connection, the error came back out of the `service_fn`, and hyper tore the connection down **without writing a response** (`Server::on_error` declines anything that isn't a parse error). The client sees a bare remote close.

Apache's default `KeepAliveTimeout` is 5s and Nextcloud Desktop polls `status.php` every 30s, so this landed on essentially every idle cycle — Connected → "Network error" → recover, with nothing in the service to explain it.

The raw-splice path has never had this: `copy_bidirectional` propagates the FIN, and client connection pools recover from an EOF invisibly. This makes the HTTP-aware path behave the same way, which is the point — the proxy should be indistinguishable from the backend apart from the headers it deliberately mutates.

## The fix

Replace the bare `try_join` with a select loop that calls `graceful_shutdown()` on the client connection once the upstream one resolves. Same change on the h2 path, which has the identical shape at `http.rs:311` and is in fact worse: its client leg has no `header_read_timeout`, so the orphan was permanent rather than 60s.

One rule carries this, and it took two rounds of review to find it: **never disable keep-alive while a response is still on its way out.** `disable_keep_alive` (`proto/h1/conn.rs`) closes the connection outright *only* when it is idle; mid-response it sets `KA::Disabled` instead, and `enforce_version` then `insert`s `Connection: close` into the head as it is written. That is a header the backend never sent, and on a 101 it lands on top of `Connection: Upgrade`, which conformant WebSocket clients reject (RFC 6455 §4.1) even though the tunnel still carries bytes.

So the shutdown is deferred until the relayed response has actually been written, tracked by a guard that the response body holds — it drops once hyper is done with that head and body, at which point the connection is idle and `disable_keep_alive` takes its clean-close branch. Measured, both failure modes are real: making the shutdown unconditional produces

```
http/1.1 200 ok / content-length: 2 / connection: close / date: ...
http/1.1 101 switching protocols / connection: close / upgrade: websocket / ...
```

The second thing worth knowing: **re-polling a resolved hyper connection is a contract violation, and the client one panics outright** — `UpgradeableConnection::poll` does `self.inner.as_mut().unwrap()` on an `Option` its upgrade path `take`s, unlike the server's, which returns `Ok(())` on `None`. `.fuse()` on the upstream leg handles that.

## Tests

Two duplex-driven tests in `net::http::tests`, both verified to fail without the change:

| Test | Reverted to | Result |
| --- | --- | --- |
| `upstream_close_is_relayed_to_the_client` | the original `try_join` | fails — 5.00s timeout waiting for an EOF that never comes |
| `relayed_101_keeps_its_upgrade_header` | latch removed (unconditional shutdown) | fails — `connection: close` in the relayed 101 |

`cargo test -p start-core --lib` is green (598 passed).

## What this does not change

- **The eager dial stays.** A client that opens a socket and sends nothing still opens a backend connection at ClientHello, so the paired `408`s in #3731 remain. That is the transparent behaviour — a client connecting directly would produce the same entry — and removing it would mean the proxy inventing a connection lifecycle the backend would not otherwise see.
- **The in-flight race remains**, and should: a request already on the wire when the FIN lands still fails, exactly as it would against the backend directly. The exposure window goes from up to 60s to the time it takes to relay a close.
- Idle client connections now live as long as the backend's keep-alive rather than up to 60s. That is the same coupling the splice path already has. Worth noting separately that the vhost builds a fresh rustls `ServerConfig` per ClientHello, whose session cache and ticketer are per-connection, so none of those reconnects can resume — caching the per-SNI config is a real follow-up, but it is not a regression introduced here.

## Noted while in here, not fixed

The ALPN reflection at `vhost.rs:1153-1155` is a no-op and always has been. `tokio-rustls`'s `connect_with` closure runs *before* the handshake, so `conn.alpn_protocol()` is unconditionally `None` there and the `extend` appends nothing. Verified against the pinned `tokio-rustls` 0.26.4 / `rustls` 0.23: with the upstream negotiating `h2`, the client-facing config still ends up with an empty ALPN list. Introduced in #3037, untouched since. Consequence worth a separate look: on the `connect_ssl: Ok` branch (the `https`/`wss` presets) we advertise no ALPN to the client while forwarding the client's whole offer upstream.

### DIFF
--- a/shared-libs/crates/start-core/src/net/http.rs
+++ b/shared-libs/crates/start-core/src/net/http.rs
@@ -519,6 +519,50 @@ mod tests {
         assert_eq!(eof.unwrap(), 0);
     }
 
+    /// The proxy relays the backend's own 101 rather than answering the
+    /// upgrade itself, so the client hears nothing until the backend has
+    /// spoken.
+    #[tokio::test]
+    async fn upgrade_is_not_answered_before_the_backend_answers() {
+        let (mut client, client_facing) = tokio::io::duplex(4096);
+        let (backend_facing, mut backend) = tokio::io::duplex(4096);
+
+        tokio::spawn(run_http1_proxy(
+            client_facing,
+            backend_facing,
+            None,
+            false,
+            None,
+        ));
+        tokio::spawn(async move {
+            read_head(&mut backend).await;
+            tokio::time::sleep(Duration::from_millis(200)).await;
+            backend
+                .write_all(
+                    b"HTTP/1.1 101 Switching Protocols\r\n\
+                      Connection: Upgrade\r\nUpgrade: websocket\r\n\r\n",
+                )
+                .await
+                .unwrap();
+            std::future::pending::<()>().await;
+        });
+
+        client
+            .write_all(
+                b"GET /ws HTTP/1.1\r\nHost: x\r\n\
+                  Connection: Upgrade\r\nUpgrade: websocket\r\n\r\n",
+            )
+            .await
+            .unwrap();
+        assert!(
+            tokio::time::timeout(Duration::from_millis(100), client.read(&mut [0u8; 1]))
+                .await
+                .is_err(),
+            "the client was answered before the backend had responded",
+        );
+        assert!(read_head(&mut client).await.starts_with("http/1.1 101"));
+    }
+
     /// Shutting the client leg down while a 101 is in flight makes hyper
     /// replace `Connection: Upgrade` with `Connection: close`, which
     /// conformant WebSocket clients reject (RFC 6455 §4.1). The tunnel still
--- a/shared-libs/crates/start-core/src/net/http.rs
+++ b/shared-libs/crates/start-core/src/net/http.rs
@@ -6,6 +6,7 @@ use std::sync::atomic::{AtomicBool, Ordering};
 use std::time::Duration;
 
 use bytes::Bytes;
+use futures::FutureExt;
 use http::{HeaderMap, HeaderValue};
 use http_body_util::combinators::BoxBody;
 use http_body_util::{BodyExt, Full};
@@ -309,16 +310,12 @@ where
             }),
         );
     let mut from = Box::pin(from);
-    let mut to = Box::pin(to);
-    let mut upstream_done = false;
+    let mut to = Box::pin(to.fuse());
     loop {
         tokio::select! {
             res = from.as_mut() => return Ok(res?),
-            // Re-polling a hyper connection after it has resolved is a contract
-            // violation, so this arm has to disarm itself.
-            res = to.as_mut(), if !upstream_done => {
+            res = to.as_mut() => {
                 res?;
-                upstream_done = true;
                 // GOAWAY plus drain, so in-flight streams still finish.
                 from.as_mut().graceful_shutdown();
             }
@@ -439,17 +436,12 @@ where
             }),
         );
     let mut from = Box::pin(from.with_upgrades());
-    let mut to = Box::pin(to.with_upgrades());
-    let mut upstream_done = false;
+    let mut to = Box::pin(to.with_upgrades().fuse());
     loop {
         tokio::select! {
             res = from.as_mut() => return Ok(res?),
-            // Re-polling a hyper connection after it has resolved is a contract
-            // violation, and the client one panics outright, so this arm has to
-            // disarm itself.
-            res = to.as_mut(), if !upstream_done => {
+            res = to.as_mut() => {
                 res?;
-                upstream_done = true;
                 // The backend is gone. Hand the client the EOF now, the way the
                 // splice path does, instead of leaving it a connection that
                 // looks reusable and aborts the next request it carries.
--- a/shared-libs/crates/start-core/src/net/http.rs
+++ b/shared-libs/crates/start-core/src/net/http.rs
@@ -2,10 +2,10 @@ use std::collections::HashMap;
 use std::convert::Infallible;
 use std::net::IpAddr;
 use std::sync::Arc;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::time::Duration;
 
 use bytes::Bytes;
-use futures::FutureExt;
 use http::{HeaderMap, HeaderValue};
 use http_body_util::combinators::BoxBody;
 use http_body_util::{BodyExt, Full};
@@ -308,9 +308,22 @@ where
                 }
             }),
         );
-    futures::future::try_join(from.boxed(), to.boxed()).await?;
-
-    Ok(())
+    let mut from = Box::pin(from);
+    let mut to = Box::pin(to);
+    let mut upstream_done = false;
+    loop {
+        tokio::select! {
+            res = from.as_mut() => return Ok(res?),
+            // Re-polling a hyper connection after it has resolved is a contract
+            // violation, so this arm has to disarm itself.
+            res = to.as_mut(), if !upstream_done => {
+                res?;
+                upstream_done = true;
+                // GOAWAY plus drain, so in-flight streams still finish.
+                from.as_mut().graceful_shutdown();
+            }
+        }
+    }
 }
 
 pub async fn run_http1_proxy<F, T>(
@@ -330,6 +343,11 @@ where
         .handshake(TokioIo::new(to))
         .await?;
     let client = Arc::new(Mutex::new(client));
+    // Set while a 101 is being relayed. Shutting the client leg down in that
+    // window makes hyper replace the upstream's `Connection: Upgrade` with
+    // `Connection: close`, which conformant WebSocket clients reject.
+    let upgrading = Arc::new(AtomicBool::new(false));
+    let svc_upgrading = upgrading.clone();
     // hyper disarms `header_read_timeout` while a body or upgrade is in
     // flight, so this can't kill an active stream — only idle keep-alive.
     let from = hyper::server::conn::http1::Builder::new()
@@ -340,6 +358,7 @@ where
             service_fn(move |mut req| {
                 let client = client.clone();
                 let gate = gate.clone();
+                let upgrading = svc_upgrading.clone();
                 async move {
                     if let Err(resp) =
                         apply_request_policy(&mut req, src_ip, add_forwarded, gate.as_ref())
@@ -358,6 +377,10 @@ where
                                     .any(|s| s.trim().eq_ignore_ascii_case("upgrade"))
                             })
                         {
+                            // Must be set before `send_request`: the upstream
+                            // connection resolves in the same poll that delivers
+                            // the 101, before this future is resumed.
+                            upgrading.store(true, Ordering::Relaxed);
                             Some(hyper::upgrade::on(&mut req))
                         } else {
                             None
@@ -365,8 +388,14 @@ where
 
                     let mut res = match client.lock().await.send_request(req).await {
                         Ok(r) => r,
-                        Err(e) => return Err(e),
+                        Err(e) => {
+                            upgrading.store(false, Ordering::Relaxed);
+                            return Err(e);
+                        }
                     };
+                    if upgrade.is_some() && res.status() != http::StatusCode::SWITCHING_PROTOCOLS {
+                        upgrading.store(false, Ordering::Relaxed);
+                    }
 
                     if let Some(from) = upgrade {
                         let kind = res
@@ -409,9 +438,27 @@ where
                 }
             }),
         );
-    futures::future::try_join(from.with_upgrades().boxed(), to.with_upgrades().boxed()).await?;
-
-    Ok(())
+    let mut from = Box::pin(from.with_upgrades());
+    let mut to = Box::pin(to.with_upgrades());
+    let mut upstream_done = false;
+    loop {
+        tokio::select! {
+            res = from.as_mut() => return Ok(res?),
+            // Re-polling a hyper connection after it has resolved is a contract
+            // violation, and the client one panics outright, so this arm has to
+            // disarm itself.
+            res = to.as_mut(), if !upstream_done => {
+                res?;
+                upstream_done = true;
+                // The backend is gone. Hand the client the EOF now, the way the
+                // splice path does, instead of leaving it a connection that
+                // looks reusable and aborts the next request it carries.
+                if !upgrading.load(Ordering::Relaxed) {
+                    from.as_mut().graceful_shutdown();
+                }
+            }
+        }
+    }
 }
 
 // Silence unused-import lints that may show up depending on feature flags.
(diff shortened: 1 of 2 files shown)


shared-libs/crates/start-core/src/net/http.rs
```diff
@@ -308,9 +308,22 @@ where
...
-    futures::future::try_join(from.boxed(), to.boxed()).await?;
-
-    Ok(())
+    let mut from = Box::pin(from);
+    let mut to = Box::pin(to);
+    let mut upstream_done = false;
+    loop {
+        tokio::select! {
+            res = from.as_mut() => return Ok(res?),
+            // Re-polling a hyper connection after it has resolved is a contract
```
dr-bonez: isn't that the kind of thing `.fuse()` was created to solve?

shared-libs/crates/start-core/src/net/http.rs
```diff
@@ -330,6 +343,11 @@ where
         .handshake(TokioIo::new(to))
         .await?;
     let client = Arc::new(Mutex::new(client));
+    // Set while a 101 is being relayed. Shutting the client leg down in that
+    // window makes hyper replace the upstream's `Connection: Upgrade` with
+    // `Connection: close`, which conformant WebSocket clients reject.
+    let upgrading = Arc::new(AtomicBool::new(false));
```
dr-bonez: why are we doing this? if the client sent `Connection: Upgrade` then immediately closed, that's exactly what the proxy should also do.

shared-libs/crates/start-core/src/net/http.rs
helix-nine: Yes — swapped both loops to `.fuse()`, pushed in 7afaf714. Drops the flag and the precondition on the arm.

Worth recording that it isn't only tidier: I checked what the guard was actually preventing by removing it, and the client-side connection doesn't just misbehave, it panics —

```
thread '...' panicked at hyper-1.9.0/src/client/conn/http1.rs:600:60:
called `Option::unwrap()` on a `None` value
```

`UpgradeableConnection::poll` does `self.inner.as_mut().unwrap()` on an `Option` its own upgrade path `take`s. (The server side handles `None` and returns `Ok(())`, which is why only the upstream leg needs fusing.) `relayed_101_keeps_its_upgrade_header` covers it — that test re-polls the upstream leg after it resolves, so it fails without the `.fuse()`.

shared-libs/crates/start-core/src/net/http.rs
helix-nine: Agreed on the principle — but this latch isn't about the client closing. It's guarding a case where the upstream connection resolving does **not** mean anything closed.

On a 101, hyper's client `Connection` future resolves `Ok(())` — that is how it hands the upgraded IO to `hyper::upgrade::on`. `Dispatched::Shutdown` and `Dispatched::Upgrade` both come out as `Poll::Ready(Ok(()))`, so from the select loop a successful upgrade is indistinguishable from the backend closing its idle leg. And on an upgrade the upstream side resolves *before* the client side, so without the latch we call `graceful_shutdown` on the client leg while the 101 is still being written.

What that does isn't "close a bit early". `graceful_shutdown` → `disable_keep_alive` → `enforce_version`, which does an `insert` of `Connection: close` — replacing the `Connection: Upgrade` in the 101 we're relaying. Measured with the latch removed:

```
HTTP/1.1 101 Switching Protocols
connection: close
upgrade: websocket
```

The tunnel still carries bytes afterwards, so it looks like it works. But RFC 6455 §4.1 requires a conformant client to fail the connection when the 101's `Connection` header has no `Upgrade` token. So the naive version has us rewriting a header on a successful handshake, which is exactly the kind of mutation we're not supposed to be making.

To your actual scenario — a real close on an upgrade connection is still propagated, just not by this branch:

- **Before the 101**, the backend closing makes the upstream future resolve `Err` (incomplete message), and `res?` returns before the latch is ever consulted. Client leg goes down with it.
- **After the 101**, the connection is a tunnel driven by `copy_bidirectional` — the same primitive as the splice path, propagating closes in both directions.
- A request that sends `Connection: Upgrade` and gets back anything other than a 101 clears the latch, so it stays on the normal shutdown path.

One residual I should be upfront about: if the backend closes in the window between us writing the request and the response arriving, on a connection that had asked for an upgrade, the latch is set and we skip the shutdown — that connection falls back to the old 60s `header_read_timeout` behaviour. It errs toward a missed shutdown rather than a corrupted upgrade, which is the right direction, but it isn't nothing. Closing it properly needs the service to signal the loop once it knows the status (a `Notify`, or moving the latch to a tri-state). I judged that more machinery than the case warrants — happy to add it if you'd rather have it exact.

`relayed_101_keeps_its_upgrade_header` is the regression test; it fails with the header rewritten if the latch goes.

dr-bonez: this seems to be the wrong way to handle upgrades, no? we shouldn't respond to the client requesting the upgrade until the downstream connection has responded.

helix-nine: We already don't — and I'd read too much into `hyper::upgrade::on(&mut req)` if I were reviewing this, so that's a fair thing to challenge. It doesn't accept or answer the upgrade; it just takes the `OnUpgrade` extension out of the request so we can collect the upgraded IO later. It writes no bytes. The only response the client ever sees is `res`, i.e. whatever the backend returned, relayed — we never synthesize a 101.

Pushed `3edeb581` so that's enforced rather than asserted: the backend holds its 101 back for 200ms, and the client must hear nothing at all in the meantime.

```rust
assert!(
    tokio::time::timeout(Duration::from_millis(100), client.read(&mut [0u8; 1]))
        .await
        .is_err(),
    "the client was answered before the backend had responded",
);
assert!(read_head(&mut client).await.starts_with("http/1.1 101"));
```

So the ordering you're describing is the ordering we have. The thing the latch guards is one step earlier and isn't about responding at all: **the upstream connection future resolves before our service future is resumed with the response.** The 101 arriving from the backend both resolves that future (`Dispatched::Upgrade`) and completes the `send_request` oneshot, but the service only wakes up on a later poll, and the head is written after that. So the select loop can observe "upstream connection finished" while the response is still in our hands, unwritten. That's the window — not us jumping ahead of the backend.

One thing I should have said in the earlier reply, because it makes the latch much narrower than it looks: the `Connection: close` rewrite is *correct* for every response except a 101. If the backend has gone away, stamping `Connection: close` on a relayed 200 is exactly the right thing to tell the client. It's only wrong on a 101, where that header field is load-bearing for the handshake itself. So the latch isn't "suppress shutdown during upgrades", it's "don't let the close stamp land on the one status where it corrupts the message" — which is also why clearing it on any non-101 response costs nothing.

If you'd still rather this were structural than a flag, the honest options are (a) defer the shutdown until the service has returned — doesn't work, the head is written after that, so the rewrite still lands; or (b) have hyper distinguish upgrade-handoff from shutdown in the connection future's output, which is an upstream change. I couldn't find a third. Happy to be wrong about that.

dr-bonez: I don't think we should be doing a "Connection: close" rewrite at all. if the downstream connection closed without responding with "Connection: close" already, that's exactly what we should do. we should behave exactly the same way the backend does *with the exception of the specific things we intentionally mutate* (i.e. auth or x-forwarded headers)

helix-nine: You're right, and I was wrong to defend it — I asserted that rewrite was correct-except-on-101 without checking whether it was even reachable. It is. Probe against a backend that answers and closes in the same breath, on the code as it stood:

```
http/1.1 200 ok
content-length: 2
connection: close        <- the backend never sent this
date: mon, 17 aug 2026 20:47:29 gmt
```

Fixed in `ab55ba93`, and finding the mechanism made the whole thing smaller rather than bigger. The relevant hyper code is `Conn::disable_keep_alive`:

```rust
if self.state.is_idle() {
    trace!("disable_keep_alive; closing idle connection");
    self.state.close();
} else {
    trace!("disable_keep_alive; in-progress connection");
    self.state.disable_keep_alive();   // -> KA::Disabled -> enforce_version inserts Connection: close
}
```

So the header only ever appears because we called it at the wrong moment. The rule is just **don't disable keep-alive while a response is still on its way out** — then hyper always takes the clean-close branch and the reply goes through untouched.

Implemented by deferring the shutdown until the relayed response has actually been written. The signal is a guard the response body holds, so it drops once hyper is finished with that head and body; the loop then shuts down against an idle connection.

That subsumes the upgrade latch, so **the flag you asked about is gone**. A 101 is simply the case where the head is written before the guard drops, and `UpgradeableConnection::graceful_shutdown` is documented as a no-op once upgraded — so there is nothing left to special-case. Net effect on the service_fn is a deletion:

```diff
-                            upgrading.store(true, Ordering::Relaxed);
                             Some(hyper::upgrade::on(&mut req))
...
-                        Err(e) => {
-                            upgrading.store(false, Ordering::Relaxed);
-                            return Err(e);
-                        }
+                        Err(e) => return Err(e),
                     };
-                    if upgrade.is_some() && res.status() != http::StatusCode::SWITCHING_PROTOCOLS {
-                        upgrading.store(false, Ordering::Relaxed);
-                    }
```

Both failure modes are pinned by tests, and I checked they bite: making the shutdown unconditional again fails `a_relayed_response_is_not_rewritten_when_the_backend_closes` *and* `relayed_101_keeps_its_upgrade_header`, while the other 8 stay green. 600 tests pass. PR description updated to describe this rather than the latch.

dr-bonez (approved):

Issue #3717, opened 2026-08-15
[bug]: domains bound to non-HTTP ports serve the local Root CA chain instead of the Let's Encrypt certificate the UI claims


stupleb:
When a domain with Let's Encrypt selected as its certificate authority is bound
to a non-HTTP port, the port serves the device's local Root CA chain instead.
The UI continues to show Let's Encrypt; there is no error. Standard web
services on sibling subdomains of the same domain receive real LE certificates.

This is not inherent to non-HTTP ports: ACME certificates are issued per name,
not per port, and serving one on an arbitrary port is standard practice. The
constraint is validation — StartOS implements only the TLS-ALPN-01 challenge,
which Let's Encrypt dials exclusively at `<name>:443` — and the platform owns
that path in both observed configurations. The failure is challenge routing,
not an ACME limitation.

## Observations

1. **Electrum port over StartTunnel** (external support case, 2026-08): a
   Fulcrum interface routed through a StartTunnel domain, CA set to Let's
   Encrypt and shown as such in the UI. Bitkey support investigated the
   endpoint independently: publicly reachable, TLS on 50002, "presenting a
   certificate issued by a private Start9/StartOS local CA, rather than a
   publicly trusted certificate chain." Bitkey validates against public roots
   and has no trust-on-first-use flow, so the pairing is hard-blocked.
2. **TURN port on a directly-routed custom domain** (2026-08-10, still
   reproducible 2026-08-15): `turns://turn.<domain>:5349` (coturn) serves
   `CN=StartOS Local Intermediate CA` while the UI shows Let's Encrypt.
   crt.sh has zero CT entries for that subdomain — issuance was never
   completed — while six sibling subdomains hold current LE certificates,
   proving 443 reaches this device. Not StartTunnel-routed, so the bug is not
   ST-specific.

## Mechanism (verified in code)

1. The CA selector offers LE for the domain (#3462), and
   `net_controller.rs` (~518) correctly attaches the domain's ACME provider to
   the vhost target of an `addSsl` binding on any port (`acme: None` is
   reserved for passthroughs).
2. `AcmeTlsHandler::get_cert` (`net/acme.rs` ~90) orders on demand when no
   healthy cached certificate exists, honoring Retry-After backoff (#3194).
3. Validation then requires answering `acme-tls/1` at `<name>:443`. The vhost
   controller runs one server per external port (`net/vhost.rs` ~261), each
   with a by-name mapping that has no fallback (~1466: "a host answers to the
[...]

PR #3734, opened 2026-08-17
fix: ACME domains on non-443 ports get a certificate, and never fall back to the local CA

helix-nine:
Closes part of #3717 — the two fixes you specified, plus the companions each of them needs and the follow-ups from review.

## What was happening

A domain's ACME provider is attached to the vhost target for whatever SSL port its binding got (`net_controller.rs:518`), and a package can never hold 443 (`may_claim` in `forward.rs`). Each `VHostServer` answers by name for its own port only, so a TLS-ALPN-01 probe to `<name>:443` reached nothing — and the resolution chain then fell through to `RootCaTlsHandler`, which mints a leaf for any name it is asked for.

## 1. An ACME domain with no certificate fails the connection

The vhost now holds the ACME and root-CA handlers itself instead of chaining them, and decides from the target the connection actually routed to: an address with an authority is served that authority's certificate or nothing; an address without one falls back to the local root CA as any LAN address does. Declining *is* the refusal — nothing is chained below — so `ChainedHandler` and the reject action an earlier draft needed are gone.

**The alert was never reaching the wire.** On the refusal path `mid.io` is still `BTBuffer::Buffering`, so `write_all` of the fatal `unrecognized_name` went into an in-memory buffer and the client saw a bare TCP close. `stop_buffering()` first. Already true on master; refusing puts far more connections through that path, so it stops being cosmetic.

**A failing renewal must not take the domain down.** `should_use_cert` turns false 30 days before expiry so a renewal is attempted; that certificate is still publicly trusted. Without this, fail-closed would turn every failing renewal into a 30-day outage with a valid certificate sitting in the DB. The cached certificate is now the answer of last resort until it actually expires — including when the provider was removed from the server's settings while a domain still names it.

## 2. A domain on another port also asks for the challenge port

`reconcile_port_maps`' desired map carried **one** internal port per `(box IP, external port)` shared by every hostname on it, so it could not express `443/SNI=a → 50002` and `443/SNI=b → 5349` at once; two ACME domains would have collapsed onto whichever was reconciled first. It now carries a box-side port per hostname, and an ACME target on a port other than 443 additionally asks for `443/SNI=<name> → <its port>`. A real 443 binding keeps that route, in any visit order.

**No relay change.** `port_map/server/mod.rs:397` already sets `target = SocketAddrV4::new(peer, internal_port)` independently of the external port, `SniRoute.target` persists an arbitrary port, and the relay's own UI is not on 443. Verified through `SniDemux::register`.

The derivation is now the free function `desired_port_maps`, mirroring `desired_listeners` — which is what makes it testable.

## 3. Answering a challenge for a name a server does not route

Not one of your two, and separable. Your fix (2) reaches a gateway that speaks PCP with the Start9 `HOSTNAME` extension — StartTunnel. The other case in the issue (coturn on `5349`, directly routed, siblings holding real LE certs so 443 demonstrably reaches the box) has no such gateway: the challenge arrives at the box's own 443 listener, which does not route that name.

So an `acme-tls/1` ClientHello for an SNI a server does not route is answered from the controller-wide cache of orders in flight. Nothing else populates that cache, so this still refuses every name we are not ourselves ordering for.

## 4. Split DNS keeps its local side (your call)

A name reachable both ways now gets **one vhost entry per side**, and only the public one carries an authority. The two are told apart by the source address — IPv4 hands a forwarded WAN dial and a local one the same arrival gateway, so a target that claims a connection as one of our own private addresses wins over one that claims it on the gateway alone (`filter_private`, preferred in the routing lookup; this is what stops the choice depending on insertion order). The local side is served the server's own certificate, or the authority's where one has been issued — the same certificate either way, as `private-domains.md` promises.

`GetVHostAcmeProvider` had to move with it: it took the *first* alive target for a name, which with two legs would have found the private one and never ordered a certificate at all. It now takes the provider from whichever leg carries one.

## 5. The refusal says why (your call)

The first failure to obtain a certificate for a name raises a `Warning` notification naming the domain and the reason. Every retry while it stands is silent, and the per-order entry is dropped once a certificate lands, so the next failure after a success speaks again. Wired as a callback from `VHostController` (which holds the concrete DB) so `AcmeTlsHandler` stays generic. New keys in all five locales.

## 6. No authority where the service holds the certificate (your call)

The Certificate Authority column read the domain's stored config for every address, including one whose container serves its own TLS and which StartOS never terminates — so a passthrough address claimed an authority that has nothing to do with the certificate a client actually checks. It now reports from the binding: `Self signed`. And `add-public-domain` refuses an authority for a host with no binding StartOS terminates, which covers the CLI as well; the form already hid the selector where no binding on the host is ours.

## Trade-offs still worth stating

- **Refusal is a fatal alert, not a degraded certificate.** During first boot, DNS propagation, the 60s failure backoff and up to a 24h `Retry-After`, a public ACME address is *down*, not degraded. Its LAN side (4) and the notification (5) are what keep that from being silent or total.
- **Remote admin lockout is still possible.** An admin who put a Let's Encrypt domain on the admin host, disabled the bare public IP, and has no private domain for it loses their only remote entry point until issuance succeeds. LAN and `.local` are unaffected.
- **No IPv6 challenge route.** A GUA belongs to the gateway and every host on the box shares it, so claiming `[gua]:443` for one domain would claim it for all of them. A dual-stack domain validates over IPv4; one public only over IPv6 has no path for its challenge.
- **The add-domain checks still test the interface's own port**, not 443, so they can go all-green on a domain that then refuses. Documented in `clearnet.md`; probing 443 for an ACME domain would change the RPC result shape and the modal, so I left it out. Worth keeping #3717 open for that.

## Testing

`cargo test --features=test --lib` — 605 pass; `check:ui` clean; `cargo check -p startwrt-core` clean (the `TlsHandlerAction` variant it constructs is untouched). New:

- `tls.rs::a_refused_connection_gets_a_tls_alert` — real loopback handshake through a `TlsListener`; asserts the client sees `AlertReceived(UnrecognisedName)` rather than a bare close (this pins the `stop_buffering` fix), and that a handler serving a certificate still completes.
- `vhost.rs::port_map_tests` — five cases over `desired_port_maps`, including two ACME domains keeping distinct box-side ports at 443, a real 443 binding winning, and nothing else claiming 443.
- `vhost.rs::a_lan_dial_of_a_dual_exposure_domain_belongs_to_the_private_leg` — pins the precedence in (4), including the fact that the public leg *also* matches a LAN dial, which is why precedence is needed.
- `vhost.rs::a_challenge_without_a_name_is_not_a_challenge` — a dial with no SNI is not a challenge. An adversarial review pass caught that without this, (3) handed the Root CA chain to `openssl s_client -alpn acme-tls/1 -noservername` against an unrouted port: the ACME handler bails on the missing name *before* it can refuse, and the chain then reached `RootCaTlsHandler` — the exact leak this PR is about.
- `ssl.rs::the_renewal_window_is_unexpired_but_due_for_renewal` — the 30-day window is unexpired but due.

Not covered in-repo: end-to-end issuance (needs public DNS + a reachable relay) and the PCP `443 → 50002` handshake against a live `tunnelbox`. I can run the latter on a VM pair if you want it before merge.

### DIFF
--- a/shared-libs/crates/start-core/src/net/host/address.rs
+++ b/shared-libs/crates/start-core/src/net/host/address.rs
@@ -408,6 +408,20 @@ pub async fn add_public_domain<Kind: HostApiKind>(
                 {
                     return Err(Error::new(eyre!("unknown acme provider {}, please run acme.init for this provider first", acme.0), ErrorKind::InvalidRequest));
                 }
+                // A host that serves its own TLS holds its own certificate:
+                // StartOS never terminates the connection, so it can neither
+                // obtain one from an authority nor present it.
+                if !Kind::host_for(&inheritance, db)?
+                    .as_bindings()
+                    .de()?
+                    .values()
+                    .any(|b| b.options.add_ssl.is_some())
+                {
+                    return Err(Error::new(
+                        eyre!("{}", t!("acme.passthrough-host", domain = fqdn)),
+                        ErrorKind::InvalidRequest,
+                    ));
+                }
             }
 
             // Adding a domain that is already present is a no-op for exposure:
--- a/shared-libs/crates/start-core/src/net/host/address.rs
+++ b/shared-libs/crates/start-core/src/net/host/address.rs
@@ -408,20 +408,6 @@ pub async fn add_public_domain<Kind: HostApiKind>(
                 {
                     return Err(Error::new(eyre!("unknown acme provider {}, please run acme.init for this provider first", acme.0), ErrorKind::InvalidRequest));
                 }
-                // A host that serves its own TLS holds its own certificate:
-                // StartOS never terminates the connection, so it can neither
-                // obtain one from an authority nor present it.
-                if !Kind::host_for(&inheritance, db)?
-                    .as_bindings()
-                    .de()?
-                    .values()
-                    .any(|b| b.options.add_ssl.is_some())
-                {
-                    return Err(Error::new(
-                        eyre!("{}", t!("acme.passthrough-host", domain = fqdn)),
-                        ErrorKind::InvalidRequest,
-                    ));
-                }
             }
 
             // Adding a domain that is already present is a no-op for exposure:
--- a/projects/start-os/CHANGELOG.md
+++ b/projects/start-os/CHANGELOG.md
@@ -55,9 +55,8 @@ file tracks notable changes since the move to the monorepo.
 - **A service that presents its own TLS certificate is shown as presenting
   it.** StartOS does not terminate such a connection, so the certificate the
   client checks is the service's — but a domain on that interface reported
-  whichever certificate authority had been chosen for it, and the choice was
-  offered in the first place. Those addresses now report `Self signed`, and
-  choosing an authority for one is refused with an explanation.
+  whichever certificate authority had been chosen for it. Those addresses now
+  report `Self signed`.
 
 - **Services that run their own containers or a VPN can reach the devices they
   were granted.** A service opting into `userspaceFilesystems` or
--- a/projects/start-os/CHANGELOG.md
+++ b/projects/start-os/CHANGELOG.md
@@ -52,6 +52,13 @@ file tracks notable changes since the move to the monorepo.
   `443` to your server as you would for a standard domain — the domain's own
   port is not enough on its own.
 
+- **A service that presents its own TLS certificate is shown as presenting
+  it.** StartOS does not terminate such a connection, so the certificate the
+  client checks is the service's — but a domain on that interface reported
+  whichever certificate authority had been chosen for it, and the choice was
+  offered in the first place. Those addresses now report `Self signed`, and
+  choosing an authority for one is refused with an explanation.
+
 - **Services that run their own containers or a VPN can reach the devices they
   were granted.** A service opting into `userspaceFilesystems` or
   `virtualNetworking` is handed `/dev/fuse` and `/dev/net/tun` inside its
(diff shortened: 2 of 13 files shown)


shared-libs/crates/start-core/src/net/host/address.rs
```diff
@@ -408,6 +408,20 @@ pub async fn add_public_domain<Kind: HostApiKind>(
...
+                    .as_bindings()
+                    .de()?
+                    .values()
+                    .any(|b| b.options.add_ssl.is_some())
+                {
+                    return Err(Error::new(
+                        eyre!("{}", t!("acme.passthrough-host", domain = fqdn)),
+                        ErrorKind::InvalidRequest,
+                    ));
+                }
```
dr-bonez: not necessary - this implies this is more of an invariant than it is. we enforce this on the frontend for UX reasons, but a domain could theoretically change from `addSsl` to passthrough, or vice versa, and we would retain the ACME config. this is fine.

dr-bonez (changes requested):
the comments need to be significantly simplified. otherwise looks good

shared-libs/crates/start-core/src/net/host/address.rs
helix-nine: Removed — you're right, and it was the wrong shape of guard: bindings can move between `addSsl` and passthrough across an update, so a stored authority outliving that is expected rather than a violation.

Kept the display side: the certificate column reads from the binding now, so a passthrough address reports `Self signed` instead of whichever authority happens to be stored.

dr-bonez (dismissed):

dr-bonez (approved):

PR #3736, opened 2026-08-17
fix(sdk): reject manifest fields the SDK does not define

MattDHill:
## The problem

`setupManifest` infers its parameter type from the argument, and TypeScript applies its excess-property check only when a fresh object literal meets a **non-generic** target. So an unknown field joined the inferred `Manifest`, compiled cleanly, and was then dropped on the way to the built manifest — it never reached the registry, and nothing ever said so.

That is how `docsUrls` came to sit in twelve packages across both registries doing nothing (now removed), and how `alerts` outlived its removal from the SDK in the packages that still declare it.

## The fix

One mapped type. Keys in `SDKManifest` keep their type; everything else is retyped, so the compiler names it:

```
error TS2322: Type 'string[]' is not assignable to type
  'string[] & "Unknown manifest field — not declared in SDKManifest"'.

24  export const bad = setupManifest({ ...base, docsUrls: ['https://x'] })
                                                ^
```

The replacement is a string literal rather than `never` purely for that message — `not assignable to type 'never'` is correct and tells a packager nothing. It does mean the literal itself would be accepted as a value, which seemed a fair trade.

`SDKManifest` has no index signature, so the check is exact. Optional declared fields (`plugins`, `osVersion`, `hardwareRequirements`, …) are unaffected.

## Verified

Typechecked against fourteen real packages with this SDK linked in — `hello-world`, `llama-cpp`, `immich`, `nextcloud`, `ollama`, `vllm`, `lnd`, `bitcoin-core/31.x`, `bitcoin-knots/29.x`, `bitcoin-cash-daemon`, `i2pd`, and three under review. Nine pass. Type inference is unchanged: nothing that depends on `Manifest['id']`, `Manifest['volumes'][number]`, or the images/dependencies threading was affected.

Of the five failures, **four are not this change** — they are `TS2554: Expected 1-2 arguments, but got 3`, from the `SubContainer.exec` signature change already landed in the unreleased 2.0.10. They are a migration those packages owe 2.0.10 regardless.

## The one thing worth a decision before merging

The fifth failure is what this change is for, and it shows where the cost lands. `navidrome-startos` fails on **its dependencies'** manifests, not its own:

```
node_modules/maloja-startos/startos/manifest/index.ts(22,3): error TS2322 …
node_modules/multi-scrobbler-startos/startos/manifest/index.ts(22,3): error TS2322 …
```

Both still declare `alerts`. A package's `tsconfig.json` includes `node_modules/**/startos`, so you typecheck your siblings' source — and a sibling that hasn't dropped a dead field breaks *your* build, for a file you don't control.

This isn't new behavior (the `exec` change breaks dependents' source the same way — `lnd` is currently failing on `node_modules/bitcoin-core-startos/…`), but it does mean this wants sequencing:

- **`alerts` should be swept first.** Two packages in our registries still declare it — `bisq2-startos` and `ppq-private-mode-startos`. Say the word and I'll clear them the same way `docsUrls` went.
- **The sibling repos we don't own** — `maloja-startos`, `multi-scrobbler-startos` — are on contributors' accounts and outside a sweep. They'd need the same edit, or the packages depending on them stay red until they get it.
- Landing this **with 2.0.10** rather than after would fold it into one migration for the packages the `exec` change already breaks, instead of two.

## No changelog entry

Omitted deliberately, per maintainer direction — flagging it so review doesn't read it as an oversight.

### DIFF
--- a/projects/start-sdk/lib/manifest/setupManifest.ts
+++ b/projects/start-sdk/lib/manifest/setupManifest.ts
@@ -9,6 +9,23 @@ import { VersionGraph } from '../version/VersionGraph'
 import { version as sdkVersion } from '../../package.json'
 
 /**
+ * Retypes every key the SDK does not define, so declaring one is a compile
+ * error at that key rather than a field that silently does nothing. The
+ * replacement is a string literal rather than `never` purely so the error names
+ * what went wrong.
+ *
+ * TypeScript applies its excess-property check only when a fresh object literal
+ * meets a non-generic target, and `Manifest` is inferred from the argument — so
+ * without this an unknown field joins the inferred type, typechecks, and is
+ * then dropped on the way to the built manifest.
+ */
+type NoExtraKeys<Manifest> = {
+  [K in keyof Manifest]: K extends keyof SDKManifest
+    ? Manifest[K]
+    : 'Unknown manifest field — not declared in SDKManifest'
+}
+
+/**
  * @description Use this function to define critical information about your package
  *
  * @param manifest Static properties of the package
@@ -20,7 +37,7 @@ export function setupManifest<
     id: Id
     volumes: VolumesTypes[]
   } & SDKManifest,
->(manifest: Manifest & SDKManifest): Manifest {
+>(manifest: Manifest & SDKManifest & NoExtraKeys<Manifest>): Manifest {
   return manifest
 }
 


helix-nine (approved):
Approving on the condition drbonez set: **extra keys stay legal at runtime.** They do, and the guard is fully type-erased.

## Runtime, in three layers

1. **The SDK adds nothing.** Compiling head and `origin/master` sources independently produces byte-identical JavaScript for all 35 `.js` files in `dist/` — `setupManifest.js` is `382e1081c48ca6dcf447a82a7afa9dba` on both sides. `diff -rq` over the two dists reports exactly one differing file, `setupManifest.d.ts`. `NoExtraKeys` and the sentinel string appear in zero emitted `.js`. `setupManifest` is still `return manifest`, returning the *same object reference*; no zod, no `.parse`, no key whitelist, no `delete`, no throw.
2. **The TS packaging path carries the key.** `buildManifest` spreads `...manifest`, so an unknown key lands in the built manifest object and in the JSON `start-cli` reads. Executed against both dists with `docsUrls` / `alerts` / `totallyMadeUp`: identical output, keys intact through `JSON.stringify`.
3. **Rust ignores it.** `#[serde(deny_unknown_fields)]` occurs **0 times** across all 527 `.rs` files. The s9pk `Manifest` (`s9pk/v2/manifest.rs:28-31`) and every type on its path — `PackageMetadata`, `PackageVersionInfo`, `PackageIndex`, `HardwareRequirements`, `DeviceFilter`, `Description` — carry only `rename_all = "camelCase"`, so serde's default applies. Verified by round-tripping a manifest with unknown keys through the real struct shape, and by re-adding `deny_unknown_fields` to confirm the permissiveness comes from the attribute's absence rather than by accident. `pack`, `install` and publish add no strict shape check — `validate_and_filter` → `Manifest::validate_for` only enforces description lengths and expected-file presence.

Where the key actually dies, for the record: `start-cli s9pk pack` writes the JS output verbatim, then calls `as_manifest_mut()` unconditionally (`pack.rs:732`), setting `manifest_dirty`, so `serialize()` (`v2/mod.rs:248-259`) rewrites `manifest.json` from the Rust struct and the unknown key is gone. Exit 0, no warning. Confirmed by packing a real s9pk: `docsUrls` present in node's output, absent from the archived `manifest.json`. It does survive inside `javascript.squashfs`, which is a verbatim copy of the ncc bundle — so a package's own in-container code can still read the key even after the OS stops seeing it. Another argument for catching this at compile time.

Escape hatches all still work with the value intact at runtime: `@ts-expect-error`, a cast, or spreading the key on *after* `setupManifest` (`buildManifest({...m, docsUrls: [...]})` typechecks on head and the built type still carries `docsUrls: string[]`).

## Regression risk to valid manifests: measured, none

Base vs head over ~120 legitimate-manifest fixtures across four independent harnesses (minimal, every optional field, multi-arch and `dockerBuild` images, real `dependencies`, multiple volumes, `plugins`, `satisfies`, `as const`, pre-declared variable, spread-from-base, helper function, class instance, getters, `Readonly<>`/`Partial<>` wrappers, `LocaleString` in both forms, `exactOptionalPropertyTypes`, `noUncheckedIndexedAccess`, the real `docs/package-template` manifest) plus the full `StartSdk.of().withManifest().build(true)` → `Mounts` → `SubContainer` → `Backups` chain: base and head agree on every fixture containing no undeclared key. Volume/image id literal unions stay enforced and wrong-id diagnostics are byte-identical.

Widened to the fleet: **131 real manifests** and **115 whole real repo trees** typechecked with each package's own tsconfig — base 0 errors, and every head-only error traces to a genuinely undeclared field. `mempool`'s base and head error sets are identical (the same 6 × TS2554 from 2.0.10's `SubContainer.exec` change), corroborating "four of five failures are not this change." `make check`, `make test` (96/96) and `make check-fmt` are green on head, as is monorepo CI (Prettier, Automated Tests, Formatting & Lockfiles, Generated Artifacts, Dependency Licenses, all three Compile Base Binaries).

## One thing worth amending before merge

The sentinel's stated purpose fails in the most common case. When the undeclared key's value is a **primitive**, `Manifest[K] & 'Unknown manifest field…'` is an intersection of disjoint literals, which reduces to `never`; that empty property collapses the whole object intersection, so every property is checked against `never`:

```
prim.ts(3,3): error TS2322: Type '"demo"' is not assignable to type 'never'.
prim.ts(4,3): error TS2322: Type '"Demo"' is not assignable to type 'never'.
…12 errors, 11 of them on perfectly valid fields, sentinel string appears 0 times
```

This is not hypothetical: `Start9-Community/specter-startos@master` declares top-level `icon: 'icon.png'` (`startos/manifest/index.ts:18`) and is published — `2.1.10:3` on the community registry, `2.1.11:0` on beta. Built with its own tsconfig and its own `typescript`, it goes from 2 errors on base to 14 on head, 12 of them `is not assignable to type 'never'`, none naming `icon`. Arrays and objects (`docsUrls`, `alerts`) give the clean single error, which is why it didn't show up in testing. Not worth attributing to TS 6 — 6.0.3, 5.9.3 and 5.7.3 are identical here, and 5.4.5 collapses too for number/boolean/null.

**The fix is one line: make the sentinel an object type rather than a bare string literal.**

```ts
type NoExtraKeys<Manifest> = {
  [K in keyof Manifest]: K extends keyof SDKManifest
    ? Manifest[K]
    : { UnknownManifestField_NotDeclaredInSDKManifest: never }
}
```

A primitive intersected with an object type is not reduced, so nothing collapses. Measured against the fixture set (`ok` = no error, `n` = error count):

| argument | master | head | with object sentinel |
| --- | --- | --- | --- |
| valid manifest, all optionals | ok | ok | ok |
| `docsUrls: ['…']` | **ok (missed)** | 1, names it | 1, names it |
| `icon: 'icon.png'` | **ok (missed)** | **12, all `never`** | **1, names it** |
| `alerts: { install: '…' }` | **ok (missed)** | 1, names it | 1, names it |
| extra key on a pre-declared const / spread | **ok (missed)** | 2 | 2 |
| unknown key smuggled through a generic wrapper | **ok (missed)** | caught | caught |
| volume / image id literal unions | enforced | enforced | enforced |

Applied to the real source: `tsc --noEmit` clean, jest 96/96, prettier clean. Putting the message in an identifier rather than a quoted string also keeps it readable — with the em dash in a property-name position tsc prints it escaped (`Unknown manifest field — not declared…`).

Caveat, measured: this localizes `string`, `number` and `boolean` extras to one clean error, but a top-level **`null`**-valued extra still collapses to the 12-error `never` form, since `null & {…}` is empty too. That covers every value shape actually present in the fleet (`icon` string, `docsUrls` array, `alerts` object), so it is a real improvement rather than a complete one.

Two things I tried that are **not** the fix, so nobody re-treads them:

- **Dropping the redundant `Manifest &` from the parameter** gives the clean single error and makes the generic-wrapper case compile — but it reopens the hole. `wrap<M extends SDKManifest>(m: M) { return setupManifest(m) }` then accepts a literal carrying `docsUrls` with no diagnostic (measured: 0 errors, vs 1 on head), where head correctly rejects it. Which also means head's TS2345 on a generic wrapper is *not* a false positive: `M extends SDKManifest` permits extra keys, so refusing an unresolved `M` is sound. Worth keeping — though since `NoExtraKeys` isn't exported, an author currently has no way to write a wrapper that satisfies it, which is worth a line of docs if anyone asks.
- **`[K in Exclude<keyof Manifest, keyof SDKManifest>]`** does remove a side effect worth knowing about (below), but leaves the 12-error collapse exactly as it is.

## Smaller things

- **The doc comment is wrong about where the field goes.** "typechecks, and is then dropped on the way to the built manifest" — `buildManifest`, three lines below, spreads `...manifest` and therefore *preserves* it. The drop happens later, when `start-cli` re-serializes at pack time. Since this sentence is the stated rationale, it is worth making precise.
- **Undocumented side effect:** `Manifest` now also appears inside a homomorphic mapped type, which triggers reverse-mapped-type inference and suppresses literal widening — `title`, `license`, `packageRepo`, `upstreamRepo`, `marketingUrl`, `donationUrl`, `osVersion` infer as string literals instead of `string` (`id` already did on master). 130 of 131 real manifests emit a different `.d.ts`, and across the packages A/B'd **zero** new errors appear — no fleet code reads any affected field, and the one pattern that would break (`manifest.title === '…'` → TS2367) exists nowhere. So this is latent rather than breaking; the `Exclude` form above eliminates it (delta drops to 1 file / 2 lines, with identical error positions on all 131) if you'd rather not carry it. Worth a line in the PR body either way.
- **What the guard does not catch**, so the claim stays bounded: an intersection-typed source (`declare const o: SDKManifest & { docsUrls: string[] }; setupManifest(o)`) typechecks clean, because TS infers `Manifest` as the constraint only and the extra member never reaches the mapped type. Same for a union-typed source, an index-signature source, `as any`, a value typed `any`/`never`, and the sentinel written out as a value (the body already concedes this one). Unknown keys nested inside `description` or `hardwareRequirements` are unchanged from master. None of these are regressions, and they are what makes the runtime concern moot in practice — but a package can lose the check just by annotating its manifest variable, with no diagnostic.
- **No regression fixture.** `NoExtraKeys` is the same technique as `ValidateOtherKeys` in `lib/version/VersionInfo.ts`, and that file pins its validator with a `function __type_tests()` block. `tsconfig.json` includes `lib/**/*` and excludes only `*.test.ts`, so such a block is covered by `make check` — I confirmed a `@ts-expect-error` on an undeclared key passes on head and fails `TS2578: Unused '@ts-expect-error' directive` against master, i.e. it would catch a silent removal of this guard. Cheap to add here.
- **Changelog.** Noted that it is deliberate. Flagging once and deferring: this is the kind of change a packager meets as a compile error on an SDK bump, and `## 2.0.10` is unreleased with a `### Changed` section already open, so it is the one place they would look.

## Sequencing

Nothing turns red on merge — every package consumes `@start9labs/start-sdk` from npm at an exact pin behind a committed lockfile (132 of 141 default branches pin `"2.0.9"`), and this lands in unreleased 2.0.10. The gate is the SDK **release** plus the per-package bump wave.

The list in the body is stale rather than wrong: it was accurate when written at 21:07:16Z, and the `bisq2` / `ppq` sweep commits (`beb3102c`, `c4994273`) landed 21:10:53Z, 3m37s later. Both are clean now. Live state across all 141 non-archived `*-startos` repos in both orgs — 13 repos carry an undeclared top-level manifest key, and 4 of them pin SDK 2.0.9, so those are the ones that would newly fail on the bump:

- `Start9-Community/specter-startos@master` — `icon`. The one that hits the collapsed diagnostics, and **the only offender whose release branch is dirty** (precise phrasing matters: `bch-asicseer`, `bch-elopool@next` and `namecoin-core@0.4.x` also carry dead fields and are published packages, but none of those *branches* publishes — each releases from a clean `master`/`main`, and namecoin's `0.4.x` has no `tagAndRelease.yml` at all).
- `Start9-Community/knuth-bch-startos@master` — `docsUrls`
- `Start9-Community/nutshell-startos@main` — `docsUrls`
- `Start9-Community/navidrome-startos@master` — `alerts`

knuth-bch and nutshell are in no registry (same-day community-intake forks with an open audit PR each), so their fix rides those PRs.

So "`docsUrls` already swept out of twelve" is true as history but not as fleet state. One correction on navidrome: it declares `alerts` in **its own** manifest at `startos/manifest/index.ts:31` on master (both remotes at `14578fc6`, and it has been there since that file's only prior commit `f85641dd`) — the "fails on its dependencies' manifests, not its own" reading holds only on the unmerged `audit/needs-review-pass` branch, which is presumably what got typechecked. Its full failure is 3 errors: its own manifest plus `Jolls/multi-scrobbler-startos` and `Jolls/maloja-startos`, the latter two pulled in only by the fleet-standard `include: ["node_modules/**/startos"]` glob. That is the only transitive chain in either org — a closure sweep over 141 repos finds no other consumer of any offender — and navidrome can mostly self-unblock, since its own `alerts` is already dropped on its open PR.

The remaining `alerts` occurrences (44 refs across 23 in-org repos) are all non-default or pre-2.0 branches; they cannot break today but each surfaces the moment that package is migrated to SDK 2.x, which is worth listing so it is not a surprise mid-migration.

## Considered and dismissed

Raised during review and refuted on re-measurement, so nobody chases them:

- *"A generic wrapper is a false positive on a valid manifest."* No — the rejection is sound (see above), the error does carry a 5-line elaboration, and there are zero instances across 141 repos.
- *"21 shipping packages will newly fail `npm run check`."* Measured against a stale pool mirror. Live count is the 4 branches listed; 6 of the 21 pin SDK 1.x, 5 were unmodified hello-world scaffolds where the fields were still declared, and `i2p`/`i2pd` were the same repo counted twice.
- *"The body's impact assessment is wrong in both directions."* Overstated — it was correct at write time and is only stale, as above.
- *"The `never` collapse is a TS 6 regression."* No — 5.4.5 collapses too for number/boolean/null. Immaterial anyway, since everything pins `^6.0.3`.
- *"The inference change is major."* Downgraded: compile-time only, no new errors anywhere in the fleet.

Not established either way: the historical "twelve packages" `docsUrls` count, and whether the ncc build fails independently of `tsc` — though `s9pk.mk:130` makes `npm run check` a prerequisite of `javascript/index.js`, so the s9pk build blocks regardless.

PR #3754, opened 2026-08-19
fix(core): reap orphaned processes in the subcontainer init

### LINKED ISSUE #3713: Orphaned subcontainer processes are never reaped — launch_init's leader branch waits on TERM without calling waitpid (sync.rs:574)
### Environment

- Area: service runtime (subcontainer process supervision)
- Found by reading `master` at `678fbb03b`; not tied to a particular build or architecture — the code path is unconditional.

### What happens

A process orphaned inside a service's subcontainer is never reaped and stays a zombie for the life of the container. Any daemon that spawns process trees — where an intermediate process exits before its descendants do — accumulates `<defunct>` entries indefinitely.

The case that surfaced it: hermes-agent runs its dashboard and messaging gateway as ordinary daemons, and both spawn agent session/tool subprocess trees. Sessions that complete or are interrupted leave descendants behind, and nothing in the subcontainer collects them.

The subcontainer's PID 1 should reap orphans, as a container init is expected to.

Found by reading source rather than from a running box, so no reproduction steps — the mechanism is:

1. `SubContainerEager` spawns the leader with **no command**: [`projects/start-sdk/lib/util/SubContainer.ts:439`](https://github.com/Start9Labs/start-technologies/blob/master/projects/start-sdk/lib/util/SubContainer.ts#L439) — `['subcontainer', 'launch', rootfs]`.
2. `launch` creates the PID namespace ([`shared-libs/crates/start-core/src/service/effects/subcontainer/sync.rs:458`](https://github.com/Start9Labs/start-technologies/blob/master/shared-libs/crates/start-core/src/service/effects/subcontainer/sync.rs#L458), `unshare(CLONE_NEWPID)`) and then spawns `start-container subcontainer launch-init`, which is therefore PID 1 of that namespace.
3. With an empty command, `launch_init` takes this branch (`sync.rs:574-578`):

   ```rust
   if params.command.is_empty() {
       signal_hook::iterator::Signals::new(signal_hook::consts::TERM_SIGNALS)?
           .forever()
           .next();
       std::process::exit(0)
   }
   ```

   It blocks on TERM signals only. There is no `waitpid`, and `SIGCHLD` is left at `SIG_DFL`, so nothing is ever reaped.
4. Daemons spawned through `SubContainer.spawn` **join** that namespace rather than creating their own — `sync.rs:691` is a `setns(proc/1/ns/pid, CLONE_NEWPID)`, not an unshare. So every non-`runAsInit` daemon's orphaned descendants reparent to that non-reaping PID 1.

The `runAsInit` path is unaffected: there `launch_init` takes the `params.exec()` branch, the package's own command is PID 1, and reaping is the image's responsibility — which is what the existing s6-overlay / tini guidance in `main.md` already covers.

Suggested fix: in the empty-command branch, add `SIGCHLD` to the signal set and drain `waitpid(-1, WNOHANG)` on each delivery, exiting when a TERM signal arrives. Setting `SIGCHLD` to `SIG_IGN` would also work — that branch never waits on anything specific — but an explicit reap loop is what an init is expected to do and won't surprise a later change that does want to wait.

Packages can work around it today by wrapping the daemon command in `tini -s`, making tini a child subreaper below PID 1. hermes-agent is doing exactly that in Start9-Community/hermes-agent-startos#17.

helix-nine:
`launch_init` is PID 1 of a subcontainer's PID namespace. With an empty command it blocked on TERM signals and never waited, with SIGCHLD left at `SIG_DFL` — so nothing reaped. Daemons `setns` into that namespace (`sync.rs:717`) rather than creating their own, so every descendant they orphan reparents to that non-reaping PID 1 and stays `<defunct>` for the life of the container.

It now registers SIGCHLD alongside the TERM signals and drains `waitpid(-1, WNOHANG)` on each delivery, exiting on the first TERM signal as before.

Closes #3713

## Verified on a real box

StartOS 0.4.0 VM, `subcontainer launch` + `subcontainer exec` driven with the shipped binary as the baseline and a build of this branch as the treatment. Zombies counted from the **host's** `/proc` by PID-namespace inode, so nothing is mounted to observe them.

| | PID 1 `SigCgt` | SIGCHLD | 80 orphans | 1000 orphans |
|---|---|---|---|---|
| shipped `0.4.0` | `…4446` | neither caught nor ignored | **80 zombies** | — |
| this branch | `…14446` | caught | **0** | **0** |

The single bit of difference in PID 1's mask is 0x10000, SIGCHLD. (The mask also gains bit 32 between the two builds; that is a `0.4.0` → `master` delta unrelated to this change — the untouched `launch` process gains it too.)

Also checked on the same box:

- `launch` still exits `0` on SIGTERM, and the leader's exit status still propagates.
- `runAsInit` is untouched: `subcontainer launch <rootfs> /bin/sh -c 'exit 7'` still execs and still returns 7.
- 5 × 200 orphans across repeated execs held at 0 zombies, so nothing accumulates.

This replaces a single blocking `.next()` with a loop, so a TERM arriving while PID 1 is busy
draining is a question the change newly raises. Measured rather than argued — a sustained orphan
storm (20 orphans per iteration, continuous) for 5s, then SIGTERM:

| under sustained load | in-namespace | zombies | SIGTERM → exit |
|---|---|---|---|
| shipped `0.4.0` | 1843 | **1840** | 35 ms, status 0 |
| this branch | 3 | **0** | 35 ms, status 0 |

The baseline accrues ~368 zombies/second under that load while this branch holds at zero, and the
TERM latency is identical, so the drain does not delay shutdown. Every number above was measured
against `dfa4d3bae`, whose `start-container` a rebuild reproduced byte for byte; the reaping code
is untouched since, so they still describe this branch — but the head now also carries a merge of
master, so the binary as a whole is no longer that one.

`make start-core-test` passes (306 tests), `make start-core-format-check` and `prettier --check` are clean.

## Two things worth a reviewer's attention

**The drain loop matches explicitly rather than `while let Ok(..)`.** `nix::sys::wait::waitpid` calls `WaitStatus::from_raw` *after* the syscall has already reaped the child, and `from_raw` → `term_signal` → `Signal::try_from` returns `Err(EINVAL)` for any signal outside nix's enum — which is every real-time signal. A `while let Ok(..)` drain therefore stops dead at the first child killed by, say, `SIGRTMIN+1`, leaving the rest of the batch unreaped until some later SIGCHLD that may never come. Measured against nix 0.30.1, with two of five children killed by signal 35:

```
-- naive drain: `while let Ok(status)` --
   reaped 0 of 5 before the loop ended
-- what the next call returns (the error that ended it) --
   Err(EINVAL)
-- hardened drain over the remainder --
   hardened drain collected 3 more
```

Every arm that continues has consumed a child, so the loop is still bounded.

**`reap_orphans()` is called once before the loop.** `SubContainerEager.waitProc` polls for `<rootfs>/proc/1`, and that path is created by the procfs mount immediately above — i.e. the mount is what unblocks a caller to `setns` in and start orphaning things, and it runs before the handler is installed. A child that dies in that window has its SIGCHLD discarded under `SIG_DFL`, so the one pre-loop drain is what collects it. Installing the handler above the mount instead would also close it; I kept the mount and the `runAsInit` path untouched rather than reshaping the function around a microsecond.

## Scope

`SIG_IGN` / `SA_NOCLDWAIT` would also drive the count to zero, but `SIG_IGN` survives `execve`, so it would be inherited by anything `launch_init` later execs and would make a bundled supervisor's own `waitpid` return `ECHILD`. The explicit loop leaves `wait` semantics alone.

`runAsInit: true` is deliberately not covered — there the package's command *is* PID 1 and reaping is the image's init's job, which is the arrangement the packaging book already steers authors to. Likewise, a daemon process orphaned because the *host-side* `exec` wrapper died has its `real_parent` outside the namespace and reparents to the outer reaper; that is a separate pre-existing path, not this one.

## Docs and changelog

`runAsInit` used to be a free choice as far as reaping went — nothing reaped either way. It is a trade-off now, so the places a package author reads say so:

- `projects/start-os/CHANGELOG.md` — a `### Fixed` entry under the unreleased `## [0.4.0.2]` (latest origin tag is `start-os/v0.4.0.1`, which carries the defect).
- `projects/start-sdk/docs/src/main.md` — one sentence on the `runAsInit` page. Nothing there was falsified (the book justifies `runAsInit` by "the supervisor must be PID 1", never by reaping), but "my daemon spawns process trees, do I need an init?" is a real authoring question that two packages answered with a workaround — `hermes-agent` with `tini -s`, `metube-startos` with `runAsInit: true`.
- `projects/start-sdk/CHANGELOG.md` — the same trade-off under the unreleased `## 2.0.10 — StartOS 0.4.0.2`, which is the SDK release that raises the minimum OS to the one carrying this fix. The SDK's own surface is unchanged; this sits alongside the existing `effects.getServicePortForward` bullet as platform behaviour a packager builds against.
- `projects/start-sdk/lib/mainFn/Daemons.ts` — the `runAsInit` tsdoc, which is the description most readers actually see: the book itself tells them to find SDK options by grepping the types.
- `projects/start-sdk/lib/backup/Backups.ts` — a comment this change falsifies. `withMysqlDump` runs in a `SubContainer.withTemp` whose leader takes exactly this branch; on the MySQL path `--daemonize` double-forks, so that mysqld reparents onto the init and is now collected by it, while the MariaDB path deliberately does not daemonize and is reaped by the `exec` that started it. The loop already handled a vanished PID — the comment had credited the zombie state as the case that mattered, and named one reaper where there are two.

`projects/start-os/docs/` needs nothing — no page there describes subcontainer process supervision.

No test: `make start-core-test` runs unprivileged in a container, so `unshare(CLONE_NEWPID)` is unavailable and nothing there can exercise the actual defect. A `reap_orphans`-only unit test would assert that `waitpid` reaps, which was never the broken half.


projects/start-sdk/lib/mainFn/Daemons.ts
```diff
@@ -120,7 +120,7 @@ export type ExecCommandOptions = {
    * @default 60_000
    */
   sigtermTimeout?: number
-  /** Run the command as PID 1 inside the container (init process) */
+  /** Run the command as PID 1 inside the container, in place of the init that reaps orphaned processes */
```
dr-bonez: revert this, this is a bad comment change, not inline with timeless documentation.

dr-bonez (dismissed):

projects/start-sdk/lib/mainFn/Daemons.ts
helix-nine: Reverted in 2106ade92 — `Daemons.ts` is now byte-identical to master, so it has dropped out of the diff entirely.

You are right about the principle: it named the init the option displaces, which describes the implementation rather than the option, and goes stale as soon as that init's responsibilities change. "Run the command as PID 1" is the part that stays true.

The trade-off I was trying to get across — that turning the option on hands child-reaping to your entrypoint — is already stated in the packaging book and the SDK changelog, both of which are tied to a release and can date. I left those alone; say the word if you want the book's sentence pared back on the same grounds.

dr-bonez (approved):

PR #3756, opened 2026-08-19
fix(start-wrt): stop stale cached UIs after firmware updates — cache headers, live detection, and an auto-reload/prompt flow

Dominion5254:
## Summary

After a firmware update, browsers could silently keep running the previous
UI bundle against the new backend — the embedded web server sent no cache
headers at all, so cache behavior was per-browser heuristics. This PR closes the whole class:

1. **Any page load now picks up the new UI** — explicit cache headers with
   per-build ETag revalidation.
2. **An already-open tab notices too** — the UI compares the firmware's
   build stamp against its own on every RPC response and reloads (or
   prompts) on mismatch.
3. The detection path is hardened for every deploy shape found on the
   bench: CLI deploys faster than the poll cadence, idle pages that make no
   requests, tabs backgrounded through an OTA reboot.

Seven commits, in dependency order:

## 1. `fix(start-wrt): stop browsers from serving a stale cached UI after updates` (2f11219df)

- Stable-named files (index.html, assets/) get `Cache-Control: no-cache`
  plus a per-build ETag (the baked-in `STARTWRT_GIT_HASH`) — every load
  revalidates, costing a 304 until the firmware actually changes.
- Angular's content-hashed bundles get `public, max-age=31536000,
  immutable` — names change with content, so forever-caching is safe and
  speeds up loads.
- The SPA fallback only serves index.html for extensionless (route-like)
  paths; a hashed chunk from an older build now 404s instead of receiving
  an HTML body that can break module loading.
- Follows start-os's `static_server.rs` (ETag + 304 + immutable) but
  deliberately diverges on index.html: start-os caches it immutable and
  compensates with a "Refresh Needed" dialog; revalidating it makes a plain
  reload sufficient, removing the stale-UI state entirely.
- Unit tests exercise the handler against the real embedded dist (self-skip
  when the dist placeholder is empty, e.g. plain `cargo test` in CI).

## 2. `feat(start-wrt): detect a stale UI after firmware updates and reload it` (9b576016d)

A tab that is already open never refetches index.html (in-app navigation is
client-side routing), so the cache fix alone leaves it stale until a manual
reload. Detect the mismatch instead:

- **One build stamp everywhere.** `build-rust.sh` and ctrl's `build.rs`
  fallback now emit the same stamp as `build/env/check-git-hash.sh` (full
  hash, `-modified` on a dirty tree; the separate short-hash `-dirty`
  format is retired), so the binary's UI ETag / `system.info` /
  `startwrt verify` output and the web bundle's baked-in `config.json`
  gitHash are the identical string.
- `system.info` (no_auth) returns `gitHash`, so detection works even when
  an update reboot wiped the session store.
- `StaleUiService` compares (dirty-marker-insensitive, skipping dev
  `@branch`/`unknown` stamps), fed from every `system.info` consumer.
- **UX split by who updated:** an update initiated from this tab reloads
  automatically once the router is back (the session is gone either way;
  the "Updated to vX" confirmation rides sessionStorage and shows after
  login). An update applied any other way — CLI deploy, another device —
  shows a "Refresh Needed" dialog with a Reload button instead: never
  reloading out from under unsaved work.

## 3. `fix(start-wrt): translate the post-update confirmation and document the update UX` (8ba6413b7)

The "Updated to vX" confirmation was the one untranslated string in the new
flow (now in all five locales), and `docs/src/updating.md` documents the
automatic reload, the post-login confirmation, and the "Refresh Needed"
dialog.

## 4. `fix(start-wrt): detect a stale UI on every RPC response, not just connection loss` (ef909110e)

Found on the bench: a CLI deploy (`make start-wrt-update`) restarts the
daemon faster than the 5s poll cadence, so no request ever failed and an
open tab never noticed the new firmware (a tab backgrounded through a whole
OTA reboot could miss it the same way). The daemon now stamps
`x-startwrt-git-hash` (CORS-exposed for the dev serve) on every `/rpc/v1`
response and `HttpService` feeds it to `StaleUiService` — any request an
open tab makes detects a deploy within seconds. The reconnect-probe check
remains as fallback for disruptive updates.

## 5. `fix(start-wrt): re-check for a stale UI every 30s on idle pages` (3004df19a)

Also found on the bench: non-FormService pages (e.g. Settings → General,
which renders from the boot-loaded info signal) make zero requests while
idle, so nothing carried the header. `SystemService` runs a 30s
`system.info` heartbeat as a detection floor (also keeps the displayed
system date fresh), skipped while the tab is hidden, the in-tab update flow
owns polling, or the reconnect loop is already probing. Detection latency:
~5s on form pages, ≤30s anywhere.

## 6. `fix(start-wrt): stop periodic info refreshes from reverting unsaved theme/language previews` (bc7b2e91e)

Audit fallout from the heartbeat: the app-level effect re-applied the saved
theme/language on every info object, stomping a live preview on Settings →
General within 30s. The effect now keys on the values themselves.

## 7. `fix(start-wrt): re-arm a dismissed stale-UI prompt when the firmware updates again` (8612a28b3)

Dismissal latched forever; it now latches the newer build's hash and
silences per build — a *second* update re-prompts, matching start-os's
refresh alert.

## Hardware validation

Iterated against a live K1 board across two bench days; commits 4 and 5 each
close a gap the bench itself exposed:

- curl header suite: correct `Cache-Control`/`ETag` per file class, 304 on
  matching ETag, stale-chunk 404.
- Browser: plain F5 picks up a new build with no hard refresh.
- Back-to-back CLI deploys with a tab open: "Refresh Needed" appears within
  seconds (per-RPC header path), including on idle non-polling pages
  (heartbeat path) — **hardware-confirmed**.
- In-tab update: automatic reload, post-login "Updated to vX" confirmation.

Commits 6–7 are logic-level fixes verified by type checks and prod build,
not re-benched. The 2026-08-18 rebase onto current master touched only
i18n ids, changelog placement, and the (auto-merged) dirty-detection
reconciliation — a post-deploy smoke of the dialog + dismissal UX and the
dev-serve CORS header is cheap insurance but nothing in those paths changed.

## Known limitations (accepted)

- **Dev-only ETag collision:** redeploying via `make start-wrt-update` from
  the *same dirty HEAD* keeps ETag `<hash>-modified` while content changed,
  so a browser can 304 into a stale UI. Deploy paths from clean commits are
  unaffected; a dist-content-hash fallback ETag would fix it if it ever
  bites.
- `serve_embedded` still lacks HEAD/Range/compression (pre-existing scope
  choice, unchanged here).
- `build.rs`'s dev fallback has no `.git/HEAD` rerun trigger, so a raw
  local `cargo build` can carry the previous commit's stamp (dev-only;
  every deploy path regenerates).

## Merge notes

- **Independent of every open PR** — binary + web change only, no image
  rebuild needed; an OTA/binary push ships it.
- Changelog heading retitle `[1.0.2]` → `[1.1.0]` is the identical hunk
  every open StartWRT PR carries — auto-resolves in any merge order. No
  manifest bump here (the feature PRs own that).
- New i18n ids are **545–548**, chosen above #3634's claimed 534–544 so the
  two merge cleanly in either order (a gap in the id sequence is harmless
  if #3634 lands later — ids only need uniqueness).
- `API_CONTRACT.md` (gitHash field + response header), docs, and changelog
  entries are in the commits; no separate docs work owed.


dr-bonez (commented):
@helix-nine should we be applying this strategy to the StartOS UI as well?

dr-bonez (dismissed):

dr-bonez (approved):

PR #3758, opened 2026-08-19
feat(start-wrt): ship the 6in4 and 6rd packages in the image

dr-bonez:
## What

Adds `CONFIG_PACKAGE_6in4=y` and `CONFIG_PACKAGE_6rd=y` to the StartWRT OpenWrt diffconfig, so both IPv6-over-IPv4 tunnel protocols ship in the image.

## Why

**6RD was already selectable and already broken.** The WAN IPv6 `6rd` mode has been in the UI and the backend since 1.0.1 (`web/src/app/routes/wan/routes/ipv6/utils.ts`, `backend/ctrl/src/wan.rs`), and it writes a valid `/etc/config/network`. But the image at `start-wrt/v1.0.1` carried neither the `6rd` netifd proto handler nor `kmod-sit`, so netifd had no handler for the proto it was given and the WAN interface simply came up without IPv6 — with nothing in the UI to indicate why.

**6in4 rides along.** It needs the same `kmod-sit` (and `kmod-iptunnel{,4}`), and it covers the case 6RD can't: 6RD requires the ISP to offer it, while a 6in4 tunnel comes from a third-party broker such as Hurricane Electric and works over any public IPv4 address. Getting it onto a router previously meant building and sideloading the packages by hand *after every update*, since a sysupgrade does not preserve separately installed packages.

There is no UI for 6in4 yet — it is configured over SSH.

## Changes

- `build/openwrt.diffconfig` — the two package selections, with a comment noting the shared `kmod-sit` dependency and which half supplies the netifd proto handler.
- `docs/src/wan.md` — a TIP under the IPv6 section covering what 6in4 is, how it differs from 6RD, that it does not work behind CGNAT, and that it is SSH-configured.
- `CHANGELOG.md` — an `### Added` entry for 6in4 and a `### Fixed` entry for 6RD, both under the unreleased `## [1.0.2]`.

## Notes for review

- **Changelog placement**: origin has tags only through `start-wrt/v1.0.1`, so `1.0.2` is unreleased and both entries go under the existing heading. No manifest bump — this is a fix plus an additive package, and `1.0.2` already covers that tier.
- **6RD is a genuine `### Fixed`**, not a fix to an unreleased feature: `git show start-wrt/v1.0.1` confirms the mode shipped to users while the diffconfig at that tag had no `6rd`/`6in4`/`sit` entries.
- **No CI `paths:` change needed** — no `build.mk` prerequisite moved.
- Image build not run locally (hours); the diffconfig lines are ordinary upstream OpenWrt package symbols, so CI's image job is the real check here.

### DIFF
--- a/projects/start-wrt/CHANGELOG.md
+++ b/projects/start-wrt/CHANGELOG.md
@@ -61,6 +61,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   RPC response, so an open tab notices within seconds of its next request even
   when the update restarted the daemon too quickly to drop a connection; pages
   that make no requests while idle re-check every 30 seconds.
+- **6in4 tunnels can now be configured on the router.** The `6in4` protocol
+  and the SIT kernel module it needs now ship in the image, so an IPv6 tunnel
+  from a broker such as Hurricane Electric can be set up over SSH — useful for
+  reaching IPv6 on an ISP that provides none. There is no UI for this yet.
+  Previously these packages had to be built and sideloaded by hand after every
+  update, since a sysupgrade does not preserve separately installed packages.
 
 ### Changed
 
@@ -101,6 +107,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   service goes live.
 
 ### Fixed
+
+- **The WAN IPv6 "6RD" mode now works.** Selecting it wrote a valid
+  configuration, but the image shipped without the `6rd` protocol handler or
+  the SIT kernel module it depends on, so the WAN interface simply came up
+  without IPv6 — with nothing in the UI to indicate why. Both now ship.
 
 - **Disabled outbound VPNs are no longer offered as a Security Profile's
   outbound route.** The VPN picker on the profile's WAN / Internet tab listed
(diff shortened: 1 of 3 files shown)


projects/start-wrt/CHANGELOG.md
```diff
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [1.0.2]
```
Dominion5254: Put this under 1.1.0; 1.0.2 isn't going to be a release. Next release will be 1.1.0 given the features that have already been merged to master.

Dominion5254 (changes requested):
Aside from the CHANGELOG comment this PR is good to merge

projects/start-wrt/CHANGELOG.md
helix-nine: Done — rebased onto master, which had already moved the top heading to 1.1.0 (the crate manifest is there too), so no bump was needed here. The 6in4 entry now sits in that section's existing `### Added` list rather than opening a second one, and the 6RD fix under its `### Fixed`. Only the changelog placement changed; the diffconfig and docs hunks are as reviewed.

Dominion5254 (approved):

PR #3768, opened 2026-08-20
fix(start-os): check the port a certificate authority validates on

### LINKED ISSUE #3760: A Let's Encrypt domain on a non-443 TLS port passes every add-domain check and then refuses every connection — the checks probe only the interface's own port (net/host/address.rs:453)
### Environment

StartOS `master`, read at `142e109bc`; area networking + web UI. Paths below are relative to `shared-libs/crates/start-core/src/` for Rust and `projects/start-os/web/ui/src/app/` for the frontend, except where given in full.

This is #3734's fourth stated trade-off — "the add-domain checks still test the interface's own port, not 443, so they can go all-green on a domain that then refuses … Worth keeping #3717 open for that" — filed on its own, since #3717 was closed at merge. The refusal that makes it user-visible is unreleased: it lands in the prospective `0.4.0.2`, and the latest tag is `start-os/v0.4.0.1`. The checks behave identically on `0.4.0.1`; there the green result precedes #3717's silent local-CA fallback rather than a refusal.

The design decisions in **What should happen instead** were made by @dr-bonez in review of this issue, and the section is written to be implementable without further context.

### What happens

Adding a public domain runs three checks — DNS, IPv4 port forwarding, IPv6 firewall. The two reachability checks test one port: the external port of the address being added (`net/host/address.rs:453-460`, `:571-584`). For most interfaces that port is 443, so the checks and reality agree.

For an interface StartOS terminates TLS for on some other port, the domain's address sits on that port and 443 is never tested. The shape is not rare: Fulcrum and electrs bind Electrum with `addSsl.preferredExternalPort: 50002`, LND its REST API on `8080`, CLN clnrest on `3010`, coturn `turns` on `5349` (and `5449` for its second listener), OwnTracks MQTTS on `8883`, Public Pool stratum TLS on `4333`. The Certificate Authority selector appears whenever any binding on the host is SSL, and defaults to the first registered provider (`gateway.component.ts:203-213`), so on a box with an ACME provider registered Let's Encrypt is pre-selected.

A Let's Encrypt domain needs 443 either way — TLS-ALPN-01 is validated at `<name>:443` whatever port the name is served on (`net/vhost.rs:537-538`). So every check can pass, and the setup modal does not open on its own unless one fails (`domain-health.service.ts:203-206`). On all-green the user sees nothing: no modal, no confirmation, just a new row whose Certificate Authority column reads `Let's Encrypt` (`interface.service.ts:63-81`). Whether the domain works then turns on the one thing the checks never looked at — whether the challenge can reach the box on 443. Where it cannot, the address refuses every connection with a fatal `unrecognized_name` alert (`net/tls.rs:238-253`).

443 reaches the box over StartTunnel, where #3734 asks the gateway for `443/SNI=<name>` alongside the interface's own port (`net/vhost.rs:579-592`), and wherever 443 is separately forwarded, which the box's own 443 listener answers from the cache of orders in flight (`net/vhost.rs:1686-1692`). A gateway speaking UPnP, NAT-PMP or PCP without the Start9 `HOSTNAME` extension gets neither: that route is a PCP `HOSTNAME` mapping, skipped with no fallback to a plain one (`net/port_map/client.rs:697`, `:782`). The checks pass identically in all three cases, because none of them probes 443.

The only signal is post-hoc. #3734 raises a Warning notification whose text does name 443 ("… a certificate authority validates on port 443 whatever port the address itself serves"), but issuance is lazy — an order starts only on an inbound TLS ClientHello for that name (`net/acme.rs:364` is the sole caller of `get_cert`). A user who adds the domain and does not immediately connect gets nothing, and by the time the notification arrives the address has already refused someone.

Two pieces of work. (1) is the defect. (2) is a change to the challenge routing #3734 already shipped, decided in review of this issue; it is independent of (1) and can land first.

#### 1. Check 443 when the domain actually needs it

Run an additional reachability probe on port 443 when **all** of:

- the domain's certificate authority is ACME (`PublicDomainConfig.acme` is `Some`, `net/host/address.rs:34-39`), and
- the port already being checked is not 443, and
- the domain does not currently hold a certificate it can serve.

The third condition is the important one: a domain whose certificate is in force does not need 443 reachable right now, and failing the check there would show red on a domain that works.

**The predicate is `should_use_cert`, not `cert_is_unexpired`** (`net/ssl.rs:47-56` and `:58-70`). `should_use_cert` demands 30 days of headroom, so it turns false at the start of the renewal window — exactly when 443 has to work again. `cert_is_unexpired` stays true until the last day of validity and would hide a failing renewal right up to the outage.

**Where the certificate is.** `AcmeCertStore` (`net/acme.rs:380-383`), reached as `db.as_private().as_key_store().as_acme()` (`net/acme.rs:390-395`), i.e. `private.keyStore.acme.certs[<provider url>][<SAN set>]`. A vhost-driven order uses the singleton SAN set `{domain}` (`net/acme.rs:361`), so the lookup is by fqdn under the domain's configured provider. The leaf to test is `fullchain[0]`, as `get_cert` does at `net/acme.rs:110-119`.

**This forces the shape.** The private tree is not synced to the UI, so the gate cannot be evaluated client-side. `add_public_domain` (`net/host/address.rs:385`) must decide and report it: add a field to `AddPublicDomainRes` (`net/host/address.rs:230-234`, ts-rs binding at `shared-libs/ts-modules/start-core/lib/osBindings/AddPublicDomainRes.ts`) carrying the 443 result, or `None`/absent when the probe was not applicable. The modal's Test button needs the same answer, so whatever carries it must reach `DomainValidationData` (`domain-validation.component.ts:31-43`).

**Cover both families.** Probe 443 over IPv4, and over IPv6 wherever the gateway has a GUA — the same shape `portAllPass` already uses, where the v6 leg is required only when a GUA exists (`utils/gua.ts:37-46`). No new RPC is needed: `net.gateway.check-port` and `net.gateway.check-port-v6` both take an arbitrary port (`net/gateway.rs:625-630`), and `checkPublicDomain` already calls `checkPort` directly on the enable path (`domain-health.service.ts:170-172`).

**A 443 probe does not false-fail the StartTunnel case**, where 443 is claimed only per-SNI: the active-mapping lookup matches on external port and protocol alone, ignoring the hostname (`net/port_map/client.rs:401-407`), so that mapping satisfies `check_port(443)`.

**Frontend surface.** The modal is three hardcoded sections, not a loop over a returned set — headings at `domain-validation.component.ts:52`, `:109`, `:133`, with per-check result and loading signals at `:285-290`, the verdict at `:301-306` (`allPass`) and the per-check probes at `:319-367`. A 443 row is a fourth section reusing `port-check-test.component.ts`, which the IPv6 Firewall row is already a second instance of. The auto-check and its modal gate are `domain-health.service.ts:143-224` (open at `:373-377`), and the same port is used on the enable path (`:117-124`) and when the modal is opened by hand from the address row (`actions.component.ts:267-277`) — all three need the new check. New UI strings go in `shared-libs/ts-modules/shared/src/i18n/dictionaries/en.ts` plus the `de`/`es`/`fr`/`pl` siblings, where the key is the English string mapped to a numeric id.

#### 2. Make the challenge route uniform, and always listen for it

In `desired_port_maps` (`net/vhost.rs:551`):

- **Box-side port becomes 443.** Change `.or_insert(*external)` (`net/vhost.rs:591`) to `.or_insert(ACME_CHALLENGE_PORT)`, so the challenge route is `443 -> 443` universally rather than `443 -> <the domain's own port>`. The ACME cache is global — one instance created at `net/vhost.rs:289` and cloned into every server at `:370`, consulted without reference to the server's own mapping (`net/acme.rs:330-340`) — so a challenge is answerable from any listener, and 443 always has one. The comment above that block (`net/vhost.rs:579-581`) explains the old choice and must be rewritten with the new one.
- **Add the IPv6 challenge pinhole.** The v6 loop (`net/vhost.rs:595-618`) currently requests a hostname-less pinhole for `GUA:<interface port>` and deliberately none for the challenge port. For an ACME target whose external port is not 443, also request `(GUA, 443) -> 443`.
- **Always listen on 443 where a challenge can arrive.** Port-443 listeners are bound per-address, and port 443's bind requirements are computed only from the targets on port 443 (`compute_bind_reqs`, `net/vhost.rs:646-658`; `desired_listeners`, `:680-707`; `ProxyTarget::bind_requirements`, `:1152-1160`). An ACME domain on another port contributes nothing to them, so `<box ip>:443` can be unbound on a gateway where that domain is public — for instance when the operator disables the admin UI's bare address there, since today it is the admin UI's own LAN address that happens to hold the socket open (`net/host/mod.rs:187-193`, `net/host/binding.rs:75-95`, `net_controller.rs:430-436`). A challenge can arrive wherever an ACME domain is public, so port 443 must bind there regardless: derive from every target with `acme.is_some()` its `public_v4` gateways and `public_v6` GUAs, and add them to the port-443 server's bind requirements.

  Implementation notes for that last bullet:

  - The seam is `reconcile_port_maps` (`net/vhost.rs:475-487`), which already receives the full per-owner `BTreeMap<VHostKey, ProxyTarget>` that `desired_port_maps` consumes. Derive the requirement set there as a free function beside `desired_port_maps`, so it is unit-testable the same way, and key it per `HostMapOwner` as `port_mappings` is (`:490-500`) so an owner whose ACME targets go away withdraws its contribution.
  - `VHostServer` recomputes `bind_reqs` from its own mapping alone (`update_bind_reqs`, `:1972-1982`, called at `:1944` and `:1967`), so it needs an externally-supplied set folded into that computation, and setting that set must trigger the same recompute. Create the port-443 server if it does not already exist (`create_server`, `:355`).
  - Binding an address exposes nothing on it. `filter` gates acceptance per address independently of the bind — the property the existing `bind_requirements` comment already relies on (`:1152-1157`) — and non-challenge traffic for a name the 443 server does not route is refused exactly as it is today (`:1686-1692`, `net/tls.rs:238-253`).

One consequence to expect: the v6 redirect block (`net/vhost.rs:620-635`) adds an 80→443 pinhole to any GUA exposing 443, so an ACME domain will now also open v6 port 80. That is the same treatment a real 443 binding gets.

Also note the behaviour change for ordinary traffic: a client dialling `https://<name>:443` for a domain published on 50002 is currently proxied to the service, and after this change is refused, since the box's 443 server does not route that name. The name is only ever published on its own port, so this is arguably the more correct answer, but it is a change.

**Tests.** `mod port_map_tests` (`net/vhost.rs:2076-2240`); run with `cargo test --features=test --lib net::vhost::port_map_tests` from `shared-libs/crates/start-core` (5 tests, all passing today).

- `an_acme_domain_off_443_also_routes_the_challenge_port` (`:2161`) and `each_acme_domain_keeps_its_own_challenge_route` (`:2177`) both assert the box-side port is the domain's own; they must flip to 443.
- `nothing_else_claims_the_challenge_port` (`:2197`) and `a_real_443_binding_owns_the_port_it_serves` (`:2210`) should be unaffected — keep them green as the regression guard.
- `ipv6_pinholes_are_unchanged` (`:2226`) covers a target already on 443. Add a case for an ACME target off 443 asserting the new `(GUA, 443)` pinhole and the 80→443 redirect it pulls in.
- Add cases for the bind-requirement derivation: an ACME target on 50002 public on a gateway contributes that gateway and its GUA to port 443; a non-ACME target on the same port contributes nothing; an ACME target already on 443 needs no extra entry.

#### Documentation

`clearnet.md` already tells the user the requirement in prose — `projects/start-os/docs/src/clearnet.md:41`, `:48-49`, `:78-79`, added by #3734, including a note under the checks step saying they cover only the port the interface is served on. Once the checks cover 443, that note is stale and must be rewritten to describe what the checks now do. Add the changelog entries under the existing `## [0.4.0.2]` heading in `projects/start-os/CHANGELOG.md` — that version has no origin tag, so it is still open.

Found by reading code; not reproduced on a box.

- `net/host/address.rs:453-460` — `add_public_domain` derives one port to test from the target binding's own address rows, preferring the SSL row: `.filter(|a| a.public && a.hostname == fqdn).max_by_key(|a| a.ssl).and_then(|a| a.port)`.
- `net/host/mod.rs:287-296` — that SSL row's port is `assigned_ssl_port`, re-projected onto the binding's preferred SSL port when one of our own TLS listeners answers there. `preferred_ssl_port()` prefers `addSsl.preferred_external_port` and falls back to the binding's own port when the container serves its own TLS (`net/host/binding.rs:459-467`). The SDK defaults the `addSsl` one to 443 for `http`/`https`, which is why most domains land on 443. A passthrough binding lands off 443 the same way but is out of scope: its vhost carries `acme: None` (`net_controller.rs:517-521`), so no authority is ever claimed for it.
- `net/host/address.rs:571-584` — that one port is what both reachability probes receive: `check_port { port: ext_port }` and `check_port_v6 { port: ext_port }`. The DNS leg takes no port at all.
- The handler knows the certificate authority — `AddPublicDomainParams.acme` (`address.rs:219-220`) — and uses it only to reject an unregistered provider (`:401-411`) before storing it (`:429`). Neither probe takes an ACME argument (`net/gateway.rs:625-630`), and neither `net/gateway.rs` nor `net/dns.rs` contains the string `acme`.
- The frontend uses the same port everywhere else it probes: enabling an existing address passes `addr.hostnameInfo.port` (`domain-health.service.ts:117-124`), opening the modal by hand from the address row passes the same (`actions.component.ts:267-277`), and the modal's Test buttons re-probe `context.data.port` (`domain-validation.component.ts:338-341`, `:357-360`). Under `projects/start-os/web/ui/src/app/routes/portal/components/interfaces/` the only occurrence of `443` is a static bullet in the pre-add picker, `'Port forwarding (usually 443)'` (`domain-type-picker.component.ts:40`).
- The verdicts carry no certificate term: `portAllPass` requires `openExternally && hairpinning` on that port, plus the v6 leg's `openExternally` where a GUA exists (`utils/gua.ts:37-46`), and `allPass` is DNS ∧ port (`domain-validation.component.ts:301-306`). The authority is never plumbed into the check path — `savePublicDomain` sends `acme` to the backend (`gateway.component.ts:280`) and omits it from the check context, which is `{ packageId, addSsl, watch }` (`domain-health.service.ts:22-26`).

helix-nine:
Adding a public domain runs three checks — DNS, IPv4 port forwarding, IPv6 firewall — and both reachability probes test one port: the external port of the address being added. For an interface StartOS terminates TLS for on some other port (Electrum on `50002`, clnrest on `3010`, `turns` on `5349`, LND REST on `8080`), that is not the port the domain's certificate depends on. TLS-ALPN-01 is validated at `<name>:443` whatever port the name is served on, so every check could pass while the one thing that decides whether the domain works went untested. On all-green the modal does not open, and the user sees only a new row reading `Let's Encrypt` — then the address refuses every connection with a fatal `unrecognized_name`.

Both halves of #3760.

## 1. Check 443 when the domain needs it

`net.acme.check-challenge` probes 443 — IPv4, and IPv6 where the gateway has a GUA — when the domain names an ACME authority, the address is served elsewhere, and no certificate is on hand to serve. That last condition uses `should_use_cert`, the predicate the cert resolver itself applies, so the check comes back at the start of the renewal window rather than on the last day of validity. The certificate store is private and never reaches the browser, so the backend answers the whole question; `add_public_domain` embeds the same answer in its response, and the enable path calls the RPC.

`CheckChallengeRes` carries each leg as an option, so three states are representable end to end: absent (the domain does not need 443), present-and-null (outstanding but unmeasured), present-and-measured. That is what keeps a probe error from counting as a pass, and what keeps a `check_port(443)` failure from failing the whole `add-public-domain` call after the domain is already committed.

The setup modal grows a **Certificate Authority** section carrying the result. One answer both renders it and holds `allPass` to it, so nothing there can be red beside an enabled **Done**. Its verdict is `openExternally` only — a certificate authority dials from outside, so whether the router hairpins says nothing, and the hairpinning warning is suppressed for those rows.

`Host::port_forwards` emits the `443` rule alongside the interface's own, so an operator who dismisses the modal and goes to **System → Gateways → Port Forwards** to configure their router is shown both rules rather than half of what they need.

## 2. A uniform challenge route, and a listener to receive it

- The IPv4 SNI route is `443 -> 443` rather than `443 -> the domain's own port`. Every listener answers a challenge out of the same cache, so the box-side port never had to be the interface's.
- IPv6 gets the matching pinhole, which it was skipping. It does **not** pull in the `80 -> 443` redirect a served 443 does: TLS-ALPN-01 never uses port 80, so that would open a port nothing serves.
- Port 443 now binds wherever an ACME domain is public. Port 443's bind requirements are computed from its own targets, which cannot see a domain served on `50002` — so its socket was held open only by whatever else happened to be there, typically the admin UI's LAN address. The requirement is derived per host-map owner beside `desired_port_maps`, unit-testable the same way.

Binding 443 exposes nothing on it: `filter` gates acceptance per address independently of the bind, and a name the 443 server does not route is refused exactly as today.

**Behaviour change worth knowing:** a client dialling `https://<name>` with no port, for a domain published on another port, is now refused rather than proxied to the service. The name is only ever published on its own port, so this is the more correct answer, but it is a change; the changelog says so.

## Review

Two rounds of the `code-review` skill at `high`, and the second one mattered.

**Round 1** reviewed commit 1 and returned 10 findings; commit 2 is the fixes. The ones worth naming, because each was a way this change could still have left a domain silently broken:

- The IPv6 challenge pinhole is hostname-less, making it the first port-map key two owners can both request. `sync_port_maps` diffed each owner against itself alone, so one ACME domain going away shut the pinhole another on the same GUA was still using — a renewal failing two months later with nothing on screen.
- The challenge probe failed open: an RPC error and "not needed" were both `null` on the client, so an errored probe counted as a pass and the modal never opened.
- The modal rendered its Certificate Authority row on one predicate and gated **Done** on another.
- Port Forwards omitted the `443` rule.
- `sync_challenge_binds` took its union under one lock and applied it under another.

**Round 2** reviewed commit 2 — which had gone out unreviewed, since round 1 was scoped to commit 1 — and found **3 of those 10 fully closed**. Three closed their stated defect and opened another; commits 3 and 4 are those:

- Collapsing onto one predicate was done by choosing "the backend says 443 is outstanding", which is only computed on a path that probes. That deleted the Certificate Authority section from the **Address Requirements** view reached from an address's own menu — the screen named after requirements, and exactly where a user lands after clicking **Later** to go configure their router. The contradiction that predicate was avoiding cannot arise there: that view renders no footer, so there is no verdict. A view with a verdict now shows what is outstanding; the manual view shows the requirement.
- The Port Forwards rule keyed on `addr.ssl`, which is true for a TLS passthrough binding, so a host with one `addSsl` binding and one passthrough advertised a `443` rule against the passthrough interface that nothing here would ever answer.
- `take_ownership` made the bookkeeping atomic but still dispatched the withdrawal after releasing the lock, so the interleaving it exists to prevent survived in a narrower window.
- Two of the comment rewrites shipped false claims, one onto the generated `CheckChallengeRes` binding: a null `port_v6` is the no-GUA case as well as the errored one, and an address serves the certificate it holds right through a renewal window rather than refusing every connection.
- And the hairpinning warning, suppressed for the challenge rows in the modal, came back on Port Forwards — same forward, two screens, contradictory verdicts. A flag on the widget could not fix it: rows group by port pair, so a challenge `443` and a served `443` coalesce. Whether anything local dials a forward is a property of the rule, so `PortForward` carries it now, defaulting true.

Round 2 also confirmed by independent trace what I had checked by hand: `Host::port_forwards` has no Rust reader — nftables and PCP go through `forward.rs`/`port_map` — so that field is display-only, and both outside callers of the two shared components omit `local` and keep the behaviour they had.

Two findings I have **not** acted on, deliberately. `add_public_domain` can still fail after committing the domain if `check_port` errors on the address's own port; that is pre-existing on the `port`/`dns` legs, absorbing it needs those response fields to become optional, and this branch only made the new 443 leg safe. And the client's fail-closed `{port: null, portV6: null}` is byte-identical to a real outstanding challenge, so one transport failure shows a spurious Certificate Authority section — fail-closed is the right default here and the rows carry a working Test button, but it is a real conflation and worth a second opinion.

## Verification

`cargo test -p start-core --features=test --lib` — 320 pass, including 24 across `net::vhost::` and 10 in `port_map_tests` specifically. The two tests asserting the old box-side port flip to 443; `nothing_else_claims_the_challenge_port` and `a_real_443_binding_owns_the_port_it_serves` stay green as the regression guard. New: the v6 challenge pinhole and that it does not earn the port-80 redirect, that a served 443 still does, the bind-requirement derivation (contributes where public, nothing for a non-ACME target or one already on 443), and that a pinhole another owner still wants is not withdrawn.

`npm run check` (3 i18n checks + 7 `tsc` projects) and `make format-check` are clean; bindings regenerated with `make start-core-ts-bindings` and stable. Not exercised on a box — no test VM has a public domain with a real authority — so the reachability probes and the 443 bind are verified by unit test and by reading, not by a live challenge.

Docs and changelog ship with it: `clearnet.md`, the `net acme check-challenge` entry in `cli-reference.md`, and the `0.4.0.2` entry for #3734 edited in place rather than a new `### Fixed` line, since that feature has not shipped yet.

Closes #3760


projects/start-os/web/ui/src/app/routes/portal/components/interfaces/addresses/gateway/domain-validation.component.ts
```diff
@@ -147,6 +163,51 @@ export type DomainValidationData = {
       />
     }
 
+    @if (showChallenge) {
+      <h3 tuiHeader="h6">{{ 'Certificate Authority' | i18n }}</h3>
+      <p>
+        {{
+          'Your certificate authority proves you control this domain by connecting to it on port 443, whatever port the address itself uses.'
```
dr-bonez:
```suggestion
          'Your certificate authority proves you control this domain by connecting to it on port 443, regardless of which port the address itself uses.'
```

projects/start-os/web/ui/src/app/routes/portal/components/interfaces/addresses/gateway/domain-validation.component.ts
```diff
@@ -147,6 +163,51 @@ export type DomainValidationData = {
...
+        [result]="challengeResult()"
+        [loading]="challengeLoading()"
+        [local]="false"
+        (test)="testChallenge()"
+      />
+
+      @if (gua) {
+        <p>
+          {{
+            'Over IPv6 there is nothing to forward — your gateway firewall must allow inbound connections to this address instead.'
```
dr-bonez:
```suggestion
            'Over IPv6 there is nothing to forward — your gateway firewall must allow inbound connections to this port instead.'
```

projects/start-os/docs/src/cli-reference.md
```diff
@@ -475,6 +475,19 @@ Remove ACME certificate configuration.
 
 - `--provider <PROVIDER>` — ACME provider to remove (required)
 
+### `start-cli net acme check-challenge <FQDN> <GATEWAY> <PORT>`
+
+Test whether a certificate authority can reach the domain to validate it. An
+ACME authority connects on port `443` whatever port the address itself is served
+on, so this tests `443` — over IPv4, and over IPv6 where the gateway has a global
+address. Reports nothing when the domain needs no such test: it names no ACME
+authority, it is already served on `443`, or it holds a certificate with life
```
dr-bonez: being already served over 443 does not seem like a good reason to short-circuit the check on the backend. seems like it should be on the frontend to deduplicate the request in that scenario

shared-libs/crates/start-core/src/net/host/mod.rs
```diff
@@ -63,6 +68,11 @@ pub struct PortForward {
     /// existing single-port `PortForward`s.
     #[serde(default = "default_port_forward_count")]
     pub count: u16,
+    /// Whether devices on your own network dial this forward too. False for one
+    /// only a certificate authority dials: no address is served on that port,
+    /// so whether the router hairpins it back says nothing.
+    #[serde(default = "default_port_forward_local")]
```
dr-bonez:
```suggestion
    #[serde(default = "const_true")]
```

helix-nine: Done — `const_true` in `util::serde`, which three other types already use. Dropped my `default_port_forward_local`.

projects/start-os/web/ui/src/app/routes/portal/components/interfaces/addresses/gateway/domain-validation.component.ts
helix-nine: Applied. It is an i18n key, so the change is the key text in `en.ts` plus the template; the four translations already read "regardless of which port" (`unabhängig davon, welchen Port`, `sea cual sea el puerto`, `quel que soit le port`, `niezależnie od portu`), so they needed nothing.

projects/start-os/web/ui/src/app/routes/portal/components/interfaces/addresses/gateway/domain-validation.component.ts
helix-nine: Applied — and the translations did need it here, since all four said "address": `zu diesem Port`, `a este puerto`, `vers ce port`, `do tego portu`.

projects/start-os/docs/src/cli-reference.md
helix-nine: Agreed, and it went further than deduplicating the request — that port was the only thing `port` decided, so removing the rule removed the parameter. `check_challenge` now takes an fqdn, a gateway and an authority, and answers the question it was asked: is the authority reachable, and does this domain still need it to be.

Both callers already knew. `add_public_domain` runs `check_port(ext_port)` a few lines earlier, which *is* the challenge probe when `ext_port` is 443, so it reuses that instead of asking twice. The web client already had the guard — a second review round had flagged it as duplicating the backend rule, which it was; now it is the only copy.

Knock-on: the `PORT` positional is gone from `net acme check-challenge`, so the man page and `cli-reference.md` are regenerated/updated and the `help.arg.address-port` key I added for it is removed.

shared-libs/crates/start-core/src/net/acme.rs
```diff
@@ -554,6 +559,106 @@ pub fn acme_api<C: Context>() -> ParentHandler<C> {
...
+#[group(skip)]
+#[serde(rename_all = "camelCase")]
+#[ts(export)]
+pub struct CheckChallengeParams {
+    #[arg(help = "help.arg.fqdn")]
+    pub fqdn: InternedString,
+    #[arg(help = "help.arg.gateway-id")]
+    pub gateway: GatewayId,
+    #[arg(long, help = "help.arg.acme-provider")]
+    pub acme: Option<AcmeProvider>,
```
dr-bonez: same issue here. why is this optional when it turns the command into a no-op if not present? that's a footgun. the caller should just *not call it* when there's no acme provider. save a round trip

helix-nine: Right, and it was the same mistake twice — I had the backend absorbing two things the caller already knew. `acme` is required now, and both callers skip the call instead of asking a question with a foregone answer: `add_public_domain` folds it into the check it already had for a domain served on 443, and the web client into the `!ctx.acme` guard it already had. So the round trip is saved in both places rather than answered with `null`.

That leaves `check_challenge` with exactly one reason to report nothing — the certificate on hand still has life in it — which is the only one it is positioned to know, since the store it reads never reaches the browser.

`--acme` is required in the man page and `cli-reference.md` now. One thing I left alone but will change if you want it: its siblings `net acme init` and `net acme remove` both spell this `--provider`, so `--acme` under `net acme` is the odd one out. Renaming it is a one-line change to the params struct plus the regenerated artifacts.

shared-libs/ts-modules/start-core/lib/osBindings/PortForward.ts
```diff
@@ -13,4 +13,10 @@ export type PortForward = {
    * existing single-port `PortForward`s.
    */
   count: number
+  /**
+   * Whether devices on your own network dial this forward too. False for one
+   * only a certificate authority dials: no address is served on that port,
+   * so whether the router hairpins it back says nothing.
```
dr-bonez:
```suggestion
   * Whether this forward receives local traffic. Used to gate hairpinning check.
```

helix-nine: Applied verbatim — at the Rust `///` on `PortForward::local` rather than in the binding, so the regeneration keeps it.

Took the same wording to the two places the frontend restates that field (`PortCheckWarningsComponent.local`, `PortForwardRow.local`), and dropped the comment on `PortCheckTestComponent.local` entirely — it is a passthrough, and the input it feeds now carries the fact.

shared-libs/crates/start-core/src/net/host/address.rs
```diff
@@ -581,13 +585,29 @@ pub async fn add_public_domain<Kind: HostApiKind>(
                 port: ext_port,
                 gateway: gateway.clone(),
             },
-        )
+        ),
+        // TLS-ALPN-01 is validated at 443, whatever port the address serves.
```
dr-bonez:
```suggestion
        // TLS-ALPN-01 is validated at 443, regardless of which port the address serves.
```

shared-libs/crates/start-core/src/net/host/mod.rs
```diff
@@ -464,6 +470,16 @@ impl Model<Host> {
                 let Some(wan_ip) = ip_info.wan_ip else {
                     continue;
                 };
+                // A certificate authority validates at `ACME_CHALLENGE_PORT`,
+                // whatever port the address serves. `add_ssl` excludes a
```
dr-bonez:
```suggestion
                // regardless of which port the address serves. `add_ssl` excludes a
```

shared-libs/crates/start-core/src/net/host/address.rs
helix-nine: Applied.

shared-libs/crates/start-core/src/net/host/mod.rs
helix-nine: Applied. Swept the branch for the same phrase while I was here — it was also in two sentences in `clearnet.md`, so both take "regardless of which port" too. One of those two predates this branch (it is from #3734, in a paragraph this PR was already editing); say the word if you would rather I left that one alone.

dr-bonez (approved):

Issue #3776, opened 2026-08-21
vhost: a container that accepts TCP but never completes TLS parks the client's handshake forever


helix-nine:
Found while reviewing the fix for #3739, which edits the await this is about. Pre-existing, and unchanged by that fix.

`net/tls.rs` bounds phase 1 (reading the ClientHello) at 5s and phase 3 (the client handshake) at 15s. Phase 2 is bare:

```rust
// Phase 2: Resolve TLS config (scoped timeout so mid
// remains accessible for sending a TLS alert on timeout)
let hello = mid.client_hello();
let action = tls_handler.get_config(&hello, &metadata).await;   // no timeout
```

The comment claims a scoped timeout that is not there. Either the timeout or the comment should go — and the timeout looks like the one that was intended, given the two neighbours have one.

It matters because `ProxyTarget::preprocess` runs inside that call and does an unbounded `TcpStream::connect` followed by an unbounded TLS handshake to the container. A backend that accepts TCP and then never speaks TLS — a deadlocked server, a socket still accepting after a worker wedged — parks the browser's connection indefinitely rather than failing it. Measured against such a backend: still hung at 40s.

Each parked connection also holds a future in `TlsListener::in_progress` with two file descriptors, and nothing bounds that set: `MAX_PROXY_CONNS_PER_TARGET` is applied in `handle_stream`, which is downstream of the handshake, so the pre-handshake stall is ungated.

Handshakes are concurrent (`FuturesUnordered` in `poll_accept`), so one stalled connection does not block others — the cost is the unbounded accumulation and the browser that never gets an answer.

Issue #3775, opened 2026-08-21
vhost: run_http2_proxy advertises extended CONNECT without checking the backend, and skips GOAWAY on an upstream error


helix-nine:
Found while reviewing the fix for #3739. Both are pre-existing defects in `run_http2_proxy`; filing separately because that fix is what makes them reachable.

`run_http2_proxy` needs `http_aware` (`vhost.rs`) **and** a client-facing ALPN of `h2` (`http.rs:265`). Today the second condition is essentially never met for an `https`/`wss` binding, because #3739 leaves the client-facing ALPN empty. Once #3739 lands, every `protocol: 'https' | 'wss'` binding whose container advertises `h2` — nginx with `http2 on`, Caddy, Traefik, Go's `ListenAndServeTLS` — takes this path.

## 1. Extended CONNECT is advertised unconditionally

`http.rs` builds the client-facing h2 server with `.enable_connect_protocol()`, which sets `SETTINGS_ENABLE_CONNECT_PROTOCOL=1` and tells the browser it may open a WebSocket as an RFC 8441 extended CONNECT. Nothing checks whether the backend supports it:

- hyper's `client::conn::http2` exposes `is_extended_connect_protocol_enabled()`; `run_http2_proxy` never calls it.
- hyper's `proto/h2/client.rs` forwards the `Protocol` extension unconditionally.
- h2's `Pseudo::request(method, uri, protocol)` does not gate on the peer's SETTINGS.
- h2's own server side rejects `:protocol` when the setting is off, with RST_STREAM.

So against a backend with no RFC 8441 support (nginx has none), the browser opens `wss://…` as extended CONNECT, we forward `:protocol`, the backend resets the stream, and the WebSocket dies. Gating `.enable_connect_protocol()` on the backend's advertised support would make the browser fall back to an HTTP/1.1 connection for the WebSocket, which works.

Worth confirming what `is_extended_connect_protocol_enabled()` reports at the moment the client handshake resolves — if the peer's SETTINGS has not arrived yet it will read `false`, which is the safe direction but would also decline extended CONNECT for backends that do support it.

## 2. An upstream error tears down the client connection with no GOAWAY

```rust
res = to.as_mut() => {
    res?;
    // GOAWAY plus drain, so in-flight streams still finish.
    from.as_mut().graceful_shutdown();
}
```

[...]

PR #3777, opened 2026-08-21
fix(start-os): reflect the upstream's ALPN across an add_ssl rewrap

### LINKED ISSUE #3739: vhost: upstream ALPN reflection in ProxyTarget::preprocess is a no-op
Found while working on #3731 / #3733 and noted at the bottom of that PR; filing it so it isn't lost now that the PR is merged. It is **not** a regression from #3733 — it has been this way since #3037.

## The reflection appends nothing

`ProxyTarget::preprocess` dials the backend during the TLS ClientHello and, on the `connect_ssl: Ok(_)` branch, reflects the upstream's negotiated ALPN into the client-facing `ServerConfig`:

```rust
// shared-libs/crates/start-core/src/net/vhost.rs:1149-1157
let target_stream = TlsConnector::from(Arc::new(client_cfg))
    .connect_with(
        ServerName::IpAddress(self.addr.ip().into()),
        tcp_stream,
        |conn| {
            prev.alpn_protocols
                .extend(conn.alpn_protocol().into_iter().map(|p| p.to_vec()))
        },
    )
```

`connect_with`'s closure is a **pre-handshake configuration hook, not a post-handshake inspection hook**. `tokio-rustls-0.26.4/src/client.rs`'s `connect_impl` is a non-async fn that does `ClientConnection::new_with_alpn(...)` → `f(&mut session);` → `Connect(MidHandshake::Handshaking(...))` — so the closure has already run by the time `connect_with` *returns*, before the future exists, before a single handshake byte moves. At that point `conn.alpn_protocol()` is unconditionally `None` (rustls initialises `alpn_protocol: None` and sets it only while processing ServerHello/EncryptedExtensions), so the `extend` appends nothing, on every connection.

Measured against the pinned versions (tokio-rustls 0.26.4, rustls 0.23), mirroring those lines against a backend advertising `[h2, http/1.1]`:

```
[backend] negotiated ALPN with the proxy:            Some("h2")
[probe]   closure ran before any handshake I/O:      true
[probe]   conn.alpn_protocol() inside the closure:   None
[probe]   prev.alpn_protocols (client-facing config): []
```

The upstream genuinely negotiated `h2` and the client-facing config still came out empty. `git log -S 'prev.alpn_protocols' -- '*vhost.rs'` returns exactly one commit, so there is no version in which this worked.

## Why it may be worth more than deleting three lines

On this branch the client-facing config therefore carries an **empty** ALPN list, so we send no ALPN extension, `tls_info.alpn` is `None`, and `run_http_proxy` always selects `run_http1_proxy`. Meanwhile the *upstream* leg is offered the client's entire ALPN list (`client_cfg.alpn_protocols = hello.alpn()...`), so the backend is free to select `h2`.

That combination — h1 framing written onto a connection the backend believes is h2 — is the part I have **not** reproduced end to end, and it is the reason I'm filing rather than just sending a deletion. Worth confirming against a real h2-capable backend behind an `https`/`wss` binding before deciding whether the fix is "drop the dead closure" or "actually reflect, after the handshake".

Note the other two branches need no dial at all for ALPN: `AlpnInfo::Reflect` copies the ClientHello's own list, `AlpnInfo::Specified` is static config.

helix-nine:
## What was wrong

`ProxyTarget::preprocess` read the backend's negotiated ALPN from the closure it
passed to `TlsConnector::connect_with`. rustls runs that closure as a *configuration*
hook — `connect_impl` builds the `ClientConnection`, calls the closure, and only then
returns the future — so it ran before a single handshake byte moved,
`conn.alpn_protocol()` was unconditionally `None`, and the client-facing
`ServerConfig` was left with an empty ALPN list on every connection.

`git log` dates the closure to `9ff0128fb` (2023-07-14). It has been inert since the
day it landed; there is no tokio-rustls version under which it worked.

## What that cost

The empty list is only half of it. The upstream leg *is* offered the client's own
ALPN list, so a backend that speaks HTTP/2 selects `h2` — while the client, offered
nothing, settles on HTTP/1.1. `run_http_proxy` keys **both** legs off the client's
negotiation, so StartOS then wrote HTTP/1 requests onto a connection the backend was
reading as HTTP/2.

The issue flagged that combination as the part not yet reproduced. It is real, and
both halves are pinned by tests that fail on `master`:

- `net::vhost::upstream_alpn_tests::the_client_lands_on_the_protocol_the_backend_chose`
  drives a real ClientHello through `TlsListener` into the real `preprocess`, in front
  of a TLS backend advertising `[h2, http/1.1]`. On `master` it fails
  `left: None, right: Some("h2")` — the backend picked `h2`, the client got nothing.
- `net::http::tests::the_negotiated_alpn_frames_the_upstream_leg` shows what that
  `None` then writes upstream: `GET / HTTP/1.1`, where `Some("h2")` writes
  `PRI * HTTP/2.0`.

## The fix

`connect` instead of `connect_with`, and read `alpn_protocol()` off the connected
stream. The client handshakes on this config only after `preprocess` returns, so the
backend's choice is still in time to reach it — and because the backend was offered
the client's own list, whatever it chose is something the client asked for. No extra
round trip: that handshake was already awaited.

Every branch now **assigns** `prev.alpn_protocols` rather than appending. rustls picks
by the server list's order, so an entry inherited from the base config would outrank
whatever the branch chose. `Reflect` and `Specified` need that guard as much as the
rewrap does — more, in fact, since every `http` and `ws` binding resolves to
`Specified`. No config on this path carries an entry today, so those two commits are
**guards, not behavior changes**; `f695cc2f3` is the only one that alters what goes
over the wire. `a_protocol_on_the_base_config_does_not_outrank_the_backend` pins it
and fails if the assignment goes back to `extend`.

A failed upstream handshake now names the container address. It was logged bare, so an
ALPN mismatch printed `received fatal alert: NoApplicationProtocol` with no address,
service or SNI, while the client got an `unrecognized_name` alert — a failure whose
entire diagnostic trail pointed at certificates.

## No configuration that worked before stops working

The upstream leg's ALPN offer is unchanged by this diff, so the backend selects the
same protocol either way. That leaves two cases:

- **Backend selects `h2`** — previously the proxy wrote HTTP/1 framing onto it, so the
  binding was already broken for *all* traffic. Now both legs frame `h2`.
- **Backend selects `http/1.1` or nothing** — the client-facing handshake lands on the
  same protocol it did before. No change on the wire.

## Two gaps this makes reachable, neither introduced here

`run_http2_proxy` was previously all but unreachable for `https`/`wss` bindings and now
becomes their primary path. Review turned up two pre-existing defects in it, filed
separately rather than folded in here:

- **#3775** — it advertises RFC 8441 extended CONNECT to the client
  unconditionally, without checking `is_extended_connect_protocol_enabled()` on the
  backend, and its upstream-error path returns before `graceful_shutdown()`, so the
  browser gets no GOAWAY.
- **#3776** — `tls.rs`'s `get_config` has no timeout despite a comment
  claiming one, and `preprocess` does an unbounded TCP connect plus TLS handshake
  inside it.

Worth a look before merging, since they decide whether WebSockets survive on a binding
whose container advertises `h2`.

## Docs

`projects/start-sdk/docs/src/interfaces.md` told package authors that a rewrap
negotiates no ALPN and pointed them at passthrough for that reason. It now describes
what the rewrap carries, and warns that advertising `h2` commits the container to
serving WebSockets over RFC 8441 extended CONNECT — which is what #3775 is about, and
the reason a container that only speaks HTTP/1.1 should say so.

The section keeps a warning that `addSsl.alpn` is not where you configure this —
setting it drops the binding to a plaintext dial into a TLS listener — because the new
prose above it urges authors to think about which protocols to advertise, and that
option looks like where to say so.

The option's **type**, and the rewrap section's claim that `addSsl` + `secure.ssl` is
the whole condition, are corrections to already-published text and go to `live-docs`
(#3778) instead, which deploys on merge and is backported to master unattended. They
touch different lines than the warning above, so there is no double-edit to conflict.

## `addSsl.alpn` no longer decides how the container is dialled

Raised by @dr-bonez on #3778, where this was originally being written down as a
documentation gap.

`connect_ssl` carried two unrelated decisions in one `Result`: `Ok` meant *dial over
TLS* and also *carry the ALPN the container negotiated*; `Err` meant *dial plaintext*
and also *apply this strategy*. Setting `alpn` could only be expressed as `Err`, so it
silently moved the binding onto the plaintext dial — a container serving its own TLS
got cleartext on a port expecting TLS, and `upstreamCertValidation` was discarded with
it.

They are now separate fields: `connect_ssl: Option<Arc<ClientConfig>>` says only how to
dial and is derived from `secure.ssl` alone, and `alpn` says only what the client is
offered, defaulting to the container's choice over a TLS leg or the client's own list
without one. Every existing binding shape resolves to what it did before; the only
behaviour that changes is `alpn` on a TLS-serving container, which could not work.

`setting_alpn_does_not_stop_the_container_being_dialled_over_tls` watches the
container's own handshake complete — the client-facing list alone cannot tell a TLS
dial from a plaintext one, so a test that only checked it passed either way.

`addSsl.alpn` narrows the protocols the binding puts forward: a dialled container
chooses out of that narrower list, and the client is offered its choice, so one
protocol still frames both legs. Offering the container the client's list while telling
the client the pinned one would let them settle on different protocols — the mismatch
this PR exists to remove — and
`a_pinned_protocol_is_what_the_container_is_offered` reports what the container settled
on, which is the only thing that separates the two designs.

## Follow-ups

Two things review surfaced that are deliberately **not** here:

- `AddSslOptions::alpn` (`net/host/binding.rs`) is a bare field while its siblings carry
  `///` docs that ts-rs emits into the published binding. That is where an author meets
  this option in their editor.
- An upstream ALPN refusal logs at `error!`, and this PR documents that refusal as
  expected. It is client-triggerable, so a junk ALPN produces one error line per
  connection. Pre-existing; `warn!` would fit better.
- `interfaces.md`'s options table still types `addSsl.alpn` as `string`. #3778 owns that
  row, against `live-docs`; it is left alone here so the two do not collide.

Closes #3739


shared-libs/crates/start-core/src/net/host/binding.rs
```diff
@@ -499,6 +508,11 @@ pub struct AddSslOptions {
     /// forwarding them upstream. Setting this implies HTTP-aware proxying.
     #[serde(default)]
     pub add_x_forwarded_headers: bool,
+    /// Narrows the application protocols this binding puts forward. Where the
+    /// container serves its own TLS it is offered the ones the client also
+    /// asked for, and the client is offered whatever it picks; otherwise the
+    /// client is offered these directly. Unset and `reflect` both put the
+    /// client's own list forward.
```
dr-bonez: this comment is inscrutable

shared-libs/crates/start-core/src/net/vhost.rs
```diff
@@ -392,7 +392,8 @@ impl VHostController {
             addr_v6: None,
             add_x_forwarded_headers: false,
             auth: None,
-            connect_ssl: Err(AlpnInfo::Reflect),
+            connect_ssl: None,
+            alpn: None,
```
dr-bonez, line 396: why was this split? what would it mean to include both?

shared-libs/crates/start-core/src/net/host/binding.rs
helix-nine: Fair. Cut to one sentence in 286089ad0, with `reflect`'s meaning moved onto the variant where it lives:

```rust
/// Application protocols to put forward, narrowed to the ones the client
/// also asked for. Unset puts the client's own list forward.
pub alpn: Option<AlpnInfo>,
```

The old version was trying to explain both arrangements (rewrap vs. StartOS terminating alone) in a field doc. That belongs in the book, which already covers it.

shared-libs/crates/start-core/src/net/vhost.rs
helix-nine: Including both is the case the old type could not express, and it is the bug.

On master the field is one value carrying two decisions:

```rust
pub connect_ssl: Result<Arc<ClientConfig>, AlpnInfo>, // Ok: yes, connect using ssl, pass through alpn; Err: connect tcp, use provided strategy for alpn
```

so "dial the container over TLS" and "which protocols to put forward" are the same choice — `Ok` means TLS **and** reflect, `Err` means plaintext **and** this list. The derivation reads the `alpn` test first:

```rust
if let Some(alpn) = ssl.alpn.clone() {
    Err(alpn)
} else if bind.options.secure.as_ref().map_or(false, |s| s.ssl) {
    Ok(ctrl.upstream_client_config(&ssl.upstream_cert_validation))
}
```

So setting `alpn` on a binding whose container serves its own TLS silently downgraded the dial to plaintext, and the container got cleartext on a port expecting TLS. There was no way to ask for both, because the type forbade it.

Both means: dial the container over TLS, and put only these protocols forward. That is `secure: { ssl: true }` + `addSsl.alpn` — rewrap the container's TLS while keeping the binding off a protocol it would otherwise select, e.g. holding an h2-capable container to `http/1.1`. All four combinations are now reachable and each means something:

| `connect_ssl` | `alpn` | |
|---|---|---|
| `None` | `None` | StartOS terminates alone, client's own list reflected |
| `None` | `Some` | StartOS terminates alone, client offered the pin — this is `protocol: 'http'` |
| `Some` | `None` | rewrap, container offered the client's list — this is `protocol: 'https'` |
| `Some` | `Some` | rewrap, container offered the pin ∩ the client's list — unrepresentable before |

`alpn_does_not_decide_whether_the_container_is_dialled_over_tls` pins the split itself: it asserts the dial is the same for `None`, `reflect`, and a pin.

The site you're on is a manual SNI demux, which is a passthrough — `passthrough: true`, so `preprocess` discards the config either way. Both fields are `None` there because neither applies, where master had to write `Err(AlpnInfo::Reflect)` to say the same thing.

dr-bonez: ok so it becomes a filter? just say that

helix-nine: Yes. Said it, in 9d3487f8f:

```rust
/// Filters the application protocols the client asked for down to the ones
/// this binding puts forward. Unset filters nothing.
pub alpn: Option<AlpnInfo>,
```

and on the variants, which is where the two cases belong:

```rust
pub enum AlpnInfo {
    /// Filter nothing.
    Reflect,
    /// Keep only these.
    Specified(Vec<MaybeUtf8String>),
}
```

`ProxyTarget::alpn` and the book's paragraph now use the same framing, and the SDK binding is regenerated.

Sorry for the wall of text — you asked why the field was split and I answered with the whole derivation when the one-line version was the point.

shared-libs/ts-modules/start-core/lib/osBindings/AddSslOptions.ts
```diff
@@ -11,6 +11,10 @@ export type AddSslOptions = {
    * forwarding them upstream. Setting this implies HTTP-aware proxying.
    */
   addXForwardedHeaders: boolean
+  /**
+   * Filters the application protocols the client asks for. Unset filters
+   * none of them.
+   */
   alpn: AlpnInfo | null
```
dr-bonez, line 18: if `null` == `"reflect"` we shouldn't have both

helix-nine: Agreed — and they're only the same because of this PR. Before it, `'reflect'` also moved the container onto a plaintext dial, so the two values did differ; the fix is what made one of them redundant.

Dropped `reflect` in 56c5fc2c7, keeping the value that needs no value:

```ts
export type AlpnInfo = { specified: Array<MaybeUtf8String> }
```

so the option is a filter or nothing. `null` is the only way to say "filter nothing", and `AlpnInfo::default()` — which returned `Reflect` and had no callers on master either — goes with it.

Two things worth your eye:

**It is a manifest-visible break.** Nothing in-tree sets `reflect` — `knownProtocols` uses `{ specified: ['http/1.1'] }` or `null`, the embassy shim uses `{ specified: [...] }`, and the OS's own bindings pin a list — but a package that hand-wrote `'reflect'` would now fail to deserialize its binding. Anything that did was on the plaintext-dial path this PR changes anyway. Logged it in the SDK changelog under 2.0.10 with the one-line migration.

**The type row in the book needs your call on ordering.** #3778 fixes `interfaces.md:177` on `live-docs`, where `'reflect'` is still a real value of the released software, and `docs-backport.yml` then pushes that commit to master — where, after this lands, the value will not exist. I have left row 177 alone here rather than write a second version of the same line and hand you a backport conflict. Either #3778 merges first and a follow-up narrows the row on master, or #3778 takes the narrowed type now and the published book is briefly ahead of the published OS. I would rather you pick than guess.

dr-bonez: search all packages in Start9Labs ans Start9Community org. is anyone using this field?

helix-nine: Searched both orgs. **Nobody sets it to anything but `null`, and nothing anywhere uses `'reflect'`.**

Method: enumerated every non-archived repo in `Start9Labs` and `Start9-Community` via the API — 204 repos, 144 of them `*-startos` — pulled each one's default-branch tarball and grepped the whole tree. Zero fetch failures. (Tarballs rather than `gh search code`, whose index has silently covered only part of the fleet before.)

Eight package repos mention `alpn` in `startos/`, and every one of them writes `null`:

| repo | |
|---|---|
| `Start9-Community/electrs-startos` | `startos/interfaces.ts:14` `alpn: null` |
| `Start9-Community/frigate-startos` | `startos/interfaces.ts:14` `alpn: null` |
| `Start9Labs/coturn-startos` | `startos/interfaces.ts:33`, `:83` `alpn: null` |
| `Start9Labs/fulcrum-startos` | `startos/interfaces.ts:14` `alpn: null` |
| `Start9Labs/lnd-startos` | `startos/interfaces.ts:58` `alpn: null` |
| `Start9Labs/owntracks-recorder-startos` | `startos/interfaces.ts:47` `alpn: null` |
| `Start9Labs/public-pool-startos` | `startos/interfaces.ts:38` `alpn: null` |

All eight are hand-rolled `addSsl` blocks on a `protocol: null` binding, where the field has to be written out. The other 136 packages never name it — they go through `protocol: 'http'`/`'https'` and let `knownProtocols` fill it in.

- **`'reflect'`: zero hits in either org.** The only occurrence of the string in 204 repos is this repo's own Rust.
- **`{ specified: [...] }`: zero hits in any package.** The only in-tree writers are `knownProtocols` (`http`/`ws`), the embassy shim, and the OS's own admin binding.
- Two false positives worth naming so the numbers reconcile: `public-pool-web-startos` matched a base64 PNG inside an SVG, and lnd's two `specified` hits are i18n prose. The remaining eleven hits are vendored upstream TLS code (`arti`, `async-acme`, `openwrt`, `DC-linux`, `fedimint`, `chama`, `start-wrt`).
- No open PR in either org touches the field except #3777 and #3778.

So the removal breaks nothing we own. It also confirms the collapse went the right way round: those eight explicit `alpn: null` sites are exactly what would have broken had we kept `'reflect'` and dropped the nullability instead.

Caveat on scope: this is default branches of the two orgs. A package in a contributor's personal repo that hand-wrote `'reflect'` would be outside it — though on the released OS that value puts a TLS-serving container on a plaintext dial, so any binding relying on it is already broken in the way #3739 describes.

dr-bonez: let's change the format then. `string[] | null` where null == reflect, and string[] gets lifted to `{ specified: string[] }`

helix-nine: Done in cb9040d34.

```ts
export type AlpnInfo = Array<MaybeUtf8String>
alpn: AlpnInfo | null
```

`#[serde(untagged)]` on the one-variant enum does the lifting: the wire carries the list, `Specified` is what it lands in, so nothing inside the OS had to change shape. Pinned by a round-trip test — `["h2","http/1.1"] ⟷ Some(Specified([...]))`, `null ⟷ None` — because that JSON is now a contract with every manifest.

Converted the in-repo writers: `knownProtocols` (×2), the embassy shim's two sites, and the OS's own admin binding.

**Bindings already on disk needed a migration**, which I nearly missed — `BindOptions` is persisted under `serverInfo.network.host` and each package's `hosts`, so a server updating with `{ specified: [...] }` in patch-db would have failed to load its own bindings. `v0_4_0_2` now rewrites them, walking the server's host and every package's; `down` re-wraps for a downgrade. A stored `"reflect"` becomes `null`, which is what it meant.

Two things worth knowing:

- **My first cut of the migration was not idempotent** — second run saw a list, found no `specified`, and wrote `null`. The house-style idempotency assertion in the test caught it. It now returns early on a list.
- **`v0_3_6_alpha_12` still writes the old shape** and I left it that way. It seeds a binding at the version where that was the format, and the chain converts it forward; rewriting history there would only make that migration's own tests describe a database it never produced.

Verified: 642 Rust tests, `tsc --noEmit` clean on both `shared-libs/ts-modules/start-core` and the container runtime (the latter needed `start-core`'s dist built first — its 686 errors before that were all unresolved-module noise, not this change).

Still your call on #3778's row, and the format change sharpens it: `live-docs` documents `'reflect' | { specified: string[] } | null` for the released OS, master will want `string[] | null`. I have left row 177 alone here rather than hand the backport a conflict.

projects/start-sdk/docs/src/interfaces.md
```diff
@@ -370,6 +370,16 @@ const origin = await multi.bindPort(443, {
...
+
+`addSsl.alpn` filters the protocols the client asks for. Unset, it filters none of them, and a filter naming nothing turns nobody away. A client whose whole list is filtered out is refused; a client that asks for no protocol at all is served as it would be without ALPN.
+
+Where your container serves its own TLS — `secure: { ssl: true }` alongside `addSsl` — StartOS offers it whatever survives the filter and hands the client back the one it picks, so both ends of the connection carry the same protocol; a container that ignores ALPN leaves the client without one. Where StartOS terminates alone, what survives the filter is what the client gets and your container is never asked, so the filter is the only thing holding a client to what your container speaks — which is why `protocol: 'http'` sets it to `http/1.1`. Set it on a binding StartOS terminates alone; on a rewrap leave it unset unless you need to keep the binding off a protocol your container would otherwise select.
```
dr-bonez: the perspective of this is weird. in the *common* case, the container isn't listening on TLS, so the OS dials the service in plaintext, and answers alpn *purely* based on this list. all of the framing of this is a filter between the client hello and the container, when very often the filter is between the client hello, and the response

helix-nine: You're right, and the cause is structural: I wrote this up inside **Rewrapping SSL to a TLS container**, so it came out describing the rare arrangement as if it were the shape of the option.

Moved it to its own `### Application protocols` under **TLS Termination**, and led with the common case (523326959):

> `addSsl.alpn` is the list of protocols StartOS will answer a client with, chosen from the ones that client asked for. Unset — the usual case — it answers with whatever the client asked for. A client left with nothing is refused; a client that asks for no protocol at all is served as it would be without ALPN.
>
> StartOS answers from that list itself, because it terminates the client's TLS and forwards plain HTTP. Your container never sees the negotiation, so the list is the only thing holding a client to a protocol your container can serve — which is why `protocol: 'http'` and `'ws'` set it to `http/1.1`.
>
> A container serving its own TLS does get a say; see [Rewrapping SSL to a TLS container](#rewrapping-ssl-to-a-tls-container).

The rewrap section keeps one sentence — that your container chooses from the list and StartOS answers with its choice — which is the only thing the rewrap adds. That also collapsed a duplicate: it had been said twice, once at the top of the section and once at the bottom.

Two more places carried the same perspective, so they take the client's side too:

- the reference row: *The ALPN protocols StartOS answers a client with, from those it asked for. `null`, the usual choice, answers with whatever it asked for.*
- `AddSslOptions::alpn`'s doc, which reaches a package author's editor through the generated binding: *The application protocols StartOS answers a client with, from those it asked for. Unset answers with whatever it asked for.* Bindings regenerated.

"Filter" is gone from all of them — it was your word from the earlier thread and I over-applied it. It describes the mechanism accurately and the perspective badly, which is what you're pointing at.

642 tests, `make start-core-ts-bindings-check` and prettier all clean.

shared-libs/crates/start-core/src/net/vhost.rs
```diff
@@ -1499,16 +1529,35 @@ impl ProxyContext {
     }
 }
 
+/// The protocols a binding keeps. Carried on the wire as the list itself.
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Deserialize, Serialize, TS)]
-#[serde(rename_all = "camelCase")]
+#[serde(untagged)]
```
dr-bonez: shouldn't be an enum anymore

helix-nine: Agreed — one variant carrying one list is a newtype wearing an enum's clothes, and `untagged` was only there to undo the wrapper the variant added. 1ed0908bd:

```rust
/// The protocols a binding answers with, carried on the wire as the list itself.
#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Deserialize, Serialize, TS)]
#[serde(transparent)]
#[ts(export)]
pub struct AlpnInfo(pub Vec<MaybeUtf8String>);
```

`AlpnInfo::Specified(v)` is `AlpnInfo(v)` at all thirteen sites, and the `Some(AlpnInfo::Specified(protos))` arm in `preprocess` is `Some(AlpnInfo(protos))`.

Nothing observable moved: `transparent` and `untagged` serialize a single-field wrapper identically, so the wire format is the same and the generated binding is still `export type AlpnInfo = Array<MaybeUtf8String>` — only the TSDoc line above it changed. The round-trip test is what holds that down, and it passes unchanged.

If you'd rather go one step further and drop the wrapper entirely — `alpn: Option<Vec<MaybeUtf8String>>`, with the binding inlining `Array<MaybeUtf8String> | null` — say so and I'll do it. I kept the named type because it is what `AddSslOptions` and `ProxyTarget` both refer to, and the name is the only thing left saying which list this is.

642 tests, `make start-core-ts-bindings-check` clean.

dr-bonez (approved):

PR #3778, opened 2026-08-21
docs(start-sdk): correct the addSsl.alpn type

helix-nine:
The table gives `addSsl.alpn` as `` `string` \| `null` `` with `'h2'` as the example. The field takes an `AlpnInfo` — `'reflect'` or `{ specified: [...] }` — so the documented value is not one it accepts. That is a plain error about published software, which is why it is here rather than on master.

The PR originally also documented the option's effect on how StartOS dials the container. @dr-bonez pointed out that is a bug, not a documentation gap, and he is right — so that half is dropped and fixed in code instead: #3777 splits the dial decision from the ALPN decision, so setting `alpn` no longer moves a TLS-serving container onto a plaintext dial.


projects/start-sdk/docs/src/interfaces.md
```diff
@@ -174,7 +174,7 @@ The key steps are:
...
 | `addSsl`                        | `object` \| `null`                                    | SSL termination options for HTTPS. Set to `null` for no SSL.                                                                                                                                                                |
-| `addSsl.alpn`                   | `string` \| `null`                                    | ALPN protocol negotiation (e.g., `'h2'`). Usually `null`.                                                                                                                                                                   |
+| `addSsl.alpn`                   | `'reflect'` \| `{ specified: string[] }` \| `null`    | Sets the ALPN protocols offered to the client. Usually `null`.                                                                                                                                                              |
```
dr-bonez: we should clarify what null means

helix-nine: Done in 960055bc6:

> The ALPN protocols this binding puts forward. `null`, the usual choice, puts none of its own forward.

One thing I deliberately did not write: that `null` and `'reflect'` mean the same thing. On the released OS they do not, and the difference is the bug in #3777 — `net_controller.rs` tests `alpn` before `secure.ssl`:

```rust
if let Some(alpn) = ssl.alpn.clone() {
    Err(alpn)
} else if bind.options.secure.as_ref().map_or(false, |s| s.ssl) {
    Ok(ctrl.upstream_client_config(&ssl.upstream_cert_validation))
}
```

so on `start-os/v0.4.0.1` writing `'reflect'` explicitly also moves a TLS-serving container onto a plaintext dial, while leaving it `null` does not. Equating them here would be wrong about the software this branch documents.

I also kept the cell to what is true of `null` in both arrangements rather than "offers whatever the client asked for" — that holds where StartOS terminates alone, but on a rewrap the released code offers the client nothing at all (the `connect_with` closure runs before the handshake, so there is no negotiated protocol to pass back). That is the other half of #3777, and it is why the description says what `null` does *not* do rather than promising a negotiation the current release doesn't perform.

Once #3777 ships, the fuller explanation of the option lands with it on master, and this row can gain the "same as `reflect`" equivalence it will have earned by then.

dr-bonez: the "put forward" language is very confusing. and make sure #3777 has the appropriate docs update

helix-nine: Both done.

**Here (8fce38cf8)** — "put forward" is gone; the row is the filter framing you asked for on #3777:

> Filters the ALPN protocols the client asks for. `null`, the usual choice, filters none of them.

**#3777 (05e9e1f1a)** — it already carried the book update for the new behaviour; I had left `interfaces.md:379-381` in the same confusing language, so that is rewritten around the filter too, and `puts forward` no longer appears anywhere in the file or in the field docs:

> `addSsl.alpn` filters the protocols the client asks for. Unset, it filters none of them, and a filter naming nothing turns nobody away. A client whose whole list is filtered out is refused; a client that asks for no protocol at all is served as it would be without ALPN.
>
> Where your container serves its own TLS — `secure: { ssl: true }` alongside `addSsl` — StartOS offers it whatever survives the filter and hands the client back the one it picks, so both ends of the connection carry the same protocol; a container that ignores ALPN leaves the client without one. Where StartOS terminates alone, what survives the filter is what the client gets and your container is never asked, so the filter is the only thing holding a client to what your container speaks — which is why `protocol: 'http'` sets it to `http/1.1`. Set it on a binding StartOS terminates alone; on a rewrap leave it unset unless you need to keep the binding off a protocol your container would otherwise select.

The type row itself stays only here, so the two branches don't both carry it and hand you a conflict on backport.

PR #3785, opened 2026-08-22
feat(start-os): send a bare service domain on to its HTTPS address

### LINKED ISSUE #3654: [feat]: Add HTTP→HTTPS redirect on port 80 for private domains (Wired / StartTunnel gateways)
### Prerequisites

- [x] I searched [existing issues](https://github.com/Start9Labs/start-technologies/issues?q=is%3Aissue) and this is not already requested.

### Project

StartOS

### Problem & use case

When accessing a service via a **Private Domain** on the LAN or over a VPN, entering the bare domain name (e.g., `cloud.mydomain.com`) without specifying `https://` causes the browser to default to HTTP (port 80). The request then hits the StartOS server dashboard instead of the intended service, because unencrypted HTTP lacks the TLS SNI hostname that StartOS uses to demux the correct backend [5]. Users must manually type `https://` or rely on perfect bookmarks every time, which is friction-heavy for households and breaks the “it just works” experience that public domains already enjoy via StartTunnel’s edge redirect [4].

### Proposed solution

Add an HTTP→HTTPS redirect listener on port 80 for the LAN/VPN ingress path. When StartOS receives an HTTP request on port 80 with a `Host` header matching any configured private domain (Wired or StartTunnel gateway), it should return a 301/302 redirect to the HTTPS equivalent, preserving the hostname and path. This would mirror the existing StartTunnel `HTTP Redirect (80 → 443)` feature that already works for clearnet public domains [4], extending the same behavior to private-domain contexts.

### Alternatives considered

- **Manual `https://` bookmarks:** Functional but poor UX for non-technical family members, and unreliable when links are typed from memory or generated by apps.  
- **Router-level redirect rules:** Could be done on the GL-X3000 or another edge device, but requires duplicating domain knowledge outside of StartOS and cannot dynamically discover which domains are registered as private on each VM.  
- **HSTS preload:** Not viable for private/local names that are not publicly preload-eligible, and does not help on the very first visit.  
- **Using public domains for everything:** Avoids the problem because StartTunnel handles the redirect, but it forces clearnet exposure for services that should remain LAN/VPN-only.

### Affected areas

Likely touches the StartOS web ingress layer (rust/web_server.rs based on prior troubleshooting traces [4]) and the gateway/service binding subsystem. The HTTP listener would need to reference the same domain↔service mapping used for TLS SNI demuxing on 443, ideally without building a second registration table.

### Anything else?

- Prior art exists within the Start9 stack: StartTunnel 1.2.x already auto-enables `HTTP Redirect (80 → 443)` for public domains [4].  
- This is especially painful for services that embed their own URLs in generated links (e.g., Nextcloud share links, Matrix client `.well-known`, Immich shares), because the same hostname must work seamlessly across LAN, VPN, and public internet. Currently, private-domain links copied from those apps fail on first click if the browser omits the scheme.

helix-nine:
Closes #3654

Typing a service's domain without a scheme now opens its web interface over HTTPS instead of the StartOS dashboard.

## What it does

The StartOS UI's port-80 router matches the request domain against installed services and returns a `307` to the service's HTTPS address, preserving path and query.

The redirect applies only when the destination is available:

- the connection arrived through a gateway assigned to the domain;
- the enabled binding and address serve a private or public domain;
- the interface declares HTTPS, or is a UI interface with no declared scheme.

Server-owned domains remain on the dashboard to avoid looping through its TLS proxy. IP addresses and the server's `.local` name also remain dashboard addresses. Port 443 is preferred; otherwise StartOS uses the lowest matching HTTPS port.

## Comment cleanup

- Replaced implementation narration with precise names for domain extraction, gateway resolution, HTTPS-port selection, and outer router layers.
- Kept short comments only for external or non-obvious constraints: request-target precedence, authority safety, unplaced gateway representation, the server redirect loop, and response-layer ordering.
- Rewrote the changelog and Private Domains guide in terms of present user behavior.

## Verification

- `cargo check -p start-core --all-targets`
- `cargo test -p start-core domain_redirect --features=test` (26 passed)
- `cargo test -p start-core lookup_tests --features=test` (4 passed)
- `cargo fmt --all -- --check`
- `npx --yes prettier@3.8.3 --check projects/start-os/CHANGELOG.md projects/start-os/docs/src/private-domains.md`

### DIFF
--- a/projects/start-os/docs/src/private-domains.md
+++ b/projects/start-os/docs/src/private-domains.md
@@ -4,7 +4,7 @@ A private domain works like your server's [mDNS address](mdns.md), except it als
 
 Private domains can be added to any gateway: wired (Ethernet), wireless (WiFi), or WireGuard ([StartTunnel](/start-tunnel/)). They can only be accessed when connected to the same network as your server — your LAN, or a VPN / tunnel — and they require trusting your server's Root CA.
 
-Type the domain on its own — `nextcloud.private` — and you land on the service over HTTPS. A browser given a bare name tries plain HTTP first, and your server answers that by sending the request on to the domain's HTTPS address. This applies to a web interface, on a network the domain was added on; a domain on an interface you reach with something other than a browser is left alone.
+Type the private domain on its own — `nextcloud.private` — to open the service over HTTPS. This works for service web interfaces on each network where the domain is assigned.
 
 ## Adding a Private Domain
 
--- a/projects/start-os/docs/src/private-domains.md
+++ b/projects/start-os/docs/src/private-domains.md
@@ -4,6 +4,8 @@ A private domain works like your server's [mDNS address](mdns.md), except it als
 
 Private domains can be added to any gateway: wired (Ethernet), wireless (WiFi), or WireGuard ([StartTunnel](/start-tunnel/)). They can only be accessed when connected to the same network as your server — your LAN, or a VPN / tunnel — and they require trusting your server's Root CA.
 
+Type the domain on its own — `nextcloud.private` — and you land on the service over HTTPS. A browser given a bare name tries plain HTTP first, and your server answers that by sending the request on to the domain's HTTPS address.
+
 ## Adding a Private Domain
 
 > [!NOTE]
--- a/projects/start-os/docs/src/private-domains.md
+++ b/projects/start-os/docs/src/private-domains.md
@@ -4,7 +4,7 @@ A private domain works like your server's [mDNS address](mdns.md), except it als
 
 Private domains can be added to any gateway: wired (Ethernet), wireless (WiFi), or WireGuard ([StartTunnel](/start-tunnel/)). They can only be accessed when connected to the same network as your server — your LAN, or a VPN / tunnel — and they require trusting your server's Root CA.
 
-Type the domain on its own — `nextcloud.private` — and you land on the service over HTTPS. A browser given a bare name tries plain HTTP first, and your server answers that by sending the request on to the domain's HTTPS address.
+Type the domain on its own — `nextcloud.private` — and you land on the service over HTTPS. A browser given a bare name tries plain HTTP first, and your server answers that by sending the request on to the domain's HTTPS address. This applies to a web interface, on a network the domain was added on; a domain on an interface you reach with something other than a browser is left alone.
 
 ## Adding a Private Domain
 
--- a/projects/start-os/docs/src/private-domains.md
+++ b/projects/start-os/docs/src/private-domains.md
@@ -4,8 +4,6 @@ A private domain works like your server's [mDNS address](mdns.md), except it als
 
 Private domains can be added to any gateway: wired (Ethernet), wireless (WiFi), or WireGuard ([StartTunnel](/start-tunnel/)). They can only be accessed when connected to the same network as your server — your LAN, or a VPN / tunnel — and they require trusting your server's Root CA.
 
-Type the private domain on its own — `nextcloud.private` — to open the service over HTTPS. This works for service web interfaces on each network where the domain is assigned.
-
 ## Adding a Private Domain
 
 > [!NOTE]
--- a/shared-libs/crates/start-core/src/net/domain_redirect.rs
+++ b/shared-libs/crates/start-core/src/net/domain_redirect.rs
@@ -7,8 +7,8 @@ use axum::body::Body;
 use axum::extract::Request;
 use axum::middleware::Next;
 use axum::response::Response;
+use http::Uri;
 use http::uri::{Authority, Scheme};
-use http::{HeaderValue, StatusCode, Uri};
 use imbl_value::InternedString;
 
 use crate::GatewayId;
@@ -16,11 +16,11 @@ use crate::context::RpcContext;
 use crate::db::model::DatabaseModel;
 use crate::net::gateway::GatewayInfo;
 use crate::net::host::binding::BindInfo;
-use crate::net::service_interface::{HostnameMetadata, ServiceInterfaceType};
+use crate::net::http::{https_redirect_uri, request_authority};
+use crate::net::service_interface::ServiceInterfaceType;
 use crate::prelude::*;
 
 const HTTPS_PORT: u16 = 443;
-
 const MAX_DNS_NAME_LEN: usize = 253;
 
 pub fn redirect_service_domains(ctx: RpcContext, router: Router) -> Router {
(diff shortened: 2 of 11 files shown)


projects/start-os/docs/src/private-domains.md
```diff
@@ -4,6 +4,8 @@ A private domain works like your server's [mDNS address](mdns.md), except it als
 
 Private domains can be added to any gateway: wired (Ethernet), wireless (WiFi), or WireGuard ([StartTunnel](/start-tunnel/)). They can only be accessed when connected to the same network as your server — your LAN, or a VPN / tunnel — and they require trusting your server's Root CA.
 
+Type the private domain on its own — `nextcloud.private` — to open the service over HTTPS. This works for service web interfaces on each network where the domain is assigned.
+
```
dr-bonez: this doesn't merit mentioning in the docs

shared-libs/crates/start-core/src/net/domain_redirect.rs
```diff
@@ -0,0 +1,618 @@
...
+                        }
+                    }
+                }
+                next.run(req).await
+            }
+        },
+    ))
+}
+
+/// An empty `GatewayId` represents an unplaced connection.
```
dr-bonez:
```suggestion
/// An empty `GatewayId` represents a connection from an unknown origin.
```

shared-libs/crates/start-core/src/net/domain_redirect.rs
```diff
@@ -0,0 +1,618 @@
...
+fn arrival_gateway_id(req: &Request) -> Option<GatewayId> {
+    req.extensions()
+        .get::<GatewayInfo>()
+        .map(|g| g.id.clone())
+        .filter(|id| !id.as_str().is_empty())
+}
+
+fn request_domain(req: &Request) -> Option<String> {
+    // RFC 9112 gives the request-target authority precedence.
```
dr-bonez: >request-target authority precedence

what?

shared-libs/crates/start-core/src/net/domain_redirect.rs
```diff
@@ -0,0 +1,618 @@
...
+    gateway: &GatewayId,
+    name: &str,
+    uri: &Uri,
+) -> Result<Option<Response>, Error> {
+    // Authority- and asterisk-form targets have no redirectable path.
+    if uri
+        .path_and_query()
+        .map_or(true, |p| !p.path().starts_with('/'))
+    {
+        return Ok(None);
```
dr-bonez: should still redirect to `/` in this case, no?

shared-libs/crates/start-core/src/net/domain_redirect.rs
```diff
@@ -0,0 +1,618 @@
...
+    }
+    // The server's mDNS name belongs to the dashboard.
```
dr-bonez: only on 443

shared-libs/crates/start-core/src/net/static_server.rs
```diff
@@ -62,6 +62,9 @@ pub trait UiContext: Context + AsRef<RpcContinuations> + Clone + Sized {
     fn extend_router(self, router: Router) -> Router {
         router
     }
+    fn apply_outer_layers(self, router: Router) -> Router {
+        router
+    }
```
dr-bonez, line 68: this seems redundant with `extend_router`

dr-bonez: got it, it's after the fallback, rather than before. that is unclear from the naming. needs at least a doc comment

projects/start-os/docs/src/private-domains.md
helix-nine: Removed the Private Domains book addition.

shared-libs/crates/start-core/src/net/domain_redirect.rs
helix-nine: Applied the suggested wording.

shared-libs/crates/start-core/src/net/domain_redirect.rs
helix-nine: Reworded this to the concrete case: absolute-form requests use the URI authority instead of `Host`. The precedence remains required by RFC 9112 and is covered by the existing conflicting-values test.

shared-libs/crates/start-core/src/net/domain_redirect.rs
helix-nine: Changed this to redirect both authority-form and asterisk-form targets to `/`, with a regression test.

shared-libs/crates/start-core/src/net/domain_redirect.rs
helix-nine: Moved the server mDNS exclusion after port resolution and limited it to port 443. A matching package address on another HTTPS port now redirects; the 443/non-443 distinction has a focused test.

shared-libs/crates/start-core/src/net/static_server.rs
helix-nine: Added `Applies layers after the UI fallback is installed.` to distinguish it from `extend_router`.

shared-libs/crates/start-core/src/net/domain_redirect.rs
```diff
@@ -55,7 +55,7 @@ fn arrival_gateway_id(req: &Request) -> Option<GatewayId> {
 }
 
 fn request_domain(req: &Request) -> Option<String> {
-    // RFC 9112 gives the request-target authority precedence.
+    // Absolute-form requests use the URI authority instead of `Host`.
```
dr-bonez: I didn't realize this was a thing. Seems like the port 443 redirect from the vhost controller should also handle this. speaking of - is there any code that can be reused between these components?

shared-libs/crates/start-core/src/net/domain_redirect.rs
```diff
@@ -82,41 +82,43 @@ async fn redirect(
...
+        .peek(|a| is_dashboard_mdns(name, port, &a.hostname.hostname));
+    if is_dashboard {
+        return Ok(None);
+    }
+    let path_and_query = uri.path_and_query().filter(|path| path.as_str() != "*");
+    https_redirect(path_and_query, name, port).map(Some)
+}
+
+fn is_dashboard_mdns(name: &str, port: u16, hostname: &str) -> bool {
+    port == HTTPS_PORT && name.strip_suffix(".local") == Some(hostname)
```
dr-bonez: I don't actually see the point of this. the redirect only happens when the bare domain is entered, without a port. it's a port 80 thing. for the .local, we shouldn't have any special handling at all, it should show the root CA installation page as it already does rn.

shared-libs/crates/start-core/src/net/domain_redirect.rs
helix-nine: Yes. I moved request-authority extraction and HTTPS URI construction into `net::http`; both the port-80 middleware and `handle_http_on_https` now use those helpers. The shared builder also sends asterisk/pathless targets to `/`.

shared-libs/crates/start-core/src/net/domain_redirect.rs
helix-nine: Agreed. I removed all server `.local` special handling; it retains the existing root-CA page behavior.

dr-bonez (approved):

PR #3787, opened 2026-08-22
feat(start-os): manifest-gated /dev/kvm access for service workloads

### LINKED ISSUE #3637: [feat]: allow manifest-gated /dev/kvm access for service workloads
## Summary

Allow a service that opts into `hardwareAcceleration` (or a future dedicated virtualization permission) to access the host's `/dev/kvm` device when present.

This is narrower than #2303: it does **not** ask StartOS to manage VMs or become a Proxmox-like hypervisor. It only extends the existing, manifest-gated device pass-through used for `/dev/dri`, `/dev/nvidia*`, and `/dev/kfd` so a confined service can run its own QEMU/KVM workload.

## Use case

Hermes can build and test Android applications inside its StartOS service container. The Android Emulator works with `-no-accel`, but x86_64 execution through QEMU TCG is extremely slow and is unsuitable for representative Macrobenchmark/Perfetto runs. The host already supports KVM, while the service container cannot see `/dev/kvm` even when its package sets `hardwareAcceleration: true`.

Other potential consumers include CI runners, device emulators, and Firecracker/QEMU-based development tools.

## Proposed behavior

- When the package opts in, add `/dev/kvm` to the existing hardware-acceleration device allowlist.
- Skip it cleanly on hosts where the device does not exist, matching current optional GPU-node behavior.
- Keep the service unprivileged, user-namespace mapped, and AppArmor-confined; do not grant broad host capabilities or VM-management APIs.
- Document that KVM exposes additional host-kernel attack surface and make the permission visible to users. A dedicated `hardwareVirtualization` flag would be preferable if `hardwareAcceleration` is intended to mean GPU access only.

## Acceptance criteria

1. An opted-in service sees `/dev/kvm` on a KVM-capable StartOS host and can open it.
2. A package without the opt-in cannot access `/dev/kvm`.
3. Installation/startup still succeeds on hosts without KVM.
4. Existing GPU, FUSE, and TUN device handling is unchanged.
5. Tests cover present, absent, and non-opted-in device cases.

## Validated implementation direction

The current LXC implementation already funnels manifest-gated host devices through `handle_devices`. Adding `/dev/kvm` to the relevant allowlist is sufficient at the code level and the resulting `start-core` change compiles. No live StartOS binary or host configuration was modified while validating this direction.

helix-nine:
Closes #3637

## Summary

Services can opt into a dedicated `hardwareVirtualization` manifest grant. On a host with KVM, StartOS exposes `/dev/kvm` through the existing confined device-node path; without the flag, no KVM device is granted. Hosts without `/dev/kvm` still start the service normally.

The grant adds no capabilities or VM-management API. The device retains its host mode and is owned by mapped container root, so the process opening it must run as container root. The packaging book documents the availability, fallback, permissions, and kernel attack surface.

The field defaults to false in the SDK, Rust compatibility layer, and legacy v1 conversion. Generated bindings, registry/UI fixtures, and StartOS, Start SDK, and start-registry changelogs are aligned.

## Verification

Automated:

- `cargo test -p start-core --features=test` — 657 passed, 3 ignored
- old-manifest default and device-selection regressions
- generated TS bindings
- UI, brochure, marketplace, and container-runtime checks
- formatting and docs build

Runtime acceptance on a KVM-capable `startos-0.4.0` VM upgraded to branch commit `c4147c99a`:

- packed two otherwise-identical x86_64 diagnostic services with the branch-built SDK and `start-cli`
- `hardwareVirtualization: true`: `/dev/kvm` appeared as a character device with mode `0660`, major/minor `10:232`; opening it as container root succeeded; `KVM_GET_API_VERSION` returned `12`
- `hardwareVirtualization: false`: `/dev/kvm` was absent (`ENOENT`)

The host-installed older `start-cli` silently dropped the new manifest field while repacking, so the final artifacts were packed with the branch-built `start-cli`; inspection confirmed the raw manifests carried `true` and `false` respectively before sideloading.


shared-libs/crates/start-core/src/lxc/mod.rs
```diff
@@ -49,6 +49,40 @@ const USERSPACE_FILESYSTEM_PATHS: &[&str] = &["/dev/fuse"];
 // services (and slirp4netns / pasta networking for nested containers) — opted
 // in via `manifest.virtualNetworking`.
 const VIRTUAL_NETWORKING_PATHS: &[&str] = &["/dev/net/tun"];
+// /dev/kvm: the kernel hypervisor interface, for services that run their own
+// virtual machines (QEMU/KVM, Firecracker, device emulators) — opted in via
+// `manifest.hardwareVirtualization`.
+const HARDWARE_VIRTUALIZATION_PATHS: &[&str] = &["/dev/kvm"];
+
+/// The patterns `path` can still lead to, or `None` if the list narrowed to
+/// nothing. An empty list is unfiltered and matches everything.
```
dr-bonez: >An empty list is unfiltered and matches everything.

huh? that's not what I'm seeing

helix-nine: You're right that it reads wrong — my sentence was ambiguous about *which* list, and read the natural way it contradicts itself.

I wrote "`None` if **the list** narrowed to nothing" (the returned list) and then "an empty **list** matches everything" meaning the `matches` parameter. Same word, two different things, one sentence apart — so it parses as "empty ⇒ None" and "empty ⇒ everything" at once.

The underlying fact is about the *input*, and it does hold. `matches.is_empty()` returns `Some(Vec::new())`, and in `handle_devices` the `Some(_)` arm is the act-on-it arm — descend the directory, or `mknod` the node in. Running the function verbatim:

```
empty `matches` input:
        /dev/kvm -> Some([])
        /dev/mem -> Some([])
        /dev/sda -> Some([])
  /dev/dri/card0 -> Some([])
narrowed to nothing (non-empty input, no hit):
        /dev/sda -> None
```

So `None` and `Some([])` both look "empty" but mean opposite things to the caller.

Fixed in fb6e4e89c by naming the parameter:

> The patterns from `matches` that `path` can still lead to, or `None` if none do. An empty `matches` filters nothing, so every path matches.

Happy to drop that second sentence entirely if you'd rather the doc not cover a branch no caller can reach — all four constants are non-empty, so it's unreachable today. It's in there because it's the one way to misuse the function after the extraction, but it's your call and I'd not argue for keeping it.

dr-bonez (approved):

PR #3791, opened 2026-08-22
feat(start-os): the server's name is its .local hostname

### LINKED ISSUE #3631: [feat]: collapse "server name" into the .local hostname

`ServerInfo` carries two fields for one identity:

- `name` — a free-form display label ("My Cool Server"), shown in the browser tab title
- `hostname` — the DNS label (`my-cool-server`), which is the Linux system hostname and the thing `.local` is derived from

They're coupled by `normalize()` / `denormalize()` (`shared-libs/crates/start-core/src/hostname.rs:91-133`): the UI collects one "Server Name" and derives the other. The model doesn't require this — `ServerHostnameInfo::new_opt` (`:144-161`) already accepts them as an independent pair — so the derivation is a UI convention layered on top.

Collapse them. **"Server name" becomes the `.local` hostname itself**, and the browser tab shows that name with the `.local` suffix stripped.

### Describe the Use Case

The name is presented as cosmetic and is anything but. `server-name.md:3` calls it "a custom label you choose," but it's the Linux system hostname: `set_hostname` (`hostname.rs:212-238`) writes `/etc/hostname` atomically, copies it to `/media/startos/config/overlay/etc/hostname` to survive the read-only squashfs, and `sync_hostname` restarts services after. The docs already carry a warning that renaming drops you off the address you're connected through.

Two spellings of one identity also means the UI can say "My Cool Server" while every address the user actually types says `my-cool-server.local`. Someone hunting for the second doesn't recognize the first.

And with per-gateway aliases landing (#3629), `.local` becomes one row in a list of hostnames on each gateway. Having that row alone derive from a differently-spelled display name makes it the odd one out in a list where every other entry is literally the name you typed.

- Remove `ServerInfo.name`; `hostname` is the single identity. Rust blast radius is small — `as_name()` / `as_name_mut()` in start-core only appear in `hostname.rs` (the hits in `tunnel/api.rs` and `registry/info.rs` are a WireGuard subnet name and the registry index name respectively).
- The "Server Name" field edits the `.local` hostname directly, surfacing `ServerHostname::validate` (`:31-49` — non-empty, lowercase, `[a-z0-9-]`) in the field rather than silently normalizing input away.
- Browser tab title = server name minus the `.local` suffix.
- `normalize()` / `denormalize()` and the `ServerHostnameInfo` pair collapse into `ServerHostname`.
- Setup wizard's name field becomes a hostname field with the same validation. The generated default is unchanged — `generate_hostname()` (adjective-noun from `src/assets/{adjectives,nouns}.txt`) still gives a fresh box a working `.local` with no user input.
- DB migration to drop `name`.
- `projects/start-os/docs/src/server-name.md` currently describes the derivation and needs rewriting, not just retitling.

**User-visible changes:** existing servers' tab titles change from "My Cool Server" to `my-cool-server`, and the field stops accepting spaces and capitals.

**Coupled pair to keep in sync:** the UI has its own `randomServerName()` (`server-name.dialog.ts:83`) mirroring the Rust `generate_hostname()`. If validation changes, both generators have to agree.

### Describe Alternatives

- **Keep both, decouple them** — let users set a display name unrelated to the hostname. Rejected: two names for one server is the source of the confusion, and the second one isn't an address you can reach the server by, so it earns nothing.
- **Keep `name`, drop `hostname`, derive on the fly.** Doesn't work — `hostname` is the system hostname with its own validation and its own persistence path; it can't be a derived view of a free-form label.

helix-nine:
Closes #3631

## Summary

The server's `.local` hostname is now its single user-visible identity. The setup wizard and admin UI edit that hostname directly, and the separate display-name field is removed.

Operator-entered hostnames are limited to 32 characters. Updates, restores, and drive transfers repair retained hostnames to the same limit.

## CLI breaking changes

- `setup execute --name` is removed.
- `server set-hostname` takes one required hostname.

The prospective start-cli version is therefore `2.0.0`; Cargo metadata, lockfile, changelog, and generated manpages are aligned.

## Migration safety

The StartOS `0.4.0.2` migration composes the existing admin-port and ALPN changes with display-name removal and hostname repair. Downgrade restores the legacy display name and ALPN wrapper. Migration, restore, and transfer tests cover these boundaries.

## Verification

- `cargo test -p start-core --features=test version::v0_4_0_2` — 15 passed
- Rust checks for start-cli, start-registry, StartOS, and StartTunnel
- generated bindings and all start-cli manpages; regeneration is byte-stable
- `npm run build:core`
- `npm run check`
- `npm run build:ui`
- `npm run build:setup`
- Start SDK bundle
- `make format-check`
- `git diff --check`


projects/start-os/web/setup-wizard/src/app/services/state.service.ts
```diff
@@ -80,22 +68,13 @@ export class StateService {
...
 
-  /**
-   * Called for fresh, restore, and transfer flows
-   * Password is required for fresh, optional for restore/transfer
-   */
-  async executeSetup(
-    password: string | null,
-    name: string,
-    hostname: string,
-  ): Promise<void> {
```
dr-bonez, line 91: seems you were over-aggressive in deleting comments in this file

AGENTS.md
```diff
@@ -147,6 +147,7 @@ Some pairs of files mirror each other by hand — nothing enforces them, so a ch
...
...y surface that states it.** `MAX_LEN` in [`shared-libs/crates/start-core/src/hostname.rs`](shared-libs/crates/start-core/src/hostname.rs) is bounded by the root CA's Common Name (`<hostname> Local Root CA`, and X.509 caps a Common Name at 64), and `hostname::test::the_longest_hostname_still_fits_the_root_ca_common_name` fails if the branding outgrows it. The number is then restated by hand in `MAX_LENGTH` in [`shared-libs/ts-modules/shared/src/util/hostname.ts`](shared-libs/ts-modules/shared/src/util/hostname.ts), the `Must be N characters or less` string in all five web dictionaries, `help.arg.hostname` in all five locales of `locales/i18n.yaml`, the `///` doc on `SetupExecuteCliParams::hostname` (which feeds its man page) and the one on `SetServerHostnameParams::hostname` (which feeds the TS binding; that command's man page comes from `help.arg.hostname`), [`projects/start-os/docs/src/server-name.md`](projects/start-os/docs/src/server-name.md), and — until each release is cut — the unreleased changelog entries stating it in `projects/start-os/CHANGELOG.md` and `projects/start-cli/CHANGELOG.md`. Move one and you must move all of them, or the form accepts a name the server rejects.
```
dr-bonez: cap it to 32. 50 is unreasonable

projects/start-os/web/setup-wizard/src/app/services/state.service.ts
helix-nine: Restored the four comments that carry state-flow contracts: kiosk initialization from the browser hostname, drive-state population sources, the kiosk keyboard prerequisite for locale navigation, and password optionality during restore/transfer.

AGENTS.md
helix-nine: Capped operator input and persisted hostnames at 32 characters across the backend, frontend, locales, docs, manpages, and generated bindings. The 0.4.0.2 migration, old-backup restore, drive transfer, and same-version startup now repair longer retained names to 32; boundary and end-to-end persistence tests cover each path.

shared-libs/crates/start-core/src/version/mod.rs
```diff
@@ -108,8 +109,9 @@ impl Current {
                 .result?;
             }
             Ordering::Equal => {
-                db.apply_function(|db| {
-                    Ok::<_, Error>((to_value(&from_value::<Database>(db.clone())?)?, ()))
+                db.apply_function(|mut db| {
+                    repair_current_hostname(&mut db);
+                    Ok::<_, Error>((to_value(&from_value::<Database>(db)?)?, ()))
```
dr-bonez: this is tech debt. there is no one on 0.4.0.2 already that has a >32char hostname

helix-nine: Removed the same-version repair path and its PatchDb regression test. Hostname repair is scoped back to the `0.4.0.2` migration module, which still trims migrated names to 32 characters.

dr-bonez (approved):

PR #3793, opened 2026-08-22
fix(start-os): isolate same-gateway address switches

### LINKED ISSUE #3628: [bug]: unconfigured domains and disabled addresses are still served over HTTPS (0.4.0.1)
### Prerequisites

### Server Hardware

Server Pure

### StartOS Version

0.4.0.1

### Client OS

Other

### Client OS Version

Tested using os x and graphene

### Browser

Other

### Browser Version

Vanadium, brave and chrome

### Current Behavior

The server appears to serve any request that reaches it, rather than only the addresses it has been configured to serve. Three cases, all over HTTPS, all returning the expected page with a certificate signed by my server's Root CA:

1. **A domain never configured on the server.** I added an A record on my router pointing an arbitrary domain at the server's LAN IP. That domain was never added to any gateway on the server. Browsing to it over HTTPS returns the StartOS login page, with a valid Root-CA-signed certificate for that name.

2. **An address disabled via its Interfaces toggle.** Toggling an address off on the Interfaces page (`System > StartOS UI`) does not stop that address being served. It remains reachable over HTTPS exactly as before.

3. **Every row on a gateway disabled.** Disabling all rows for my wired gateway also leaves the server reachable over HTTPS, both at a disabled addresses and at the unconfigured domain from case 1.

### Expected Behavior

In all three cases I expected the connection to be refused.

### Steps to Reproduce

1. On your router, add an A record for a domain that has never been added to the server (e.g. `notconfigured.example`) pointing at the server's LAN IP.
2. From a LAN device, browse to `https://notconfigured.example`.
   - Observed: StartOS login page, certificate signed by the server's Root CA.
   - Expected: connection refused.
3. Open `System > StartOS UI` and toggle one address row off. Let it settle and reload the page to confirm the toggle persisted.
4. Browse to that address over HTTPS.
   - Observed: still served.
   - Expected: connection refused.
5. Toggle off every row on the wired gateway.
6. Browse again, both to a disabled address and to the domain from step 1.
   - Observed: still served in both cases.
   - Expected: connection refused.

### Anything else?

_No response_

helix-nine:
Closes #3628

## Problem

#3692 fixed unconfigured hostnames being served by the StartOS UI, but a private IP disabled on the Interfaces page could remain reachable when another address on the same gateway was still enabled.

For SSL, the fallback route for bare IP addresses was built from gateways rather than individual IPs. One enabled address caused StartOS to add every IP on that gateway to the route, including IPs the user had disabled. The always-enabled `.local` address could trigger this even after every visible IP row was switched off.

Directly forwarded ports had the same control-plane bug: nftables rules already matched exact destination IPs, but the controller requested a rule for every IP on a gateway whenever any private address there was enabled.

## Fix

- Build the fallback SSL route and private direct-forward rules from enabled private IPs instead of expanding each enabled address to its entire gateway. This applies to single ports and port ranges.
- Treat an mDNS (`.local`) address as reachable only when a matching LAN IP is enabled on the same gateway, port, and SSL leg. The UI and SDK address selection now use the same rule.
- Keep exact SSL-route additions idempotent. A predecessor-aware replacement preserves active connections when only accepted IP addresses change; backend, authentication, TLS, or ALPN changes still replace the connection lifecycle.
- Keep a shared route alive until its last lease is dropped, while allowing one binding to diverge without changing another binding's route.
- Recognize the StartOS bridge's IPv6 address as an internal address.
- Repair server-hostname private-domain validation after the hostname type change merged on `master`.
- Document address-switch behavior and its remaining exceptions on the Interfaces page.

## Scope

Public IPv4 exposure remains gateway-wide because traffic forwarded by a router arrives under the gateway's local IPv4 address. On a non-SSL port, a domain and the IP rows that can carry it remain linked because the connection carries no hostname.

## Verification

- `make start-core-test` — 434 passed, 3 ignored
- focused private direct-forward tests — 2 passed
- focused vhost lifecycle tests — 7 passed
- focused vhost reconciliation tests — 2 passed
- `cargo check -p start-core`
- start-core TypeScript tests, including focused mDNS and address-enabled cases
- root TypeScript and i18n checks
- StartOS UI production build
- documentation site build
- formatting and `git diff --check`


projects/start-os/docs/src/interfaces.md
```diff
@@ -20,14 +20,20 @@ Each inbound gateway on your server has its own table. The rows in each table ar
...
+
+> [!NOTE]
+> Switching an IP address off stops it serving the interface over SSL. An SSL connection names the address it is for, so a [public domain](clearnet.md) or a [private domain](private-domains.md) on the same gateway is unaffected — each is an address in its own right, with its own row. Two cases keep an address answering anyway:
+>
+> - A gateway that has an enabled **public IPv4** address goes on serving every IPv4 address there. Your router presents traffic from the Internet under the same local address that gateway already has, so your server cannot tell a local connection from a forwarded one. Switch the public IPv4 address off as well to take that gateway's IPv4 addresses down.
+> - An address served without SSL goes on answering — a plain-HTTP address, and the addresses of a raw-TCP interface such as a P2P or ZMQ endpoint. Your server forwards a plaintext port for a whole gateway at once, so every address on that gateway shares one rule.
```
dr-bonez: "goes on answering" is is very unclear. I cannot tell just from reading this sentence when the ip would be listening and when it wouldn't for a non-SSL binding. that means this is a bad doc

projects/start-os/CHANGELOG.md
```diff
@@ -185,6 +185,10 @@ file tracks notable changes since the move to the monorepo.
   now serves its `.local` address, the domains you have assigned to it, and
   direct connections to its IP address.
 
+- **Interface address switches now apply to SSL connections.** Switching off a
```
dr-bonez: the same-gateway nature needs to be highlighted in the headline. this is a very rare circumstance, to have multiple ips for the same gateway, so the fact that this pr is fairly edge case should be clear

projects/start-os/docs/src/interfaces.md
helix-nine: Reworded this to state the exact condition: a non-SSL port is forwarded per gateway, remains reachable at every IP on that gateway while any private address using the port is enabled, and closes after the last such address is disabled. It also now says that rows sharing a non-SSL domain switch together. Updated in a6481719e.

projects/start-os/CHANGELOG.md
helix-nine: Updated the changelog headline, PR title, and PR description to foreground the same-gateway condition. The changelog now says that another IP or `.local` address on the same gateway could restore the disabled IP. I included `.local` because it is an always-enabled same-gateway contributor on an ordinary LAN gateway, so this was not limited to a manually configured multi-IP gateway. Updated in a6481719e.

projects/start-os/docs/src/interfaces.md
```diff
@@ -20,14 +20,20 @@ Each inbound gateway on your server has its own table. The rows in each table ar
...
+
+> [!NOTE]
+> Switching an IP address off stops it serving the interface over SSL. An SSL connection names the address it is for, so a [public domain](clearnet.md) or a [private domain](private-domains.md) on the same gateway is unaffected — each is an address in its own right, with its own row. Two gateway-level exceptions apply:
+>
+> - While a gateway has an enabled **public IPv4** address, every IPv4 address on that gateway accepts SSL connections. Your router presents traffic from the Internet under the same local address that gateway already has, so your server cannot tell a local connection from a forwarded one. Switch the public IPv4 address off to close SSL access through the gateway's other IPv4 addresses.
+> - For an interface served without SSL, StartOS forwards its port per gateway rather than per IP. While any private address using that port is enabled, the port accepts connections at every IP on that gateway. It closes on the gateway after the last such address is disabled. This applies to plain HTTP and raw-TCP interfaces such as P2P or ZMQ; rows that share a non-SSL domain are switched together.
```
dr-bonez: this doesn't seem inherent. we should be able to do nftables filters per-ip, no?

helix-nine: Yes. The nftables rule was already exact (`ip daddr $sip`); only the controller input was gateway-expanded. I changed both single-port and port-range forwarding to derive private destinations from enabled IPv4 rows directly, and removed the non-SSL exception from the docs. Non-SSL domain-linked rows still switch together because those packets carry no hostname. Updated in 683cca921.

shared-libs/crates/start-core/src/net/vhost.rs
```diff
@@ -327,6 +327,29 @@ impl VHostController {
         })
     }
 
+    #[instrument(skip_all)]
+    pub(super) fn replace(
```
dr-bonez, line 331: why can't this be handled with an idempotent add?

helix-nine: An exact-target add can be idempotent, and I restored that behavior in d9e3bf85e: re-adding the same target now returns the existing shared `Arc` lease instead of allocating another owner.

It cannot replace the predecessor-aware operation for a routing-only change, because the target key changes and several distinct targets on one host/port can have the same connection lifecycle. A predecessor-free add cannot tell which target’s `ProxyContext` should move. `replace` carries that identity: it transfers the context only when this is the predecessor’s last lease, leaves a shared predecessor intact when another binding still owns it, and joins an already-live destination when present.

I also changed reconciliation to borrow the old lease so `Arc::strong_count` is not inflated, and added cases for exact-add pointer identity, shared-route divergence, unrelated lifecycle-compatible targets, live/dead destinations, and failed replacement. `make start-core-test` passes (434 passed, 3 ignored) on the rebased head.

PR #3794, opened 2026-08-22
feat(ci): fail a pull request that carries conflict markers

### LINKED ISSUE #3730: Conflict markers can reach master — nothing checks for them and only Prettier fails, incidentally (.github/workflows/docs-backport.yml:240)
### What happens

`docs-backport.yml`'s fallback path deliberately commits unresolved conflict markers so a maintainer can resolve them in the PR ([`docs-backport.yml:228-241`](.github/workflows/docs-backport.yml#L228-L241) — *"One commit at a time, and every conflict resolved by keeping the markers"*, then `git add -A` and `cherry-pick --continue`). That design is fine on its own.

What is missing is a check that those markers cannot go any further. Nothing in CI looks for them. The only job that fails on such a PR is **Prettier**, and it fails incidentally: it parses a `>>>>>>> <sha> (subject)` line as blockquote syntax and wants to reformat it as `> > > > > > >`. The failure message says nothing about a conflict.

That means:

- The signal is indirect, so the failure reads as a formatting nit rather than "this branch contains conflict markers".
- Any file type Prettier does not cover, or any repo/path where the Prettier job does not run, has **no guard at all**.
- The same hole applies to a human who commits markers by hand; it is not specific to the backport bot.

Observed live on #3729, the backport of #3725: three files landed on the branch carrying `<<<<<<< HEAD` / `=======` / `>>>>>>> 88a3520c4`, and Prettier was the only red check.

A dedicated CI step fails on any conflict marker in the diff, with a message that says so. Roughly:

```bash
if git diff --unified=0 "$BASE"...HEAD | grep -nE '^\+(<{7} |={7}$|>{7} )'; then
  echo "::error::Conflict markers found in the diff — resolve before merging."
  exit 1
fi
```

Cheap, repo-wide, independent of file type, and it fails for the actual reason. It also gives the backport fallback a proper gate: the PR it opens is *meant* to be unmergeable until a human resolves it, and right now nothing enforces that.

- [`.github/workflows/docs-backport.yml:228-241`](.github/workflows/docs-backport.yml#L228-L241) — the fallback stages conflicted files with `git add -A` and continues the cherry-pick, by design, so markers are committed to the branch.
- [`.github/workflows/test.yaml:43`](.github/workflows/test.yaml#L43) — `prettier --check .`, the only job that currently fails on a marker-carrying branch, and only because of how `>>>>>>>` parses in Markdown.

No job anywhere greps for conflict markers.

### Logs & evidence

On #3729 before it was resolved, `prettier --check` reported only:

```
[warn] projects/start-sdk/docs/package-template/AGENTS.md
[warn] projects/start-sdk/docs/src/project-structure.md
[warn] projects/start-sdk/docs/src/service-to-service.md
[warn] Code style issues found in 3 files. Run Prettier with --write to fix.
```

Running `prettier --write` on those files turned `>>>>>>> 88a3520c4 (...)` into `> > > > > > > 88a3520c4 (...)` — i.e. Prettier's proposed "fix" would have silently normalised the marker into valid Markdown rather than flagging it.

Resolving the three conflicts also showed why an unmergeable-by-default gate matters: none of them was a "take one side" resolution, and on `service-to-service.md` taking the incoming side wholesale would have deleted a paragraph that exists only on master.

helix-nine:
Closes #3730

A dedicated check greps the files a pull request changes for a conflict marker, annotates every line it finds, and fails. It is file-type agnostic, so it also covers the trees `.prettierignore` excludes — `**/dist/`, `**/target/`, lockfiles, `__fixtures__`, `**/locales/` — which have no formatting gate at all today.

```
::error::Conflict markers found in the files this pull request changes — resolve them before merging.
::error file=projects/start-sdk/docs/src/service-to-service.md,line=66::Unresolved conflict marker.
```

### Why a separate workflow rather than a job in `test.yaml`

- `test.yaml`'s jobs all skip drafts. This one shouldn't: a branch carrying markers should say so from its first push, and the job is a grep.
- Making the fast `prettier` gate `needs:` it would put a checkout in front of the ~20s feedback loop that file was deliberately tuned for.
- `docs-backport.yml`, `docs-deploy.yml` and `docs-sync-on-tag.yml` are byte-identical on `master` and `live-docs`; `test.yaml` is not. A standalone file can be carried across the same way — see below.

### Detection

`$'^(<{7}|>{7}|\|{7})( |\r?$)|^={7}\r?$'` over each changed file — the opening and closing markers, the diff3 base marker, and the divider.

- **The divider is included, and that is a trade worth stating.** A conflict git writes always has all three markers, so the outer pair alone suffices for an unresolved file; the divider catches a half-resolution that deletes the fences and keeps the middle. The cost is that a bare `=======` line in a fenced example or a vendored licence fails the check, and since the scan is whole-file, it keeps failing for every later PR touching that file. The tracked tree on `master` and `live-docs` has no line matching any of these patterns today, and no `.rst`/`.adoc` at all — but if a page ever needs to show a conflict, indenting the example is the way out.
- **`$'…'` quoting is load-bearing**: it makes `\r` a carriage return. Without it a CRLF file's divider and any unlabelled marker slip through the `$` anchor — the repo tracks two CRLF text files today. Verified through the real runner, not just locally.
- **A file that cannot be scanned fails the check.** `xargs` reports "grep matched nothing" and "grep failed" with the same status, so the only evidence a file went unscanned is stderr; a non-empty stderr is an error rather than a pass. Markers are reported *before* that gate, so a run that hits both still names every marker it found.
- **`--binary-files=text` is load-bearing.** git decides a file is text from its first 8000 bytes and will happily write markers into it; grep sniffs a much larger buffer, so with the default a file whose first NUL sits past 8000 bytes is silently skipped — markers and all. Reproduced at 9000, 50000 and 98000 bytes.
- Non-regular paths and symlinks are filtered before grep. `[ -f ]` alone dereferences, so a symlink to a marker-carrying file would be reported a second time under the link's name; `[ ! -L ]` keeps each file reported once, and both tests together keep the five `docs/theme` symlinks and any dangling link from tripping the stderr gate.
- `filter: blob:none` keeps `fetch-depth: 0` from transferring ~150 MB of blobs on every push for a grep over the working tree; `--no-renames` keeps the diff off the blobs too, and yields the identical path set.

### Verification

- **Replayed the real failures.** At `7673938a7` (the marker-carrying commit of #3729) it names all three files; at `0d38d26be` (the human resolution) it passes. Against the real `backport/live-docs-7695ab1ff` it flags 260 marker lines across 14 files — including `.ts` files, which prettier's blockquote accident would never have caught.
- **Ran it on GitHub**, on a throwaway repo, with this file verbatim: a *draft* PR carrying markers fails with per-line annotations on the diff; a CRLF conflict reports all three markers including the divider; a clean PR touching a symlink-to-directory, a binary file and a filename with a space passes. Marking a draft ready keeps the failing check and fires no second run, which is why `ready_for_review` is not in `types:`.
- Also exercised locally: a divider-only half-resolution, a diff3 base marker, markers behind a NUL at 9000+ bytes, a symlink pointing at a marker-carrying file, an injected grep failure, markers *and* a scan error together, and an empty `base.sha`.
- actionlint, shellcheck and `prettier --check` are clean.

### Two things to know

- **The check is advisory.** The org ruleset on `master` has a `pull_request` rule but no required-status-checks rule, so a red `Conflict Markers` informs a reviewer rather than blocking the merge button — same as `Prettier` today. Making it required is a ruleset change, not a repo one.
- **Green is not the same as a clean backport, and a PR based on `live-docs` is not covered at all.** A file git cannot merge line by line — a page deleted on one side and edited on the other, or an image changed on both — comes through as one side with no marker, and the published change can be missing from the diff altogether; `CONTRIBUTING.md` and `AGENTS.md` both now say to compare a backport PR against `live-docs`. Separately, a workflow has to exist on the refs a pull request merges, and `live-docs` carries its own `.github/workflows` tree; the three most recent `live-docs` PRs have zero check runs. That leaves one route by which markers still reach `master` without a PR against it: merged into `live-docs`, cherry-picked cleanly, written to `master` by `createCommitOnBranch`. Closing it means carrying this file onto `live-docs` in a follow-up PR, after this lands — in that order, so the backport finds the commit already in master and no-ops. Happy to open it on the word.

One thing I deliberately left alone: `docs-backport.yml:6-8` carries the same inaccuracy this PR fixes in `CONTRIBUTING.md` ("A conflict is the one case that still needs a human" — there are four fallback reasons). That file is byte-identical on `master` and `live-docs`, so editing it here would break a mirror this PR cannot fix on both sides. It belongs with the `live-docs` follow-up.

A review pass also turned up a repo-wide CI gap while looking at this file: `edited` is not a default `pull_request` activity type, so retargeting a PR's base leaves every check's conclusion computed against the old base. It affects all seven `pull_request` workflows, so it belongs in one change across the set rather than here — filed as #3798, same shape as #3769.

<details>
<summary>Why not <code>git diff --check</code>?</summary>

git has its own leftover-marker detection and it does find all three markers, so it is the obvious alternative. I measured it rather than assuming, and it trades two things this check keeps:

- **It reports whitespace errors in the same breath.** Out of the box a single trailing space anywhere in the diff fails the gate alongside a real conflict. Suppressing that needs a hand-maintained `core.whitespace=-trailing-space,-space-before-tab,-indent-with-non-tab,…` list that has to stay complete as git adds checks — and if it ever falls behind, clean pull requests start failing for the wrong reason.
- **It is scoped to added lines, not to the changed files.** A marker already sitting in a file the pull request touches goes unreported. That is the narrower of the two scopes, and the wider one was a deliberate choice.

Neither is fatal, and on this repo today the two produce the same verdict on every branch. If you would rather have the smaller step and take those trades, it is a clean swap — but it is a decision to make with both losses visible, not a strictly better version of this.
</details>


AGENTS.md
```diff
@@ -54,6 +54,7 @@ Each product lives under `projects/` as a thin wrapper; the bulk of the code liv
...
+- **A backport PR carrying conflict markers is still reported mergeable.** The markers are ordinary file content, so [`conflict-markers.yml`](.github/workflows/conflict-markers.yml) fails any pull request into master whose changed files carry one, and names every line in the run log. A green check is not a clean backport: a file git cannot merge line by line comes through as one side with no marker, and the published change can be missing from the diff altogether, so compare the PR against `live-docs` before merging it.
```
dr-bonez: what?

helix-nine: Fair — it was three ideas stacked in a branch-policy bullet: that markers don't block a merge, that a check now catches them, and that a green check still isn't a clean backport. Only the last one is a rule an agent can't learn from the check itself, so I cut it to that:

> - **A green `Conflict Markers` check does not mean a backport is clean.** Where git cannot merge a file line by line — a page deleted on one side, an image changed on both — the backport keeps one side and leaves no marker, so compare the PR against `live-docs` before merging it.

Pushed as 012c894. If you meant the bullet shouldn't be here at all, say the word and I'll drop it — `CONTRIBUTING.md` carries the same caveat for humans, and this was only here because the preamble says a pointer from `AGENTS.md` doesn't get followed.

dr-bonez (approved):

PR #3796, opened 2026-08-23
fix(exver): evaluate ranges against represented release versions

### LINKED ISSUE #3797: Alias union is existential, so a negated versionRange is escapable via satisfies (dependencies.ts:131, dep-error.service.ts:129)
### Environment

`@start9labs/start-sdk` 2.0.9 (and `master`), `@start9labs/start-core`, StartOS web UI. Found by reading code while confirming Start9Labs/nextcloud-startos#130; reproduced against the published SDK.

### What happens

A dependency version and the versions it declares in `satisfies` are unioned with an existential test — "does *any* candidate satisfy the range?". That is only sound for a positive range. Against a **negated** range the alias lets the installed version through the exclusion:

```
range '>=2.0:0 && !=2.0:5', installedVersion '2.0:5', satisfies ['2.0:4']

  ExtendedVersion.parse('2.0:5').satisfies(R)  ->  false   (correctly excluded)
  ExtendedVersion.parse('2.0:4').satisfies(R)  ->  true    (escapes the exclusion)
  union .some(...)                             ->  true
```

So a dependent that writes `!=2.0:5` to exclude a revision it knows is broken is told the dependency is satisfied, and starts against exactly that revision — provided the dependency ships an ordinary backwards-compatibility alias, which is what `satisfies` is for.

The same shape defeats a flavor exclusion: `#quantum:1.5.2:0` against `>=2.0:0 && !#quantum` is correctly `false`, but its alias `2.63.23:0` returns `true`, so the union readmits the flavor the range excluded.

A negated atom needs universal quantification, not existential — no candidate may match `!=X`. Either evaluate negated atoms over all candidates, or refuse the alias union for a range containing a `Not`, rather than treating "some candidate passes" as sufficient for every range shape.

I do not think the right resolution is obvious, which is why this is an issue rather than a PR: it is a semantics decision that has to hold across the OS, the UI and the SDK at once.

Three independent implementations of the same union, all existential:

- `shared-libs/ts-modules/start-core/lib/dependencies/dependencies.ts:131` — `![installedVersion, ...satisfies].find(...)` (and `:71`, which #3796 brings into line with it)
- `projects/start-os/web/ui/src/app/services/dep-error.service.ts:129-130` — `!exver.satisfies(version, expected) && !depManifest.satisfies.some(...)`
- `projects/start-os/web/ui/src/app/utils/dry-update.ts:31-38` — `versionSatisfies(version, satisfies, range, exver)`

`shared-libs/ts-modules/marketplace/src/components/preview.component.ts:187` has a fourth (`satisfiesWithAliases`). None of the four is shared, so a fix has to land in all of them — extracting one helper into `start-core/lib/exver/` would make that a single edit.

`versionRange` is a plain `string` at `setupDependencies.ts:22` with no compile-time gate, and `!` / `!=` are first-class in the exver grammar, so the range shape is expressible today.

### Logs & evidence

Measured against published `@start9labs/start-sdk@2.0.9`, not inferred:

```js
const R = VersionRange.parse('>=2.0:0 && !=2.0:5')
ExtendedVersion.parse('2.0:5').satisfies(R)  // false
ExtendedVersion.parse('2.0:4').satisfies(R)  // true
```

I have not found a package in either org that currently declares a negated `versionRange`, so I believe this is latent rather than live — but I did not sweep for it exhaustively, so treat that as unverified.

Two smaller findings from the same read, recorded here rather than as separate issues:

- **`CheckDependencies` type drift.** `dependencies.ts:24` declares `satisfied: () => boolean` while the implementation is `(packageId?: DependencyId)`, so `deps.satisfied('bitcoind')` is `TS2554` despite working at runtime; `throwIfNotSatisfied(packageId?)` accepts it. Likewise `:20-23` declares `healthCheckSatisfied`'s `healthCheckId` required while the implementation and `pkgSatisfied`'s own internal call make it optional. Both pre-existing since `0e598660`; both are the same parallel-surface asymmetry, on the type axis.
- **Unknown `packageIds` are vacuously satisfied.** `checkDependencies(effects, ['typo']).satisfied()` returns `true` — Rust `filter_map`s unknown ids out (`dependency.rs:438-444`) and `:49-51` filters them from `dependencies`, so `.every()` runs over an empty list. Every other method raises `Unknown DependencyId` for the same typo.

### LINKED ISSUE #3799: exver: compareLexicographic returns null across flavors, so compareForSort returns undefined from a 1|0|-1 signature
`ExtendedVersion.compareLexicographic` tests `>` in both branches, so `'less'` is unreachable:

```ts
// shared-libs/ts-modules/start-core/lib/exver/index.ts
compareLexicographic(other: ExtendedVersion): 'greater' | 'equal' | 'less' {
  if ((this.flavor || '') > (other.flavor || '')) {
    return 'greater'
  } else if ((this.flavor || '') > (other.flavor || '')) {   // should be <
    return 'less'
  } else {
    return this.compare(other)!
  }
}
```

When `this.flavor < other.flavor` both tests fail and control reaches
`this.compare(other)!`, whose first statement returns `null` for differing
flavors. `compareForSort`'s `switch` then matches no case and falls off the end.

Measured against the shipped source:

```
compareLexicographic(unflavored, flavored) = null       // return type says 'greater'|'equal'|'less'
compareLexicographic(flavored, unflavored) = "greater"
compareForSort(unflavored, flavored)       = undefined  // signature says 1 | 0 | -1
compareForSort(flavored, unflavored)       = 1
```

Two consequences:

- The JSDoc directly above says "Unlike `compare`, this never returns `null`:
  different flavors are ordered alphabetically." Both halves are false in the
  `<` direction.
- The comparator is asymmetric, so `Array.sort` leaves mixed-flavor input in
  whatever order it arrived in (`sort` coerces `undefined` to `+0`):

```
sort([flavored, unflavored]) -> ['#quantum:1.0.0:0', '1.0.0:0']
sort([unflavored, flavored]) -> ['1.0.0:0', '#quantum:1.0.0:0']
```

No user-visible failure today: every caller either buckets by flavor first or
coerces with `|| 0`, and `controls.component.ts` is guarded by `sameFlavor()`
one line above the `@switch` that would otherwise render no install/update
control. It is a published return type that lies, and it will bite the first
cross-flavor sort.

Fix is one character at the second comparison. Worth a `compareForSort` case in
`exver.test.ts` covering both directions across flavors — the suite has none.

Found while reviewing #3796; out of that diff (`exver/index.ts` is untouched
there), so filing separately rather than folding it in.

### LINKED ISSUE #3800: exver: Rust pest and TS peg grammars disagree on prerelease segments, so the OS can emit a version TS cannot parse
The Rust and TypeScript exver grammars disagree on prerelease segments, so the
OS can hand TypeScript a version string TypeScript cannot parse.

`shared-libs/crates/exver/src/grammar.pest`:

```
prerelease_seg = @{ ( '0'..'9' | 'a'..'z' | 'A'..'Z' | "-" )+ }
```

One segment may mix letters, digits and hyphens.

`shared-libs/ts-modules/start-core/lib/exver/exver.pegjs`:

```
PreReleaseSegment = "."? segment:(Digit / String) { return segment }
Digit  = [0-9]+
String = [a-zA-Z]+
```

A segment is *either* digits *or* letters, and never contains a hyphen.
Measured against the shipped parser:

```
ExtendedVersion.parse('1.0.0-rc.1:0')    -> ok
ExtendedVersion.parse('1.0.0-rc1:0')     -> Expected ".", ":", or [a-zA-Z] but "1" found.
ExtendedVersion.parse('1.0.0-beta2:0')   -> Expected ".", ":", or [a-zA-Z] but "2" found.
ExtendedVersion.parse('1.0.0-alpha-1:0') -> Expected ".", ":", or [a-zA-Z] but "-" found.
ExtendedVersion.parse('1.0.0-1a:0')      -> Expected ".", ":", or [0-9] but "a" found.
```

All four are accepted by the pest grammar. `rc1` and `beta2` are ordinary
upstream spellings.

Why it matters: `CheckDependenciesResult.satisfies` is written verbatim from the
dependency's manifest (`service/effects/dependency.rs`), and both
`checkDependencies(...).satisfied()` and `throwIfNotSatisfied()` parse every
entry. A dependency published with such a version makes a dependent's
`satisfied()` — typed `=> boolean` — throw a raw `peg$SyntaxError` naming
neither the package nor the string. `.satisfies()` on the version builder is
typed through `ValidateExVer`, which does not reject it at author time.

Not currently reachable. I pulled the production, community and community-beta
registry indexes (193 packages, 595 distinct version and `satisfies` strings)
and parsed all of them with the TS parser: **0 failures**. Every published
prerelease uses the dotted form both grammars accept:

```
0.16.1-alpha:4  0.17.0-alpha:7  0.20.1-beta:13  0.21.1-beta:12
0.21.1-rc.0:1   0.21.2-beta:1   0.23.1-rc.0:10  2.4.3-rc.4:1
```

So this is latent, not live — but the failure mode is a crash-loop in a
dependent, triggered by a package the dependent does not control.

The fix belongs in the grammars, not in a `try`/`catch` at one call site:
catching would make `satisfied()` silently report a dependency unsatisfied
forever while `throwIfNotSatisfied()`, the web UI's `Exver` service and every
other consumer still throw. Either widen `PreReleaseSegment` to match
`prerelease_seg`, or narrow the pest rule and validate at publish time — worth
deciding which grammar is normative, and a shared conformance corpus exercised
from both sides.

Found while reviewing #3796, which widens the boolean path to parse `satisfies`
(the throwing path already did). Out of that diff's scope, so filing separately.

helix-nine:
Closes #3797
Closes #3799
Closes #3800

## Summary

Dependency range checks now evaluate the installed version together with every version declared in `satisfies`.

- one represented version must satisfy every positive term in a conjunction
- different represented versions may satisfy different disjunction branches
- negation and `!=` veto their branch when a represented version satisfies the complete excluded range
- an empty represented-version set satisfies no range

Rust and TypeScript use the same release evaluator. Negated ranges are evaluated as complete exclusions rather than distributed through conjunctions or disjunctions. Normalization and reduction preserve release semantics, including nested negation and caret/tilde/default anchors. Numeric prerelease identifiers retain exact unbounded decimal ordering beyond JavaScript's safe-integer limit without changing the public `(string | number)[]`/JSON shape.

The same release evaluator now backs SDK dependency checks, StartOS dependency warnings and updates, and marketplace filtering. The TypeScript prerelease grammar also matches Rust for alphanumeric/hyphenated segments, and two existing comparison defects are corrected.

## Verification

- `cargo test -p exver` — 30 passed
- `make start-core-test`
- start-core TS tests — 206 passed, 9 skipped; typecheck/build passed
- Peggy regeneration produced no drift
- root TypeScript checks and UI build
- container-runtime tests/typecheck
- formatting, docs, and `git diff --check`


shared-libs/crates/exver/src/exver.rs
```diff
@@ -37,15 +37,57 @@ impl fmt::Display for ParseError {
 }
 impl std::error::Error for ParseError {}
 
-#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Clone, Debug)]
 pub enum PreReleaseSegment {
     Number(usize),
+    Numeric(InternedString),
```
dr-bonez: what is this for?

helix-nine: It preserves numeric prerelease identifiers that exceed `usize` without overflow. I renamed the variant to `BigNumber`, added it to property generation, and fixed the generator’s own overflow path.

dr-bonez: my concern is that `PreReleaseSegment::Number(123) != PreReleaseSegment::Numeric("123")`. this isn't good

helix-nine: Agreed. `Number(n)` and `BigNumber(n.to_string())` now compare equal in both directions, order as `Equal`, and hash identically. The new property test pins all three contracts across arbitrary `usize` values.

shared-libs/crates/start-core/exver.md
```diff
@@ -213,20 +213,22 @@ range.satisfiedBy(version) // boolean (convenience)
 
 Also available on `Version` (wraps in `ExtendedVersion` with downstream=0).
 
+A release may be tested against its installed version and every version in its manifest's `satisfies` list with Rust's `VersionRange::satisfied_by_release` or TypeScript's `range.satisfiedByRelease`. Evaluation first puts the range in negation normal form: De Morgan's laws push `!` to literals and double negations cancel. One declared version must satisfy every literal in a conjunction. Disjunctions may match different declared versions. A negative literal passes only when no declared version matches its positive form. An empty version list satisfies no range.
```
dr-bonez: we need to be very clear about the definition of a "negative literal". is it just `!=`?

helix-nine: No. After conversion to negation normal form, a literal is an atomic comparison, flavor test, or wildcard. A negative literal is any such atom under `!`; `!=v` is specifically evaluated as `!(=v)`.

I defined that in the reference and mirrored the wording on both public APIs in `7c8a01f12`, with `!=2:0`, `!>=2:0`, and `!#knots` as examples. Verified with `cargo test -p exver` (28 passed), Rust formatting, Prettier, and `git diff --check`.

dr-bonez: when does !>=2:0 ever show up? wouldn't that be normalized to <2:0? my concern is the only intentional exclusion that could ever happen with a comparison is probably going to specify a range: `!(>= 2:0 && < 3:0)`. which I'm concerned is normalized before evaluation to `<2:0 || >= 3:0`. doesn't seem very sound

helix-nine: You’re right—the previous evaluator effectively distributed `!` through composite ranges, so it would reject `[1:0, 4:0]` for `!(>=2:0 && <3:0)`. Fixed in `2f90a5bd9`: `!R` now excludes a release only when one represented version satisfies the complete `R`. Thus `[1:0, 4:0]` passes that exclusion, while `[2.5:0, 4:0]` fails it; positive disjunction branches can still use separate represented versions. `!=` remains the whole-release exclusion `!(=v)`, and double negation cancels. Rust and TypeScript now carry the same vectors, including raw Rust AST/reduction cases and a normalization case that actually changes shape. `cargo test -p exver` passes 30 tests; start-core Jest passes 206 with 9 skipped, plus typecheck and formatting.

dr-bonez (approved):

PR #3801, opened 2026-08-23
feat(start-os): open a UI at the address its service nominates

### LINKED ISSUE #3715: Let a package nominate a preferred launcher address for a UI interface
### Prerequisites

- [x] I searched existing issues and this is not already requested.

### Project

StartOS (with an SDK-side surface — see Affected areas)

### Problem & use case

The Launch button opens the best address for **how the admin is currently reaching StartOS** — `launchableAddress()` matches the session's `accessType` (LAN IP / mDNS / domain / Tor / WAN IP) against the interface's enabled addresses, then falls back best-private-if-on-LAN → best-public → best-private. That is the correct default: opening a `.local` address in a Tor session, or a `.onion` in a LAN session, yields a link the browser cannot resolve.

But some services are origin-sensitive: they are configured with one canonical origin and behave incorrectly when reached at any other. CryptPad is the sharp case — its `httpUnsafeOrigin` participates in client-side account/key derivation, so launching at a different (perfectly reachable) address surfaces as "invalid username/password" with correct credentials. Ghost, Gitea, Vaultwarden, Synapse and DocuSeal are milder versions of the same shape: generated links, callback URLs, federation identity, emailed signing requests.

Those packages already let the user pick a canonical URL — that is what `recipe-primary-url.md` describes — but the choice is stored in the **package's own file model** and passed to the app as env/config. Nothing about it is visible to StartOS: `ServiceInterface` is `{id, name, description, masked, addressInfo, type}`, `AddressInfo` is `{username, hostId, internalPort, scheme, sslScheme, suffix}`, and no patch-db field carries a nominated address. So there is currently no value for the launcher to honor even in principle.

### Proposed solution

Let a package nominate one of its own StartOS-generated addresses as the preferred launcher target for a given UI interface, and have the launcher prefer it when it is currently valid.

- Nomination, **not** override. The launcher uses it only when the nominated address is still enabled and plausibly reachable from the admin's current access route; otherwise it falls back to today's chain. Otherwise we trade "opens a working URL the service rejects" for "opens a URL that does not load at all."
- Per **interface**, not per package — CryptPad alone has Main and Sandbox on deliberately distinct origins, and the launch button already iterates every `type: 'ui'` interface.
- The value references a StartOS-generated address rather than a free-form string, so it stays meaningful as domains are added and gateways are toggled. The reactive-watcher variant in `recipe-primary-url.md` already handles "my selected URL went away"; the launcher preference wants the same lifecycle.
- Absent preference = exactly today's behavior, so this is additive.

Open question worth settling as part of the design: when the nominated address is **not** reachable from the current route, is silent fallback right? For a service like CryptPad, "this service must be opened at `<X>`, which isn't reachable over your current connection" may serve the user better than quietly opening an origin the service will reject.

### Alternatives considered

- **Admin-side only, no SDK surface** — an explicit "preferred" marker per interface in the StartOS UI. Solves the same problem with zero package-facing API, but it cannot stay in sync with the package's own canonical-URL setting, so the two drift the moment the user re-runs the package's action.
- **Do nothing; rely on the existing mitigation.** Today an admin can disable every Main-interface address except the canonical one — `filledAddress()` only considers enabled addresses, so the launcher is left with a single candidate. This works, but it is undiscoverable and it forces the admin to give up otherwise-useful addresses to get a correct launch.
- **Make the launcher always prefer a public domain.** Wrong for the general case and does not fix CryptPad — the canonical origin may well be an mDNS or Tor address.

### Affected areas

- `projects/start-os/web/ui/src/app/routes/portal/components/interfaces/interface.service.ts:408` — `launchableAddress()`; try the nomination first, validate enabled + route-reachable, else current chain.
- `shared-libs/crates/start-core` — the field on the exported interface type and the effect that sets it; then `make start-core-ts-bindings`.
- `projects/start-sdk/lib` — the packager-facing effect / `createInterface` option.
- `projects/start-sdk/docs/src/recipe-primary-url.md` — the recipe currently stops at "pass it to the app"; it would gain the launcher-nomination step.

### Anything else?

Split out of #3714, which reported this as a launcher bug.

helix-nine:
Closes #3715

## Summary

`createInterface` accepts an optional `preferredLauncherAddress`, allowing an origin-sensitive UI to nominate the enabled address used by StartOS's Open UI control. Missing, malformed, disabled, unmatched, or transport-incompatible nominations fall through to the existing address selection.

The backend stores the optional string as interface metadata without interpreting it. Launcher selection parses it and matches scheme, normalized hostname, and effective port against an address StartOS assigned and kept enabled. The opened URL still comes from the generated interface address, preserving its path, query, and username without creating a general redirect primitive.

Transport restrictions remain session-aware for Tor and plugin addresses. Existing interface records and packages that omit the field retain the prior behavior.

## Verification

- committed 35-case launcher-selection suite
- `make start-core-test`
- `cargo check -p start-core`
- generated TS bindings
- start-core TS tests/typecheck/build
- root TypeScript and i18n checks
- UI production build
- container-runtime tests/typecheck
- Rust and Prettier formatting checks
- docs build

### DIFF
--- a/shared-libs/crates/start-core/src/service/effects/net/interface.rs
+++ b/shared-libs/crates/start-core/src/service/effects/net/interface.rs
@@ -13,11 +13,7 @@ use crate::service::effects::prelude::*;
 use crate::service::rpc::CallbackId;
 use crate::{HostId, PackageId, ServiceInterfaceId};
 
-// Every service interface lives under the binding it was exported from
-// (`hosts/{hostId}/bindings/{internalPort}/interfaces/{id}` for single-port
-// `Origin.export`, `hosts/{hostId}/bindingRanges/{internalStartPort}/interface`
-// for `RangeOrigin.export`). The flat `PackageDataEntry.serviceInterfaces` map
-// is gone — these effects read/write the host tree directly.
+// Service interfaces are stored under the binding that exported them.
 #[derive(Debug, Clone, Serialize, Deserialize, TS)]
 #[ts(export)]
 #[serde(rename_all = "camelCase")]
@@ -28,7 +24,11 @@ pub struct ExportServiceInterfaceParams {
     masked: bool,
     address_info: AddressInfo,
     r#type: ServiceInterfaceType,
+    /// The interface address Open UI should prefer.
+    #[ts(optional = nullable)]
+    preferred_launcher_address: Option<String>,
 }
+
 pub async fn export_service_interface(
     context: EffectContext,
     ExportServiceInterfaceParams {
@@ -38,6 +38,7 @@ pub async fn export_service_interface(
         masked,
         address_info,
         r#type,
+        preferred_launcher_address,
     }: ExportServiceInterfaceParams,
 ) -> Result<(), Error> {
     let context = context.deref()?;
@@ -52,6 +53,7 @@ pub async fn export_service_interface(
         masked,
         address_info,
         interface_type: r#type,
+        preferred_launcher_address,
     };
 
     context
--- a/projects/start-os/CHANGELOG.md
+++ b/projects/start-os/CHANGELOG.md
@@ -18,6 +18,10 @@ file tracks notable changes since the move to the monorepo.
   behavior. The server's own `.local` name remains reserved for StartOS. See
   [Private Domains](https://docs.start9.com/start-os/private-domains.html).
 
+- **Open UI can honor a service's configured address.** Services that depend on
+  one canonical origin can now direct Open UI to that address while it remains
+  enabled and compatible with the current browser session.
+
 - **A service can permanently retire a network host or a port it no longer
   uses, and the port numbers it held become available again.** A service that
   reorganizes its interfaces across an update — renaming a host, dropping a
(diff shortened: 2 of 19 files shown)


shared-libs/crates/start-core/src/service/effects/net/interface.rs
```diff
@@ -28,7 +24,47 @@ pub struct ExportServiceInterfaceParams {
...
+    } else {
+        LauncherAddress::Malformed
+    }
+}
+
+fn usable_launcher_address(
+    package_id: &PackageId,
+    interface_id: &ServiceInterfaceId,
+    address: Option<String>,
+) -> Option<String> {
```
dr-bonez: why does the backend care about this?

helix-nine: It does not need to interpret this value. I removed the backend parsing and warning; it now stores the optional interface metadata verbatim. The launcher parses it, matches only against enabled OS-generated addresses from that interface, and silently uses the existing fallback for blank, malformed, missing, or unmatched values. I added explicit whitespace fallback coverage and updated the docs and PR description accordingly.

dr-bonez (dismissed):

dr-bonez (dismissed):

dr-bonez (approved):

PR #3802, opened 2026-08-23
fix(start-os): one certificate per name, not per route to it

### LINKED ISSUE #3718: [bug]: a distinct leaf certificate is minted per (SNI + listener address + WAN IP), so dual-stack clients see the certificate change — pinning clients get recurring MITM warnings
Field report (2026-08): Sparrow on Windows 10, connected to Fulcrum on 50002
via `<hostname>.local`, shows recurring "SSL Handshake Failed — the certificate
provided by the server appears to have changed… may indicate a man-in-the-middle
attack" dialogs. Deleting Sparrow's pinned certificate helps only until the
warning returns days later. The user captured the server alternating between
two different valid leaf certificates (NotBefore 2026-07-26 and 2026-08-11),
both issued by their device's intermediate CA. A Linux machine on the same LAN
shows no warnings.

## Mechanism

`RootCaTlsHandler::get_config`
(`shared-libs/crates/start-core/src/net/ssl.rs` ~740) builds the certificate
lookup key per connection as:

    {SNI hostname} ∪ {IP address the connection arrived on} ∪ {gateway WAN IP}

and `cert_for` lazily mints and persists a certificate per distinct key set.
The IPv4 and IPv6 listeners of the same hostname therefore hold different
certificates, minted at different times. Which one a client sees depends on
which address family its resolver happens to return — mDNS answer order is
nondeterministic — so the served identity flaps.

Reproduced on a 0.4.0.x device, same SNI on both probes:

| Listener | Serial | NotBefore | SAN contains |
|---|---|---|---|
| IPv4 (192.168.0.x) | AF66… | Jun 25 | hostname, WAN IP, the v4 address |
| IPv6 (fe80::…) | A309… | Aug 7 | hostname, WAN IP, the v6 address |

Two further destabilizers follow from the same keying: a WAN IP change
(tunnel up/down, ISP renumber) changes the key set and mints another
certificate, and each entry carries two leaves (Ed25519 and NIST P-256)
selected by the client's advertised signature schemes, so a client TLS-stack
update can change the pinned bytes as well.

## Impact

Every trust-on-first-use client — Sparrow, and the Electrum-protocol client
ecosystem generally (cf. Start9Labs/fulcrum-startos#19) — pins one identity
per name. Serving several identities per name presents as a possible MITM at
random intervals, and recurring false MITM warnings train users to click
through real ones. Clients that validate against the installed root CA are
unaffected, which is why only some machines show the dialog.

## Suggested direction

One stable identity per hostname: SAN = the union of the host's addresses
(the SNI-less/IP certificate already unions v4 + WAN today), regenerate only
when the union grows, one leaf keypair — or at minimum key the cache by
hostname rather than by connection path. Interim workarounds for affected
users: connect by IP literal (one listener → stable certificate), or install
the device root CA so the client validates instead of pinning.

helix-nine:
Closes #3718

## Summary

StartOS now reuses one server certificate per served name instead of deriving its certificate-store key from each route and listener address. The existing certificate-store model remains unchanged: server listeners request the selected name as a singleton set, while package certificate grants continue through the same API.

## Network selection

A DNS SNI value selects that server name. A connection without SNI is certified for the destination address unless an Internet-side IPv4 connection was translated through the gateway, in which case it uses the validated WAN address. IPv4-mapped listener addresses are canonicalized before classification, and unusable echo-IP results are rejected.

## Verification

- `cargo test -p start-core --features=test` — 735 passed, 3 ignored; 52 doctests passed
- focused certificate-store, served-name, gateway lookup, echo-IP, and address-classification tests
- `cargo check -p start-core --all-targets`
- `mdbook build projects/start-os/docs`

### DIFF
--- a/shared-libs/crates/start-core/src/net/ssl.rs
+++ b/shared-libs/crates/start-core/src/net/ssl.rs
@@ -36,10 +36,12 @@ use crate::SOURCE_DATE;
 use crate::account::AccountInfo;
 use crate::context::{CliContext, RpcContext};
 use crate::db::model::Database;
+use crate::db::model::public::IpInfo;
 use crate::db::{DbAccess, DbAccessMut};
 use crate::init::check_time_is_synchronized;
 use crate::net::gateway::GatewayInfo;
 use crate::net::tls::{TlsHandler, TlsHandlerAction};
+use crate::net::utils::is_global_ip;
 use crate::net::web_server::{Accept, ExtractVisitor, TcpMetadata, extract};
 use crate::prelude::*;
 use crate::util::serde::{HandlerExtSerde, Pem};
@@ -120,7 +122,7 @@ impl CertStore {
     }
 }
 impl Model<CertStore> {
-    /// This function will grant any cert for any domain. It is up to the *caller* to enusure that the calling service has permission to sign a cert for the requested domain
+    /// The caller must establish entitlement to every requested name.
     pub fn cert_for(
         &mut self,
         hostnames: &BTreeSet<InternedString>,
@@ -724,6 +726,30 @@ pub async fn generate_certificate(
     })
 }
 
+fn served_name(
+    server_name: Option<&str>,
+    tcp: Option<TcpMetadata>,
+    gateway: Option<&IpInfo>,
+) -> Option<InternedString> {
+    if let Some(host) = server_name {
+        return Some(InternedString::from(host));
+    }
+    let tcp = tcp?;
+    let dst = tcp.local_addr.ip().to_canonical();
+    let src = tcp.peer_addr.ip().to_canonical();
+    let translated = !is_global_ip(dst)
+        && is_global_ip(src)
+        && !gateway.is_some_and(|gw| gw.subnets.iter().any(|s| s.contains(&src)));
+    let wan = gateway
+        .and_then(|gw| gw.wan_ip)
+        .filter(|wan| is_global_ip(IpAddr::V4(*wan)));
+    let dialed = match (dst, wan) {
+        (IpAddr::V4(_), Some(wan)) if translated => IpAddr::V4(wan),
+        _ => dst,
+    };
+    Some(InternedString::from_display(&dialed))
+}
+
 pub struct RootCaTlsHandler<M: HasModel> {
     pub db: TypedPatchDb<M>,
     pub crypto_provider: Arc<CryptoProvider>,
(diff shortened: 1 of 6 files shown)


projects/start-os/docs/src/installing-startos.md
```diff
@@ -8,7 +8,7 @@ This guide is for flashing StartOS to a USB drive, then installing it onto a des
 
 ## Download
 
-1.  Visit the [Github release page](https://github.com/Start9Labs/start-technologies/releases/tag/start-os/v0.4.0.2) to find the latest version of StartOS.
+1.  Visit the [Github release page](https://github.com/Start9Labs/start-technologies/releases/tag/start-os/v0.4.0.3) to find the latest version of StartOS.
```
dr-bonez: 0.4.0.2 hasn't been tagged yet

shared-libs/crates/start-core/src/net/ssl.rs
```diff
@@ -92,6 +94,21 @@ impl CertBranding {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Deserialize, Serialize)]
+#[serde(untagged)]
+pub enum CertKey {
+    Package(BTreeSet<InternedString>),
+    Server(InternedString),
+}
```
dr-bonez: I don't see any reason to change the db model for this

projects/start-os/docs/src/installing-startos.md
helix-nine: Kept this at 0.4.0.2 and removed the 0.4.0.3 version/migration changes. The install links now remain on 0.4.0.2.

shared-libs/crates/start-core/src/net/ssl.rs
helix-nine: Removed `CertKey` and restored the existing `BTreeMap<JsonKey<BTreeSet<InternedString>>, CertData>` model and `cert_for` API. The TLS handler now requests the selected served name as a singleton set, which provides one cached certificate per name without a schema change or migration.

shared-libs/crates/start-core/src/net/ssl.rs
```diff
@@ -724,6 +726,30 @@ pub async fn generate_certificate(
     })
 }
 
+fn served_name(
+    server_name: Option<&str>,
+    tcp: Option<TcpMetadata>,
+    gateway: Option<&IpInfo>,
+) -> Option<InternedString> {
+    if let Some(host) = server_name.filter(|name| name.parse::<IpAddr>().is_err()) {
```
dr-bonez: filter seems unnecessary. if the SNI is an IP, that's against spec, but it's still definitely the ip the user is connecting via - so is what we should sign for.

shared-libs/crates/start-core/src/net/utils.rs
```diff
@@ -147,6 +147,17 @@ pub fn is_private_ip(addr: IpAddr) -> bool {
     }
 }
 
+/// Outside private, loopback, link-local, and shared address space.
+pub fn is_global_ip(addr: IpAddr) -> bool {
+    match addr.to_canonical() {
+        IpAddr::V4(v4) => {
+            let [a, b, ..] = v4.octets();
+            !is_private_ip(v4.into()) && !(a == 100 && (64..128).contains(&b))
```
dr-bonez: use a helper for cgnat, so it's clear that's what's being excluded here

shared-libs/crates/start-core/src/net/ssl.rs
helix-nine: Removed the filter. Any supplied SNI now selects that value verbatim, including IPv4 and IPv6 literals; destination-address inference runs only when SNI is absent. Updated the regression test for both literal families, and the served-name tests plus `cargo check -p start-core --all-targets` pass.

shared-libs/crates/start-core/src/net/utils.rs
helix-nine: Extracted the range check into `ipv4_is_cgnat`, so `is_global_ip` now reads as private-address and CGNAT exclusion directly. Renamed the boundary test accordingly; the global-IP, served-name, and echo-IP tests pass.

dr-bonez (approved):

PR #3808, opened 2026-08-24
fix(ui): render an action's message as markdown, and unify how results copy

### LINKED ISSUE #3792: StartOS action result fields do not support newline characters
### Prerequisites

### StartOS version

0.4.0.1

### Architecture / image

x86_64 (AMD64) — Slim (FOSS-only)

### Server hardware

Intel n150 / 16GB RAM / 512GB SSD

### Browser & OS

_No response_

### Which surface?

Web UI

## Description

  StartOS does not preserve newline characters when displaying multiline text returned by service actions.

  For example, Vikunja’s “Run Diagnostics” action returns formatted diagnostic output. In the StartOS UI, the result is
  displayed as one continuous line:

  Vikunja Doctor ============== System ✓ Version: v2.4.0 ✓ Go: go1.26.5 ✓ OS: linux/amd64 ✓ User: vikunja (uid=1000) ✓
  Working directory: /app/vikunja ✓ User namespace: active (0-65535→100000-165535) UIDs inside this container are
  remapped. See directory ownership check for details. Configuration ✓ Config file: none (using defaults/environment) ✓
  Public URL: https://phoenix.local:54839/ ✓ JWT secret: configured (auto-generated) Database (sqlite) ✓ Connection: OK
  ✓ Server version: 3.53.3 Files (local) ✓ Path: /data/files ✓ Directory exists: yes ✓ Directory permissions: 0755 ✓
  Directory owner: vikunja:vikunja (uid=1000, gid=1000) ✓ Ownership match: uid 1000 matches WARNING: user namespace
  active — uid 1000 maps to host uid 101000, which may differ from the actual host directory owner. If writes fail,
  ensure the host directory is owned by the mapped host uid, or run the container with --user 0:0. ✓ Writable: yes ✓
  Disk space: 424.6 GB available ✓ Stored files: 1 files, 18 B total All checks passed

It should retain the original formatting:

  Vikunja Doctor

  ==============

  System
  ✓ Version: v2.4.0
  ✓ Go: go1.26.5
  ✓ OS: linux/amd64
  ✓ User: vikunja (uid=1000)
  ✓ Working directory: /app/vikunja
  ✓ User namespace: active (0-65535→100000-165535)

  UIDs inside this container are remapped. See directory ownership check for details.

  Configuration
  ✓ Config file: none (using defaults/environment)
  ✓ Public URL: https://redacted.local:54839/
  ✓ JWT secret: configured (auto-generated)

  Database (sqlite)
  ✓ Connection: OK
  ✓ Server version: 3.53.3

  Files (local)
  ✓ Path: /data/files
  ✓ Directory exists: yes
  ✓ Directory permissions: 0755
  ✓ Directory owner: vikunja:vikunja (uid=1000, gid=1000)
  ✓ Ownership match: uid 1000 matches

  WARNING: user namespace active — uid 1000 maps to host uid 101000, which may differ from the actual host directory
  owner. If writes fail, ensure the host directory is owned by the mapped host uid, or run the container with --user
  0:0.

  ✓ Writable: yes
  ✓ Disk space: 424.6 GB available
  ✓ Stored files: 1 files, 18 B total

  All checks passed

### Steps to reproduce

  1. Install Vikunja.
  2. Open the Vikunja service page.
  3. Run the “Run Diagnostics” action.
  4. Observe that the returned text is rendered as one continuous line.

### Logs & evidence

```shell

```

MattDHill:
Closes #3792.

## The report, and what's actually true

[#3792](https://github.com/Start9Labs/start-technologies/issues/3792) says "action result fields do not support newline characters," using Vikunja's `doctor` action as the example. That's half right, and not in the half the title names — there are three fields here and they want different answers.

Vikunja returns the raw report in **both** `message` and `result.single.value`, which is two rendering paths:

| Field | Rendered as | Newlines |
| --- | --- | --- |
| `message` | `<p>{{ data.message }}</p>` | HTML whitespace collapse — `\n` becomes a space |
| `single.value` | `<input tuiInput>` | the input value-sanitization algorithm *strips* CR/LF outright |

The screenshot in the issue has spaces where the line breaks were, so what the reporter is looking at is the collapsed `message`.

## What this changes

**`message` renders as Markdown.** It's prose, the collapse was an HTML default nobody chose, and `ActionResultV1`'s `Display` impl already prints the same payload with the breaks intact — so the UI and `start-cli` disagreed about one response. It now goes through the path service instructions and release notes already use: `| markdown | dompurify` with `safeLinks`, so package-authored HTML is sanitized and external links get `target=_blank rel=noreferrer`.

It renders with **`breaks: true`**, and that part is load-bearing rather than a flourish. CommonMark folds a single newline into a space, so plain Markdown would have left #3792 exactly where it was: the payloads in question separate lines with a single `\n` — Vikunja's report, and garage's `` `${message}\n\nFailed to delete:\n${errors.join('\n')}` ``. A lone `\n` therefore stays a line break and everything else is normal Markdown, which is the convention GitHub renders comments under and matches how packagers write these — a template literal, not a document. It's passed per call rather than made the `MarkdownPipe` default, because `instructions.md` and release notes *are* authored as documents and hard-wrapped source lines there would sprout `<br>`s.

I scanned every `message:` under `startos/**` in `start9-registry`, `community-registry` and `needs-review` for text Markdown would now reinterpret. **One hit, and it argues for the change**: nextcloud's `disableUnstableApps` builds `` `… <ul>${apps.map(a => `<li>${a}</li>`)}</ul>` `` into its message, which today shows the user literal angle brackets. Everything else is plain prose with `${}` interpolation.

**The two dialog paths become one.** A message-only result opened a plain alert; a result-bearing one opened `ActionSuccessPage`. Two renderings of the same field, and only the alert put it through the i18n pipe. `ActionSuccessPage` now takes the whole response and tolerates a null `result`, so `message` is rendered in exactly one place.

**Markdown body styling is one global class.** The HTML the pipe produces is unencapsulated, so each site that styled it reached for `::ng-deep` and carried its own copy of the `pre` and table rules — `instructions.component.ts` had one set, shared `MarkdownComponent` another, and this dialog would have been a third. `.g-markdown` now holds them once in `shared-libs/ts-modules/shared/styles/shared.scss`, beside the other global classes the shared libs depend on and that every host loads, and all three sites use it. Net −61/+39 lines and three `::ng-deep` blocks gone. One behavior change falls out: code blocks in the dialogs `MarkdownComponent` opens (release notes, service About, sideload, marketplace preview) now keep their line structure and scroll instead of wrapping, matching the instructions page, and pick up its padding/radius/background.

**Copy is unified on `CopyService`.** This one is a plain bug independent of the newline question: `ActionSuccessSingleComponent` copies via CDK Clipboard, while `ActionSuccessMemberComponent` selected its `<input>` and called `execCommand` — copying the value *after* the browser stripped its line breaks, and showing no confirmation toast. So the same `copyable: true` produced different clipboard contents depending on whether the value sat at the top level or inside a group. Both now copy the string the action returned, and the DOM juggling (`ViewChild`, the `type` swap around masked fields) goes away with it.

## What this deliberately does not change

**A `single` value stays single-line.** It carries `copyable`, `qr` and `masked` and renders as a labeled textfield: masking a multi-line blob is meaningless, QR-encoding a diagnostic report is absurd, and a `<textarea>` breaks the whole affordance set. `ActionResultV1` exists to be *structured* output — it replaced V0's free-text `message` + `value`. Dumping terminal output into it is misuse.

Both facts are now stated where a packager actually meets them, which is the part that would have prevented this: the `///` doc comments on `ActionResultV1::message` and `ActionResultValue::Single::value`. ts-rs carries those into `ActionResultV1.ts`, `ActionResultValue.ts` and `ActionResultMember.ts`, which ship inside `@start9labs/start-sdk` — so they are the editor hover when someone types `message:`. The regenerated bindings are in this PR (`make start-core-ts-bindings`; the only deltas are those three files). `projects/start-sdk/docs/src/actions.md` § Result Types says the same thing in prose, with an example.

**`start-cli` still prints the raw source.** Markdown source is legible in a terminal and rendering it there would mean choosing an ANSI dialect for tables and headings; the payload is what it is.

**No new result member type.** An earlier draft of this description flagged that action *inputs* have a `textarea` spec (`inputSpecTypes.ts`) with no counterpart on results, and proposed one for "show the user this block of text." Markdown closes that: a fenced code block in `message` is monospace, preformatted and copyable without a new type.

## Verification

`npm run check` (all eight projects), `npm run check:i18n`, and AOT builds of both hosts that render `MarkdownComponent` — `npm run build:ui:dev` and `npm run build:brochure` — all clean; prettier clean. Confirmed `.g-markdown` lands in each app's compiled global stylesheet. Bindings regenerated rather than hand-written, so the `Generated Artifacts` job has nothing to catch. The default mock (`Mock.ActionResMessage`) now carries a heading, a list, emphasis and a bare `\n`, so `npm run start:ui` → any service action shows the behavior.

## Downstream

[Start9-Community/vikunja-startos#6](https://github.com/Start9-Community/vikunja-startos/pull/6) drops the duplication and the false comment; the report can now be a fenced code block in `message`. It should land after this ships in 0.4.0.2 — before then it isn't yet an improvement.

### DIFF
--- a/projects/start-sdk/docs/src/actions.md
+++ b/projects/start-sdk/docs/src/actions.md
@@ -94,7 +94,23 @@ export const actions = sdk.Actions.of().addAction(setAdminPassword)
 
 Actions return structured results that the StartOS UI renders for the user.
 
-`message` is prose shown under the title, and **line breaks in it are preserved** — put anything with its own line structure there. `result` holds discrete values the user acts on: each one is a single-line field with optional copy, QR and masking, so a newline in a `value` is not rendered. A multi-line report goes in `message`, not in a `value`.
+`message` is prose shown under the title. It is rendered as **Markdown** — headings, lists, tables, code blocks, emphasis and links all work — and a single newline is kept as a line break, so text written as plain lines arrives as plain lines. Anything with its own line structure goes here.
+
+`result` holds discrete values the user acts on: each one is a single-line field with optional copy, QR and masking, so a newline in a `value` is not rendered. A multi-line report goes in `message`, not in a `value`.
+
+```typescript
+return {
+  version: '1',
+  title: 'Diagnostics',
+  message: `### Checks
+
+- database: **ok**
+- search index: **rebuilding**
+
+Restart the service once the rebuild finishes.`,
+  result: null,
+}
+```
 
 ### Single Value
 
--- a/projects/start-sdk/docs/src/actions.md
+++ b/projects/start-sdk/docs/src/actions.md
@@ -94,6 +94,8 @@ export const actions = sdk.Actions.of().addAction(setAdminPassword)
 
 Actions return structured results that the StartOS UI renders for the user.
 
+`message` is prose shown under the title, and **line breaks in it are preserved** — put anything with its own line structure there. `result` holds discrete values the user acts on: each one is a single-line field with optional copy, QR and masking, so a newline in a `value` is not rendered. A multi-line report goes in `message`, not in a `value`.
+
 ### Single Value
 
 ```typescript
--- a/projects/start-os/CHANGELOG.md
+++ b/projects/start-os/CHANGELOG.md
@@ -248,17 +248,6 @@ file tracks notable changes since the move to the monorepo.
   settings excluded. StartOS now rejects the overlapping claim and names both
   ports, so the service reports the conflict instead of serving it.
 
-- **The message an action returns is rendered as Markdown, and its line breaks
-  are kept.** A service action that reports back over several lines — a
-  diagnostic report, a list of what succeeded and what failed — reads as the
-  service wrote it, and can use headings, lists, tables, code blocks, emphasis
-  and links.
-
-- **The copy button beside a value in a grouped action result copies the whole
-  value.** Copying from a group dropped any line breaks the value contained and
-  gave no confirmation that anything had been copied. It now copies exactly
-  what the action returned, and confirms, like every other copy button.
-
 ### Security
 
 - **Service mount paths are validated and confined to their intended
--- a/projects/start-os/CHANGELOG.md
+++ b/projects/start-os/CHANGELOG.md
@@ -248,6 +248,17 @@ file tracks notable changes since the move to the monorepo.
   settings excluded. StartOS now rejects the overlapping claim and names both
   ports, so the service reports the conflict instead of serving it.
 
+- **Line breaks in the message an action returns are kept.** A service action
+  that reports back over several lines — a diagnostic report, a list of what
+  succeeded and what failed — had every line run together into one paragraph.
+  The message now reads as the service wrote it, matching what `start-cli`
+  shows for the same action.
+
+- **The copy button beside a value in a grouped action result copies the whole
+  value.** Copying from a group dropped any line breaks the value contained and
+  gave no confirmation that anything had been copied. It now copies exactly
+  what the action returned, and confirms, like every other copy button.
+
 ### Security
 
 - **Service mount paths are validated and confined to their intended
--- a/projects/start-os/CHANGELOG.md
+++ b/projects/start-os/CHANGELOG.md
@@ -248,11 +248,11 @@ file tracks notable changes since the move to the monorepo.
   settings excluded. StartOS now rejects the overlapping claim and names both
   ports, so the service reports the conflict instead of serving it.
 
-- **Line breaks in the message an action returns are kept.** A service action
-  that reports back over several lines — a diagnostic report, a list of what
-  succeeded and what failed — had every line run together into one paragraph.
-  The message now reads as the service wrote it, matching what `start-cli`
-  shows for the same action.
+- **The message an action returns is rendered as Markdown, and its line breaks
+  are kept.** A service action that reports back over several lines — a
+  diagnostic report, a list of what succeeded and what failed — reads as the
+  service wrote it, and can use headings, lists, tables, code blocks, emphasis
+  and links.
 
 - **The copy button beside a value in a grouped action result copies the whole
   value.** Copying from a group dropped any line breaks the value contained and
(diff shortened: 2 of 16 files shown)


projects/start-sdk/docs/src/actions.md
```diff
@@ -94,6 +94,8 @@ export const actions = sdk.Actions.of().addAction(setAdminPassword)
 
 Actions return structured results that the StartOS UI renders for the user.
 
+`message` is prose shown under the title, and **line breaks in it are preserved** — put anything with its own line structure there. `result` holds discrete values the user acts on: each one is a single-line field with optional copy, QR and masking, so a newline in a `value` is not rendered. A multi-line report goes in `message`, not in a `value`.
```
dr-bonez: why not support full markdown?

MattDHill: Done — `message` renders as Markdown now.

One wrinkle worth naming, because it decides the implementation: plain CommonMark on its own would have re-broken #3792. A single newline folds to a space in CommonMark, and single newlines are exactly what the reported payloads use — Vikunja's `doctor` report, and garage's `` `${message}\n\nFailed to delete:\n${errors.join('\n')}` ``. So it renders with `breaks: true`: a lone `\n` stays a line break, everything else is normal Markdown. Same convention GitHub uses for comments, and it matches how packagers actually write these — a template literal, not a document.

It goes through the path service instructions and release notes already use — `| markdown | dompurify` with `safeLinks` — so package-authored HTML is sanitized and external links get `target=_blank rel=noreferrer`.

`breaks: true` is passed per call rather than made the `MarkdownPipe` default: `instructions.md` and release notes *are* authored as documents, and hard-wrapped source lines there would sprout `<br>`s.

I scanned every `message:` under `startos/**` in start9-registry, community-registry and needs-review for text Markdown would now reinterpret. One hit, and it argues for the change: nextcloud's `disableUnstableApps` builds `` `… <ul>${apps.map(a => `<li>${a}</li>`)}</ul>` `` into its message, which today shows the user literal angle brackets. Everything else is plain prose.

This also closes the gap I flagged at the bottom of the description — "show the user this block of text" is a fenced code block now, so it doesn't need a new result member type.

Unchanged: `start-cli` still prints the raw source, which is the right thing in a terminal; and a `single` value stays single-line, for the reasons in the description.

dr-bonez (approved):

PR #3826, opened 2026-08-25
docs(start-os): per-service outbound gateway overrides the system-wide default

helix-nine:
## Summary

The StartOS user book claimed that pinning a system-wide outbound gateway "forces _all_ outbound traffic for everything on the server" — which contradicts the per-service **Set Outbound Gateway** override documented on the same page. This PR makes the intended precedence explicit everywhere it's described:

- `outbound-vpn.md`: pinning a gateway sets a *default*; services with their own override keep their own gateway. The per-service section now states outright that the service's setting always takes precedence over the system-wide default (Auto or pinned), and how to clear an override ("System default" in the dialog).
- `cli-reference.md`: `package set-outbound-gateway` says it takes precedence over the system-wide default; `net gateway set-default-outbound` says it's a default used by services without their own override.

Docs-only change (`projects/start-os/docs/`); no code touched.

## Follow-up worth flagging

While verifying before writing this, I checked the actual routing and **the code currently does not implement this precedence** when a global gateway is *pinned*:

- Per-package overrides install `ip rule add from <service-ip> lookup <table> priority 100` (`net_controller.rs`).
- A pinned global default installs a selector-less catch-all at `priority 75` (`gateway.rs` `apply_default_outbound`).

Linux evaluates policy rules in ascending priority order, so the catch-all at 75 matches first and its table lookup succeeds — the priority-100 rule is never reached. Verified empirically on a VM: with rules `75: from all lookup 2001` and `100: from 203.0.113.10 lookup 2002`, `ip route get` from the package IP resolves via table 2001, not the package's table. In Auto mode (no pin) there is no 75 rule, so per-service overrides work as documented.

Happy to take the routing fix (e.g. installing per-package rules below 75) as a follow-up if you want it.

### DIFF
--- a/projects/start-os/docs/src/outbound-vpn.md
+++ b/projects/start-os/docs/src/outbound-vpn.md
@@ -22,16 +22,16 @@ Both options hide your home IP address, and in both cases the provider knows who
 
 ## Set System-Wide Default Gateway
 
-By default, StartOS dynamically selects which gateway to use for outbound traffic for optimal performance ("Auto" mode). You can override this under `System > Gateways > Outbound Traffic` by switching from "Auto" to a specific gateway. This forces _all_ outbound traffic for everything on the server through the selected gateway.
+By default, StartOS dynamically selects which gateway to use for outbound traffic for optimal performance ("Auto" mode). You can override this under `System > Gateways > Outbound Traffic` by switching from "Auto" to a specific gateway. This sets the system-wide default: every service uses it, except services with their own [per-service override](#route-individual-services-through-vpn), which keep their own gateway.
 
 ## IPv6 leak prevention
 
-StartOS treats IPv6 outbound routing the same way as IPv4: the default gateway is chosen by route metric, and you can pin all traffic to a specific gateway under `System > Gateways > Outbound Traffic`.
+StartOS treats IPv6 outbound routing the same way as IPv4: the default gateway is chosen by route metric, and you can set the system-wide default under `System > Gateways > Outbound Traffic`.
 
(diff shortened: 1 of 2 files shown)


projects/start-os/docs/src/outbound-vpn.md
```diff
@@ -22,16 +22,16 @@ Both options hide your home IP address, and in both cases the provider knows who
 
 ## Set System-Wide Default Gateway
 
-By default, StartOS dynamically selects which gateway to use for outbound traffic for optimal performance ("Auto" mode). You can override this under `System > Gateways > Outbound Traffic` by switching from "Auto" to a specific gateway. This forces _all_ outbound traffic for everything on the server through the selected gateway.
+By default, StartOS dynamically selects which gateway to use for outbound traffic for optimal performance ("Auto" mode). You can override this under `System > Gateways > Outbound Traffic` by switching from "Auto" to a specific gateway. This sets the system-wide default: every service uses it, except services with their own [per-service override](#route-individual-services-through-vpn), which keep their own gateway.
```
dr-bonez: note that this also sets how the OS itself reaches out over the internet. i.e. for downloading s9pks / connecting to registries

helix-nine: Added — the paragraph now reads: pinning a gateway "sets the system-wide default: it covers everything on the server — every service, and the OS itself (registry connections, package downloads) — except services with their own per-service override, which keep their own gateway." Pushed as 3d451f371.

dr-bonez (approved):

PR #3840, opened 2026-08-26
fix(start-os): align PCP and UPnP port-check behavior behind a second NAT

### LINKED ISSUE #3784: Port Forwarding check flips pass/fail behind double NAT depending on the granting protocol — PCP fast path trusts a private external IP, UPnP-held mappings discard it and probe the true public IP (net/gateway.rs:714)
### Environment

StartOS 0.4.0.2-unreleased (master-lineage build, `wrt/upnp-hostname-action` branch tip 64953eb8c; the behavior is master's — see Root cause; line numbers cited as of that tip), x86_64, bench server behind a StartWRT gateway that is itself behind another NAT (double NAT). Area: networking (gateways / domain validation).

### What happens

With automatic port forwarding active on a double-NAT gateway, the domain dialog's "Port Forwarding" check gives opposite verdicts for identical reachability, depending only on which protocol granted the mapping:

- A **PCP-held** mapping passes: `check_port`'s fast path sees an active mapping with a gateway-reported external IP and reports `open_externally: true` — even though that IP is RFC1918 (the inner gateway's WAN) and the port is not reachable from the internet.
- A **UPnP-held** mapping on the same gateway fails: the UPnP client discards the private `GetExternalIPAddress` result as non-routable, so the mapping carries no external IP, the fast path misses, and the check falls through to the echoip probe — which tests the true public IP (the outer NAT), where nothing forwards, and honestly reports unreachable.

Bench observation: two SNI hostname routes on the same gateway, one granted over PCP (check green), one over the UPnP vendor action (check red), both serving identically from the outer network.

The verdict should not depend on the granting protocol. The check conflates two questions — "did my gateway forward this port" (what the PCP fast path answers) and "is this port reachable from the public internet" (what the echoip probe answers). Either:

- both protocols trust the active mapping (define `open_externally` as "the gateway forwards it"), or
- both fall through to the echoip probe when the mapping's external IP is not a public candidate (define it as internet reachability — the minimal fix: apply the same `is_wan_candidate` filter to the PCP fast path), or
- (best UX) surface both facts separately so the UI can say "your gateway forwards this port, but your network appears to be behind a second NAT" — actionable, where a bare red check is not.

### Steps to reproduce

1. Put a gateway (e.g. StartWRT) behind another NAT; connect a StartOS server to its LAN and enable automatic port forwarding for it.
2. Bind a private domain through the gateway; let the mapping be granted over PCP. Run the domain dialog's Port Forwarding test → passes.
3. Force the same mapping over UPnP (e.g. drop UDP/5351 at the gateway and re-add the domain). Run the test again → fails. Reachability from the outer network is unchanged in both cases.

- `shared-libs/crates/start-core/src/net/gateway.rs:714-739` — `check_port`'s fast path returns `open_externally: true` whenever `port_map.mapped_external_ip()` yields an IP, with no public-address check.
- `shared-libs/crates/start-core/src/net/port_map/client.rs:415-416` — a PCP-held mapping reports `m.external_ip()` verbatim (private or not); a UPnP-held mapping reports its cached `external_ip`, which
- `client.rs:876-890` / `client.rs:1069-1082` populate from `upnp::external_ipv4()` → filtered through `is_wan_candidate` (`net/port_map/upnp.rs`), which deliberately discards RFC1918/CGNAT — so it is `None` behind double NAT and the fast path never triggers for UPnP.

The asymmetry: PCP trusts the gateway's word unfiltered; UPnP filters it and then the check probes the real internet. Each is defensible alone; together the same topology yields opposite verdicts by protocol.

### Logs & evidence

Bench 2026-08-21: gateway WAN 192.168.10.97 (inner, RFC1918), UDP/5351 blackholed at the gateway (nft counter confirmed 20 eaten PCP packets). `curl -vk --resolve <domain>:443:192.168.10.97 https://<domain>/` from the outer network serves the correct certificate and content for both domains; the Port Forwarding check showed green for the PCP-held route and red for the UPnP-held one.

Dominion5254:
*(Replaces the previous description, which described a wider rework: always-probe, mapping re-assertion, an RPC/sync-task ordering wait, and an `Option<bool>` tri-state. Per review — "we want to limit use of external echoip services where possible; sounds like we just need to fix the mismatch between upnp and pcp behavior when it comes to double-nat" — this is now the mismatch alone.)*

## What was wrong

`check_port` short-circuits to "open" whenever the port-map client holds a mapping for the port, and the two protocols disagreed about what the address in that record had to be:

- **PCP** reported whatever external address the gateway returned. Behind a second NAT that address is private, so the check passed a port nothing outside could reach.
- **UPnP** discarded a non-routable address at record time and fell through to the echo probe, so the identical topology answered red.

Same reachability, opposite verdicts by protocol — the issue as filed.

## What this does

The shard's `Command::ExternalIp` handler (`port_map/client.rs`) is where both protocols' addresses converge. The routability filter moves there, so it applies to whichever protocol reported the address. A gateway with no public address of its own falls through to the probe on both protocols, which is what UPnP already did.

v6 passes through untouched — the GUA is the box's own address, and `check_gua_port` only tests that a pinhole exists, never reading its address.

`gateway.rs` and `upnp.rs` are untouched; `is_wan_candidate` is the filter UPnP already used, applied unchanged.

### On echoip usage

No new echoip traffic beyond what a UPnP gateway already generates on this topology. The fall-through is the path UPnP has always taken; PCP now joins it only where the gateway's own reported address rules out a direct answer. The only alternative that would cut echoip further is answering `false` outright on a non-routable address — but that is wrong when the outer router carries a manual forward, and it would push a user whose port already works into the manual-forward modal.

## Verified

- `cargo test -p start-core --features=test port_map`: 80 passed. `cargo fmt --check`: clean.
- `Command::ExternalIp`'s body was inline in the shard's select loop and unreachable from a test; it is extracted as `external_ip_of`, with both protocol arms funnelling through one `routable_external_ip`. New tests cover the UPnP arm, key selection (right port, TCP only), a v6 pinhole passing through unfiltered, an addressless NAT-PMP grant, and the filter over private and public v4.
- The PCP arm is not constructible in a test — `crab_nat::PortMapping` has private fields and no public constructor — so it reaches the filter by construction: `external_ip_of` is the single call site and wraps both arms.
- **Bench (double NAT, PCP):** StartOS on the inner LAN of a router whose WAN holds an address from the outer router's LAN. With a PCP grant held for the port (confirmed on the router side), `check_port` reports the port not open. That is the premise of #3784 confirmed — a real PCP gateway behind a second NAT does report a private external address — and the verdict this changes. Master's green on the same setup is by inspection of the unfiltered short-circuit, not an A/B run.
- **The single-NAT accept path was not bench-run**, and cannot be on that hardware: the inner router always reports a private address, and the outer router speaks neither PCP nor UPnP, so no gateway there reports a routable one. It rests on the unit tests, which is a short reach — `check_port` is byte-identical to master and the only difference is one `is_wan_candidate` call on the reported address. `a_routable_address_is_reported_as_given` pins that public addresses pass, and `a_upnp_mapping_answers_only_for_its_own_tcp_port` exercises it end to end through `external_ip_of` — the function `check_port` calls.
- **The UPnP arm was not bench-run** (no IGD on that router) and cannot regress: master already filtered at record time (`upnp.rs:109`), and this adds a second, idempotent filter at read time.

## Not in this PR

Real but separate; each wants its own issue:

- **A withdrawn mapping stays green** (router reboot, StartWRT permission revocation) until the client's next renewal is refused. This is the follow-up comment on #3784; it needs re-assertion machinery.
- **Enable-then-check ordering.** Nothing sequences a host-mutating RPC's db write against the network sync task that applies it, so a check fired immediately after can measure the world from before the change. Theorized from the code; not observed in the field.
- **`is_wan_candidate` does not reject CGNAT** (`100.64.0.0/10`) despite its doc comment saying it does, so a gateway holding a carrier-assigned address passes the shortcut on both protocols. Left out here because the same function feeds the gateway poll's WAN-IP discovery, and rejecting it there changes what the box shows as its public IP and when it asks echoip.

Fixes #3784

https://claude.ai/code/session_01RwXsV4feB6gSVXTedUrzxT


dr-bonez (changes requested):
wrong direction. we want to limit use of external echoip services where possible. sounds like we just need to fix the mismatch between upnp and pcp behavior when it comes to double-nat

dr-bonez (dismissed):

dr-bonez (approved):

PR #3865, opened 2026-08-28
Curated known-registries list with pinned identity

MattDHill:
Start9 publishes a manifest of alternative registries at `marketplace.start9.com`, pinning the name and icon each one is expected to present. Users pick from that list instead of only typing a URL. StartOS displays the pinned name and uses the pinned icon while loading or when a live icon is unavailable.

## The manifest

`projects/brochure-marketplace/well-known/startos/registries.json`, copied to `.well-known` by an `angular.json` assets entry — the deploy rsyncs with `--delete`, so it has to come out of the build. It ships empty; nothing is listed until Start9 populates it.

Two readers, independently: the brochure fetches it same-origin, and StartOS fetches it through a new `marketplace.known-registries` RPC. Going through the backend keeps the request on the server's connection rather than adding a third-party request from the user's browser, and avoids a CORS header on a prod nginx none of us can change from here. Recorded as a coupled change in the root `AGENTS.md`, along with the wrinkle that the site answers an unknown path with `index.html` and a 200, so a stale path fails as a JSON parse error rather than a 404.

## Identity pinning

`registries$` resolves each entry through `resolveIdentity` rather than the name a registry reports. That closes a hole that exists today: `updateRegistryName` writes whatever `registry.info` returns straight into `ui.registries`, so a custom registry could label itself "Start9 Registry" and sit in the picker next to the real one. A live name colliding with a pinned one now renders as the host instead.

`identityMatches` detects the other direction — selecting a listed registry whose live name or icon has drifted from the pin shows a red banner for as long as that registry remains selected. The warning does not block registry use. Icons compare by MIME type and decoded bytes, since `DataUrl` formatting isn't stable enough for string equality. Missing metadata on only one side is a mismatch.

Once `registry.info` resolves, the shared icon directive honors a valid live `data:image/*` icon. Listed registries use it only when it matches the pin; failed image loads fall back through the pin to the generic storefront. The StartOS UI exposes icons for every fetched saved registry, while the brochure exposes its current registry. The existing `start-registry info set-icon` command is the registry-side configuration path; the five deployed Start9 registries currently return `icon: null` and still need provisioning.

Start9's own five registries carry compiled-in pins (`defaultIdentities`) so the protection holds offline and on first boot, and so the manifest can stay purely additive.

## Add dialog

"Add" now opens a dialog with two sections rather than a bare URL prompt: a select over the curated list (icon, name, URL), and the custom URL field. It carries a `tuiNotification` saying Start9 neither operates these registries nor supports their services.

That per-switch dialog is removed. Its operation-and-support caveat now remains visible in a yellow banner while any registry outside Start9's five is selected, including Community and listed or custom third-party registries. Identity drift on a listed registry adds the red banner described above.

## Drive-by

Three components style `tui-notification` — Taiga **4**'s element form, which matches nothing against v5's `<div tuiNotification>` host directive. All three were dead rules, which is why a dialog's banner sits flush against whatever follows it. One global rule in `shared.scss` replaces the two that were only about spacing.

Still broken and left alone: `routes/metrics/time.component.ts` has the same dead selector, and there it isn't cosmetic — the `:host-context(tui-root._mobile)` rule meant to swap the banner for an icon on mobile has never fired. Page component with visible mobile behavior, so it wants its own change.

## Bindings

`KnownRegistry.ts` and its `index.ts` line are hand-written to match what ts-rs would emit. `make start-core-ts-bindings` does not fit on the machine this was written on, so `Generated Artifacts` is the first thing to compile `marketplace.rs` at all — if the Rust or the binding is wrong, that job is where it shows.

## Not included

- The publish-time drift check — a Start9-side job that re-fetches each listed registry and drops one whose identity moved. Ops, not repo code, and it needs a decision about where it runs. Without it the client-side warning is the only guard, and a legitimate rename remains usable under that warning until someone updates the manifest.
- Signing the manifest. It is served from the same VPS as start9.com and docs.start9.com, so a compromise there could re-pin identities. `registry/signer.rs` already has the primitives; it needs a key-management decision.

## Verification

The original change passed `npm run check` across every project, `check-i18n` (all five dictionaries), `web-format-check`, `start-core-format-check`, and AOT builds of `ui` and `brochure`. After rebasing onto current `master`, adding live-icon support, and moving warnings to persistent marketplace banners: `npm run build:deps`, `check:marketplace`, `check:ui`, `check:brochure`, `check:i18n`, focused Prettier, and production builds of `ui` and `brochure`.


projects/start-os/web/ui/src/app/services/marketplace.service.ts
```diff
@@ -286,8 +309,18 @@ export class MarketplaceService extends AbstractMarketplaceService {
...
-    await this.api.setDbValue<string | null>(['registries', url], name)
+    const info = await firstValueFrom(this.fetchInfo$(url))
+    const pin = findKnown(url, await firstValueFrom(this.knownRegistries$))
+
+    if (pin && !identityMatches(pin, info)) {
+      throw new Error(
+        this.i18n.transform(
+          'This registry no longer presents the name and icon Start9 published for it, so it cannot be added from the list. It may have changed hands.',
+        ),
+      )
```
dr-bonez: is throwing the right thing to do here?

helix-nine: No. I changed this to a persistent, dismissible warning (`appearance: warning`, `autoClose: 0`) and let the add continue. Registry-info failures and duplicate adds still error normally; only identity drift is non-blocking. Applied in a1f45f27b to both StartOS and the brochure.

projects/brochure-marketplace/src/app/services/marketplace.service.ts
```diff
@@ -140,8 +160,25 @@ export class MarketplaceService extends AbstractMarketplaceService {
...
-    const { name } = await firstValueFrom(this.fetchInfo$(url))
-    this.setCustom({ ...this.custom$.value, [url]: name })
+    const info = await firstValueFrom(this.fetchInfo$(url))
+    const pin = findKnown(url, await firstValueFrom(this.knownRegistries$))
+
+    if (pin && !identityMatches(pin, info)) {
+      this.alerts
+        .open(
+          this.i18n.transform(
+            'This registry does not present the name and icon Start9 published for it. It may have changed hands.',
```
dr-bonez:
```suggestion
            'This registry does not present the name and icon Start9 published for it. Start9 has not validated that it is what it claims to be.',
```

helix-nine: Applied this wording in be41c881d in both frontends and updated all five translations.

dr-bonez (approved):

PR #3869, opened 2026-08-30
fix(start-os): bound service-side TLS establishment

### LINKED ISSUE #3776: vhost: a container that accepts TCP but never completes TLS parks the client's handshake forever

helix-nine:
## Summary

- Cap the ordinary backend TCP connect and optional service-side TLS handshake at one shared 15-second deadline.
- Preserve source-preserving transparent-connect setup and fallback behavior outside that deadline.
- Add a paused-time regression test for a backend that accepts TCP and never speaks TLS.
- Document the observable timeout in the StartOS interface guide and changelog.

## Why

TLS listener phase 2 resolves the service-side connection before completing the client handshake. A service could accept TCP and then never answer its own TLS handshake, leaving the client parked indefinitely and retaining both sockets in the listener's in-progress set.

The deadline belongs inside `ProxyTarget::preprocess`, around backend establishment. Bounding all TLS config resolution would also cancel ACME certificate work and root-CA state updates, so those remain outside it.

## Verification

- `cargo test -p start-core upstream_alpn_tests --features=test`
- `cargo test -p start-core --features=test --lib` — 750 passed, 3 ignored
- `cargo check -p start-core --features=test --all-targets`
- Prettier checks for the changelog and interface guide
- Taplo check for `shared-libs/crates/start-core/Cargo.toml`
- Timeout ablation: replacing the deadline wrapper with a direct await makes the regression test fail on its outer guard

`cargo check -p start-os --all-targets` reaches the existing embedded-UI prerequisite and cannot run in this checkout because `projects/start-os/web/dist/static/{ui,setup-wizard}` has not been built.

Closes #3776

### DIFF
--- a/shared-libs/crates/start-core/src/net/vhost.rs
+++ b/shared-libs/crates/start-core/src/net/vhost.rs
@@ -743,20 +743,6 @@ fn compute_bind_reqs<A: Accept + 'static>(mapping: &Mapping<A>) -> VHostBindRequ
 const BIND_RETRY_BACKOFF: Duration = Duration::from_secs(2);
 const BACKEND_DIAL_TIMEOUT: Duration = Duration::from_secs(15);
 
-async fn with_backend_deadline<T>(
-    deadline: tokio::time::Instant,
-    target: SocketAddr,
-    future: impl Future<Output = T>,
-) -> Option<T> {
-    match tokio::time::timeout_at(deadline, future).await {
-        Ok(value) => Some(value),
-        Err(_) => {
-            tracing::warn!("no usable connection to {target} within {BACKEND_DIAL_TIMEOUT:?}");
-            None
-        }
-    }
-}
-
 /// Listener that manages its own TCP listeners with IP-level precision.
 /// Binds ALL IPs of public gateways and ONLY matching private IPs.
 pub struct VHostBindListener {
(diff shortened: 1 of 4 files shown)


shared-libs/crates/start-core/src/net/vhost.rs
```diff
@@ -1254,23 +1269,29 @@ where
...
         let peer = extract::<TcpMetadata, _>(metadata).map(|m| m.peer_addr);
         let plain_connect = || async {
-            TcpStream::connect(self.addr)
-                .await
+            let deadline = tokio::time::Instant::now() + BACKEND_DIAL_TIMEOUT;
+            let stream = with_backend_deadline(deadline, self.addr, TcpStream::connect(self.addr))
+                .await?
                 .with_ctx(|_| (ErrorKind::Network, self.addr))
-                .log_err()
+                .log_err()?;
```
dr-bonez, line 1264: why not just tokio::time::timeout??

helix-nine: No good reason for the wrapper. I removed it and now use Tokio’s timeout API directly at both awaits. I kept `timeout_at` rather than two independent `timeout` calls so the ordinary TCP connect and optional TLS handshake share one 15-second budget instead of each receiving 15 seconds. Rebased and re-ran the ALPN tests plus the start-core all-targets check in ab81885af.

dr-bonez (approved):

PR #3872, opened 2026-08-31
feat(startos): persist a CPU energy/performance preference and apply it at boot

MattDHill:
## Why

`intel_pstate` and `amd-pstate` in active/EPP mode expose only the `performance` and `powersave` governors:

```
$ cat /sys/devices/system/cpu/cpu0/cpufreq/scaling_available_governors
performance powersave
```

So `GOVERNOR_HEIRARCHY` (`ondemand`, `schedutil`, `conservative`) never matches and `init.rs` sets nothing — silently, the log line being inside the `if let Some`. That is harmless in itself: `powersave` is the right governor on those drivers and the machine is already on it. What matters is that **nothing ever touched EPP**, the hint that actually decides how hard HWP chases a burst, so every such server runs on whatever its firmware left in the request register — and StartOS had no way to change it.

On a Server Pro (Librem Mini v2, i7-10710U) PureBoot leaves `performance` on all 12 CPUs. A 3% aggregate poll took the package from 55 °C to 90 °C in under a second and back; 34 days of uptime had logged **24,881,470 package-throttle events**, and the chassis fan chased every excursion, cycling every few seconds indefinitely.

## Measured

Same box, steady-state workload (mempool's 8-second poll plus bitcoind/electrs/Fulcrum), 50-second windows:

| EPP | idle / median | worst seen | clock ceiling |
|---|---|---|---|
| `performance` (PureBoot default) | 55–58 °C | **92 °C** | 4.09 GHz |
| `balance_performance` (stock Linux default) | 47–48 °C | **82 °C** steady state, **90 °C** under load | 3.91 GHz |
| `balance_power` | 47–48 °C | **58 °C** | 2.0 GHz |

`balance_performance` fixes the median but still lets the clock reach 3.9 GHz, so a real burst still spikes the die into the 80s — and the fan tracks spikes, not medians. `balance_power` is the only value measured to stop the cycling on this hardware.

## The change

- **`serverInfo.epp`** persists a preference (`#[serde(default)]`, so existing databases deserialize). Init applies it at boot, warning if the driver does not offer that value.
- **`start-cli server epp [<value>]`** reads the live sysfs value and sets the persisted one, mirroring `governor`. It reports the live value because the persisted one is `null` on every box that has not set it.
- **When no preference is saved, StartOS defaults Librem Mini v2 systems to `balance_power` if the driver offers it.** Other systems retain their current EPP. A saved preference always takes precedence.
- Machines whose driver exposes no EPP attribute — aarch64, VMs, `intel_pstate` passive, acpi-cpufreq, pre-HWP Intel — read an empty set and are left alone.

The governor is deliberately untouched: the hierarchy falling through costs nothing on these drivers, and appending `powersave` would pin acpi-cpufreq boxes to minimum frequency.

## Review changes

Per dr-bonez: every fallible cpupower call at init — governor and EPP — goes through `log_err()` rather than `?`, so a sysfs failure logs and moves on instead of dropping the server into Diagnostic Mode. `get_preferred_governor` became `preferred_governor`, taking the available set the caller already read.

The CPU governor and EPP commands now live directly under `start-cli server`; `server experimental` retains only the ZRAM command.

## Drive-by

`display_serializable(format, params)` at `system/mod.rs` and `ssh.rs` serialized the command's arguments instead of its result — `server governor --format json` and `ssh list --format json` both returned the wrong object. Verified against `action.rs`.


shared-libs/crates/start-core/src/init.rs
```diff
@@ -334,6 +334,22 @@ pub async fn init(
         tracing::info!("{}", t!("init.setting-cpu-governor", governor = governor));
         cpupower::set_governor(governor).await?;
     }
+
+    let selected_epp = server_info.as_epp().de()?;
+    let epp = if let Some(epp) = &selected_epp {
+        if cpupower::get_available_epps().await?.contains(epp) {
```
dr-bonez: I would prefer this to warn on failure

shared-libs/crates/start-core/src/init.rs
```diff
@@ -334,6 +334,22 @@ pub async fn init(
...
+            Some(epp)
+        } else {
+            tracing::warn!("{}", t!("init.cpu-epp-not-available", epp = epp));
+            None
+        }
+    } else {
+        cpupower::get_preferred_epp().await?
```
dr-bonez: Same here

shared-libs/crates/start-core/src/init.rs
```diff
@@ -334,6 +334,22 @@ pub async fn init(
...
+    };
+    if let Some(epp) = epp {
+        tracing::info!("{}", t!("init.setting-cpu-epp", epp = epp));
+        cpupower::set_epp(epp).await?;
```
dr-bonez: And here

shared-libs/crates/start-core/src/init.rs
MattDHill: Agreed — an `Err` escaping `init` drops the server into Diagnostic Mode, and neither setting is required for the server to run. Fixed in 1a6ca4af7, covering all three sites.

Every fallible cpupower call in the block now goes through `log_err()` rather than `?`: an unreadable available-set degrades to empty (which selects nothing), and a failed write is logged and passed over.

Two things worth flagging, since they go slightly beyond what you marked:

**I extended it to the governor block.** It carries the identical `?`s three lines up, and leaving the governor able to fail the boot while the EPP cannot seemed incoherent. Say the word if you would rather I left that half alone — it is a clean revert.

**`get_preferred_{governor,epp}` became `preferred_{governor,epp}`**, taking the available set the caller already read instead of reading it a second time. That fell out of hoisting the read so a single `log_err()` covers it; it also drops a redundant `cpupower` invocation on the no-override path.

dr-bonez (dismissed):

dr-bonez (approved):

PR #3873, opened 2026-08-31
feat(registry): stream package-index changes over a websocket

MattDHill:
Adds `db.subscribe` to the registry RPC surface, so a client can follow the package index instead of re-fetching it.

`POST …/rpc/registry/db/subscribe` returns `{ dump, guid }`; connecting to `GET …/ws/rpc/<guid>` then streams a `Revision` (`{ id, patch }`) each time the index changes.

## Why

Nothing on a registry pushes. A consumer that wants to react when a package or a version lands has to poll `package.get` or `package.index`, and those responses are large and uncacheable — measured against production, with `otherVersions: "none"` and no ETag or gzip on any of them:

| registry | bytes |
| --- | --- |
| `registry.start9.com` | 1.9 MB |
| `beta-registry.start9.com` | 2.0 MB |
| `community-registry.start9.com` | 8.6 MB |
| `community-beta-registry.start9.com` | 21.3 MB |

That is ~34 MB per sweep across the fleet, for a signal that fires a few times a week.

## Shape

The registry already carried the machinery: `RegistryContext` holds a `TypedPatchDb<RegistryDatabase>` and `RpcContinuations`, and `registry_router` mounts `/ws/rpc/{*path}`. No registry method had ever called `rpc_continuations.add`, so no GUID was ever issued and that route was unreachable — this is the first thing to use it.

The handler mirrors `db::subscribe` in `src/db/mod.rs`, minus the `sync_db` merge, which has no registry equivalent.

## Scope and auth

Pinned to `/index/package`, not a caller-supplied pointer, and tagged `authenticated: false`.

`package.index` already returns that whole subtree unauthenticated (`get_package_index`, `authenticated: false`), so the stream carries nothing a client could not already fetch. A caller-supplied pointer would reach `/admins` and `/index/signers` on a public handler, which is why the pointer is fixed. patch-db scopes a subscriber's patches to its pointer (`ScopedSender::send` → `Revision::for_path`), so patches arrive relative to `/index/package` and revisions that do not touch it are dropped rather than delivered empty — a consumer must not assume contiguous revision ids.

No CLI surface (`no_cli`), so no manpage change; no `#[ts(export)]`, so no bindings regeneration.

## Consumer

support-server will hold one subscription per registry (alpha, beta, prod, community-beta, community-prod) and diff successive states to announce new packages and new versions to Matrix and X.

### DIFF
--- a/shared-libs/crates/start-core/src/registry/db.rs
+++ b/shared-libs/crates/start-core/src/registry/db.rs
@@ -1,5 +1,7 @@
 use std::path::PathBuf;
+use std::time::Duration;
 
+use axum::extract::ws;
 use clap::Parser;
 use itertools::Itertools;
 use patch_db::Dump;
@@ -14,7 +16,13 @@ use crate::context::CliContext;
 use crate::prelude::*;
 use crate::registry::RegistryDatabase;
 use crate::registry::context::RegistryContext;
+use crate::rpc_continuations::{Guid, RpcContinuation};
 use crate::util::serde::{HandlerExtSerde, apply_expr};
+
+lazy_static::lazy_static! {
+    /// Must stay within what `package.index` already serves unauthenticated.
+    static ref PACKAGE_INDEX: JsonPointer = "/index/package".parse().unwrap();
+}
 
 pub fn db_api<C: Context>() -> ParentHandler<C> {
     ParentHandler::new()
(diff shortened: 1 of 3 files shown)


shared-libs/crates/start-core/src/registry/db.rs
```diff
@@ -96,6 +110,61 @@ pub async fn dump(ctx: RegistryContext, DumpParams { pointer }: DumpParams) -> R
         .await)
 }
 
+#[derive(Deserialize, Serialize, TS)]
+#[serde(rename_all = "camelCase")]
+pub struct SubscribeRes {
+    #[ts(type = "{ id: number; value: unknown }")]
+    pub dump: Dump,
+    pub guid: Guid,
+}
```
dr-bonez: should be able to reuse this type

MattDHill: Done — `use crate::db::SubscribeRes`, local copy dropped. cb60904

dr-bonez (approved):

PR #3783, opened 2026-08-21
feat(start-wrt): SNI hostname routes end to end — UPnP vendor actions, StartWRT dataplane, Remote Access coexistence

Dominion5254:
## Summary

Adds hostname-based shared-port mappings to StartTunnel and StartWRT. A device can register a hostname over PCP or the new UPnP vendor actions, and the gateway routes TLS by ClientHello SNI without terminating TLS.

## Changes

- Adds `X_START9_AddHostnameMapping` and `X_START9_DeleteHostnameMapping` to the shared UPnP IGD server and client fallback path.
- Runs the shared SNI demultiplexer on StartWRT, including reply-path diversion, WAN admission rules, lease cleanup, WAN re-keying, and Remote Access fallback on port 443.
- Preserves source addresses except for same-subnet hairpin connections.
- Makes route grants transactional: listener binding, reply diversion, firewall activation, persistence, lease renewal, and rollback are coordinated before success is returned.
- Serializes StartTunnel forward mutations and StartWRT WAN re-keying with route updates.
- Treats manual/automatic collisions by transport, so UDP-only mappings do not block TCP SNI routes.
- Reports hostname-route owners in the Published Ports UI and uses consistent automatic-forward/WAN-port terminology across the API and frontend.
- Rewrites the UPnP proposal as the initial Internet-Draft `rfcs/draft-start9-upnp-hostname.md`; it does not claim external acceptance.
- Updates StartWRT and StartTunnel documentation and the StartWRT changelog.
- Repairs the hostname-validation compile regression carried in by the latest master merge.

## Validation

- `cargo test -p start-core --lib` — 739 passed, 3 ignored
- `cargo test -p startwrt-core --lib` — 550 passed
- `npm run check:wrt`
- `npm run check:i18n:wrt`
- `mdbook build projects/start-wrt/docs`
- `mdbook build projects/start-tunnel/docs`
- `cargo fmt --all -- --check`
- `git diff --check`

The original branch was bench-tested on a K1 router with a StartOS server behind it. PCP and UPnP fallback, SNI routing, Remote Access coexistence, manual-forward conflicts, WAN re-keying, restart reclaim, lease expiry, and same-/cross-subnet hairpin behavior were exercised. The StartWRT review-fix paths were retested: first-grant firewall activation, UPnP add/delete validation, and manual-forward displacement of an SNI route passed. A final StartTunnel bench pass for transactional persistence and the forward write lock remains.

## Deployment note

[...]

PR #3888, opened 2026-09-02
fix(start-wrt): hairpin published ports for every profile with Internet or LAN access

Dominion5254:
## Problem

A published port was only reachable at the router's WAN address from the target device's own Security Profile. fw4 defaults a redirect's `reflection_zone` to its `dest` zone, so the hairpin rules matched the target's subnet alone. A client on any other profile sent its packet to the router's INPUT chain instead of the DNAT and got the router's own web UI. Found by running Synapse on a public domain behind StartWRT: the domain worked from the server's profile and from the Internet, but not from a phone on the guest profile.

## Policy

A published port is a public resource, so it is reachable at the router's public address from every profile that qualifies on either ground:

- **Internet access.** WAN Access is All or Blacklist and no blackout window is active. Whitelist and None do not qualify: someone who cut a profile off from the Internet has cut it off from public resources too.
- **LAN access.** The profile's Access setting includes the target's profile, so it can already reach the device at its LAN address.

Only the published port opens. A profile without Access still cannot reach the device at its LAN IPv4 address or on any unpublished port. A Source-restricted rule is never hairpinned: fw4's reflection rules drop `src_ip`, so it would have been reachable by every device on the zone (the `### Security` changelog entry).

The hairpin is local. The reflected DNAT rewrites the packet inside the router and it goes out the target's VLAN; nothing touches the WAN interface, and the server sees the client's real address.

The first version of this PR honored profile isolation on the public address, listing only the target's zone and the zones with Access to it. Review pointed out that blocking a profile from a device the operator has deliberately published buys nothing, and the Internet ground was added.

## What changed

- **Hairpin list** (`b6d91cbe0`, `7cd7e150d`, `04c703526`). `reflection_zone` is derived from the firewall config: the target's zone, every zone with a `config forwarding` into it, and every zone with a forwarding into a masquerading zone that no unqualified REJECT toward that zone cancels (the Whitelist catch-all and the blackout rule are both such rejects). Only existing, non-masquerading zones are emitted, because fw4 drops the whole redirect over one unknown name, WAN DNAT included.
- **Forward accept** (`04c703526`). Listing a zone is not enough on its own: fw4 sets the flag that emits `ct status dnat accept` only on the redirect's src and dest zones (`fw4.uc:2748`), never on the reflection zones, so a reflected flow from an isolated zone reached its forward chain and was rejected by policy. The daemon now installs one chain-pre include for fw4's top-level `forward` chain, `ct status dnat accept`, at boot, so binary-updated routers converge without a reflash. It matches only flows a DNAT rule rewrote, and the only DNAT into the LAN is a published port or an automatic forward.
- **IPv6** (`04c703526`). No DNAT to reflect, so each unrestricted `pp_*_v6` WAN rule is copied once per hairpin zone with only `src` changed, tagged `_pp_hairpin`, and rebuilt wholesale on every sync. `list` ignores the copies; `reconcile` retargets them; forgetting the device removes them with the rule.
- **Resync** (`9da23109a`, `5318981d4`, `04c703526`). The projections are derived state, so a `sync_hairpin` pass runs at the end of `rewrite_firewall` and `delete_config`, in `published-ports set`, in the automatic-forward apply, in `evaluate_and_apply_schedules`, and once at daemon start. The WAN-schedule crontab rewrites the firewall outside the daemon, so both cron edges now run `startwrt-cli published-ports sync-hairpin` between `uci commit` and the reload. Automatic UPnP/PCP forwards get the same scoping through the pass.
- **Target zone by address** (`c5b817f6f`, `982a2227e`). The zone is resolved from the device's IPv4 against the profile subnets, with the static reservation covering offline devices and the neighbor table as fallback. A device that still cannot be placed keeps `dest 'lan'` for the WAN DNAT and gets `reflection '0'`.
- `dcc98898c` types `FirewallRedirect.reflection` as `Option<bool>`; `fa7fba5e5` trues up the docs.

Cross-zone reflection is DNAT-only by design: the reflection SNAT lands in the emitted zone's `srcnat` chain, which a hairpinned flow from another zone never enters. Same-zone flows are SNATed to the router's zone address as before.

## Verification

566 `startwrt-core` unit tests, `make format-check` clean. New tests cover the eligibility predicate over All/None/Whitelist/Blacklist/blackout/Access-only zones, the IPv6 copy set and its regeneration, that `list` and forget treat the copies as plumbing, and that both cron edges carry the resync.

Hardware, K1 behind an upstream router, deployed as a binary update, StartOS server on Admin, manual TCP 443 rule:

| Case | Result |
|---|---|
| Guest with Access to Admin: hairpin at the WAN address (the reported bug) | pass, earlier revision |
| Source restricted: `reflection '0'`, `dstnat_` chains empty; restoring Source brings the hairpin back | pass, earlier revision |
| Profile created with Access joins the list; deleted profile leaves it, `fw4 check` clean, WAN DNAT intact | pass, earlier revision |
| IoT with Same profile only and WAN All: `https://<WAN address>/` serves the StartOS login, not the router UI | pass |
| Same profile, WAN None: the router UI answers; back to All: StartOS again | pass |
| Unpublished port 80 at the WAN address: router answers itself | pass |

Behind an upstream router only addresses on the router's own WAN interface hairpin, never the public IP; the docs say so. IPv6 copies and the blackout cron edge are unit-tested only.

## Merging

Rebased onto master after #3783 landed. The conflicts were all additive on both sides (the `_apf_label` and `_pp_hairpin` fields, the `wan-changed` and `sync-hairpin` subcommands, one API table row), plus the `LABEL_PCP` to `KIND_PCP` rename in one test.

Docs and `CHANGELOG.md` (under the unreleased 1.1.0) are updated in the same change.

https://claude.ai/code/session_016bxSq9BshATVCVZy3eoKu9
https://claude.ai/code/session_01Ctt2FvqXEcPCJCtmGayHee


projects/start-wrt/docs/src/published-ports.md
```diff
@@ -58,6 +58,15 @@ Each published port rule shows a status indicator in the table:
...
+
+It works from the target device's own [Security Profile](security-profiles.md) and from every other profile whose **Access** setting includes that profile — the same access those profiles already grant on the LAN, just by a different address. A profile that is not allowed to reach the device is not given this route either: for its connections to the router's public address, the router answers instead of the device. Ports opened through [automatic port forwarding](#automatic-port-forwarding) are served the same way.
+
+> [!NOTE]
+> A rule whose **Source** is restricted to specific addresses is never served this way. That route cannot distinguish one local device from another, so serving the rule over it would let any device on your network past the restriction. Reach a restricted rule from inside your network at the device's own LAN address instead.
```
dr-bonez, line 70: I'm not sure it's coherent to block a device from others on the lan, but grant public access via a port forward. in that case, we should allow the other devices on the lan to access that port, no?

Dominion5254: Agreed, and `04c703526` adopts it. A published port is a public resource, so the hairpin now serves every profile that could reach it from the Internet (WAN Access All or Blacklist, outside a blackout window) or that already has Access to the device's profile. A profile with neither is refused, since it couldn't reach the port from outside either.

Two things worth knowing. Listing the zone in `reflection_zone` wasn't enough: fw4 only emits its `ct status dnat accept` on the redirect's own src and dest zones, so the daemon now installs a chain-pre accept for the forward chain. And only the published port opens: the device's LAN address and unpublished ports stay closed to a profile without Access, because nothing DNATs them. The hairpin itself is local, never leaving the router. The docs section and this note are rewritten to say so.

dr-bonez (approved):

PR #3904, opened 2026-09-04
fix(scripts): resolve the personal signing key from the repo and overwrite partial signatures

MattDHill:
## Summary

- `cmd_sign` and `cmd_cosign` resolve `user.signingkey` after `enter_release_dir` has `cd`'d into the release staging directory, so the repo-local key is never seen and the global one is used. With a global SSH signing key that is a `.pub` path, and gpg aborts with `No secret key` after writing the first Start9 signature. This happened cutting start-cli 2.0.0.
- Read the key from `$REPO_ROOT`, and keep the personal signature only when the repo's `gpg.format` is unset or `openpgp` **and** `gpg --list-secret-keys` on PATH knows the key. Otherwise `sign` skips it with the existing warning and `cosign` fails before downloading anything. Signing stays on `gpg` from PATH, the contract the Start9 key, the apt suite signature and pre-check already rely on; Git's `gpg.program` is deliberately not consulted, since Git guarantees only its `-bsau`/`--verify` interface for it and resolves a relative path from the worktree.
- `sign` rebuilds `signatures/` from empty on every run. `cosign` extracts the release's current archive into an empty directory and removes the current user's entries before regenerating them, so other maintainers' signatures actually present in the archive survive and nothing left in the staging directory from an earlier run is republished. The `start9` signer name is reserved for the organizational signature files.
- Pass `--yes` to the detach-sign calls so re-running `sign` or `cosign` after a partial failure overwrites the half-written `signatures/*.asc` instead of stopping at gpg's overwrite prompt.
- The install guide says additional maintainer signatures may be included in `signatures.tar.gz`, without asserting who they belong to.

Verified by signing start-cli 2.0.0, start-tunnel 1.3.0 and start-registry 1.1.0 with the patched script.

### DIFF
--- a/projects/start-os/docs/src/installing-startos.md
+++ b/projects/start-os/docs/src/installing-startos.md
@@ -124,7 +124,7 @@ Everything rests on that fingerprint being the right one, so confirm it somewher
 > [!WARNING]
 > The download in the next step contains a copy of the key, as `start9.key.asc`. Do not use that one. A key that arrives alongside the signature it is checking proves nothing.
 
-**2. Download `signatures.tar.gz`** from the release page and unpack it, into the same folder as your image. It holds one signature per image, named after the image with `.start9.asc` on the end. A second signature from the person who cut the release is in there too — the `.start9.asc` one is the one to check.
+**2. Download `signatures.tar.gz`** from the release page and unpack it, into the same folder as your image. It holds one signature per image, named after the image with `.start9.asc` on the end. A second signature from the person who cut the release may be in there too — the `.start9.asc` one is the one to check.
 
 **3. Check your image.** Replace `FILENAME` with the name of the file you downloaded, in both places:
 
--- a/projects/start-os/docs/src/installing-startos.md
+++ b/projects/start-os/docs/src/installing-startos.md
@@ -124,7 +124,7 @@ Everything rests on that fingerprint being the right one, so confirm it somewher
 > [!WARNING]
 > The download in the next step contains a copy of the key, as `start9.key.asc`. Do not use that one. A key that arrives alongside the signature it is checking proves nothing.
 
-**2. Download `signatures.tar.gz`** from the release page and unpack it, into the same folder as your image. It holds one signature per image, named after the image with `.start9.asc` on the end. A second signature from the person who cut the release may be in there too — the `.start9.asc` one is the one to check.
+**2. Download `signatures.tar.gz`** from the release page and unpack it, into the same folder as your image. It holds one signature per image, named after the image with `.start9.asc` on the end. Additional maintainer signatures may be included — the `.start9.asc` one is the one to check.
 
 **3. Check your image.** Replace `FILENAME` with the name of the file you downloaded, in both places:
 
--- a/scripts/manage-release.sh
+++ b/scripts/manage-release.sh
@@ -313,7 +313,8 @@ release_files() {
 
 resolve_gh_user() {
     GH_USER=${GH_USER:-$(gh api user -q .login 2>/dev/null || true)}
-    GH_GPG_KEY=$(git config user.signingkey 2>/dev/null || true)
+    GH_GPG_KEY=$(git -C "$REPO_ROOT" config user.signingkey 2>/dev/null || true)
+    [ "$(git -C "$REPO_ROOT" config gpg.format 2>/dev/null)" != ssh ] || GH_GPG_KEY=
 }
 
 require_kind() {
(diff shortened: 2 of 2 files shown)


helix-nine (changes requested):
Reviewed `git diff origin/master...HEAD` with a high-effort fan-out and independent verification of each candidate. `bash -n scripts/manage-release.sh` passes.

Two changes are needed:

1. **`scripts/manage-release.sh:316-317` — respect the configured OpenPGP format and program.** The guard excludes only `ssh`, but Git also supports `x509`; an X.509 `user.signingkey` therefore reaches the literal OpenPGP `gpg` calls and aborts signing/cosigning with “No secret key.” The new repo-local key lookup can also select a key provided through repo-local `gpg.openpgp.program`/`gpg.program`, while the script still invokes literal `gpg`. Please retain the personal key only when `gpg.format` is unset or `openpgp`, and use the configured OpenPGP program (falling back to `gpg`) for signing and export.

2. **`projects/start-os/docs/src/installing-startos.md:127` — align the documented archive contents with SSH signing.** This change intentionally omits the releaser signature and key when the repo uses SSH signing, but the user guide unconditionally promises a second releaser signature. Remove that assertion so an expected omission is not mistaken for an incomplete archive.

The comment pass was clean.

helix-nine (changes requested):
Re-reviewed the fix as `git diff 61ade9ea8b5404b689524bca093d56b030d9f947..37010e8cd3eb849de0668709c36f7045c9101079` with a high-effort fan-out and independent verification. `bash -n scripts/manage-release.sh` passes. My first review's recommendation to use Git's configured program was underspecified; the fix exposed three additional configuration semantics that need handling.

1. **`scripts/manage-release.sh:321-323,1340,1346,1376,1378` — the configured program resolution does not match Git's contract.** Three valid configurations fail:
   - `gpg.program` and `gpg.openpgp.program` are aliases applied in configuration order, but these separate lookups always prefer any modern-name value. A global `gpg.openpgp.program` therefore incorrectly beats a repo-local `gpg.program`.
   - Git runs a slash-relative program from the worktree root; the script invokes it later from the release directory, producing exit 127.
   - Git only guarantees that a custom OpenPGP program supports its `-bsau`/`--verify` interface. It need not support the full `--detach-sign --armor -o` and `--export -a` interface now assumed here.

   Please preserve Git's alias ordering and relative-path semantics, and do not infer full GnuPG compatibility from the Git setting. The release flow needs either a separately explicit full-GPG contract/override or signing through Git's guaranteed interface with export capability handled separately.

2. **`scripts/manage-release.sh:1336-1350` — skipped personal signing can republish stale files.** On a `sign` rerun after the personal signer becomes unavailable or changes to SSH/X.509, the guarded writes are skipped but the previous `*.${GH_USER}.asc` and `${GH_USER}.key.asc` files remain in `signatures/` and are archived again. If the staged artifacts changed, those retained signatures are invalid. Ensure skipped regeneration cannot silently republish the current signer's stale files while preserving intentionally retained cosigners.

3. **`projects/start-os/docs/src/installing-startos.md:127` — the remaining attribution is still too narrow.** `cmd_cosign` can append signatures from additional maintainers after the release was cut, so there may be multiple signatures and they need not belong to the release cutter. “Additional maintainer signatures may be included” states the archive shape without asserting unsupported provenance.

The comment pass was clean.

helix-nine (changes requested):
Re-reviewed `git diff 37010e8cd3eb849de0668709c36f7045c9101079..b70b50f044ff741f86406003060bd31ed08ecb65` with a high-effort fan-out and independent verification. The configured-program and documentation findings are addressed, and `bash -n scripts/manage-release.sh` passes. I also added the required `StartOS` label for the book change.

Two retry-path issues remain:

1. **`scripts/manage-release.sh:316,1338-1344,1359-1376` — a configured key is assumed to be available to PATH `gpg`.** Keeping GnuPG-on-PATH as the release contract is coherent, but a nonempty Git `user.signingkey` does not prove that this `gpg` can access it. A stale key or one backed only by Git's custom program passes the guards, then optional personal signing aborts `sign` after the release has already been tagged/published. `cosign` downloads the archive before failing the same way. Please preflight the personal key against PATH `gpg`: skip it with the existing warning in `sign`, and fail before downloading anything in `cosign`.

2. **`scripts/manage-release.sh:1333-1348,1365-1378` — persistent staging can republish stale signatures.** The new cleanup removes only the resolved current user's files. On a forced re-pull, signatures from other maintainers can remain against replaced artifact bytes; if `GH_USER` resolution fails, no personal files are removed. `cosign` also extracts the downloaded archive over the persistent local directory, so local-only or obsolete files can be resurrected. Rebuild `sign`'s signature directory from empty. For `cosign`, extract the current remote archive into an empty directory, then remove the current user's entries before regenerating them; that preserves other signatures actually present in the downloaded archive without retaining local residue.

The comment pass was clean.

helix-nine (approved):
Re-reviewed through `effd10284`. All requested changes are addressed; the final reserved-namespace guard passes the focused tests, shell syntax check, and diff check.

dr-bonez (approved):

PR #3910, opened 2026-09-06
fix(start-cli): say when the workspace's start-technologies checkout is off live-docs

MattDHill:
## Summary

`s9pk init-package` scaffolds from the package template in whatever branch the workspace's `start-technologies` checkout is on, and never looks at that branch. A workspace scaffolded by start-cli 1.x cloned `master`, whose template has pinned `@start9labs/start-sdk@3.0.0` since #3900 — a version npm does not have — so `npm install` fails with `ETARGET` and nothing names the cause:

```
$ start-cli s9pk init-package "Hello World"
Installing dependencies (npm install)…
npm error notarget No matching version found for @start9labs/start-sdk@3.0.0.
```

start-cli 2.0.0 clones `live-docs` for new workspaces and documents a manual `git checkout live-docs` for existing ones, which a 1.0.1 workspace cannot even run (its clone is single-branch, shallow, and sparse).

### Diagnosis, no git

`init-package` — right after workspace discovery, before the package-name and destination checks, so the scaffold directory a failed `npm install` left behind cannot hide it — and `init-workspace`, when it finds an existing checkout, print one line when `start-technologies` is on a branch other than `live-docs`. It names the checkout's path and branch and says what to do:

- **A checkout the workspace owns:** move it out of the workspace, then run `start-cli s9pk init-workspace '<workspace>'`, which clones a fresh checkout on `live-docs`. Moving rather than deleting keeps any local work; the discovered workspace path is single-quoted, and when it is not valid UTF-8 the notice says to run `init-workspace` from the workspace root instead of printing a path.
- **A symlinked checkout** (the "already have the monorepo" layout): the notice names the link and its target and says to remove the link and run the same command, so the workspace gets a checkout of its own and the owner's branch is never touched.

Nothing prescribes or performs a git operation. No single command sequence is right for every checkout state (single-branch refspecs, shallow and sparse clones, stale or diverged local branches, dirty trees, forks), and a fresh blobless clone costs seconds.

### Detection

The branch is read from `start-technologies/.git/HEAD`, following a `gitdir:` file for a linked worktree or submodule. No git process runs, so nothing walks up into an enclosing repository, no path passes through a subprocess's stdout, and a detached HEAD (no `ref:` line) says nothing. A missing checkout says nothing either; `init-workspace` clones it and `init-package` reports the missing template as before.

### Also

- **The notice that `start-cli` is behind the published release is given only when the checkout is on `live-docs`.** The gate lives inside `warn_if_start_cli_outdated`, so `CliContext::build_key()` (e.g. `s9pk pack`) is covered too; on any other branch the manifest names an unreleased version.
- **Unit tests** in the module cover reading a plain `.git/HEAD`, following a `gitdir:` file by absolute and by relative path, a detached HEAD, a missing checkout, and the printed command's quoting and UTF-8 gate. They need no git binary.
- Strings in all five locales; the start-cli changelog under `## [2.0.1]` with the matching bump in `Cargo.toml`, `Cargo.lock`, and the `.TH` header and VERSION section of `man/start-cli.1`; one sentence in the packaging guide's environment-setup page.

### DIFF
--- a/shared-libs/crates/start-core/src/s9pk/init.rs
+++ b/shared-libs/crates/start-core/src/s9pk/init.rs
@@ -150,6 +150,8 @@ pub async fn init_workspace(
             .capture(false)
             .invoke(ErrorKind::Git)
             .await?;
+    } else {
+        warn_if_guide_off_branch(&root);
     }
 
     // Symlink (not a copy) so a guide sync keeps the workspace AGENTS.md current.
@@ -216,6 +218,8 @@ pub async fn init_package(
             ErrorKind::InvalidRequest,
         )
     })?;
+    warn_if_start_cli_outdated(&root);
+    warn_if_guide_off_branch(&root);
 
     // Normalize to a candidate ID, then validate it through the manifest's own
     // rules rather than re-implementing them.
@@ -240,8 +244,6 @@ pub async fn init_package(
             ErrorKind::InvalidRequest,
         ));
     }
-
-    warn_if_start_cli_outdated(&root);
 
     let template = root.join(MONOREPO_DIR).join(TEMPLATE_SUBPATH);
     if !template.exists() {
@@ -296,6 +298,10 @@ pub async fn init_package(
 pub fn warn_if_start_cli_outdated(workspace: &Path) {
     static WARNED: std::sync::Once = std::sync::Once::new();
     WARNED.call_once(|| {
+        // Off `live-docs` the manifest names a version that has not shipped.
+        if guide_branch(workspace).as_deref() != Some(MONOREPO_BRANCH) {
+            return;
+        }
         let Ok(manifest) =
             std::fs::read_to_string(workspace.join(MONOREPO_DIR).join(CLI_MANIFEST_SUBPATH))
         else {
@@ -330,6 +336,61 @@ pub fn warn_if_start_cli_outdated(workspace: &Path) {
             );
         }
     });
+}
+
+fn guide_branch(workspace: &Path) -> Option<String> {
+    let dot_git = workspace.join(MONOREPO_DIR).join(".git");
+    let git_dir = match std::fs::read_to_string(&dot_git) {
+        Ok(gitfile) => dot_git
+            .parent()?
+            .join(gitfile.strip_prefix("gitdir:")?.trim()),
+        Err(_) => dot_git,
+    };
+    let head = std::fs::read_to_string(git_dir.join("HEAD")).ok()?;
+    Some(head.trim().strip_prefix("ref: refs/heads/")?.to_owned())
+}
+
+fn init_workspace_command(workspace: &Path) -> Option<String> {
+    Some(format!(
+        "start-cli s9pk init-workspace '{}'",
+        workspace.to_str()?.replace('\'', "'\\''")
+    ))
+}
+
+fn warn_if_guide_off_branch(workspace: &Path) {
+    let Some(branch) = guide_branch(workspace).filter(|branch| branch != MONOREPO_BRANCH) else {
+        return;
+    };
+    let docs = workspace.join(MONOREPO_DIR);
+    let next = match init_workspace_command(workspace) {
+        Some(command) => t!("s9pk.init.run-init-workspace", command = command).to_string(),
+        None => t!("s9pk.init.run-init-workspace-in-root").to_string(),
+    };
+    if std::fs::symlink_metadata(&docs).is_ok_and(|meta| meta.file_type().is_symlink()) {
+        let target = std::fs::canonicalize(&docs).unwrap_or_else(|_| docs.clone());
+        eprintln!(
+            "{}",
+            t!(
+                "s9pk.init.guide-linked-off-branch",
+                path = docs.display().to_string(),
+                target = target.display().to_string(),
+                branch = branch,
+                expected = MONOREPO_BRANCH,
+                next = next
+            )
+        );
+    } else {
+        eprintln!(
+            "{}",
+            t!(
+                "s9pk.init.guide-off-branch",
+                path = docs.display().to_string(),
+                branch = branch,
+                expected = MONOREPO_BRANCH,
+                next = next
+            )
+        );
+    }
 }
 
 /// Walk up from `start` (inclusive) for the nearest workspace — a directory whose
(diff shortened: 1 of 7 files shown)


helix-nine (changes requested):
Reviewed `git diff origin/master...HEAD` at `c346fcad` with the high-effort fan-out and independent verification of each candidate. The current CI checks are green, but these cases are not covered:

1. `projects/start-cli/CHANGELOG.md:12` — `start-cli/v2.0.0` already exists, so this must begin the prospective `2.0.1` release rather than add a bare `Unreleased` heading. Please bump the manifest and synchronize its generated artifacts; otherwise the release pre-check rejects this tree.

2. `shared-libs/crates/start-core/locales/i18n.yaml:173` — this remediation cannot work for workspaces created by start-cli 1.0.0/1.0.1. Those versions cloned `master` with `--depth 1`, which implied a single-branch refspec with no `origin/live-docs`; `git checkout live-docs` therefore fails. The remediation needs to fetch or otherwise make `live-docs` available before checkout.

3. `shared-libs/crates/start-core/locales/i18n.yaml:173` — `%{dir}` is always the caller-relative literal `start-technologies`, although `init-workspace` accepts an arbitrary workspace path and `init-package` can run from a nested directory. The printed command then fails or targets another checkout. It needs the discovered checkout path, safely quoted for shell use.

4. `shared-libs/crates/start-core/src/s9pk/init.rs:247` — emit this warning immediately after workspace discovery, before the package-name and destination checks. An older CLI's failed `npm install` leaves the scaffold directory behind; after the user upgrades, retrying exits at `package-exists` before reaching the new diagnosis.

5. `shared-libs/crates/start-core/src/s9pk/init.rs:154` — this also warns for the supported layout where `start-technologies` is a symlink to an owner-managed development checkout. The existing agent-context guidance explicitly says not to switch that repository's branch; detect the symlink and direct the user to remove it and rerun `init-workspace` instead.

6. `shared-libs/crates/start-core/src/s9pk/init.rs:344` — verify that `start-technologies` itself is the Git worktree root before reading its branch. Git's parent discovery means an existing non-repository/corrupt guide directory inside another repository reports the enclosing repository's branch, and the proposed checkout command can switch that unrelated repository.

I added the required `start-cli` and `StartSDK` labels. The code-comment pass was clean.

helix-nine (changes requested):
I re-reviewed the full `git diff origin/master...HEAD` and the fix delta from `c346fcad` at head `4cbd07767`. The original six findings are addressed, but this pass found three blocking issues:

1. `projects/start-cli/man/start-cli.1:123` — the generated header now says 2.0.1, but the explicit VERSION section still says `v2.0.0`. I ran `make manpages-check`; its four generator tests pass, then it exits 2 on exactly this missing `v2.0.0` → `v2.0.1` change. Please regenerate and commit the complete man page.

2. `shared-libs/crates/start-core/locales/i18n.yaml:180` — the new command makes `live-docs` fetchable for 1.0.0/1.0.1 workspaces, but those released versions also created a sparse, shallow checkout. Switching branches preserves the sparse patterns, so `projects/start-sdk/lib`, `projects/start-os`, `shared-libs`, and the CLI manifest remain absent after the warning disappears. Please include the existing migration steps for that population: disable sparse checkout and conditionally unshallow it (the 1.0.2 changelog already documents both).

3. `shared-libs/crates/start-core/src/s9pk/init.rs:358` — `--show-toplevel` proves only that `start-technologies` is a repository root, not that it is this monorepo. `init-workspace` accepts any existing normal repository at that path, provisions the workspace, and can now tell the user to alter that unrelated repository's `origin` refspec and branch. Please validate stable monorepo sentinel paths/content before provisioning or printing Git remediation; checking the origin URL would incorrectly reject supported forks.

Two nonblocking cleanup items:

- `shared-libs/crates/start-core/src/s9pk/init.rs:339-363` combines a filesystem path and branch into newline-delimited UTF-8, then lossily decodes it. Valid Unix paths containing a newline or non-UTF-8 byte silently suppress the warning. A root-relative identity probe plus a separate branch query avoids serializing the path.
- `shared-libs/crates/start-core/src/s9pk/init.rs:358` — reduce the explanatory `so` comment to the invariant, e.g. ``// `start-technologies` must be the Git worktree root.``

helix-nine (changes requested):
I re-reviewed the full `git diff origin/master...HEAD` and the amended delta from `4cbd07767` at head `7c2cd816`. The second-round findings are fixed, and `make manpages-check` now passes. Four blocking issues remain:

1. `shared-libs/crates/start-core/src/s9pk/init.rs:397-406` — the generated command still hardcodes `origin` and performs a plain `git checkout live-docs`. It fails when `clone.defaultRemoteName` gave the workspace another remote name, when a supported fork does not publish `live-docs`, or when `checkout.guess=false`. If a stale local `live-docs` already exists, checkout succeeds without fast-forwarding it, then future warnings are suppressed. Please resolve the fork-source policy and emit an explicit fetch/create-or-update/upstream sequence that converges on the intended published ref.

2. `shared-libs/crates/start-core/locales/i18n.yaml:172-177` — the symlink remediation is still caller-relative. From a package directory, “remove `start-technologies`” targets the wrong path and bare `init-workspace` is rejected; with `init-workspace /other/workspace`, the bare rerun targets the caller's directory. The branch migration also leaves v1.0.0/1.0.1's old root `AGENTS.md` link dangling unless initialization runs again. Please print quoted discovered link/workspace paths and finish repair with `start-cli s9pk init-workspace '<workspace>'`.

3. `shared-libs/crates/start-core/src/s9pk/init.rs:372-377` — requiring the current package-template path suppresses the warning for real older monorepo branches that predate that path. `init-workspace` reports success, while `init-package` then says to rerun `init-workspace`, which repeats the same no-op. Please use a durable identity check and emit a generic off-branch warning when Git mutation cannot be prescribed safely; the changelog and guide currently overstate coverage.

4. `shared-libs/crates/start-core/src/s9pk/init.rs:221-222` — `init-package` runs the outdated-CLI check against the working branch before diagnosing it as off `live-docs`. A supported development checkout or symlink at this PR therefore says released 2.0.0 is behind “published 2.0.1,” even though 2.0.1 is prospective. Gate the version warning on the guide actually being on `live-docs`.

Nonblocking hardening from the same paths:

- Parse `core.sparseCheckout` as a Git boolean, and avoid checking out `live-docs` before a later sparse-disable failure can leave an incomplete state that suppresses future repair.
- Require a `040000 tree` from `ls-tree`; `-d` also accepts a `160000` gitlink.
- `docs.display().to_string()` is lossy for valid non-UTF-8 Unix paths, so omit the executable command when the path cannot be represented losslessly.

The code-comment pass is clean.

helix-nine (changes requested):
I re-reviewed the full `git diff origin/master...HEAD` and the design-change delta from `7c2cd816` at head `1dcc3dcf`. The prior findings are addressed and `make manpages-check` passes, but automatic mutation of the persistent checkout introduces these blocking issues:

1. `shared-libs/crates/start-core/src/s9pk/init.rs:452-564` — refuse a dirty checkout before any mutation. `checkout -B` preserves non-conflicting tracked edits and untracked files, so this can report success and then scaffold packages from a locally modified template.

2. `shared-libs/crates/start-core/src/s9pk/init.rs:168-184` / `locales/i18n.yaml:186-191` — every Git failure advises removing the checkout, including dirty state, unpushed commits, corruption, and transient network errors. Following that advice can destroy user work. Tell the user to resolve the Git error or preserve/move the checkout before replacing it; do not recommend deletion generically.

3. `shared-libs/crates/start-core/src/s9pk/init.rs:474-509,564-570` — use fully qualified refs throughout. The fetch writes `refs/remotes/<remote>/live-docs`, but validation, ancestry, checkout, and upstream setup use DWIM shorthands. A colliding tag or branch named `live-docs` or `<remote>/live-docs` can validate/reset against the wrong object.

4. `shared-libs/crates/start-core/src/s9pk/init.rs:500-524` — unshallow before ancestry validation and distinguish `merge-base --is-ancestor` exit 1 from operational failure. `.is_err()` currently reports missing/corrupt objects and other Git failures as local divergence, discarding the real error.

5. `shared-libs/crates/start-core/src/s9pk/init.rs:526-545` — refspec coverage must validate the effective source-to-destination mapping and negative exclusions. A positive source wildcard can be accepted even when it maps elsewhere or `^refs/heads/live-docs` excludes it, leaving the configured upstream stale after the one-shot explicit fetch.

6. `shared-libs/crates/start-core/src/s9pk/init.rs:564-573` — branch replacement and upstream setup are separate operations. Failure or interruption after checkout leaves HEAD named `live-docs` with missing/wrong tracking; the next run skips repair because the branch name matches. Establish tracking with the checkout or validate/repair it on an already-`live-docs` branch.

7. `shared-libs/crates/start-core/src/s9pk/init.rs:385-573` — this persistent, roughly eleven-command Git migration has no committed regression tests. Please add real temporary-repository coverage for clean success/idempotence, dirty refusal, local commits, shallow+sparse migration, ref collisions, effective refspecs, and final tracking state. The existing module already has temp-directory and real-Git test patterns.

8. `shared-libs/crates/start-core/src/s9pk/init.rs:440-447` — the non-UTF-8 path fallback prints bare `start-cli s9pk init-workspace`. From `init-package` inside a package directory that command is rejected or targets another directory. Omit executable recovery guidance when the intended workspace path cannot be represented losslessly.

Two nonblocking documentation corrections:

- `projects/start-cli/CHANGELOG.md:23-24` says the outdated notice is only emitted on `live-docs`, but `CliContext::build_key()` still calls it unconditionally (for example from `s9pk pack`). Narrow this entry to `init-package`, or gate the shared caller too.
- `projects/start-sdk/docs/src/environment-setup.md:310` should say `init-workspace` moves a workspace-owned checkout; symlinked checkouts are only diagnosed.

Comments:

- `shared-libs/crates/start-core/src/s9pk/init.rs:385` — delete; the private helper's body and return type state these cases.
- `shared-libs/crates/start-core/src/s9pk/init.rs:450-451` — rewrite to the caller-visible result, e.g. `Moves the guide to MONOREPO_BRANCH without discarding local commits.`

Given the state space exposed here, a diagnosis-only design remains a valid smaller alternative to making `init-workspace` mutate the checkout.

helix-nine (approved):
Re-reviewed the full diff and the rewrite from `1dcc3dcf` at head `2cc13f0dd`. The diagnosis-only design resolves the prior findings: it does not mutate or prescribe Git state, preserves workspace-owned checkouts by telling the user to move them aside, handles symlinked checkouts separately, gates the outdated-client notice inside the shared helper, and gives non-UTF-8 workspace paths usable cwd-based guidance.

I also verified the direct and gitfile `HEAD` parsing, path quoting, version/manpage/changelog/docs synchronization, and all five locale placeholder sets. Detached HEAD remains outside the branch-specific notice's stated contract.

Checks:
- `git diff --check origin/master...HEAD`
- `make manpages-check`
- `CARGO_TARGET_DIR=/tmp/pr3910-target cargo test -p start-core s9pk::init --features=test` (8 passed)

No blocking findings.

shared-libs/crates/start-core/src/s9pk/init.rs
```diff
@@ -216,6 +218,8 @@ pub async fn init_package(
             ErrorKind::InvalidRequest,
         )
     })?;
+    warn_if_start_cli_outdated(&root);
+    warn_if_guide_off_branch(&root);
```
dr-bonez, line 222: these seem backwards. I imagine `warn_if_start_cli_outdated` bases that decision off of what's checked out?

dr-bonez: I guess it doesn't matter since it does not correct it

dr-bonez (approved):

PR #3911, opened 2026-09-07
fix(core): stop IPv6 stalls on a gateway with no IPv6 address of its own

MattDHill:
An OS update from the beta registry sat at 0 bytes for over four minutes before completing. The download itself then ran at 17 MB/s. Two independent defects combine to produce that stall, and this fixes both.

## The network state that reveals it

The dev box's LAN router advertises an IPv6 default route (router lifetime 1800s) with two on-link prefixes, both `A=0`, and `M=1 O=1` — SLAAC off, get your address from DHCPv6. Its DHCPv6 server then issues no lease. So `eno0` ends up with a default route, prefix routes, and only a link-local address. The box's one global IPv6 address belongs to a StartTunnel gateway on `wg0`.

That is not an exotic configuration. Any server on a LAN whose router advertises IPv6 without handing out an address lands in it, and so does any server behind an ISP whose IPv6 is silently broken.

## `gateway.rs` — the leak

`apply_policy_routing_v6` installs a `blackhole default` in an interface's policy table when that interface cannot carry v6, so that selecting a non-v6 gateway as default outbound drops v6 rather than leaking it. The condition was `has an IPv6 gateway OR has a global address`. NetworkManager reports a gateway for `eno0` straight from that RA, so `eno0` passed with no address to send from and got a real default route.

Routing was correct throughout — the priority-75 rule pointed at `eno0`'s table, and every SYN left on `eno0`. What leaked was the *source address*: a global destination needs a global-scope source, `eno0` had none, so the kernel fell back to the only global address on the host, `wg0`'s. Those packets left `eno0` carrying a StartTunnel address the LAN discards, and hung until the kernel gave up.

The check is now whether the interface has a non-link-local address of its own — global where there is no gateway, ULA accepted where there is, so a NAT66 LAN still qualifies. The gateway only decides whether the route needs `via`. On this box `eno0`'s table now holds the blackhole, every IPv6 connect fails instantly, and nothing can carry the `wg0` address out `eno0`.

## `socks.rs` — the two-minute wait per address

`startd` routes its outbound HTTP through its own SOCKS proxy (`socks5h`, so the proxy resolves). The CONNECT handler called `TcpStream::connect((domain, port))`, which tries the resolved addresses **sequentially** with no timeout of its own, moving on only once the previous attempt fails. The CDN host has two AAAA records and one A record, so the download did: AAAA #1, ~2 min SYN timeout; AAAA #2, ~2 min; A record, connected in 50 ms.

The proxy now resolves first and races the addresses RFC 8305 style: attempts start 250 ms apart alternating address families, a failure starts the next immediately, and the first to connect wins while the rest are dropped. There is deliberately no cap on attempts in flight: with two dead addresses ahead of a live one, a cap holds the live one back until a kernel timeout, which is the stall being removed. Sockets held while nothing answers are bounded by the RRset. IPv6 stays preferred where it works. The raw-IP, `.local` and `.onion` paths are unchanged, and no overall deadline is introduced — a host that is simply slow still gets as long as it did before.

The blackhole fix alone would have covered this box. This one covers what no local guard can detect: a v6 route that is valid on the server but dead further along the path.

## Also here: the gateway watcher's `ip route` usage spam — and the end of the route mirror

`apply_policy_routing` / `apply_policy_routing_v6` used to mirror the main table into each gateway's policy table by feeding every line of `ip route show` back into `ip route replace`, dropping only `linkdown` and `dead`. Show output isn't replace input: a multipath route prints as a header plus indented `nexthop …` continuation lines, so `ip -6 route replace nexthop via … table N` ran on every `watch_ip` pass for every gateway and iproute2 printed its full usage. On a Server One whose LAN router advertises a two-next-hop IPv6 default, that was 140 usage dumps — 40% of a 10,000-line OS log export.

Five review rounds of a text parser found five more ways that text fails to round-trip (lifetimes, errno, display flags, nexthop-id expansions, encap payloads carrying their own `tos`, multi-line option output, interface names that collide with keywords), so the mirror is gone. **A per-gateway table now holds only its default route, or the v6 leak-guard blackhole.** Specific routes come from `main` directly: a `lookup main suppress_prefixlength 0` rule ahead of the table rules consults `main` for every route except a default, which is exactly the set the mirror copied (its loops skipped `default`). wg-quick relies on the same rule for full-tunnel mode.

Verified in a container with iproute2 6.9, tables holding only their default or blackhole:

```
192.168.8.50   -> dev lan                      on-link, from main
10.0.3.7       -> dev lxcbr0                   container subnet, from main
10.8.0.9       -> dev wg0                      VPN-peer route (an encap route the parser could never replay), from main
8.8.8.8        -> dev wg0 table 1004           the gateway's default
fd00:3::9      -> dev lan                      ULA on-link, from main
2606:4700::1   -> RTNETLINK answers: Invalid argument   v6 leak guard (blackhole) holds
mark 0x3ec -> 192.168.8.50 -> dev lan          reply mark, on-link still wins
mark 0x3ec -> 8.8.8.8      -> dev wg0 table 1004
```

Rule order is now `49` transparent divert, `50` `lookup main suppress_prefixlength 0`, `51` per-interface reply mark, `60` IPv6 source, `74` WireGuard encap mark, `75` selected outbound. The reply-mark rules moved from 50 to 51 so a marked reply consults `main` first; one found at any other priority is deleted, so a hot restart of `startd` converges too.

Two semantics differ from the mirror and are accepted: a `throw` in `main` falls through to the outbound table's default, and a reject-type default in `main` ends the lookup before it can be suppressed. StartOS installs neither into `main`. Tables survive a hot restart of `startd` but not the reboot every OS update performs, so no cleanup of previously mirrored entries is carried; on a dev box hot-replaced via `make start-os-update-startbox`, entries mirrored by the old build linger in the tables until reboot and can only matter for a destination whose route has since left `main`.

The IPv6 `RouteData` subscription existed to reconcile the mirror and goes with it.

## Verification

Four unit tests for `carries_v6` covering the router-without-address, ULA-with-router, global-without-router, and ULA-without-router cases; four for the SOCKS path — family interleaving, and three on paused Tokio time with an injected connector that assert exactly when each attempt starts: the second one delay after a first that never resolves, the third on the cadence past two that never resolve, and the next immediately after an earlier failure. The rule logic was exercised against a command shim (adds both rules when absent, leaves present ones alone, moves a stale reply-mark rule, and does not mistake `lookup 10020` for `lookup 1002`), and the routing semantics in the container above. The `start-core` suite is heavy to build locally, so these were exercised in scratch crates against the same dependency versions and left for CI to run in place.

### DIFF
--- a/projects/start-os/CHANGELOG.md
+++ b/projects/start-os/CHANGELOG.md
@@ -461,6 +461,15 @@ file tracks notable changes since the move to the monorepo.
   you also reach on your local network keeps answering there with your server's
   own certificate.
 
+- **Marketplace, service and OS-update downloads no longer stall for minutes on
+  a server whose IPv6 cannot reach the Internet.** StartOS connects over
+  whichever of a host's IPv6 or IPv4 addresses answers first.
+
+- **An outbound gateway that has an IPv6 router but no IPv6 address of its own
+  carries no IPv6.** Such traffic fails at once and falls back to IPv4, instead
+  of leaving with an address that belongs to another of the server's
+  interfaces.
+
 ## [0.4.0.1]
 
 ### Changed
--- a/projects/start-os/CHANGELOG.md
+++ b/projects/start-os/CHANGELOG.md
@@ -441,14 +441,13 @@ file tracks notable changes since the move to the monorepo.
   flavors produced no answer, so a list mixing Bitcoin Core with Bitcoin Knots
   kept whatever order it arrived in.
 
-- **Marketplace, service and OS-update downloads proceed over IPv4 at once when
-  a host's IPv6 is unreachable.** StartOS connects over whichever of a host's
-  IPv6 or IPv4 addresses answers first.
+- **Marketplace, service-package and OS-update downloads fall back to IPv4
+  within a fraction of a second when a host's IPv6 does not answer.** StartOS
+  tries a host's IPv6 and IPv4 addresses a quarter of a second apart and uses
+  the first connection that succeeds.
 
-- **Routes with more than one next hop, a lifetime, or an `unreachable` or
-  `prohibit` type are carried into each gateway's routing table.** The OS log
-  stays clear of `ip route` usage text on a network whose router advertises
-  such routes.
+- **The OS log stays free of `ip route` usage text on a network whose router
+  advertises a route with more than one next hop.**
 
 ### Security
 
@@ -472,7 +471,8 @@ file tracks notable changes since the move to the monorepo.
 
 - **Outbound IPv6 leaves only through an address of the selected gateway's
   own.** A gateway with an IPv6 router but no IPv6 address of its own carries
-  no IPv6; such traffic fails at once and falls back to IPv4.
+  no IPv6; such traffic is refused on the spot rather than sent with another
+  interface's address.
 
 ## [0.4.0.1]
 
--- a/projects/start-os/CHANGELOG.md
+++ b/projects/start-os/CHANGELOG.md
@@ -470,6 +470,11 @@ file tracks notable changes since the move to the monorepo.
   of leaving with an address that belongs to another of the server's
   interfaces.
 
+- **The gateway watcher no longer logs `ip route`'s usage text every few
+  minutes** on a network whose router advertises a route with more than one
+  next hop. Such routes, routes with a lifetime, and `unreachable` or
+  `prohibit` routes are now carried into each gateway's routing table too.
+
 ## [0.4.0.1]
 
 ### Changed
--- a/projects/start-os/CHANGELOG.md
+++ b/projects/start-os/CHANGELOG.md
@@ -446,7 +446,7 @@ file tracks notable changes since the move to the monorepo.
   tries a host's IPv6 and IPv4 addresses a quarter of a second apart and uses
   the first connection that succeeds.
 
-- **The OS log stays free of `ip route` usage text on a network whose router
+- **The OS log stays focused on actionable errors on a network whose router
   advertises a route with more than one next hop.**
 
 ### Security
@@ -469,10 +469,9 @@ file tracks notable changes since the move to the monorepo.
   you also reach on your local network keeps answering there with your server's
   own certificate.
 
-- **Outbound IPv6 leaves only through an address of the selected gateway's
-  own.** A gateway with an IPv6 router but no IPv6 address of its own carries
-  no IPv6; such traffic is refused on the spot rather than sent with another
-  interface's address.
+- **Outbound IPv6 uses an address assigned to the selected gateway.** Traffic
+  through a gateway that has an IPv6 router but no IPv6 address of its own
+  fails immediately.
 
 ## [0.4.0.1]
 
--- a/projects/start-os/CHANGELOG.md
+++ b/projects/start-os/CHANGELOG.md
@@ -441,6 +441,15 @@ file tracks notable changes since the move to the monorepo.
   flavors produced no answer, so a list mixing Bitcoin Core with Bitcoin Knots
   kept whatever order it arrived in.
 
+- **Marketplace, service and OS-update downloads proceed over IPv4 at once when
+  a host's IPv6 is unreachable.** StartOS connects over whichever of a host's
+  IPv6 or IPv4 addresses answers first.
+
+- **Routes with more than one next hop, a lifetime, or an `unreachable` or
+  `prohibit` type are carried into each gateway's routing table.** The OS log
+  stays clear of `ip route` usage text on a network whose router advertises
+  such routes.
+
 ### Security
 
 - **Service mount paths are validated and confined to their intended
@@ -461,19 +470,9 @@ file tracks notable changes since the move to the monorepo.
   you also reach on your local network keeps answering there with your server's
   own certificate.
 
-- **Marketplace, service and OS-update downloads no longer stall for minutes on
-  a server whose IPv6 cannot reach the Internet.** StartOS connects over
-  whichever of a host's IPv6 or IPv4 addresses answers first.
-
-- **An outbound gateway that has an IPv6 router but no IPv6 address of its own
-  carries no IPv6.** Such traffic fails at once and falls back to IPv4, instead
-  of leaving with an address that belongs to another of the server's
-  interfaces.
-
-- **The gateway watcher no longer logs `ip route`'s usage text every few
-  minutes** on a network whose router advertises a route with more than one
-  next hop. Such routes, routes with a lifetime, and `unreachable` or
-  `prohibit` routes are now carried into each gateway's routing table too.
+- **Outbound IPv6 leaves only through an address of the selected gateway's
+  own.** A gateway with an IPv6 router but no IPv6 address of its own carries
+  no IPv6; such traffic fails at once and falls back to IPv4.
 
 ## [0.4.0.1]
 
--- a/projects/start-os/docs/src/outbound-vpn.md
+++ b/projects/start-os/docs/src/outbound-vpn.md
@@ -30,7 +30,7 @@ StartOS treats IPv6 outbound routing the same way as IPv4: the default gateway i
 
 If the gateway you select for outbound traffic can't carry IPv6 — for example a commercial VPN whose WireGuard config has no IPv6 address, or a LAN whose router advertises IPv6 without assigning your server an address — StartOS **drops** the server's outbound IPv6 rather than letting it fall back to your ISP connection, so your real IPv6 address never leaks around the VPN. (The drop is a blackhole in that gateway's own routing table.) A gateway that does provide IPv6 (such as a StartTunnel with a [delegated prefix](/start-tunnel/ipv6.html)) carries IPv6 normally; on a server with no native ISP IPv6, such a tunnel can also become your IPv6 default before you pin it, so select an outbound gateway explicitly if you want to control which one.
 
-When an Internet host offers both address families, StartOS tries IPv6 and IPv4 close together and uses the first connection that succeeds.
+For its own connections — marketplace, service-package and OS-update downloads — StartOS tries a host's IPv6 and IPv4 addresses a quarter of a second apart and uses the first connection that succeeds. Services make their own connections and follow their own behavior.
 
 ## Route Individual Services Through VPN
 
--- a/projects/start-os/docs/src/outbound-vpn.md
+++ b/projects/start-os/docs/src/outbound-vpn.md
@@ -28,7 +28,9 @@ By default, StartOS dynamically selects which gateway to use for outbound traffi
 
 StartOS treats IPv6 outbound routing the same way as IPv4: the default gateway is chosen by route metric, and you can set the system-wide default under `System > Gateways > Outbound Traffic`.
 
-If the gateway you select for outbound traffic can't carry IPv6 — for example a commercial VPN whose WireGuard config has no IPv6 address — StartOS **drops** the server's outbound IPv6 rather than letting it fall back to your ISP connection, so your real IPv6 address never leaks around the VPN. (The drop is a blackhole in that gateway's own routing table.) A gateway that does provide IPv6 (such as a StartTunnel with a [delegated prefix](/start-tunnel/ipv6.html)) carries IPv6 normally; on a server with no native ISP IPv6, such a tunnel can also become your IPv6 default before you pin it, so select an outbound gateway explicitly if you want to control which one.
+If the gateway you select for outbound traffic can't carry IPv6 — for example a commercial VPN whose WireGuard config has no IPv6 address, or a LAN whose router advertises IPv6 without assigning your server an address — StartOS **drops** the server's outbound IPv6 rather than letting it fall back to your ISP connection, so your real IPv6 address never leaks around the VPN. (The drop is a blackhole in that gateway's own routing table.) A gateway that does provide IPv6 (such as a StartTunnel with a [delegated prefix](/start-tunnel/ipv6.html)) carries IPv6 normally; on a server with no native ISP IPv6, such a tunnel can also become your IPv6 default before you pin it, so select an outbound gateway explicitly if you want to control which one.
+
+When an Internet host offers both address families, StartOS tries IPv6 and IPv4 close together and uses the first connection that succeeds.
 
 ## Route Individual Services Through VPN
 
--- a/shared-libs/crates/start-core/src/net/gateway.rs
+++ b/shared-libs/crates/start-core/src/net/gateway.rs
@@ -2138,6 +2138,17 @@ async fn apply_policy_routing(
 /// reroute, but the reply that opens a connection is routed before the output
 /// hook runs, so the source rule is what lets a v6 service reached through a
 /// tunnel answer.
+/// Sending needs an address of the interface's own beyond link-local; without
+/// a gateway it must be global.
+fn carries_v6(gateway: Option<Ipv6Addr>, addrs: impl IntoIterator<Item = Ipv6Addr>) -> bool {
+    let (mut own, mut global) = (false, false);
+    for addr in addrs {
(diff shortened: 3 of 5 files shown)


helix-nine (changes requested):
Reviewed `git diff origin/master...HEAD` (`f7dcde71`…`1fab5e89`) with the high-effort fan-out and independent per-location verification.

### Findings

1. `shared-libs/crates/start-core/src/net/gateway.rs:1927-1929,2018-2023,2279-2288` — **[CONFIRMED] Finite IPv6 routes can expire in policy tables while remaining renewed in `main`.**
   `expires` is copied into each policy table, but lifetime-only RA renewal does not produce a NetworkManager route-data change and the only guaranteed reconciliation is 600 seconds later. A route with a shorter lifetime can disappear from the selected policy table and send marked/source-rule traffic through its fallback default or blackhole. Please avoid replaying a decaying lifetime unless reconciliation is guaranteed before it expires.

2. `shared-libs/crates/start-core/src/net/socks.rs:43-60,105-112` — **[CONFIRMED] One SOCKS request can retain an unbounded number of live connection attempts.**
   Another connect starts every 250 ms while all earlier attempts remain in `FuturesUnordered`; the full resolver result is accepted without a cap. Large nonresponsive RRsets multiplied across concurrent requests retain sockets until kernel timeout. Please cap each request at two live attempts while continuing the staggered fallback.

3. `shared-libs/crates/start-core/src/net/gateway.rs:1971-1982,2008-2028,2045-2050` — **[CONFIRMED] The route parser removes valid values and omits an emitted route type.**
   Filtering is context-free, so valid interfaces named `trap`, `error`, or `expires` are removed after `dev`, producing malformed replay commands. `ROUTE_TYPES` also omits accepted/emitted `xresolve`, so stale cleanup treats the type as the destination and fails to delete that route. Please make cleanup position-aware, recognize `xresolve`, and add collision tests.

4. `shared-libs/crates/start-core/src/net/socks.rs:234-250` — **[CONFIRMED] The fallback test need not exercise the timer path.**
   `192.0.2.1` may reject immediately, and the `< 5s` assertion then passes through immediate-error fallback even if the 250 ms stagger is broken. Please use a deterministically pending first attempt with paused Tokio time and assert when the second attempt starts.

5. `projects/start-os/CHANGELOG.md:464-476` — **[CONFIRMED] The release notes misclassify operational fixes and violate the changelog rules.**
   Move the download and log entries under **Fixed**, leaving the cross-interface address leak under **Security**, and replace the negative/internal wording with the user-visible outcomes, for example:
   - `Internet downloads promptly continue over IPv4 when IPv6 is unreachable.`
   - `OS logs stay focused on actionable errors when gateways advertise complex routes.`
   - `Outbound IPv6 traffic uses an address assigned to its selected gateway.`

6. `projects/start-os/docs/src/outbound-vpn.md:27-31` — **[CONFIRMED] The new dual-stack connection behavior is absent from the StartOS book.**
   The existing text covers leak prevention, but not staggered IPv6/IPv4 fallback. Root `AGENTS.md` requires the product book to change with user-visible behavior. A concise addition is: `When an Internet host offers both address families, StartOS tries IPv6 and IPv4 close together and uses the first connection that succeeds.`

### Comments

- `gateway.rs:1998,2001,2003,2005` — delete the private `ShownRoute`/field docs; the names and body already say this.
- `gateway.rs:2014` — `// \`error\` is output-only; the route type determines errno.`
- `gateway.rs:2060` — `// \`table\` must precede a multipath \`nexthop\` list.`
- `gateway.rs:2092-2094` — delete the now-false first-token destination comment; typed routes put their type first.
- `gateway.rs:2206-2225` — the old `apply_policy_routing_v6` rustdoc now attaches to `carries_v6`. Give the predicate its own short contract and move/shorten the policy-routing contract back above `apply_policy_routing_v6`.
- `socks.rs:40-41` — the fixed-delay claim is false after an early failure. Suggested wording: `Connects to the first address that answers, alternating families.` / `Starts another attempt after CONNECT_ATTEMPT_DELAY or an earlier failure.`

Please also remove the generated-with-Claude footer from the PR body.

Verification run on this head: `cargo check -p start-core`; the 4 `carries_v6_tests`, 7 `route_show_tests`, and 5 `net::socks::test` tests; `make start-core-format-check`; and `git diff --check`. All passed. The first test invocation required `PLATFORM=x86_64 make metadata`; no tracked files changed.

helix-nine (changes requested):
Reviewed the follow-up as `git diff 1fab5e89..fce5ee555`, then rechecked the integrated PR with `git diff origin/master...fce5ee555`.

The previous findings are mechanically addressed, but the fix pass exposed additional correctness problems. Two are consequences of remedies I requested; my earlier requests were incomplete there.

### Findings

1. `shared-libs/crates/start-core/src/net/socks.rs:53-74,282-310` — **[CONFIRMED] Two pending attempts prevent progress to every later address.**
   With `[dead IPv6, dead IPv4, reachable later address]`, both slots remain occupied until a kernel timeout, so the reachable address is never started on the 250 ms cadence. The new third-attempt test codifies waiting for a failure and misses `[pending, pending, reachable]`. My request for a two-attempt cap was incomplete: please add bounded rotation/per-attempt expiry and a paused-time liveness test, or remove the cap.

2. `shared-libs/crates/start-core/src/net/gateway.rs:2000-2094` — **[CONFIRMED] Compatibility-expanded `nhid` routes are replayed in a form the kernel rejects.**
   Real output can be `PREFIX nhid ID via ADDRESS dev IFACE ...`; replay preserves both the nexthop ID and its expanded gateway/OIF. Linux rejects that combination as mutually exclusive, so an ordinary non-default RA route is omitted from every policy table on each pass. Retain `nhid` while stripping its compatibility expansion, and test the actual shown form.

3. `shared-libs/crates/start-core/src/net/gateway.rs:1842-1864,1918-1929,2037-2046,2288-2369` — **[CONFIRMED] Lifetime-free IPv6 mirrors are not promptly reconciled when the route set changes.**
   Dropping `expires` correctly fixes lifetime-only renewal, but this watcher has no IPv6 `RouteData` subscription. A withdrawn/changed route can therefore remain live for up to 600 seconds. Since every gateway table mirrors `main`, waking only the interface that emitted a route signal would still leave the other tables stale. My earlier lifetime finding did not account for this second half; please use a shared route-set wakeup that reconciles all active tables while continuing to omit sampled expiry values.

4. `shared-libs/crates/start-core/src/net/gateway.rs:2117-2187,2290-2365` — **[CONFIRMED] A failed `main` snapshot is treated as an authoritative empty route set.**
   `desired_prefixes` remains empty when the first `ip route show` fails, but cleanup still runs if the policy-table read succeeds and deletes every non-default mirrored route. Gate cleanup on a successful source snapshot.

5. `shared-libs/crates/start-core/src/net/gateway.rs:2117-2187,2290-2365` — **[CONFIRMED] Prefix-only desired state retains stale FIB aliases.**
   Routes with the same destination and different metrics coexist. Replaying the desired alias leaves another alias intact, and prefix membership exempts it from deletion; a stale lower-metric route remains selected. Compare normalized route identities and delete stale aliases with distinguishing selectors.

6. `shared-libs/crates/start-core/src/net/gateway.rs:1999-2054` — **[CONFIRMED] Positional cleanup still omits replayable nested valued keywords.**
   `link_dev`, and seg6local `iif`/`oif`, can carry interface values named `trap`, `error`, or `expires`; those values are still stripped. Add these forms with collision tests or parse the nested grammar explicitly. BPF `endpoint` should be handled separately because its shown text may not itself round-trip.

7. `projects/start-os/docs/src/outbound-vpn.md:33` and `projects/start-os/CHANGELOG.md:444-451,473-475` — **[CONFIRMED] The user-facing copy overstates the implementation.**
   Happy Eyeballs applies to StartOS's internal SOCKS/reqwest traffic, not arbitrary package applications. Scope it to marketplace, service-package, and OS-update downloads. The changelog's “at once”/universal IPv4 fallback claims likewise exceed the 250 ms stagger and limited client scope, while the route entry is implementation-heavy and implies the lifetime is carried even though it is deliberately removed. State the scoped outcomes instead.

### Comments / cleanup

- `gateway.rs:1999` — delete `Keywords followed by their value`; it restates `VALUED_KEYWORDS`.
- `socks.rs:235-245,366-374` — generalize/reuse `echo_round_trip` instead of keeping the identical exchange open-coded.
- Keep the `pervasive` handling/comment: the actual iproute parser rejects it despite the contradictory help grammar.

The generated-with-Claude footer also remains in the PR body; please remove it as previously requested.

Verification on `fce5ee555`: `cargo check -p start-core`, 4 `carries_v6_tests`, 9 `route_show_tests`, 6 `net::socks::test` tests, `make start-core-format-check`, and both scoped `git diff --check` commands all pass. The missing cases above are not covered by those tests.

helix-nine (changes requested):
Reviewed `git diff fce5ee555..8e0041f1c` and rechecked the integrated PR. The SOCKS liveness fix, failed-snapshot guard, scoped docs, nested keyword handling, cleanup, and footer removal are all addressed. I also accept the cross-table route wakeup as adjacent work: this PR preserves the pre-existing worst-case 600-second bound while improving the owning-interface path, although short-lived sibling copies remain a follow-up concern.

Three route-reconciliation blockers remain:

1. `shared-libs/crates/start-core/src/net/gateway.rs:2036-2106` — **[CONFIRMED] `nhid` normalization is still lexical and incomplete.**
   - A supported IPv4 route through an IPv6 nexthop can print `via inet6 ADDRESS`; the stripper removes `via inet6` but leaves the address as an invalid bare token.
   - A nexthop object with `encap mpls ...` retains that expansion, and the kernel rejects `nhid` together with `RTA_ENCAP`.
   - A valid interface named `nhid` (`dev nhid`) is mistaken for the keyword and loses its device.

   Detect an actual `nhid ID` field positionally, strip its complete compatibility expansion, and add all three fixtures. The present simple/group tests do not cover these forms.

2. `shared-libs/crates/start-core/src/net/gateway.rs:2040-2050,2112-2147,2178-2241,2349-2417` — **[CONFIRMED] `(prefix, metric)` is not a complete route identity or deletion selector.**
   IPv4 routes can alias on `tos`, and IPv6 routes on `from`. Routes sharing prefix/metric but differing by those selectors collapse into one desired identity; on a metric transition the new deletion can fail to match the stale alias or target the wrong one. The metric fix is directionally correct, but please carry `tos`/`from` through identity and deletion with family-specific tests. Route type is not required for deletion.

3. `shared-libs/crates/start-core/src/net/gateway.rs:2175-2241,2338-2417` — **[CONFIRMED] A failed replacement can delete the working predecessor.**
   A desired identity is recorded even when `ip route replace` fails. If a route changes metric from 100 to 200, replacement of 200 can fail and cleanup then deletes the still-working metric-100 route, leaving no specific route until the next pass. If any non-default replay fails, skip stale cleanup for that family/pass. Recording only successful identities is insufficient for a failed same-identity replacement.

Cleanup:
- `gateway.rs:2036` should say `Attributes expanded from nhid cannot be replayed with the id`; the current comment mentions only the gateway while the implementation must remove every expanded nexthop attribute.
- `gateway.rs:2166` promises uninterrupted connectivity, which finding 3 currently disproves; tighten it after the failure path is fixed.
- `projects/start-os/CHANGELOG.md:449-450,472-475` still uses negative/internal framing. Positive outcome wording would be: `OS logs stay focused on actionable network errors when a router advertises routes with multiple next hops.` and `Outbound IPv6 uses an address assigned to the selected gateway. Traffic through a gateway without a usable IPv6 address fails immediately.`

Verification on `8e0041f1c`: `cargo check -p start-core`, 4 `carries_v6_tests`, 12 `route_show_tests`, 7 `net::socks::test` tests, `make start-core-format-check`, and both scoped `git diff --check` commands pass. The remaining cases are absent from the test set.

helix-nine (changes requested):
Reviewed `git diff origin/master...16ddaa111` with the high-effort seven-angle pass and independent verification.

I pushed `7314d6df8` to this branch for two findings I could close directly:

- failed replacements now protect only stale routes in the same destination/`tos`/`from` domain, so a metric change keeps its working predecessor without one poison route suppressing all unrelated cleanup;
- iproute2's displayed `ttl-propogate` spelling is normalized to the accepted `ttl-propagate` replay spelling, with coverage for both values.

One blocker remains:

`shared-libs/crates/start-core/src/net/gateway.rs:2092` — **[CONFIRMED] the `nhid` encap scan discards a route-level `tos`.** iproute2 prints `nhid`, the expanded encap payload, then the route's top-level `tos`, followed by `via`/`dev`. Scanning until `via`/`dev` therefore turns a TOS-specific route into a TOS-zero replay, which can capture traffic intended for another route. A whitespace stop word is insufficient because an IP encap can itself contain `tos`; this needs type-aware parsing or structured route metadata. Please add the actual print-order case (`nhid … encap … tos … via/dev`) to the parser tests.

The cross-table wakeup remains nonblocking: it preserves the existing nominal 600-second worst-case reconciliation bound while improving the owning interface. The comment pass was clean after `7314d6df8`.

Validated at `7314d6df8` with `cargo check -p start-core`, 14 route-show tests, 4 `carries_v6_tests`, 7 SOCKS tests, `make start-core-format-check`, and both scoped `git diff --check` commands.

helix-nine (changes requested):
Reviewed `git diff 7314d6df8...6cd4093ea` with the high-effort seven-angle pass, then checked the integrated PR against `origin/master`. The measured IP-encap rule is sound: Linux dumps one payload `tos` even when zero, and a later `tos` is the route selector. Two related textual-parser cases still break it:

1. `shared-libs/crates/start-core/src/net/gateway.rs:2149-2179` — **[CONFIRMED] physical continuation lines split one route into multiple records.** iproute2 prints Geneve, VXLAN, and ERSPAN options on an indented line before the route-level `tos`/`via`/`dev`. In an `nhid` compatibility expansion, the first line becomes a replay without the route selector—broadening a TOS-specific route to all traffic for that prefix—and the option line becomes a bogus second route. `ip -oneline` on all four route-show invocations, or equivalent logical-line joining that preserves real `nexthop` continuations, closes this case.

2. `shared-libs/crates/start-core/src/net/gateway.rs:2107-2120` — **[CONFIRMED] opaque encap values are still interpreted as route keywords.** SEG6LOCAL reproduces this end to end when an `iif`/`oif` interface is legally named `tos`, `via`, or `dev`; XFRM `link_dev` has the same source-confirmed form. The scan then invents a TOS selector or terminates the expansion early and emits malformed replay arguments. More stop words cannot distinguish payload values from top-level route syntax; this part needs encap-type-aware consumption or structured metadata.

Comments: delete the version-anchored fixture comment near `gateway.rs:3889`; the test name and cases carry that information.

Validation on `6cd4093ea`: `cargo check -p start-core`, all 14 route-show tests, `make start-core-format-check`, and both scoped `git diff --check` commands pass. The two actual-output forms above are absent from the fixtures.

helix-nine (approved):
Reviewed `git diff 6cd4093ea...74a7443a1` with the high-effort seven-angle pass and checked the integrated PR. The native `main` lookup removes the unbounded text-round-trip problem, priority 50/51 ordering preserves transparent diversion and marked replies, and the accepted `throw`/reject-default differences are explicit.

I pushed `3c1e04ddc` to close the one normal-boot race found in the pass: the global suppress rule is now reconciled once by the coordinator before concurrent interface watchers start, rather than check-then-added by every interface. Detection also requires the canonical unrestricted rule, and the stale scope-exit comment is removed. Added coverage rejects source- and fwmark-restricted lookalikes.

The no-cleanup hot-binary-replacement case is accepted as the documented development-only transition; production OS updates reboot and start with empty kernel tables. Legacy priority-50 rules on absent interfaces fall under the same accepted boundary.

Validated with `cargo test -p start-core policy_rule_tests --features=test`, `cargo check -p start-core`, `cargo fmt --all`, and `git diff --check`. CI is running on `3c1e04ddc`.

dr-bonez (approved):

PR #3915, opened 2026-09-08
fix(sdk): add MariaDB 11 dump backup support

### LINKED ISSUE #3766: Backups.withMysqlDump is inoperative against MariaDB >= 11 images: the mysql* binaries no longer exist
## Summary

`sdk.Backups.withMysqlDump({ engine: 'mariadb' })` cannot back up or restore a package that uses a current MariaDB image. It shells out to five `mysql*`-named binaries, and MariaDB 11.x ships none of them. The backup fails, and — this is the bad part — **it fails as a skipped package rather than a hard error**: the run reports success and the package is silently absent from the archive.

## Evidence

Measured inside the running `mariadb:11.8.8` container of a package on StartOS 0.4.0:

```
mysqld               MISSING          mariadbd             /usr/sbin/mariadbd
mysqladmin           MISSING          mariadb-admin        /usr/bin/mariadb-admin
mysqldump            MISSING          mariadb-dump         /usr/bin/mariadb-dump
mysql                MISSING          mariadb              /usr/bin/mariadb
mysql_install_db     MISSING          mariadb-install-db   /usr/bin/mariadb-install-db
```

MariaDB dropped the `mysql*` compatibility symlinks; only the `mariadb*` names remain.

`shared-libs`-side, `lib/backup/Backups.js` hardcodes the missing names in both directions:

- `startMysql()` → `mysqld --user=mysql --datadir=... --bind-address=127.0.0.1` (fire-and-forget, its rejection goes to `.catch(console.error)`)
- default `readyCmd` → `mysqladmin ping -u <user> -p<pw> --silent`
- dump → `mysqldump ...`
- `setPostRestore` → `mysql_install_db`, then `mysqladmin ping`, then `mysql -u root -e <grants>`

Because `startMysql` never awaits, the missing binary surfaces only as the ready loop timing out:

```
Error: MySQL/MariaDB failed to become ready within 30 seconds
```

and the resulting archive contains `"packageBackups": {}` while `backup create` exits 0.

## Impact

Any package pinning MariaDB 11.x and following the [Run a MySQL/MariaDB Sidecar](https://docs.start9.com/packaging/recipe-mysql.html) recipe gets a backup that appears to succeed and contains nothing. For an accounting or bookkeeping package that is the worst possible failure mode — the user believes their ledger is backed up.

Found while packaging ERPNext (Start9Labs/erpnext-startos#1), which pins `mariadb:11.8.8` because that is what upstream `frappe_docker` uses.

## Suggested fix

1. Resolve the binary name at runtime rather than hardcoding — prefer `mariadbd`/`mariadb-admin`/`mariadb-dump`/`mariadb`/`mariadb-install-db` when `engine === 'mariadb'`, falling back to the `mysql*` names for older images (or probe with `command -v`).
2. **`await`/surface the server-start failure.** A missing binary should fail loudly and immediately, not as a 30-second readiness timeout.
3. Make a failed package backup fail the run, or at minimum report the package as failed — silently omitting it from `packageBackups` while exiting 0 is the dangerous part, independent of this bug.

Happy to open a PR for (1) and (2) if that is useful.

MattDHill:
Closes #3766.

## What

`Backups.withMariadbDump(...)` now owns MariaDB logical backup and restore through a dedicated `MariadbDumpConfig`. It runs MariaDB's native `mariadbd`, `mariadb-admin`, `mariadb-dump`, `mariadb`, and `mariadb-install-db` tools, tracks the server process, and stores a compressed dump.

`Backups.withMysqlDump(...)` remains a separate MySQL implementation using `mysqld --daemonize`, `mysqladmin`, `mysqldump`, `mysql`, and `mysqld --initialize-insecure`. The two builders keep independent startup, readiness, shutdown, initialization, option, and dump-format logic. They share only the SQL escaping/grant and parameter-bound import primitives that behave identically in both clients.

The 3.0 migration is explicit:

- MySQL callers keep `withMysqlDump`; `engine: 'mysql'` is accepted but optional for source compatibility.
- MariaDB callers switch to `withMariadbDump`, remove `engine: 'mariadb'`, and rename `mysqldOptions` to `mariadbdOptions`.

The retained MySQL discriminator rejects a legacy MariaDB config passed to `withMysqlDump`. MariaDB configs reject the stale MySQL option name instead of dropping those startup options.

The `backup` package namespace also exposes `backup.Backups` and `backup.mountBackupTarget`.

## MariaDB behavior

- Backup dumps with `mariadb-dump` as the configured user, compresses the dump, and copies it to the backup target.
- Restore clears the data directory, initializes it with `mariadb-install-db`, creates the database and accounts, and replays the dump.
- Server exit is tracked around readiness probes; shutdown uses bounded SIGTERM and SIGKILL waits.
- SQL literals, identifiers, and grant patterns are escaped; database operands are bound with `--database`; dump/import client authentication uses `MYSQL_PWD` rather than client `-p` arguments.

## Scope

This PR fixes #3766 and keeps each database engine's lifecycle local to its own builder. Broader lifecycle hardening remains collected in #3917; #3918 proposes moving dump/import work under the service's supervised daemon lifecycle.

## Fleet

The current consumers are Ghost on MySQL and RomM/Dojo on MariaDB. Their 2.0.9 source remains valid for the released SDK and will use the migration above when adopting SDK 3.0.

## Verification

- `make check`
- `make test` — 122 tests
- `make check-fmt`
- `make bundle`
- Five-book docs build
- External public-surface typecheck for both builders and both 2.x migration guards
- Docker round trip against MariaDB 11.8.9
- MySQL 8.4.10 and MariaDB 11.8.9 parser checks for a leading-hyphen database operand

The full StartOS LXC/FUSE lifecycle was not run end to end.


helix-nine (changes requested):
Reviewed `git diff origin/master...HEAD` (`f7dcde71` → `0d35a70e`) with a high-effort parallel pass.

## Findings

1. `projects/start-sdk/lib/backup/Backups.ts:86-92` — **[CONFIRMED] SQL identifiers and literals are interpolated without escaping.** A database containing a backtick, or a password/user/host containing `'`, breaks restore; crafted values can append statements executed as database root. The new `rootPassword` and `userHost` inputs expand this existing hazard.

2. `projects/start-sdk/lib/backup/Backups.ts:605,792` — **[CONFIRMED] Both restore paths construct an unquoted root shell command.** A valid database such as `my db` reads the wrong path and fails. Shell metacharacters in the database, or a quote in the MySQL root password, can inject commands as container root.

3. `projects/start-sdk/lib/backup/Backups.ts:695-701` — **[CONFIRMED] MariaDB shutdown can wait forever after an unchecked signal failure.** Both `pkill` results are discarded, and the final `await running` has no bound. An image without `pkill`, or another failed signal attempt, leaves backup/restore hung indefinitely. The same block reduces the former graceful-shutdown allowance from about 120 seconds to 60, so a 61–120 second flush is now SIGKILLed.

4. `projects/start-sdk/lib/backup/Backups.ts:407-411,795-810` — **[CONFIRMED] Zero tables does not prove that a MariaDB import failed.** A legitimate empty database dumps and imports successfully, then the new count check rejects it. The shared JSDoc also promises this invalid invariant for both engines even though MySQL performs no such check.

5. `projects/start-sdk/lib/backup/Backups.ts:682-693` — **[CONFIRMED] `readyTimeout` counts probes rather than bounding elapsed time.** Each probe inherits `SubContainer.exec`'s 30-second timeout, so the documented 30-second timeout can take about 15½ minutes. The same duplicated defect remains in the PostgreSQL and MySQL loops at lines 289–300 and 456–464.

6. `projects/start-sdk/lib/backup/Backups.ts:625,746-748` — **[PLAUSIBLE] Defaulting `rootPassword` to a lazy `password` supplier evaluates that supplier twice.** A stateful supplier can return two secrets, assigning different application and root passwords despite the documented default. Reusing the first resolved value when `rootPassword` is omitted avoids that ambiguity.

7. `projects/start-sdk/docs/src/recipe-mysql.md:7` — **[CONFIRMED] The image prerequisite claim omits required utilities and incorrectly requires gzip for MySQL.** The MariaDB path additionally relies on `mkdir`, `chown`, `cp`, `find`, `sh`, `gunzip`, and `pkill`; MySQL relies on shell/core utilities but never invokes gzip. A custom/minimal image satisfying the stated requirement can therefore hang or fail.

8. `projects/start-sdk/CHANGELOG.md:133-134` — **[CONFIRMED] The changelog names an ambiguous API surface and describes prior absence.** The built SDK facade has `sdk.Backups`, not `sdk.backup`. The package namespace is `import { backup } from '@start9labs/start-sdk'`, exposing `backup.Backups` and `backup.mountBackupTarget`. State that positive public behavior rather than saying the namespace “was empty.”

9. `projects/start-sdk/lib/backup/Backups.ts:428-454,618-643` — **[CONFIRMED] Engine dispatch duplicates config normalization and mount construction.** Both branches use the same config type and immutable mount builder; normalizing once before dispatch removes a drift point while leaving engine-specific commands separate.

## Comment pass

- `Backups.ts:63` — `/** Root password assigned during restore (default: the application user's password). */`
- `Backups.ts:65` — `/** Host pattern assigned to the restored user (default: 'localhost'). */`
- `Backups.ts:67` — `/** Database engine used to initialize and run the restored data directory. */`
- `Backups.ts:407-416` — keep the durable restore contract, but move the official-image entrypoint/healthcheck/`MARIADB_AUTO_UPGRADE` guidance to the recipe; it is third-party operational detail rather than timeless API documentation.
- `Backups.ts:648` and `765-766` — delete; these restate the helper flow or duplicate the staging comment.
- `Backups.ts:756` — state the retry invariant directly rather than arguing from what StartOS does not do.

## Verification

- `make test` — 119 tests passed
- `make check` — passed
- `make check-fmt` — passed
- `npx prettier CHANGELOG.md docs/src/recipe-mysql.md --check` — passed
- GitHub checks are green at `0d35a70e`.

helix-nine (changes requested):
Re-reviewed the full PR at `2f855dde8` with a parallel high-effort pass, including the remediation delta from `0d35a70e`.

The amendment closes the ordinary SQL quoting, shell splitting, unbounded shutdown, empty-database, duplicate-password-resolution, export wording, and normalization findings. These issues remain:

1. `projects/start-sdk/lib/backup/Backups.ts:130-138,619-628,791-800,856-860` — **[CONFIRMED] Database operands beginning with `-` remain client options.** A valid database named `--help` makes the dump/import clients display help and exit 0, allowing an empty backup or no-op restore to appear successful. Insert a standalone `--` before each database operand.

2. `projects/start-sdk/lib/backup/Backups.ts:104-105` — **[CONFIRMED] The database target in `GRANT` is a wildcard pattern, but existing backslashes are not escaped before `_` and `%`.** A database `a\b` grants `ab`; `a\_b` leaves `_` active as a wildcard and can grant access to unintended databases. Escape existing backslashes before the wildcard characters.

3. `projects/start-sdk/lib/backup/Backups.ts:368-380,539-547,750-761` and `projects/start-sdk/CHANGELOG.md:128-131` — **[CONFIRMED] `readyTimeout` still does not bound elapsed readiness time.** Each probe can consume `SubContainer.exec`'s 30-second default and then sleep beyond the deadline; a one-second budget can take about 31 seconds or accept late success. PostgreSQL also starts its deadline after `pg_ctl` has already waited. Use one budget, pass the remaining time to each operation, cap sleeps, and include startup in the budget.

4. `projects/start-sdk/lib/backup/Backups.ts:763-769` — **[CONFIRMED] The SIGKILL fallback treats a no-match `pkill` as fatal.** If MariaDB exits after the 120-second race resolves but before `pkill` scans, `pkill` returns 1 and `execFail` turns a successful graceful shutdown into a failed backup/restore. Tolerate no-match and determine failure from the final bounded settlement check.

5. `projects/start-sdk/lib/backup/Backups.ts:487-502` and `projects/start-sdk/docs/src/recipe-mysql.md:7` — **[CONFIRMED] The public image-compatibility contract is still broader than the implementation.** Both paths require a `mysql` user/group and system utilities; MariaDB restore directly invokes `gunzip`, which the recipe omits. A binary-only custom image satisfying the current JSDoc can fail at ownership, startup, or decompression. The changelog's “official or not” should likewise be scoped to conventionally packaged images.

6. `projects/start-sdk/lib/backup/Backups.ts:26,70-71` — **[CONFIRMED] `LazyPassword` documentation still says callbacks resolve only during restore.** Both MySQL and MariaDB pre-backup hooks invoke the callback, so callback errors or side effects occur during backups too.

### Comment pass

- `Backups.ts:840` names the wrong MariaDB root-auth mechanism: fresh initialization uses local `unix_socket` authentication with an invalid native password, not an ordinary passwordless root account.
- Delete mechanical narration at `Backups.ts:19`, `130`, and `716`.
- `Backups.ts:564-569` includes an impossible non-daemonized MySQL branch; retain only the surprising zombie-is-stopped invariant.
- Reduce `Backups.ts:823` to the failed-restore retry invariant.
- Compress `StartSdk.ts:899-903` to the public behavior rather than builder/chaining instructions.

### Verification at `2f855dde8`

- `make test` — 119/119 passed
- `make check` — passed
- `make check-fmt` — passed
- changed-doc Prettier check — passed
- scratch-built and linked SDK consumer typecheck — passed for both `backup.Backups.withMysqlDump(...)` and `sdk.Backups.withMysqlDump(...)`
- GitHub checks are green except `Run Automated Tests`, which is still pending.

helix-nine (changes requested):
Re-reviewed the full PR at `8ed46ddb5` and the amendment from `2f855dde8` with parallel angle and verification passes.

The option terminators, GRANT backslash escaping, SIGKILL no-match handling, image prerequisites, lazy-password timing, and prior comment findings are substantially fixed. These issues remain:

1. `projects/start-sdk/lib/backup/Backups.ts:25-35,368-394,553-569,734-764` and `CHANGELOG.md:131-132` — **[CONFIRMED] `readyTimeout` is still not a strict shared startup/probe deadline.** `pollUntil` always starts a probe and accepts success after expiry. PostgreSQL rounds milliseconds up for `pg_ctl -t` while disabling the outer timeout, permitting up to 999 ms beyond budget. Check remaining time before and after each probe, and bound `pg_ctl` with the exact remaining outer timeout.

2. `projects/start-sdk/lib/backup/Backups.ts:67-69,93-94,164-165` — **[CONFIRMED] `readyTimeout` accepts values Node cannot use as the documented budget.** `0`, negatives, `NaN`, `Infinity`, and values above `2^31-1` become one-millisecond timers, immediate termination, or—in MariaDB's `Infinity` case—an indefinitely polling hook. Reject non-finite, non-positive, and out-of-range values before starting.

3. `projects/start-sdk/lib/backup/Backups.ts:79-86,109-138` — **[CONFIRMED] `root@localhost` can be configured with two conflicting passwords.** When `user === 'root'` and `userHost === 'localhost'`, `CREATE USER IF NOT EXISTS` leaves the initialized root account unchanged and `ALTER USER` applies `rootPassword`. A distinct or null `rootPassword` lets restore succeed while the application later authenticates with `password` and fails. The omitted/default case used by Ghost is safe. Reject conflicting resolved credentials or explicitly define and implement which value wins.

4. `projects/start-sdk/lib/backup/Backups.ts:550-698` — **[CONFIRMED] MySQL still stops gracefully only on the success path.** Readiness, dump, copy, grant, or import failure skips `stopMysql`; temporary-container teardown then kills daemonized `mysqld` through namespace teardown, leaving the persistent mounted database requiring crash recovery. This predates the PR but is in the touched hook; wrap startup/work in the same `try/finally` lifecycle now used by MariaDB.

5. `projects/start-sdk/lib/backup/Backups.ts:758-769` — **[CONFIRMED, low severity] A MariaDB exit during the final readiness probe is reported as a generic timeout.** The actual exit code/stderr is captured but never rechecked after `pollUntil` returns. Recheck `exited` before throwing the timeout.

6. `projects/start-sdk/lib/backup/Backups.ts:499-515` — **[CONFIRMED] The public JSDoc remains partly inaccurate.** It says root's configured password is always created although `rootPassword` may be null, and says staging needs one copy although MariaDB restore holds compressed and uncompressed forms concurrently. Keep the API gloss concise and leave the operational walkthrough in the recipe.

7. `projects/start-sdk/docs/src/recipe-mysql.md:13` — **[CONFIRMED] Mempool no longer demonstrates `withMysqlDump`.** Its current backup uses `Backups.ofVolumes`; link the active MariaDB consumer at `Start9-Community/romm-startos/startos/backups.ts` instead.

8. `projects/start-sdk/docs/src/recipe-mysql.md:7` and `projects/start-sdk/CHANGELOG.md:128-132` — **[CONFIRMED] New prose violates the repository's positive, timeless-copy rule.** Replace “Neither engine needs…” / “has no…” with positive requirements and instructions; replace “current” / “new” with the concrete MariaDB 11 compatibility and added fields.

### Comment pass

- `Backups.ts:25` currently promises deadline termination that the implementation does not enforce.
- `Backups.ts:89` must say `MYSQL_PWD` is supplied only when the password is non-null.
- Delete mechanical narration at `Backups.ts:101`, `671`, and `687-688`.
- Tighten the external-invariant comments at `Backups.ts:111`, `115`, and the zombie predicate at `577-578`.
- `Backups.ts:831` should state that `mariadb-install-db` requires an empty data directory; successful restore rsyncs can also populate it, so “only failed restore” is false.
- `Backups.ts:848` can state directly that the initialized root account authenticates over the local socket.

### Verification at `8ed46ddb5`

- `make test` — 119/119 passed
- `make check`, `make check-fmt`, `make bundle` — passed
- Prettier on all changed files — passed
- linked-consumer typecheck for both backup API surfaces — passed
- complete docs-site build — passed
- GitHub checks are still running; no failures are reported.

helix-nine (changes requested):
Assessed the full PR at `c11485b64` and the delta from `8ed46ddb5` with parallel angle and verification passes.

The prior strict-deadline flow, ordinary timeout validation, MySQL `finally`, MariaDB exit recheck, RomM example, and positive recipe/changelog edits are mostly addressed. These issues remain:

1. `projects/start-sdk/lib/backup/Backups.ts:122-153` — **[CONFIRMED] `CREATE USER IF NOT EXISTS` does not establish configured credentials for initializer-created accounts.** The root conflict check covers only `root@localhost`: with `user: 'root'`, `userHost: '%'`, and distinct passwords, the helper's next socket-based backup authenticates to `root@localhost` using `password`, but that account was assigned `rootPassword`. MariaDB also precreates `mysql@localhost`; choosing that user leaves its socket authentication unchanged, so the configured password never works. Explicitly alter the configured account, with root-specific handling based on the account used by local clients.

2. `projects/start-sdk/lib/backup/Backups.ts:419-432` — **[CONFIRMED] PostgreSQL error-path shutdown discards stop failure.** If `pg_ctl stop` fails after readiness, dump, copy, or restore fails, temporary-container teardown can kill PostgreSQL while it still uses the persistent volume. Preserve the primary operation error, but verify orderly shutdown or retain both failures.

3. `projects/start-sdk/lib/backup/Backups.ts:101-102,590-630` — **[CONFIRMED] `mysqldOptions` can override the PID file that shutdown assumes.** A caller-supplied `--pid-file` wins because custom options are last; `stopMysql` reads only `/var/run/mysqld/mysqld.pid`, treats the missing PID as successful shutdown, and teardown kills mysqld abruptly. Make SDK-owned lifecycle options take precedence or reserve the PID-file option.

4. `projects/start-sdk/lib/backup/Backups.ts:782-825` — **[CONFIRMED] MariaDB lifecycle errors are not preserved.** An exit during a blocking readiness probe is delayed until that probe returns, and a shutdown error thrown from `finally` replaces the original dump/import/startup error. Preserve the primary error (or aggregate both) and race readiness against process settlement if immediate exit reporting is promised.

5. `projects/start-sdk/lib/backup/Backups.ts:65-66,355-361,445-489` — **[CONFIRMED] PostgreSQL database names remain unsafe as filesystem components.** Valid names such as `tenant/acme` target missing directories; `../escape` leaves the intended `/tmp` and `/backup-target` paths, causing failed or misplaced backups. Use fixed staging names and an encoded or validated persisted filename.

6. `projects/start-sdk/lib/backup/Backups.ts:166-226,698-709,885-890` — **[CONFIRMED] `datadir` may overlap the reserved `/backup-target` mount.** The later read-only backup mount can hide the database mount, making recursive restore cleanup target the backup tree and fail deterministically. Canonicalize and reject equal, ancestor, or descendant overlap; no backup erasure path was confirmed.

7. `projects/start-sdk/lib/backup/Backups.ts:25-32,75,103` — **[CONFIRMED] Timeout validation still accepts values below the documented 1 ms minimum.** `0.5` passes, then Node clamps its timers to 1 ms while PostgreSQL rounds its internal timeout to one second. Reject `ms < 1`.

8. `projects/start-sdk/lib/backup/Backups.ts:34-43` — **[CONFIRMED, modest] The elapsed budget uses adjustable wall time.** A late StartOS clock step can expire startup early or extend it beyond `readyTimeout`; use a monotonic clock for elapsed-time accounting.

9. `projects/start-sdk/ARCHITECTURE.md:315-327`, `Backups.ts:93,537-551`, `CHANGELOG.md:128-136`, and the PR body — **[CONFIRMED] Public documentation and PR metadata remain out of sync.** Architecture omits the new `backup` namespace; `rootPassword` defaults to the resolved application-user password, not literal `password`; the PR body still claims arbitrary binary-bearing images, immediate exit detection, signal-command failure semantics, a nonexistent `sdk.backup` facade property, and says `userHost` is resolved. The public JSDoc remains an implementation walkthrough with the overbroad “official or not” claim.

### Comment pass

- Delete private-helper narration at `Backups.ts:34` and `114`.
- Reduce `Backups.ts:47,75,89-103` to the non-obvious lifecycle/default/range contracts.
- Replace the implementation and rejected-alternative walkthroughs at `Backups.ts:324-336` and `537-551` with concise logical-dump contracts.
- Keep the backup-fs hazard local and concise at `Backups.ts:356-360` and `579-581`; remove the outward duplicate at `483-484`.
- Keep the accurate external invariants at `Backups.ts:127,613,705,886,903`.

### Verification at `c11485b64`

- `make test` — 119/119 passed
- `make check`, `make check-fmt`, `make bundle` — passed
- Prettier on all changed files — passed
- linked-consumer typecheck for both backup API surfaces — passed
- complete five-book docs build — passed
- all GitHub checks — passed

helix-nine (changes requested):
The narrower scope is coherent, and #3917/#3918 adequately separate the broader lifecycle work. The MariaDB 11 path itself passed an independent Docker round trip on 11.8.9. One regression introduced by this PR still blocks approval:

1. `projects/start-sdk/lib/backup/Backups.ts:110-116,623` — **[CONFIRMED] Oracle MySQL does not accept `--` as an option terminator before the `mysql` client's database operand.** An executable check against `mysql:8.4.10` created the valid database `` `--help` ``: the patched `mysqldump ... -- --help` succeeded, but the exact restore command `mysql -u root -- --help` failed with `unknown option '--help'` (exit 2), leaving the database empty. MariaDB accepts the current form, so this escaped the MariaDB round trip. Bind the operand instead, e.g. `mysql -u root --database="$1" < "$2"`; the same form is supported by `mariadb`.

Please add focused coverage for the import argv on both engine clients if practical. All 119 existing tests remain green with this regression.

Two metadata/docs corrections can ride the same amendment:

- The PR body and commit say passwords travel through `MYSQL_PWD` rather than command-line arguments, but restore still passes `mysqlGrantSql(...)`, including the password, via `-e`. `SubContainer.exec` also serializes env overrides into the transient `start-container` argv. Narrow this to the behavior actually changed: dump/import client authentication no longer uses client `-p` arguments.
- The body/commit say both readiness loops bound elapsed time while strict deadline handling is explicitly deferred to #3917; remove that claim. `ARCHITECTURE.md:323` also says “Three builder patterns” above four bullets.

Verification at `01229abfa`: `make test` (119), `make check`, `make check-fmt`, `make bundle`, linked-consumer typechecking, six-file Prettier, and the five-book docs build all pass. GitHub checks are green.

helix-nine (dismissed):
Reviewed the exact final delta with:

`git diff 01229abfad235f467b97661e154d0b389c6311ff..134df443df6aae3c015c223e2dfa3c255625c1a4`

A high-effort angle fan-out plus independent candidate verification found no remaining correctness issues. The production command now binds the database as `--database="$1"`; both restore paths exercise the shared helper, and the form has been reproduced against MySQL 8.4.10 and MariaDB 11.8.8. The architecture correction is accurate. Comment pass was clean.

One nonblocking cleanup observation: `projects/start-sdk/lib/test/backup.mysql.test.ts:89-90` asserts `mountBackupTarget`'s `execFile`/`mkdir` internals even though the focused test already proves it reached the import command. Those assertions and their lines 1-2 imports can be removed later to avoid coupling this regression test to a behavior-preserving mount refactor; they do not weaken or block this change.

Verification is green: 121 tests, typecheck, formatting, bundle, five-book docs build, both client reproductions, and all GitHub checks.

dr-bonez (changes requested):
NACK. doesn't look like mariadb and mysql share enough to justify using a single helper for both

helix-nine (dismissed):
Split in `f9a562861` after @dr-bonez's request. MariaDB now has its own public `withMariadbDump` builder and `MariadbDumpConfig`; MySQL keeps `withMysqlDump`, and each owns its lifecycle, tools, options, and dump format. The 3.0 migration, API docs, architecture notes, and tests now cover both separately. Legacy extracted configs are rejected rather than silently selecting MySQL or dropping MariaDB options.

Rebased onto current `master` and re-reviewed the exact final diff with a high-effort angle/verification pass; no findings survived. Verification: `make check`, 121 tests, `make check-fmt`, `make bundle`, five-book docs build, external public-API typecheck, and `git diff --check`.

dr-bonez (approved):

PR #3916, opened 2026-09-08
fix(web): skip zod's eval probe under the ui csp

MattDHill:
## Summary

Loading the StartOS UI, setup wizard, or StartTunnel UI logs a Content-Security-Policy violation in the browser console:

```
Content-Security-Policy: The page's settings blocked a JavaScript eval (script-src) from being executed because it violates the following directive: "script-src 'self'" (Missing 'unsafe-eval') main-4HUHR7JK.js:40:28063
```

The source is Zod 4.4.3's `allowsEval` probe (`try { new Function("") } catch {}`), run when the first object schema is built to decide whether to JIT-compile object parsers. Under `script-src 'self'`, the constructor throws, Zod catches it and uses its interpreted parser, and the browser still reports the rejected eval. Nothing is functionally broken, and the CSP stays as it is.

Calling `config({ jitless: true })` through Zod's public `zod/v4/core` entry skips the probe. Each affected app imports a local `zod-config.ts` from its polyfills entry, which Angular loads before the application bundle. Zod stores the setting on `globalThis`, so it reaches the Zod copies in the separately-built polyfill and application bundles. The obsolete later calls in the StartOS UI and StartTunnel application configs are removed.

StartTunnel's new polyfill path is extensionless so Angular continues to classify the app as zoneless and preserves native async functions. The WRT UI does not import start-core and needs no configuration.

The initial implementation imported the configuration first from each `main.ts`, but production-bundle interception showed that Angular still ordered Zod schema construction ahead of the call in the StartOS UI and StartTunnel bundles. Moving the configuration into the earlier polyfill phase reduced all three bundles to zero trapped `Function` calls.

## Verification

- `npm run check:ui`
- `npm run check:setup`
- `npm run check:tunnel`
- `npm run build:ui`
- `npm run build:setup`
- `npm run build:tunnel`
- Executed each generated polyfill followed by its application bundle with the `Function` constructor trapped: zero calls in all three apps
- Confirmed StartTunnel's generated bundle retains native async functions
- Prettier and `git diff --check`

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -248,7 +248,7 @@
             },
             "index": "projects/start-tunnel/web/src/index.html",
             "browser": "projects/start-tunnel/web/src/main.ts",
-            "polyfills": [],
+            "polyfills": ["./projects/start-tunnel/web/src/polyfills"],
             "optimization": {
               "scripts": true,
               "fonts": true,
--- a/projects/start-os/web/setup-wizard/src/main.ts
+++ b/projects/start-os/web/setup-wizard/src/main.ts
@@ -1,3 +1,6 @@
+// Keep first: Zod must be configured before any schema is built.
+import './zod-config'
+
 import { enableProdMode } from '@angular/core'
 import { bootstrapApplication } from '@angular/platform-browser'
 import { AppComponent } from 'src/app/app.component'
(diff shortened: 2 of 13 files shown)


helix-nine (approved):
Reviewed `origin/master...ff5d65eff`. No actionable findings remain after the production-bundle ordering fix. All three apps configure Zod before their application bundles, StartTunnel remains zoneless, and the typechecks and production builds pass.

dr-bonez (approved):

Issue #3921, opened 2026-09-10
package attach: a wrong -s reports "no matching subcontainers are running", then promises a choices list and prints nothing


stupleb:
## Summary

`start-cli package attach` selects a subcontainer by Guid (`-s`), name (`-n`) or image (`-i`). When `-s` is given something that is not a subcontainer Guid, the resulting error names the wrong cause — it reads as though the service is down — and then ends with *"some possible choices are:"* followed by nothing. The list exists: it is attached to the error and discarded before display.

The help text that invited this particular mistake is fixed separately in #3922. Both defects below survive that fix — any wrong `-s` value lands here, and as shown under *How a user is meant to find a Guid*, correct-**looking** values are easy to come by.

## Repro

```
start9@monsacro-casa:~$ start-cli package attach nextcloud -s nextcloud-sub
Not Found: package.attach: no matching subcontainers are running for nextcloud; some possible choices are:
start9@monsacro-casa:~$ start-cli package attach nextcloud -n nextcloud-sub
root@2ZX3NWIL5ZQPFFCARGR4B7D3KUEF6HCW:/var/www/html#
```

The first command's output ends at that colon. `nextcloud-sub` is a real, running subcontainer, as the second command shows.

## Why a non-Guid gets that far

`CliAttachParams.subcontainer` is `Option<InternedString>`, so clap validates nothing. Server-side `AttachParams.subcontainer` is `Option<Guid>` — but `Guid` is a newtype with a **derived** `Deserialize` (`rpc_continuations.rs:24-28`), so the length/alphabet check in `Guid::from` (`:34`) is reachable only through `FromStr`/`ValueParserFactory` (`:56`, `:62`), never over the wire.

`Guid("nextcloud-sub")` is therefore constructed without complaint and substring-matched against real Guids (`service/mod.rs:977`). A wrong-flag error degrades into an empty search.

## 1. The error names the wrong problem

No-match returns `ErrorKind::NotFound` with *"no matching subcontainers are **running**"* (`service/mod.rs:1013`). That points at service health, which is the wrong debugging path — the subcontainer is running fine. Nothing in the message suggests that the value, or the choice of flag, is at fault.

## 2. The choices list is never rendered

`Display for Error` (`error.rs:210`) writes only `kind` + `display_src()`; `info` is dropped. No site under `shared-libs/crates/start-core/src/bins/` or `projects/` prints it. The captured output above is the proof — the sentence sets up a list and stops.

[...]

PR #3922, opened 2026-09-10
fix(cli): describe attach's --subcontainer as the Guid it matches

stupleb:
`start-cli package attach -s/--subcontainer` matches the subcontainer's internal **Guid** (`service/mod.rs:977` — the map key, uppercase base32). Its help string read *"Name of the subcontainer"*, which is the one value it will not accept. `-n/--name`, which does take the name, read the vaguer *"Name of the container"*.

So the flag advertised as the name-taker rejects names, and the flag that accepts them doesn't say so:

```
$ start-cli package attach nextcloud -s nextcloud-sub
Not Found: package.attach: no matching subcontainers are running for nextcloud; some possible choices are:
$ start-cli package attach nextcloud -n nextcloud-sub
root@2ZX3NWIL5ZQPFFCARGR4B7D3KUEF6HCW:/var/www/html#
```

## The change

String table only. Both keys are used at exactly one site each, `CliAttachParams` (`service/mod.rs:1343`, `:1345`).

- `help.arg.container-name` (`-n`) takes over the strings that were under `help.arg.subcontainer-name`. All five values change, but none is newly authored — each is an existing, already-reviewed translation moved onto the flag it actually describes.
- `help.arg.subcontainer-name` (`-s`) gets new text naming the Guid and pointing at `--name`. Those five strings are the only new translations in the diff.

Prefixes are mentioned because the match is `contains`, not equality, so a Guid prefix works.

## Deliberately not done

The key `help.arg.subcontainer-name` is now a misnomer for text about a Guid. Renaming it to `…-guid` would also want `container-name` → `subcontainer-name`, which collides with the old key and means relocating entries to hold the file's alphabetical ordering. Happy to do that swap if a reviewer prefers it — left out here to keep the diff to values.

The matching code itself is correct and deliberate: for `name` both sides are uppercased, while for `subcontainer` only the needle is, because Guids are uppercase by construction. This PR changes no logic.

## Verified

- YAML parses; 965 keys; both entries carry all five locales.
- `npx prettier --check` clean on the file.
- `rust_i18n::i18n!` loads this file at compile time; the change is values-only inside a file that still parses, with no keys added or removed.
- Rendered `--help` **is** confirmed: the man pages are clap-generated and committed, so `make manpages-check` in CI built the binary and regenerated `start-cli-package-attach.1`. Its only drift was these two strings, taken in the second commit.

Fixing the help text does not fix the failure it currently invites: a wrong `-s` still reports the wrong cause and still prints an empty choices list. That is tracked separately in #3921.

### DIFF
--- a/projects/start-cli/man/start-cli-package-attach.1
+++ b/projects/start-cli/man/start-cli-package-attach.1
@@ -13,10 +13,10 @@ Execute commands in container
 Force TTY mode for I/O
 .TP
 \fB\-s\fR, \fB\-\-subcontainer\fR \fI<SUBCONTAINER>\fR
-Name of the subcontainer
+Guid of the subcontainer, or a prefix; use \-\-name to select by name
 .TP
 \fB\-n\fR, \fB\-\-name\fR \fI<NAME>\fR
-Name of the container
+Name of the subcontainer
 .TP
 \fB\-u\fR, \fB\-\-user\fR \fI<USER>\fR
 User name to run as
--- a/shared-libs/crates/start-core/locales/i18n.yaml
+++ b/shared-libs/crates/start-core/locales/i18n.yaml
@@ -3315,11 +3315,11 @@ help.arg.config-file-path:
   pl_PL: "Ścieżka do pliku konfiguracyjnego"
 
 help.arg.container-name:
-  en_US: "Name of the container"
-  de_DE: "Name des Containers"
-  es_ES: "Nombre del contenedor"
-  fr_FR: "Nom du conteneur"
-  pl_PL: "Nazwa kontenera"
+  en_US: "Name of the subcontainer"
+  de_DE: "Name des Subcontainers"
+  es_ES: "Nombre del subcontenedor"
+  fr_FR: "Nom du sous-conteneur"
+  pl_PL: "Nazwa podkontenera"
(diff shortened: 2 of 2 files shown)


MattDHill (approved):
Checked against the code and it holds:

- `service/mod.rs` `attach`: `-s` substring-matches the subcontainer map key (the Guid), `-n` matches `wrapper.name`. The new strings describe what each flag actually accepts.
- Both keys are referenced only from `CliAttachParams` (`service/mod.rs:1343`, `:1345`), so moving the "Name of the subcontainer" set onto `container-name` changes no other command's help.
- `start-cli-package-attach.1` matches the regenerated output and `Generated Artifacts` is green. No other man page carried the old strings.
- The five translations read correctly.

Nit, non-blocking: the match is `contains`, so any fragment of the Guid works, not only a prefix. "Or a prefix" is still true and is how people abbreviate ids, so fine as written.

Leaving the key names alone is the right call; a values-only diff is what this wants to be.

The StartOS book had the same inversion in `cli-reference.md` and a nonexistent positional `<SUBCONTAINER>` form in `service-containers.md`; that is #3924 on `live-docs`, which backports to master, so this PR correctly leaves the book untouched.

PR #3927, opened 2026-09-10
feat(startos): welcome a server with the release notes it updated to

MattDHill:
## Summary

- Raise one post-update notification for the version a server lands on, using the same curated release-notes file as the GitHub release and registry entry.
- Localize the notification title and summary, and document where users can reopen it.
- Package the notes as data in the signed OS image rather than compiling them into `startbox`.

## How it works

`version::post_init` already drains `postInitMigrationTodos` after an update. When that set is non-empty, it now raises one notification for the current version rather than one notification per migration crossed.

`projects/start-os/build.mk` selects `projects/start-os/release-notes/<version>.md` from `VERSION.txt` and installs it as `/usr/lib/startos/release-notes.md`. `version/release_notes.rs` reads that packaged file after migration. A missing source file fails the image build, while a notes-only edit rebuilds the package/image without recompiling unchanged Rust binaries.

The changelog link is appended to the notification body. The existing `v0_4_0.rs` welcome remains unchanged for servers arriving from a 0.4.0 beta.

## Verification

- `cargo check -p start-core --lib`
- `cargo test -p start-core --features test version::` (31 passed)
- pinned `cargo fmt --check` and Prettier checks
- `make -n PLATFORM=x86_64 start-os-install` resolves the current notes file and installs it at the runtime path
- a synthetic missing-version notes path fails the `start-os` Make target
- workflow retarget checker plus trigger/image-filter assertions


helix-nine (dismissed):
Verified `8531bac1`: the English and French quoted labels now match the UI dictionaries, and the guide points to the actual Notifications menu entry. The earlier build-graph, localization, docs/changelog, and comment issues remain resolved. `cargo check -p start-core --lib` passed on the preceding implementation head; this final commit changes only those verified strings.

helix-nine (approved):
Approved at `833857363`. The notes now ship as a signed image data file rather than binary content; the Make dependency, runtime fallback, workflow trigger/image classification, localization, and user documentation are consistent. Local `cargo check`, all 31 version tests, formatting, Make dependency checks, and the completed CI jobs pass. The two RISC-V images are still running.

PR #3932, opened 2026-09-12
fix(networking): reconcile and drain forwarded ports reliably

### LINKED ISSUE #3912: A "Local" exposure means two different things: the vhost accepts any RFC1918 source, a raw-port DNAT accepts only the on-link subnet
## Summary

A `Local` exposure is enforced by two different rules depending on whether the binding terminates TLS, and the two disagree about what "local" means.

- **vhost / SSL path** accepts a source that is on-link **or** any RFC1918 address.
- **Raw-port DNAT path** accepts only the exact on-link subnet, plus the container bridge.

So a client on a routed-but-separate subnet — a second VLAN, a wired/wireless split, an IoT SSID the admin has deliberately routed — can reach the StartOS admin UI and every service web UI, but cannot reach a non-SSL forwarded port.

## The two rules

`shared-libs/crates/start-core/src/net/vhost.rs:1151`

```rust
fn accepts_as_private(&self, subnets: &OrdSet<IpNet>, src: IpAddr, dst: IpAddr) -> bool {
    self.private.contains(&dst)
        && (subnets.iter().any(|s| s.contains(&src)) || is_private_ip(src))
}
```

`is_private_ip` (`shared-libs/crates/start-core/src/net/utils.rs:143`) is true for `10/8`, `172.16/12`, `192.168/16`, loopback and link-local.

`shared-libs/crates/start-core/src/net/forward.rs:708`

```rust
let src_filter = if public {
    None
} else if reqs.private_ips.contains(&IpAddr::V4(ip)) {
    Some(subnet.trunc())
} else {
    continue;
};
```

which `build/lib/scripts/forward-port:79` renders as just the on-link subnet plus the bridge:

```
add rule ip startos prerouting ip saddr <on-link subnet> ip daddr <sip> meta l4proto { tcp, udp } th dport <p> dnat to <container>:<p>
add rule ip startos prerouting ip saddr 10.0.3.0/24      ip daddr <sip> meta l4proto { tcp, udp } th dport <p> dnat to <container>:<p>
```

The admin UI reconciles through the same host-update path as a service (`shared-libs/crates/start-core/src/net/net_controller.rs:1139`, `HostId::admin()`), so it is gated by `accepts_as_private` and inherits the looser rule.

## Practical effect

From a host on a second, routed subnet:

| target | enforced by | reachable |
| --- | --- | --- |
| StartOS admin UI on 443 | vhost | yes |
| any service web UI | vhost | yes |
| any non-SSL forwarded port | DNAT | no |

The least sensitive surface is the only one restricted.

## How it surfaced

Start9Labs/datum-gateway-startos#14, and the [forum thread](https://community.start9.com/t/problem-with-datum-and-bitaxe-resets-peer-and-startos-v-0-4-0-1/5534) behind it.

Bitaxe miners are Wi-Fi-only and frequently land on an IoT or guest SSID, so DATUM Gateway's stratum port is the binding most likely to trip the DNAT rule. The failure presents as the service being down: the host refuses the connection and nothing is logged by the service. In the thread a user confirmed a hand-added `nft` rule restores connectivity, which is the workaround people are now running. Those rules do not survive a reboot.

## Neither side is a recent regression

The RFC1918 clause has been in the vhost since #3049 (alpha.12, November 2025). The DNAT subnet filter predates the monorepo reorg in #3352. This is a longstanding divergence rather than something that broke recently, which is part of why I'd rather not pick a direction unilaterally.

## @dr-bonez — which way should this go?

Whichever we choose, `Local` should mean one thing. As I see the options:

1. **Loosen the DNAT path to match the vhost.** Emit an RFC1918 set alongside the on-link subnet in `forward-port`. Smallest change, no new config surface, and it resolves datum#14 without any UI work. It is a loosening.

2. **Tighten the vhost to on-link only.** Drop the `|| is_private_ip(src)`. Hardens the admin UI and every service UI, and matches the stricter reading of "Local". It would silently cut off anyone reaching a service UI across subnets today, and it does not resolve datum#14.

3. **Per-gateway allowed client subnets.** Default stays on-link-only and the user opts in per subnet. This is what datum#14 literally asks for, and it is much the largest change: DB model, migration, API, UI, and bindings.

My default read is option 1, on the grounds that the vhost clause looks like a deliberate accommodation for multi-subnet LANs that the raw-port path never received. Option 2 is a coherent security position and the call is yours. Happy to open the PR either way once you pick.

Verified by reading `master` at `f7dcde71b`, not by a live two-subnet test.

helix-nine:
Closes #3912

## Why this PR is broader than #3912

The issue itself requires one policy change: a private IPv4 raw-port forward must accept clients routed from another private subnet, not only clients on the server's directly attached subnet.

Adding packet-level regression coverage exposed separate pre-existing forwarding defects: setup failures could lose the desired request, teardown failures could lose the applied rule or router handle, IPv6 state could claim a failed operation had succeeded, nft inspection failures could look like an empty ruleset, and process shutdown depended on dropped references and detached cleanup tasks.

Those defects share one root problem: desired intent, applied external state, and cleanup ownership were conflated. Retrying safely requires retaining both the desired specification and the exact applied rule or mapping until setup or removal succeeds. The lifecycle refactor below is not required merely to widen the #3912 source match; it is included to give every forwarding side effect a durable owner and a bounded, explicit teardown path.

## The #3912 behavior change

- Private IPv4 `Local` forwards now accept the directly attached subnet plus routed `10/8`, `172.16/12`, `192.168/16`, loopback, and link-local sources.
- The separate bridge-source rule is removed because the StartOS bridge is already covered by `10/8`.
- Public/WAN forwards remain source-unrestricted.
- IPv6 `Local` forwards remain restricted to the client's on-link subnet.

A separate packet-filter correction binds every forward-chain accept to `ct status dnat`, the original destination address and external port, and the translated destination address and internal port. Port ranges preserve the exact external-to-internal offset. Traffic from an unrelated DNAT operation can no longer match merely because it reaches the same translated destination tuple.

### Before

- `NetServiceData` attempted setup while processing a host. An IPv4 failure could return before a durable lease retained the request.
- IPv6 rules were kept in a caller-owned map that was replaced even after failed additions or removals.
- Low-level IPv4 state could discard a mapping after failed removal.
- PCP/NAT-PMP and UPnP teardown could discard the active handle needed to retry; UPnP cleanup rediscovered a gateway instead of retaining the gateway that created the mapping.
- Reconciliation stopped at the first failed host, and interface changes did not rerun host-forward reconciliation.
- Shutdown relied on `Drop`, detached tasks, and best-effort cleanup while new work could still be admitted.

### After

Forwarding has a durable ownership chain:

1. `NetServiceData` derives host-level IPv4 and IPv6 specifications; `HostBinds` keeps strong leases for the desired requests.
2. `InterfacePortForwardController` resolves those requests against the live interface snapshot, owns desired-versus-applied IPv6 state, and owns leases on successfully applied IPv4 rules.
3. `PortForwardController` owns the applied IPv4 nft mappings and removes them from state only after `unforward` succeeds.
4. `PortMapController` separately owns desired, active, and stale PCP/NAT-PMP/UPnP mappings, including the exact handles and UPnP gateways required for cleanup retries.
5. Initialization removes forwarding-owned kernel state for which no in-memory owner can survive; shutdown closes admission and drains every applied layer before exit.

Applied state advances only after successful setup or removal. Failed teardown retains the old identity; an IPv6 replacement removes the old rule before adding the new one; and router-map replacement cannot overlap an old mapping whose withdrawal failed.

A 30-second periodic pass retries retained setup and cleanup failures without requiring another database event. Interface changes now rerun host reconciliation, target prefixes are selected from the current most-specific containing subnet, and the prefix participates in rule identity so topology changes replace rather than reuse an incompatible masquerade rule. Reconciliation attempts every host/mapping before returning accumulated failures, prunes empty state, and diagnostics report only forwards confirmed applied.

## nft transaction execution

The IPv4 and IPv6 scripts now validate inputs and render family-specific rules, while the shared `forward-port-nft` helper owns their common transaction lifecycle:

- list tagged rules in all four dynamic chains;
- atomically delete the prior rule set and install its replacement in one `nft -f -` batch;
- relist and retry stale-handle `ENOENT` races up to three times;
- propagate base-setup, chain-listing, and transaction failures instead of treating them as success;
- run conntrack cleanup only after nft teardown succeeds.

IPv4 and IPv6 use strict, family-separated deterministic tags derived from the complete forward identity. Startup cleanup accepts only that generated tag shape, so it does not remove rules owned by another subsystem.

Each forwarding-script invocation has a 30-second limit and a dedicated process group. Timeout or cancellation kills descendant `nft`, `awk`, and `conntrack` processes rather than dropping only the parent shell.

### StartOS

- Forwarding initialization owns base rules, forwarding sysctls, and stale dynamic-rule cleanup; requests received during initialization remain queued for reconciliation.
- Every initialization error after the network controller begins owning forwarding state passes through the same drain boundary.
- Runtime shutdown stops cron producers, then waits for service shutdown and forwarding shutdown concurrently before marking the RPC context closed.
- Interface forwarding and router mappings drain concurrently under one shared 90-second deadline, retaining both errors if both paths fail.
- `startd` supervises and joins its signal and metrics tasks, always invokes `RpcContext::shutdown` after a primary task error, and preserves the primary error while logging secondary cleanup failures.
- `startd.service` intentionally has no fixed stop-time override because package service shutdown can legitimately exceed two minutes.

### StartTunnel

StartTunnel uses the same forwarding primitives and now owns the producer side of their lifecycle:

1. Reserve the HTTP listener before restoring forwarding state, remove stale pinholes, and drain partially restored state if initialization fails.
2. Spawn PCP, IGD, and lease-reaper workers only after successful initialization and retain their task handles.
3. On shutdown, stop and join signal/HTTPS/redirect work, stop RPC admission and drain web connections, then give PCP/IGD/lease workers up to 15 seconds to exit.
4. Close forwarding admission across IGD and every mutating RPC, wait for admitted operations, stop SNI listeners, clear IPv4 ownership, and drain IPv4 rules and IPv6 pinholes with retry under the shared 90-second deadline.

`start-tunneld.service` receives a 120-second stop allowance for that bounded 90-second forwarding drain. Reboot and poweroff requests keep their requested machine action live if cleanup fails; ordinary tunnel shutdown reports the cleanup failure.

`WebServer::shutdown` now drives Hyper's graceful-shutdown future while draining connection tasks and returns server-task panics instead of unwrapping them. StartOS, StartTunnel, StartRegistry, and StartWRT propagate the now-fallible shutdown result.

## Concrete pre-existing bugs fixed by the lifecycle work

- A transient IPv4 setup error no longer permanently suppresses a requested forward.
- Failed IPv4/IPv6 removal no longer loses the rule identity and leaks it indefinitely.
- A failed IPv6 addition is not recorded as applied, and a failed replacement cannot install over an old rule that remains live.
- Changed interface prefixes rebuild their masquerade rule instead of reusing an incompatible mapping.
- Failed PCP/NAT-PMP or UPnP withdrawal retains the exact cleanup object and retries with bounded backoff.
- Undesired or stale router mappings no longer answer reachability queries as though they were current.
- New work cannot race terminal cleanup, queued ordinary port-map commands cannot revive drained state, and concurrent drain callers share one cancellation-safe completion.
- A crashed process's generated nft rules are removed on the next initialization.

## Unrelated fixes and cleanup included

These changes are real but are not required by #3912 or by the forwarding ownership model:

- Correct the StartTunnel workflow to read its declared `arch` dispatch input instead of the nonexistent `platform` input.
- Correct malformed non-release `PROFILE` comparisons in the StartTunnel build and start-core test runners.
- Consolidate StartTunnel contributor guidance into its scoped `AGENTS.md`, remove the superseded `CONTRIBUTING.md`, and refresh architecture links/terminology.

The user-visible StartTunnel shutdown change is released as 1.3.1 with matching changelog and release notes. The StartOS book, changelog, and 0.4.0.2 release notes document routed-private reachability and forwarding reliability.

## Verification

- `make start-core-test` — 532 passed, 3 ignored on the final rebased head; this also runs the mocked nft shell suite.
- `cargo test -p start-core net::forward::tests --features=test --locked` — 26 passed.
- `cargo test -p start-core net::port_map::client::tests --features=test --locked` — 27 passed.
- `cargo test -p startwrt-core -p uciedit -p uciedit_macros --locked` — 570 startwrt-core and 18 uciedit tests passed.
- `cargo check -p start-core --locked`.
- Generated singleton, equal-range, offset-range, and IPv6 transactions pass `nft --check --file <capture>` in an isolated Debian bookworm container.
- `cargo +nightly-2026-05-28 fmt --all -- --check` and `make start-core-format-check`.
- Prettier checks for changed Markdown and workflow YAML.
- `mdbook build` in `projects/start-os/docs` and `projects/start-tunnel/docs`.
- `bash -n` and ShellCheck on the production/regression shell scripts and test runners.
- `git diff --check`.

`./scripts/manage-release.sh pre-check start-os` passed the version, changelog, release-note, docs-link, crate-label, tag, registry, and GitHub checks; it then reported the expected absence of release-operator GPG/S3 credentials and `s3cmd` in this development environment.


projects/start-os/startd.service
```diff
@@ -7,6 +7,7 @@ Environment=RUST_LOG=warn,start_core=debug
 ExecStart=/usr/bin/startd
 Restart=always
 RestartSec=3
+TimeoutStopSec=120
```
dr-bonez: no - startd waits for all running services before exiting, and their timeouts can be well over 2min

helix-nine: Agreed. Removed the `TimeoutStopSec` override from `startd`; its shutdown must retain enough time for every running service to exit. The 120-second override remains scoped to `start-tunneld`, whose own forwarding drain has a 90-second deadline and does not wait for package services.

dr-bonez:
Closing in favour of three smaller PRs:

- #3962 — the #3912 change on its own: the private-set DNAT rule in `forward-port`, changelog and docs.
- #3963 — stale forward rules removed at controller init, which covers the crash/restart case without a shutdown drain.
- #3964 — plain UPnP mappings request a finite lease, so every router grant self-expires like PCP's already do.

The rest of this PR is declined:

- The lifecycle rewrite reverses deliberate best-effort semantics (the port-map module doc, the v6 forward reconcile in `net_controller.rs`, no forwarding cleanup at exit) and adds a drain of up to 90 s to shutdown, reboot and the init-failure path into Diagnostic Mode, for state the kernel discards on reboot and routers expire on their own. It also logs a failed network update for every service on every shutdown, because admission closes before the services' own teardown reconciles.
- The forward-chain accept tightening to `ct status dnat` plus the exact tuple is a separate security change with no changelog line and no live-network test on this branch. If wanted, it needs its own PR.
- The StartTunnel lifecycle changes, the fallible `WebServer::shutdown` and the unrelated CI/profile fixes likewise belong in their own PRs with their own justification.

PR #3933, opened 2026-09-12
fix(start-os): stabilize installation disk identity

### LINKED ISSUE #3923: [bug]: installer writes kernel device names to /etc/fstab with no `nofail` — a re-enumerated disk boots the server into unrecoverable emergency mode
### Prerequisites
- [x] I searched existing issues and this is not already reported.

### StartOS version
0.4.0.2 (field report); defect verified in `master` @ `7b9521d9f`

### Architecture / image
x86_64 (AMD64) — Full

### Server hardware
DIY Apple machine, 3 SATA disks (one failing SMART — 1815 reallocated sectors)

### Which surface?
Other / unsure — boot / installer

### What happens

`os_install` writes `/etc/fstab` from `fstab.template`:

```
{boot}  /boot       vfat    umask=0077  0   2
{efi}   /boot/efi   vfat    umask=0077  0   1
{root}  /           btrfs   defaults    0   1
```

The placeholders are filled from `part_info.boot` / `extra_boot["efi"]` / `root`, all `PathBuf`s holding **kernel device names** — `partition_for()` builds them as disk path + index, e.g. `/dev/sdb` + `1` → `/dev/sdb1`. So an installed fstab reads:

```
/dev/sdb1  /boot      vfat   umask=0077  0  2
/dev/sdb2  /boot/efi  vfat   umask=0077  0  1
/dev/sdb3  /          btrfs  defaults    0  1
```

No `UUID=` / `PARTUUID=`, and **no `nofail` on any entry**.

`/dev/sdX` assignment is SATA probe-order dependent and is not stable across boots. On a multi-disk box — especially one with a marginal drive that is slow to respond — the OS disk can enumerate under a different letter for a single boot. Then:

1. `dev-sdb1.device` never appears; `boot.mount` waits the default 90 s and fails.
2. `local-fs.target` fails, so systemd enters `emergency.target`.
3. `emergency.service` runs `sulogin`. StartOS locks `root`, so sulogin prints `Cannot open access to console, the root account is locked` and exits immediately.
4. `multi-user.target` is never reached, so **`ssh.service` and `startd.service` never start**. The server is unreachable on LAN and remotely, and no console login is possible — a headless brick until someone physically power-cycles it.
5. Root is a squashfs image with a **tmpfs** upper (`startos-initramfs-module`), and persistent journal storage is not mounted, so journald stays in `/run/log/journal`. **The failed boot's journal is destroyed by the next reboot.** `journalctl -b -1` then shows a *healthy* boot, so the failure erases its own evidence and gets misattributed.

A warm `systemctl reboot` largely preserves SATA link state, so a bad enumeration tends to reproduce across warm reboots; only a cold power-off (full link re-training) reshuffles it. That makes the failure look intermittent and hardware-random.

Separately: the `/` line names a btrfs block device, but the runtime root is the overlay the initramfs assembles. `OsPartitionInfo::from_fstab()` documents exactly this — *"`/` is an overlayfs the initramfs sets up, so its fstab source (`overlay`) names no block device"* — and skips every non-`/boot` entry when reading. The generator still writes a device there, so what is on disk contradicts what the reader assumes.

- fstab should use `PARTUUID=`, which survives re-enumeration. `resolve_fstab_source()` already resolves `UUID=` / `PARTUUID=` when reading fstab, so nothing downstream needs to change.
- `/boot` and `/boot/efi` should carry `nofail`. Neither is needed to reach `multi-user.target` — they matter for OS updates and GRUB. With `nofail`, a mis-enumerated or unhealthy boot partition degrades to *"the next update fails with a clear error"* instead of *"the server is unreachable and cannot be logged into"*.
- The `/` entry should be dropped or written as `overlay`, matching what `from_fstab()` already assumes.
- Worth considering separately: `emergency.service` on a box with a locked root and no usable console is a dead end by construction. A fallback that still brings up sshd would turn a brick into a support case.

### Steps to reproduce

1. On any installed x86 box, `cat /etc/fstab` — confirm the entries name `/dev/sdXN`.
2. Add a SATA disk that enumerates ahead of the OS disk, or move the OS disk to a different port.
3. Boot. `boot.mount` times out after 90 s; the box lands in emergency mode and is unreachable over SSH.
4. Power-cycle into a good boot, then run `journalctl -b -1` — the failed boot left no journal.

Field report: DIY Apple hardware, 3 SATA disks, one failing SMART. An OS update was run over SSH; the box rebooted into the emergency console below and was unreachable for ~90 minutes. Two warm reboots did not clear it; a cold power-off did.

### Logs & evidence

```shell
You are in emergency mode. After logging in, type "journalctl -xb" to view
system logs, "systemctl reboot" to reboot, or "exit" to continue bootup.
sulogin: tcgetattr failed: Input/output error
Cannot open access to console, the root account is locked.
```

Code refs (`master` @ `7b9521d9f`):

- `shared-libs/crates/start-core/src/os_install/fstab.template`
- `shared-libs/crates/start-core/src/os_install/mod.rs:522-533` — template fill
- `shared-libs/crates/start-core/src/os_install/mod.rs:47-62` — `partition_for()`
- `shared-libs/crates/start-core/src/disk/mod.rs:50-112` — `from_fstab()` / `resolve_fstab_source()`
- `projects/start-os/build/lib/scripts/startos-initramfs-module:113-121` — tmpfs upper
- `projects/start-os/startd.service` — `WantedBy=multi-user.target`

helix-nine:
## Summary

- retain deterministic `/dev/disk/by-path` handles alongside canonical disk names and carry those handles from `setup.disk.list` through the setup UI, CLI, request validation, partitioning, and quiesce operations
- require a stable path from the freshly enumerated disk list at the destructive install boundary; a kernel device name such as `/dev/sda` is rejected, and the CLI sends `stablePath`
- match the partition Preserve protects by canonical device node, so the match no longer depends on which by-path alias the disk list picked
- write fresh-install `/etc/fstab` entries with `PARTUUID=` sources for the boot and EFI partitions, and no longer write a `/` entry, which nothing reads
- normalize existing `/boot` and `/boot/efi` device sources to `PARTUUID=` at the start of an ordinary 0.4 update, before the boot files are touched; identity comes from the live mounts, each must share a disk with the installed root, and the ID must resolve uniquely back to that device; otherwise fstab is left as it is, the refusal is logged, and the update proceeds
- resolve `PARTUUID=` and other tagged fstab sources through their `/dev/disk/by-<tag>/` link, the one systemd mounts the entry through, so a duplicated ID names the partition actually in use instead of a two-line blkid result that matches nothing
- size block devices from `/sys/class/block/<name>/size` in one helper, replacing both the `blockdev` fork and a second sysfs reader; derive `-partN` names under every `/dev/disk/` directory; canonicalize the pool target before the pvscan lookup so `pvremove` runs again; keep `list_partitions` on the kernel nodes sysfs reports
- install the incoming updater before a CI-assisted deploy, matching the production update path; retain the running kernel while pruning stale boot files, and pin each built image's newest kernel as GRUB's top-level entry so ordinary updates select the target kernel
- document the installation identity contract and update the StartOS and CLI changelogs, the 0.4.0.2 release notes, and the CLI manpage
- format the Start SDK action example that the merged base had left failing the repository-wide Prettier gate

Mount failures remain fatal; this does not add `nofail`. `/dev/disk/by-path` identifies the selected device while it remains connected, so setup assumes selected drives are not unplugged or replaced before installation completes. `PARTUUID` is the persisted boot identity after installation.

The 0.3.5.1 migration does not normalize fstab. Its `update-grub2` step runs in the payload chroot on the 0.3.5.1 userland, which cannot see the EFI mount and whose `lsblk` lacks the columns a layout derivation needs, so a migrated server keeps its kernel-name fstab for its first 0.4 boot and is normalized on its first 0.4 update. The migration payload is unchanged and does not need republishing.

## Validation

- `cargo test -p start-core --features=test -- os_install:: disk::` — 19 passed, on a clean non-incremental build after each change
- `make start-os-scripts-test`
- Prettier on the changed YAML and Markdown
- `bash -n` and ShellCheck on the changed scripts
- focused x86_64 ISO and squashfs build: [Actions run 35148021501](https://github.com/Start9Labs/start-technologies/actions/runs/35148021501)
- final-head x86_64 ISO and squashfs build: [Actions run 35163177965](https://github.com/Start9Labs/start-technologies/actions/runs/35163177965)
- fresh ISO install on a BIOS VM: `setup.disk.list` returned `/dev/disk/by-path/pci-0000:04:00.0`; `/dev/vda` was rejected before partitioning; the stable path installed, completed setup, and rebooted with `PARTUUID=` for `/boot`, no root fstab entry, both encrypted data volumes mounted, active `startd`, and no failed units
- ordinary update from a released 0.4.0 installation using the final-head artifact: the incoming updater normalized fstab, retained the running 7.0.13 kernel as a fallback, pinned the target 7.1.8 kernel with `GRUB_TOP_LEVEL`, and rebooted into `8147047` on 7.1.8; `/boot` and both encrypted data volumes mounted, `startd` was active, and systemd had no failed units
- the migration payload's own `lsblk` (util-linux 2.38.1) run against a live disk, confirming `PKNAME` works there and `PARTN` does not

Closes #3923


shared-libs/crates/start-core/src/os_install/mod.rs
```diff
@@ -54,13 +54,68 @@ pub fn partition_for(disk: impl AsRef<Path>, idx: u32) -> PathBuf {
...
+    if disk_path.parent() == Some(Path::new("/dev/disk/by-path")) {
+        root.join(format!("{}-part{}", leaf, idx))
+    } else if leaf.ends_with(|c: char| c.is_ascii_digit()) {
         root.join(format!("{}p{}", leaf, idx))
     } else {
         root.join(format!("{}{}", leaf, idx))
     }
 }
 
+fn parse_partuuid(output: &[u8]) -> Result<&str, color_eyre::eyre::Error> {
```
dr-bonez: why are we using raw `eyre` errors?

helix-nine: It was only carrying an internal parse failure before conversion to the localized API error; there was no reason to expose `color_eyre::Error` from the parser. I removed it in `e26a5f249`: parsing now returns `Option`, and the call site produces `crate::Error` with `BlockDevice` while preserving the original command diagnostic for `blkid` failures. The focused installer tests, `cargo check`, and format check pass.

dr-bonez (dismissed):

helix-nine:
Reviewed the current head, including the maintainer follow-ups and the master merge. A high-effort inline pass over `git diff origin/master...HEAD` found no additional code findings.

The refreshed CI exposed one inherited formatting failure in `projects/start-sdk/docs/src/actions.md`; `f192a6417` applies the repository's pinned Prettier output. Verification on the resulting head passed:

- focused start-core disk/install tests — 19 passed
- `make start-os-scripts-test`
- `make start-core-ts-bindings-check`
- `make manpages-check`
- `npm run check`
- setup-wizard and UI production builds
- repository-wide Prettier check
- Bash syntax and ShellCheck on the changed normalizer/test scripts

The master merge landed after the approval and automatically dismissed it, so the PR now needs a fresh approving review after checks complete.

helix-nine:
End-to-end VM pass completed against the focused x86_64 build ([run 35148021501](https://github.com/Start9Labs/start-technologies/actions/runs/35148021501)).

Fresh BIOS install passed:

- `setup disk list` returned `/dev/disk/by-path/pci-0000:04:00.0`
- submitting `/dev/vda` failed with `Disk Management Error` and left the disk unpartitioned
- submitting the listed stable path installed the OS and data pool, completed setup, and rebooted
- the installed fstab contained only `PARTUUID=<boot> /boot` plus the commented EFI row; `/` had no entry
- `/boot`, main data, and package data all mounted; `startd` was active and systemd had no failed units

The OTA/rollback pass found two integration gaps and fixed them in `57b5423ea`:

- `update-from-gha.sh` used the installed updater instead of extracting the incoming one first, unlike production `do_update`; it now follows the production order, so pre-reboot migrations run on the first CI-assisted update
- rollback left the newer running kernel on `/boot`, so GRUB selected it against the older target rootfs; `upgrade` now stages the target boot tree, removes kernel sets newer than the target, then installs the target files while preserving older fallback kernels

Final verification passed in both directions:

- released 0.4.0.1 → PR image: incoming updater installed, normalized fstab retained, booted `8dcf9da`, `/boot` resolved to `/dev/vda2`, zero failed units
- PR image → released 0.4.0.1 (whose image has no normalizer): only the target 7.1.3 kernel remained, booted `fdb27c7`, normalized fstab mounted `/dev/vda2`, zero failed units

Local gates rerun on the pushed head: `make start-os-scripts-test`, Bash syntax, ShellCheck, and `git diff --check`.

helix-nine:
Retest of `745963eb7` against the actual production updater ordering: **rollback to 0.4.0.1 fails again**.

Reproduction on the fresh-install VM:

1. Boot the PR image on kernel 7.1.8.
2. Run the new `prune-boot`; `/boot` correctly retains only the running 7.1.8 kernel.
3. Upload released `startos-0.4.0.1-fdb27c7_x86_64.squashfs`.
4. Match `do_update` by extracting that target image's `/usr/lib/startos/scripts/upgrade` before invoking it.
5. The old updater extracts its 7.1.3 kernel alongside 7.1.8 and runs its own `update-grub`.

The released target has `GRUB_DEFAULT=0` and no `GRUB_TOP_LEVEL`. Its generated first/default entry is therefore 7.1.8:

```
Found linux image: /boot/vmlinuz-7.1.8+deb13-amd64
Found linux image: /boot/vmlinuz-7.1.3+deb13-amd64
menuentry 'StartOS v0.4.0.1 GNU/Linux, with Linux 7.1.8+deb13-amd64' ...
menuentry 'StartOS v0.4.0.1 GNU/Linux, with Linux 7.1.3+deb13-amd64' ...
```

After reboot, serial confirms `Linux version 7.1.8+deb13-amd64` with the 0.4.0.1 rootfs, then missing `/lib/modules/7.1.8+deb13-amd64/modules.devname`, a wait for the boot PARTUUID, and the locked-root emergency path. SSH did not return within four minutes.

`GRUB_TOP_LEVEL` fixes transitions to images built after `745963eb7`, but cannot select the kernel in a pre-pin rollback target. The earlier staged-kernel implementation passed this exact rollback because it removed 7.1.8 before the old target ran `update-grub`.

Maintainer decision: rollback to pre-`GRUB_TOP_LEVEL` images is outside this PR's supported scope, so this result does not block the change.

helix-nine:
Supported-path retest on `745963eb7` passes.

I used the final-head x86_64 squashfs from [run 35163177965](https://github.com/Start9Labs/start-technologies/actions/runs/35163177965) (`8147047` is GitHub's PR merge SHA; BLAKE3 `ee4e953e0977dc8a244c28a0d92e5caf2a6a6cc9c1715c63302ec00af18cc651`) and followed production `do_update` ordering on a released 0.4.0 VM:

- pruned first, uploaded the image, extracted the incoming updater, then invoked it
- fstab changed from `/dev/vda2` to `PARTUUID=501cab40-5f5c-45c7-b5f5-e862e5367168` on the first update
- `/boot` retained the running 7.0.13 fallback and received target kernel 7.1.8
- the target image carries `GRUB_TOP_LEVEL="/boot/vmlinuz-7.1.8+deb13-amd64"`
- reboot selected 7.1.8 and reached the final image successfully
- `/boot` resolved to `/dev/vda2`; main and package-data mounts were active; `startd` was active; systemd reported zero failed units

The earlier fresh-ISO installation remains valid because the disk-install paths have not moved since that artifact; this retest exercises the subsequent boot-selection cleanup directly. The pre-pin rollback result is documented separately and is non-blocking per the maintainer scope decision.

All 43 reported checks on the current head are complete and passing (or intentionally skipped); none are pending or failed. The PR now only needs a fresh approval after the maintainer pushes.

dr-bonez (approved):

PR #3934, opened 2026-09-12
fix(web): revalidate stable UI assets

### LINKED ISSUE #3765: StartOS UI is served entirely `immutable`, and the SPA fallback caches HTML under asset URLs — a stale UI can only be cleared by a hard refresh (static_server.rs:835, :191-193)
Split out of #3756, where the same class of bug was fixed for StartWRT. dr-bonez asked whether the strategy should apply to the StartOS UI. Two parts of it should; the rest is already covered.

All references are against `master` at `9bb382ac2`.

## 1. Every UI file is served `immutable`, including `index.html`

`shared-libs/crates/start-core/src/net/static_server.rs:832-835` applies one policy to every file that gets an ETag — it does not branch on filename, extension, or hashed-ness:

```rust
if let Some(e_tag) = &self.e_tag {
    builder = builder
        .header(ETAG, &**e_tag)
        .header(CACHE_CONTROL, "public, max-age=21000000, immutable");
}
```

21,000,000 s ≈ 243 days. That covers `index.html`, `manifest.webmanifest`, everything under `assets/**`, and the service-worker artifacts — not just the content-hashed bundles, which are the only files for which `immutable` is actually safe.

`If-None-Match` → 304 is already implemented (`static_server.rs:863-870`) and the ETag does move every build — it is `sha256(path + mtime)` (`:889-898`), with mtime baked in at compile time by `include_dir`. The revalidation machinery is all there. `immutable` is what makes it unreachable: it tells the browser not to ask.

The user-visible consequence is in `projects/start-os/web/ui/src/app/components/refresh-alert.component.ts:44-61`. Because a plain reload is not guaranteed to fetch the new bundle, the non-PWA branch of "Refresh Needed" cannot offer a Reload button and instead prints hard-refresh keystrokes per OS:

> Your user interface is cached and out of date. Hard refresh the page to get the latest UI.

## 2. The SPA fallback serves HTML under asset URLs, with that same immutable header

`static_server.rs:191-193`, no extension check:

```rust
let file = C::ui_dir()
    .get_file(uri_path)
    .or_else(|| C::ui_dir().get_file("index.html"));
```

`mime` and `e_tag` are then computed from the *resolved* path (`:665-681`). So a stale tab requesting `/main-OLDHASH.js` gets:

- `200 OK`
- the `index.html` body
- `Content-Type: text/html`
- `Cache-Control: public, max-age=21000000, immutable`

That is a poisoned cache entry under a chunk URL, held for 243 days, that a reload cannot clear. `not_found()` is only reachable when `index.html` itself is absent.

#3756 guards exactly this for StartWRT at `projects/start-wrt/backend/ctrl/src/embedded_web.rs:44-46` — extensionless paths are routes and get `index.html`; anything with an extension 404s.

## What is already fine — scope guard for whoever picks this up

**Detection does not need porting.** #3756 added a per-response build-stamp header and a 30 s heartbeat because StartWRT polls and has no push channel. StartOS already has one: `refresh-alert.component.ts:87-100` watches patch-db `serverInfo.version` over the websocket, so an open tab learns about an update when the post-restart dump lands. The comparison is semver (`ConfigService.version`, the root `package.json` baked into the bundle, vs `serverInfo.version`) with `!== 'equal'`, so downgrades fire too. It misses only a *same-version* rebuild — dev deploys and same-version re-flashes. `ServerInfo` has no `gitHash` field today, though `version::COMMIT_HASH` exists (`shared-libs/crates/start-core/src/version/mod.rs:703`) if that gap is ever worth closing. Not user-facing.

**The service worker limits the blast radius on `ui`, and this is the part to think about before changing headers.** ngsw is registered in production (`app.config.ts:111-116`, `environments/environment.prod.ts:3`), and Angular's worker is already defensive about the HTTP cache: it always cache-busts the manifest (`ngsw-worker.js:1986`, `ngsw.json?ngsw-cache-bust=…`) and, for hashed assets, SHA-1s the body and re-fetches with a cache-bust on mismatch (`ngsw-worker.js:562-580`). So `immutable` is largely neutralized for SW-controlled clients.

The exposure is where the SW isn't:

- the window before `registrationStrategy: 'registerWhenStable:30000'` fires,
- clients where the SW is unregistered or unavailable,
- **`setup-wizard`, which has no service worker at all** — `angular.json` wires `serviceWorker` only for the `ui` project — yet is served by the same handler under the same immutable header.

Also worth noting: nothing subscribes to `SwUpdate.versionUpdates` or `unrecoverable`. `activateUpdate()` inside `pwaReload()` is the only service-worker API the app uses, so the SW's own update signal is unused and the OS version diff is the sole detector.

## Suggested fix

1. Split the cache policy in `static_server.rs`: keep `public, max-age=31536000, immutable` for content-hashed bundles, use `no-cache` + ETag for stable-named files (`index.html`, `manifest.webmanifest`, `assets/**`) so they revalidate into a cheap 304. `embedded_web.rs:52-65` in #3756 is the worked version.
2. Extension-guard the SPA fallback so a stale hashed chunk 404s instead of receiving an HTML body.
3. With (1) in place, replace the non-PWA hard-refresh instructions in `RefreshAlertComponent` with a Reload button.

(1) and (2) are independent — (2) is a defect on its own terms and could land alone.

## What I have not shown

I have not demonstrated that either defect produces the symptom the PWA branch of the dialog describes ("If you continue to see this message, uninstall and reinstall the PWA"). Both are wrong on the code's own terms; whether they are *the* cause of that report needs a repro. Item (3) in particular should be benched against a real PWA install, since the service-worker interaction is the part I have read carefully but not run.

helix-nine:
## Summary

- Split the UI cache policy: files the production Angular build declares in `immutable-assets.txt` are served `public, max-age=31536000, immutable`; everything else (`index.html`, `assets/**`, the service worker) is `no-cache` and revalidates against an ETag. The manifest is generated from the build's esbuild metafile (`--stats-json`), which lists only content-hashed outputs; a development build declares none.
- The SPA fallback serves `index.html` for extensionless routes only, so a stale hashed chunk gets a 404 instead of a cached HTML body.
- ETags differ per content encoding and every static response carries `Vary: Accept-Encoding`. A ranged request always serves identity bytes; an unparsable or unsatisfiable Range is ignored.
- Signed S9PK file hashes are exposed as an encoding-independent `File-Digest` (the BLAKE3 of the decoded file) on full, ranged, gzip, and HEAD responses, and HEAD no longer reads the archive. The s9pk routes answer GET and HEAD only.
- The Refresh Needed dialog has one Refresh button in both browser and PWA modes: it checks for and activates a waiting service-worker update, then reloads.
- StartWRT reuses the manifest lookup and route predicate from start-core.
- StartOS changelog and docs; StartTunnel 1.3.1 changelog, docs, and release notes.

## Verification

- `cargo test -p start-core net::static_server --features=test` — 9 passed
- `cargo check -p startwrt-core`
- `cargo +nightly-2026-05-28 fmt --all --check`, prettier on every touched file
- `npm run check:i18n`, `npm run check:i18n:tunnel`, `npm run check:ui`
- `npm run build:tunnel`: the production manifest holds 34 entries, every one a `chunk-*`/`main-*`/`polyfills-*`/`styles-*` hashed bundle; `stats.json` is removed; `index.html`, `assets/**` and `3rdpartylicenses.txt` are absent
- production Angular service-worker A/B (helix-nine): a plain reload retained the old main bundle, while `checkForUpdate()` + `activateUpdate()` followed by one reload loaded the new bundle

Closes #3765


shared-libs/crates/start-core/src/net/static_server.rs
```diff
@@ -194,7 +199,34 @@ pub fn rpc_router<C: Context + Clone + AsRef<RpcContinuations>>(
         )
 }
 
-fn serve_ui<C: UiContext>(req: Request) -> Result<Response, Error> {
+fn is_content_hashed(path: &Path) -> bool {
```
dr-bonez: I'd prefer the frontend build self-declare what files to cache immutably. this seems like a clunky heuristic.

shared-libs/crates/start-core/src/net/static_server.rs
```diff
@@ -194,7 +198,36 @@ pub fn rpc_router<C: Context + Clone + AsRef<RpcContinuations>>(
         )
 }
 
-fn serve_ui<C: UiContext>(req: Request) -> Result<Response, Error> {
+/// Matches top-level Angular bundles whose filenames carry content hashes.
+pub fn is_ui_content_hashed(path: &Path) -> bool {
```
dr-bonez: I'd prefer the frontend build self-declare what files to cache immutably. this seems like a clunky heuristic.

helix-nine: Implemented at 8ff967e5a. Each Angular production build now writes a small `immutable-assets.txt` containing the exact physical outputs declared by its build stats, then deletes `stats.json`. The StartOS and StartWRT servers exact-match requested identity paths against that file; missing, empty, or invalid manifests fail closed to `no-cache`, and development builds declare no immutable assets. The generator’s boundary tests now run through the CI `make test` target. I rebuilt all four frontends and verified their manifests plus the focused and full Rust suites.

helix-nine:
VM-tested head `8e3996449` using the x86_64 image from [run 35263416331](https://github.com/Start9Labs/start-technologies/actions/runs/35263416331). I upgraded a `0.4.0-beta.10` VM to this build, rebooted into `0.4.0.2` / kernel `7.1.8`, and confirmed the installed FileBrowser package survived.

Passed live checks:

- `/` and `assets/img/icon.png`: `Cache-Control: no-cache`, ETag, and `Vary: Accept-Encoding`.
- `main-366FZ54A.js`: `public, max-age=31536000, immutable`; it is listed exactly in the generated 74-entry `immutable-assets.txt`, while `index.html` and `assets/**` are absent.
- Extensionless `/services/filebrowser/about` falls back to `index.html`; missing `.js`, nested `.svg`, and `.hidden` paths return 404.
- `If-None-Match: "stale", <current-etag>` returns 304 with no body. Root and main-bundle ETags remained byte-identical across a `startd` restart.
- `Range: bytes=0-15` plus `Accept-Encoding: gzip` returns a 16-byte identity 206 with the correct `Content-Range`; an unsatisfiable range is ignored and returns the full 200. HEAD returns no body with the full length; POST returns 405.
- Through the real authenticated UI, opening FileBrowser's license fetched `/s9pk/installed/filebrowser.s9pk/LICENSE.md` as gzip with `File-Digest`. The advertised decoded-file BLAKE3 and an independent `b3sum` of Playwright's decoded 11,355-byte response were identical (`79a6b58a…8c066a`), and the frontend accepted it without a digest-mismatch error.
- Forced a live `serverInfo.version` mismatch to exercise the new non-PWA refresh dialog: exactly one **Refresh** button, no **Ok** button, and clicking it caused a top-level navigation back to `/services`.

The workflow-dispatch image used `ENVIRONMENT=dev`; Chromium therefore had no service-worker registration, so this VM pass covers the non-PWA refresh branch. The production service-worker A/B remains the separate verification recorded in the PR body.

Test setup caveat unrelated to this diff: `scripts/update-from-gha.sh` staged and installed the image but did not reboot despite saying it would; I issued the reboot manually.

helix-nine:
Merged current `master` (`018535ac8`) into this branch and resolved the sole conflict in `package.json`. The immutable-asset generation remains on the normal StartWRT build, while master's Pages demo build now calls that same build before creating `404.html`; master's brochure `404.html` generation is also retained.

Verified:

- `npm run build:deps`
- `npm run build:wrt:demo` (manifest generated, stats removed, `404.html` matches `index.html`)
- UI, StartWRT, StartTunnel, and brochure typechecks
- all three web i18n checks
- full Prettier check
- `cargo test -p start-core net::static_server --features=test` — 9 passed

dr-bonez (approved):

Issue #3466, opened 2026-07-11
StartWRT WiFi settings: regulatory-domain-aware channel selection, channel-width selector, band-aware fields


helix-nine:
## Summary

The WiFi settings page presents channel options from **hardcoded, US-centric arrays** with no notion of the radio's actual capabilities or the active regulatory domain. Rework the page so channel options are **derived from a user-selectable regulatory jurisdiction** (plus what the radio supports), add a **channel-width** selector, and only show the fields relevant to the currently-selected band.

## Current behavior

In `web/src/app/routes/wifi/routes/settings/index.ts`:

- `channels24 = ['Auto', '1' … '11']` and `channels5 = ['Auto', '36' … '165']` are static literals rendered straight into the dropdowns. There's an existing `@TODO: Implement channel optimization (requires backend channel scan endpoint)` right above the fields.
- Both the 2.4 GHz and 5 GHz channel fields render **unconditionally**, regardless of the selected Frequency Band (`'2.4 GHz' | '5 GHz' | 'Both'`).
- There is **no channel-width control** — width (`htmode`) is not modeled anywhere.
- There is **no regulatory-jurisdiction selector** in the StartWRT UI at all. (The only country-code handling in the tree is StartOS's *client-side* wpa_supplicant flow in `start-os/core/src/net/wifi.rs`, which is a separate concern from the AP's reg domain.)

Backend (`backend/ctrl/src/wifi.rs`, `backend/uciedit/src/openwrt.rs`) reads radios purely from UCI: `WifiRadio { band, channel, enabled, broadcast }` and `WifiDevice { device_type, path, disabled, band, channel }`. There's no `country`/`htmode` field and no path that queries the hardware for permitted channels.

**Consequences:** the static list is wrong in both directions — it's **missing** valid channels (5 GHz UNII-4 `169/173/177`; 2.4 GHz `12/13/14` outside the US) and it **offers** DFS/US channels a non-US radio/reg-domain will refuse. No single hand-maintained array can be correct because the valid set is `radio capabilities ∩ active regulatory domain`.

Note (for scoping, not a bug): even-numbered "channels" like `142` are **center-frequency indices** for wide (40/80/160 MHz) channels, not primary/control channels — `142` is the 40 MHz center of the `140`+`144` bond. LuCI omits them for the same reason. Width is a *separate* axis from the primary channel, which is why this issue splits them into two controls rather than flattening center indices into the channel list.

## Proposed changes

[...]

PR #3939, opened 2026-09-14
feat(start-wrt): set the Wi-Fi regulatory country from the UI

MattDHill:
## Summary

StartWRT never wrote `option country`, so every router — US or otherwise — runs on the kernel's world regulatory domain: 2.4 GHz channels 1–11 and 5 GHz channels 36–48, all at 20 dBm. The Wi-Fi settings page nonetheless offered 5 GHz channels 52–165, which the world domain forbids for an AP; picking one left that radio down with nothing in the UI to say why.

**Backend**
- `WifiConfig` gains `country: Option<String>` (ISO 3166-1 alpha-2). `wifi.set` writes it to every `wifi-device`, validates it against the country records in `/lib/firmware/regulatory.db`, restarts the radios, and waits for `iw reg get` to report the new domain before returning, so a follow-up read sees it. Clearing it runs `iw reg set 00`, since netifd only ever sets a domain and never unsets one.
- New `wifi.regulatory`: the codes the regulatory database defines (what `wifi.set` accepts) and, per band, the channels an AP may currently use — parsed from `iw phy`, skipping `disabled` and `no IR` frequencies. iwinfo's `freqlist` was not usable for this: it masks NO-IR on radar channels, so under the world domain it would report 52–144 as usable.
- The UCI writer leaves options it doesn't model untouched, so a `country` set by hand over SSH already survived UI edits; it is now read back and shown.

**Web**
- A Country combobox at the top of Wi-Fi > Settings (searchable, names from `Intl.DisplayNames` in the UI language), with a warning while unset explaining the worldwide subset.
- Channel dropdowns list only what the current domain permits. Changing the country resets both to Auto — the previous selection may not be legal under the new domain, and the new list isn't known until the domain is applied.
- Help Mode and the five dictionaries carry the new strings.

**Docs / release**: `wifi.md` documents the setting and drops the fixed "36–165" channel claim; `initial-setup.md` points at it from Next Steps; CHANGELOG `## [1.2.0]`, manifest bump, `release-notes/1.2.0.md`.

Verified locally: `cargo check -p startwrt-core --tests`, `cargo test -p uciedit`, `npm run check:wrt`, `check:i18n:wrt`, prettier and rustfmt clean. The two new parser unit tests run in CI.


Dominion5254:
Bench-tested on a K1 and pushed three commits on top. Two are behavior fixes found on hardware.

First, typing in the Country combobox was wiped by the settings poll: a strict combobox holding null stays pristine while you type, so the five-second poll reset the form and Taiga's writeValue cleared the search text. The control now holds 00, the kernel's world domain code, shown as "Not set", the way Timezone holds UTC, so typing dirties the form and the poll leaves it alone. Null still travels on the wire, and the warning banner keys on 00.

Second, with a country set, automatic channel selection picked DFS channel 100 on its own and the 5 GHz radio was silent for the 62-second radar check on every boot and full restart. That cannot happen today because the world domain marks DFS channels no-transmit, so it is a regression the feature introduces for anyone who sets a country and leaves Auto. The backend now writes acs_exclude_dfs '1' on every managed radio, keeping Auto on non-DFS channels while a channel picked by hand still works. A test covers the country write, the clear, and the new option.

The docs follow those two: the help text in all five languages, the user guide's Channel bullet, the 1.2.0 changelog entry and the release notes say that Auto skips radar-detection channels and that picking one by hand costs a minute or more at each start. The third commit is release tidying: versions.conf to 1.2.x so the tag publishes the right book, a warning when the domain wait times out, the InvalidValue and upper-case rules for country in the contract, and the stale channel-scan TODO removed. This PR is part of #3466 rather than closing it, since the width selector, band-gated fields and manufacturing default remain.

dr-bonez (approved):

PR #3940, opened 2026-09-15
refactor(net): simplify forwarding ownership and close cleanup gaps

helix-nine:
## Scope

Stacked on #3932, targeting `fix/local-forward-private-subnets` at `8ce6a4aa94a2076f75620990c4ea8789824e19ca`. That branch is unchanged.

This implements the approved correctness-first follow-up. There are real ownership/plumbing reductions, **not an overall line-count reduction**: the incremental diff is +4,257/−1,051 across 33 files, including regression tests and a disposable-VM harness.

## Simplifications

- Make `HostBinds` lease-only; remove its duplicate forwarding/proxy specifications while preserving acknowledgment-before-replacement and request-time prefix fallback.
- Use named, owned `FuturesUnordered` task collections with abort handles instead of parallel optional-handle bookkeeping.
- Reuse `GeneralGuard` for subprocess groups, consolidate per-family forwarding command construction, and reuse IPv6 reconciliation during drain.
- Remove asynchronous forwarding-state Drop cleanup, the initialization auxiliary queue, and the singleton forwarding wrapper. Keep controller layers whose acknowledgment contracts differ.
- Correct adjacent architecture descriptions and remove redundant comments in the touched code.

## Correctness repairs

- Retain exact attempted IPv4/IPv6 cleanup identities before installation awaits. Failed or timed-out commands remain unconfirmed and unadvertised; reuse and overlapping replacements wait for successful retirement.
- Give terminal drain one cancellation-resistant completion owner and the original absolute deadline. Close admission, finish admitted work, and abort/join outstanding owned workers before returning an incomplete-cleanup error. Preserve the IPv6 initialization barrier and prefix-sensitive reconciliation.
- Bound subprocess stdin/output/wait as one operation. Kill inherited process-group descendants and reap the immediate child on timeout.
- Retain rejected router grants until withdrawal succeeds, renew UPnP hostname mappings through their retained gateway, reserve footprints across shards, and validate actual initial/renewed grants. Preserve supported same-client TCP fallback/SNI sharing and try independent gateway candidates when a candidate is occupied.
- Preserve range counts and offsets through repeated toggles. Serialize manual IPv4 add/remove; repeated disables retry cleanup. Atomically remove exact-tag IPv6 rules before deleting their database/lease state.
- Record automatic pinhole expiry before applying rules and recheck expired SNI ownership under the mutation lock, protecting renewed leases and manual routes.
- Propagate PCP/UPnP removal failures, retaining DNAT entries and leases for retry. Adapt StartWRT's shared backend, including configuration-error reporting; prepare its next patch version **1.1.2**, changelog, book, and release notes. Its firewall reload remains asynchronous.

Preserved: IPv4 Local on-link/RFC1918/loopback/link-local access, public forwarding and IPv6 Local policy, exact original/translated DNAT tuples, stale-handle retries, and producer shutdown ordering. **No systemd unit changes:** `startd.service` receives no fixed stop timeout; `start-tunneld.service` retains its existing 120 seconds.

## Verification

- `make start-core-test`: **572 passed, 4 ignored**; canonical target skips generated `export_` tests.
- `make start-wrt-test`: **588 passed**, including a rerun after the version bump.
- Disposable StartOS VM: explicitly ran the ignored real nftables/WireGuard API test. Covered range offsets, repeated failed disables, PCP/IGD deletion failures, initial pinhole failure/expiry, failed pinhole removal and retry, SNI renewal/reaper races, concurrent manual add/remove, and final rule withdrawal.
- Locked Cargo checks: `start-core`, `start-os`, `start-tunnel`, and `startwrt-core`/`startwrt`.
- Real StartOS UI/setup-wizard, StartTunnel UI, and StartWRT UI builds.
- `make format-check`, `git diff --check`, forwarding shell regression suite, Bash syntax, ShellCheck on forwarding scripts and VM runners, Python wrapper syntax.
- StartOS, StartTunnel, and StartWRT mdBook builds.
- Independent targeted verification of the final uncertain-install, router-sharing/actual-grant, and protocol/lease corrections found no remaining actionable defect.

### Limits and observed test behavior

- Withdrawal deadlines bound cleanup effort, not external router cooperation: an unreachable router can keep a grant until its lifetime expires. Process-group termination covers ordinary descendants, not deliberate session escape or portable reaping of non-child grandchildren.
- ShellCheck on the unchanged core test runner reports the same baseline SC1091/SC2034/SC2001/SC2086 diagnostics; forwarding-script checks pass.
- Earlier runs exposed an overly short simulated-script fixture deadline, corrected before the final passing runs. One earlier run also hit the existing SNI `bind_failure_refuses_grant_and_rolls_back` fixture's intermittent port-binding failure; subsequent full runs passed.
- No physical-router, macOS, full OS-image, or OpenWrt-image run was performed. Networking experiments ran only in the disposable VM.


dr-bonez:
Closing with #3932; the replacements are #3962, #3963 and #3964.

Beyond the reasons given on #3932: this stack is +488 non-test lines over its base and not a simplification, and it reintroduces the regression the base explicitly avoided. `RpcContext::shutdown` now returns the forwarding-drain error when services stop cleanly, `inner_main` fails on `cleanup?`, and a requested reboot lands in Diagnostic Mode instead of executing; a router that ignores PCP deletes or a hung `nft` is enough to hit the 90 s deadline. `forward.rs` also replaces `t!("net.forward.error-initializing-controller")` with an English literal, orphaning the key in all five dictionaries. The StartWRT error-code change would need its own PR if wanted.

PR #3958, opened 2026-09-15
fix(core): preserve file modes in dir_copy

chrisguida:
dir_copy_inner restored the mode and owner of every directory it created but only chowned regular files, so each file was created through File::create and kept the umask default of 0644 regardless of the source mode.

The setup wizard's Transfer copies the previous drive's main/ tree with dir_copy, and main/system-connections is bind-mounted over /etc/NetworkManager/system-connections. After a transfer every keyfile there was 0644; NetworkManager rejects keyfiles that are not 0600 ("File permissions (100644) are insecure"), so an imported WireGuard gateway such as StartTunnel stayed disconnected until someone fixed the mode by hand.

Create each file with the source mode and apply set_permissions after the chown, since chown can clear setuid/setgid and the umask can narrow the creation mode. The same path copies the LUKS folder during backups.


MattDHill:
@helix-nine take a look

helix-nine (approved):
Reviewed `git diff origin/master...HEAD` (two files) at high effort. The implementation creates files with the source mode and restores permissions after `chown`, covering both the transfer and backup callers.

I pushed `ea8df010b` to exercise the two ordering hazards in the regression test: umask-narrowed `0666` and chown-cleared `06755`. It also reduces the ordering comment to the underlying constraint.

Verified with:
- `cargo test -p start-core dir_copy --features=test`
- `git diff --check`

No remaining findings.

dr-bonez (approved):

dr-bonez:
@chrisguida please set up GPG signing for future contributions

chrisguida:
@dr-bonez I actually have it set up, just don't have it on automatically haha. Will do it for the next one

PR #3965, opened 2026-09-16
feat(start-os): restart the server when it hangs

MattDHill:
## Summary

A DIY box (ThinkCentre M90q Gen 6, Samsung 990 Pro) went unreachable every one to three days with the kernel and NIC still up: the NVMe had dropped off the bus from its deepest APST idle state, the root squashfs became unreadable, and the box sat there until someone power-cycled it. A second occurrence was a full hard lockup. StartOS neither noticed, recovered from, nor recorded either. Disabling APST is what fixed it.

- **Hardware watchdog.** `RuntimeWatchdogSec=60` in a `system.conf.d` drop-in; PID 1 pets it, so a kernel lockup or a wedged init resets the box within a minute. Where there is no `/dev/watchdog` (VMs without one) systemd logs and moves on.
- **OS-drive watch in startd.** Every 30 s startd reads one sector of the OS root partition with `O_DIRECT` (the mount source of `/media/startos/root`, so the check is skipped on the live installer). After four consecutive failures it writes the reason to `/dev/kmsg` and triggers a sysrq crash; with `kernel.panic=10` that reboots in seconds and leaves a pstore record where the firmware keeps one. This is the only thing that fires in the dropped-NVMe shape, where the kernel and PID 1 are healthy. `kernel.sysrq=1` admits the crash key (Debian's default mask allows sync and reboot but not crash).
- **`hardlockup_panic=1`** turns a silent hard lockup into a panic, for the same record and reboot. Prefixed `-` so kernels without the detector ignore it.
- **Crash records survive.** systemd-pstore archives into `/var/lib/systemd/pstore`, which sits on the root overlay's tmpfs, so startd copies the archive onto the data drive (`<main>/logs/pstore`) at init. `gather-debug-info` now collects it, plus the previous boot's journal, the kernel log, the cmdline and each NVMe drive's APST state — everything this month-long thread had to be asked for by hand.
- **NVMe APST off on every platform** via `nvme_core.default_ps_max_latency_us=0` on the kernel cmdline (postinst, the 0.3.5.1 migration grub, and the Pi's `grub.d` override, which replaces `GRUB_CMDLINE_LINUX` outright). The cmdline rather than `modprobe.d` because the Pi kernel builds `nvme` in. Matches the kernel's own `NVME_QUIRK_NO_APST` treatment of the drives it already knows about; an idle NVMe draws about a watt more.
- Changelog, release-notes highlight, FAQ entry "My server restarted on its own" (what the restart means, where the evidence is, the BIOS settings to check), and two lines in the physical pre-ship checklist so shipped SKUs verify the watchdog arms and holds.

## Notes for review

- The new Rust module was type-checked in isolation against tokio/libc/tracing and the `O_DIRECT` aligned read exercised against a real file; the crate itself was not built locally — CI is the compile check. `libc::O_DIRECT` does not exist on darwin, so the module and its spawn are `cfg(target_os = "linux")`.
- The one thing to watch on alpha hardware: a watchdog driver that fails to register pets would reboot the box every minute. `iTCO_wdt`, `sp5100_tco` and `bcm2835_wdt` have been stable for years, and the pre-ship checklist now checks for exactly this, but it is the failure mode this change introduces.
- A hand-applied APST fix does not survive an OS update — `upgrade` regenerates grub from the pristine new image and ships its initramfs — which is why this lands as the default rather than as documentation. The 0.4.0.1 workaround and its caveat are an advisory in `Start9Labs/advisories`, to be retired when this ships.


MattDHill:
On the concern that disabling APST by default can cause problems — what I can and can't find:

**The documented downside is heat and power.** With `default_ps_max_latency_us=0` the drive never leaves PS0. Idle draw goes from ~5–50 mW to ~1–2 W depending on the controller, and the drive runs roughly 10–15 °C hotter. In a fanless mini PC or a Pi 5 with an M.2 HAT, a drive that already idles at 55 °C can reach its throttle range (~70–80 °C), where it stalls I/O and ages the NAND faster. It also keeps the CPU package out of its deep C-states, so the box idles a few watts higher.

**I can't find a documented case of a drive misbehaving because APST was disabled.** The kernel's quirk table only ever restricts APST (`NVME_QUIRK_NO_DEEPEST_PS`, `NVME_QUIRK_NO_APST`); nothing forces it on. Drop-offs and `controller is down; will reset: CSTS=0xffffffff` are the failure mode of APST *on*. If there is a specific drive or platform that breaks with it off, that's the case to name here.

**Fleet-wide risk** is real regardless: this changes a kernel power default for every install on the strength of one platform.

**Middle ground:** cap the latency instead of zeroing it — `nvme_core.default_ps_max_latency_us=5500`. That is what the kernel itself does for drives it knows are bad (`NO_DEEPEST_PS`): only the deepest state is dropped. On the reporter's 990 Pro, PS4 has a ~38 ms exit latency and goes; PS3 (~1 ms, ~40 mW) stays, so the thermal argument mostly disappears. The catch: the reporter turned APST off entirely, so PS4 alone is not proven to be the culprit — though that is where nearly every reported drop-off lives (PHY powered down in PS4 + PCIe L1.2, host fails to re-train the link).

**What would settle it:** have the reporter run `5500` for two weeks. Holds → propose `5500` as the default here. Doesn't → it needs `0`, and a per-model udev rule (set the module parameter, then `reset_controller`) is the alternative to a fleet default.

**Open question:** what drives and cooling the Server Pure / Server One ship with. If either is fanless with NVMe, that's the box to measure before this merges in either form.

I'll switch this PR to `5500` if that's the preferred shape.

PR #3966, opened 2026-09-16
docs(start-os): FAQ entry for a server that hangs after hours or days

MattDHill:
## Summary

A new FAQ entry, **My server becomes unreachable after hours or days**, for boxes on 0.4.0.1 today. It comes out of a community thread where a ThinkCentre M90q Gen 6 with a Samsung 990 Pro dropped off the network every one to three days with the power and link lights on. A monitor showed `SQUASHFS error` lines: the NVMe had gone to its deepest APST idle state and never woke, so StartOS could no longer read its own root image. BIOS ASPM/C-state changes stretched the interval; disabling APST on the kernel command line stopped it.

The entry gives the symptom, the console tell, the BIOS steps, and the by-hand kernel setting over SSH via `/etc/default/grub.d` (sourced by `grub-mkconfig` after `/etc/default/grub`; named to sort after the Pi's `raspberrypi.cfg`, which replaces `GRUB_CMDLINE_LINUX` outright). `update-grub` writes the boot partition, so the setting survives reboots — but an OS update regenerates it from the new image, so the entry tells the reader to re-check `/proc/cmdline` after updating.

#3965 makes the setting the default and adds self-recovery on master; this entry stays true after that ships (the re-check simply finds the setting already there), and that PR adds its own, separately placed, entry for the new behavior.

Book builds with mdBook 0.5.2; anchor `#my-server-becomes-unreachable-after-hours-or-days` is what #3965 links to.


MattDHill:
Moving this to an advisory in Start9Labs/advisories: the by-hand APST step is 0.4.0.1-bound and #3965 makes it the default, so it does not belong in the published book. #3965's own FAQ entry carries the permanent part (what the restart means and the BIOS settings).

PR #3975, opened 2026-09-17
fix(cli): take a binding's --address as the UI shows it

### LINKED ISSUE #3952: host binding set-address-enabled --address only accepts a JSON HostnameInfo, not an address — the arg is CliFromJsonString<HostnameInfo> and the parse error names neither (net/host/binding.rs:691)
### Environment

start-cli 2.0.0 (`git-info` `b27b5331`) against StartOS 0.4.0.2 (`a30d84d2`), x86_64; same code on `master` @ `81c0b5a08`. Failing invocation: `start-cli package host stalwart binding mail set-address-enabled --address 64.23.194.12 --enabled true 25`.

### What happens

The `--address` argument of `package host <pkg> binding <host> set-address-enabled` is documented by `--help` as "Network address", but it only accepts the JSON encoding of a whole `HostnameInfo` record:

```
$ start-cli package host stalwart binding mail set-address-enabled --address 64.23.194.12 --enabled true 25
error: Deserialization Error: invalid type: floating point `64.23`, expected struct HostnameInfo at line 1 column 5
```

To enable a public IPv4 from the CLI I had to dump the DB, copy the matching entry out of `bindings.<port>.addresses.available`, and pass it back verbatim:

```
--address '{"ssl":false,"public":true,"hostname":"64.23.194.12","port":2525,"metadata":{"kind":"ipv4","gateway":"wg0"}}'
```

Nothing else about the CLI works this way (`domain public add` takes an FQDN and a gateway id), and the error message — a serde parse error on `64.23` — does not say what shape is expected.

`--address` accepts the address as the UI shows it — `64.23.194.12`, `demo.local`, `mail.example.com`, an IPv6 literal — and the command resolves it against the binding's `addresses.available` (erroring with the candidate list if it is ambiguous or unknown). Accepting the JSON form as well is fine for scripts.

### Steps to reproduce

Any installed package with a raw-port binding: `start-cli package host <pkg> binding <host> set-address-enabled --address <lan ip> --enabled true <internal port>`.

`shared-libs/crates/start-core/src/net/host/binding.rs:686-694`: `BindingSetAddressEnabledParams.address` is `CliFromJsonString<HostnameInfo>`, so clap hands the raw string to serde as JSON. `set_address_enabled_on` (`:897`) already matches by the full record; a lookup by `hostname` (plus `port` where one hostname is served on several) over `bind.addresses.available` would give the CLI a usable argument. `set-range-address-enabled` and `set-gua-wan` should be checked for the same shape.

### Logs & evidence

Transcripts above, 2026-09-15.

MattDHill:
## Summary

`package host <pkg> binding <host> set-address-enabled --address` only accepted the JSON encoding of a whole `HostnameInfo` record, and its parse error (`invalid type: floating point \`64.23\`, expected struct HostnameInfo`) named neither that nor any usable form. `set-range-address-enabled` and `set-gua-wan` shared the shape, under `server host` as well.

- The three server handlers drop their clap side (`.no_cli()`, `address: HostnameInfo`); the RPC shape and the generated TS bindings are unchanged.
- Each gets a CLI-side handler (`from_fn_async_local`, the `cli_install` / `cli_attach` pattern) that takes `--address` as a string, reads just the binding's `addresses.available` set through `db.dump` at the pointer `HostApiKind::host_pointer` derives for the package or server host, and matches it by hostname — `:port` (or `[v6]:port`) is the tiebreak where one address serves several ports. Several matches prompt for one; none is refused with the binding's addresses listed, each row translated. A `{…}` argument is still parsed as `HostnameInfo` for scripts.
- Because resolution is client-side, a 2.1.0 CLI does this against every 0.4.x server.
- `help.arg.address` rewritten in the five locales; the candidate-row labels and two headings are new `net.host.binding.*` strings; the six affected man pages updated.

```
$ start-cli package host stalwart binding mail set-address-enabled --address 64.23.194.12 --enabled true 25
$ start-cli package host stalwart binding mail set-address-enabled --address 10.0.0.5 25
10.0.0.5 is not an address of binding 25. Its addresses are:
  192.168.8.170:2525  IPv4 via eth0, private
  64.23.194.12:2525  IPv4 via wg0, public
  demo.local:2525  mDNS via eth0, private
```

Closes #3952


MattDHill:
@helix-nine review

shared-libs/crates/start-core/src/net/host/binding.rs
```diff
@@ -1139,6 +1173,216 @@ pub async fn set_gua_wan<Kind: HostApiKind>(
...
+    address.port.unwrap_or(if address.ssl { 443 } else { 80 })
+}
+
+fn describe(address: &HostnameInfo) -> String {
+    let host = if address.hostname.parse::<Ipv6Addr>().is_ok() {
+        format!("[{}]", address.hostname)
+    } else {
+        address.hostname.to_string()
+    };
+    let kind = match &address.metadata {
```
helix-nine: These labels, plus `via`, `public`/`private`, and `ssl` below, are new user-facing CLI text but remain hard-coded in English. With a non-English locale the translated prompt/error heading is followed by English candidate rows. Please route the row (or each label) through `i18n.yaml` for all five locales.

helix-nine (changes requested):
The address resolution and RPC handoff work, but the new candidate output introduces an untranslated user-facing path. I verified the binding test module, regenerated the start-cli man pages with no drift, checked both changed help trees, and exercised the CLI against a mock RPC server to confirm the resolved `HostnameInfo` reaches the original method.

shared-libs/crates/start-core/src/net/host/binding.rs
MattDHill: Done in ac9dd8527 — kind, `via`, `public`/`private` and `SSL` each come from `i18n.yaml` (five locales) now; the row layout stays in code.

helix-nine: Confirmed on `ac9dd8527`: the binding test module passes, the fix compiles in `start-cli`, and the German unknown-address path renders the complete candidate row in German (`private Domain über eth0, privat, SSL`). Resolved.

helix-nine (dismissed):
Re-reviewed `ac9dd8527`. The requested candidate-row labels are localized in all five locales; targeted binding tests pass and a German CLI smoke test confirms the rendered row is localized.

dr-bonez (commented):
Dumping all of public seems excessive

helix-nine:
Changed in `e5e546082`: the CLI now dumps only the selected binding’s `addresses.available` subtree (for package bindings, range bindings, and server bindings) rather than `/public`. I verified the three generated pointers with a mock RPC server, ran `cargo test -p start-core net::host::binding::test --features=test` (22 passed), and rebuilt `start-cli`.

dr-bonez (approved):

Issue #3819, opened 2026-08-25
start-cli: print a service's plaintext bridge URL by package id


MattDHill:
### Problem & use case

A user wiring an ad-hoc integration between two installed packages — a Cal.diy webhook posting to n8n on the same box, in the case that prompted this — needs the target service's **plaintext bridge address**, `10.0.3.1:<assignedPort>`. That address works today with no dependency declaration and no clearnet/Tor exposure (see #3782), but there is no ergonomic way for a human to find the port.

Package authors don't have this problem: `sdk.host.getBridgeAddress` resolves it, correctly and reactively, and heals when the port moves. That API isn't reachable by a person pasting a URL into a webhook field.

What's left for a user today is reading it out of the database or the forward table:

```bash
start-cli db dump --format json | jq '.value.packageData["n8n"].hosts[].bindings[].net'
start-cli net forward dump-table   # find the row whose target port is the service's internal port
```

Both work, but the first requires knowing the db shape and that `assignedPort` — not `assignedSslPort` — is the one you want, and the second requires already knowing the target's internal port in order to identify its row. Neither names the service.

We've decided **not** to surface bridge addresses in the web UI (see #3782): for the overwhelming majority of users a `10.0.3.1:<port>` row in the Interfaces tab is noise, and the predictable outcome is people pasting it into a browser and filing bugs when nothing loads. A CLI-only affordance closes the discoverability gap without adding a surface a non-technical user can trip over.

### Proposed solution

Original suggestion from @7thgenerationdesign in #3782:

> a small convenience like a `start-cli` command that prints a given service's plaintext bridge URL by name (something like `start-cli net bridge-url n8n`)

Shape, scope, and naming are @dr-bonez's call — he owns the CLI. Things worth deciding:

- Which binding to print when a package has several hosts or bindings, and how to select one (host id? internal port? print them all?).
- Whether to print a bare `host:port`, a scheme-qualified URL, or both — and what to do for a binding that carries only an `assignedSslPort` (`secure: { ssl: true }`, container serves its own TLS).
- Whether it belongs under `net` at all, vs. `package`.

### Alternatives considered

- **Surface it in the web UI.** Declined in #3782, for the reasons above.
- **Leave it at `db dump` / `net forward dump-table`.** Works, but requires knowing the db shape or the target's internal port — which is the gap being reported.
- **`.startos` DNS names.** Deprecated and being removed, and they resolve to the container IP plus internal port, which is a different and less stable addressing model.

[...]

PR #3977, opened 2026-09-17
feat(cli): completions, log in on demand, bridge address in binding list

### LINKED ISSUE #662: Add bash autocompletion to start-cli
Create Tab completion scripts for various commands in embassy cli tools (cli/sdk)

### LINKED ISSUE #1302: [feat]: [cli]: Auto-prompt for login creds upon auth failure
### Prerequisites

- [X] I have searched for [existing issues](https://github.com/start9labs/embassy-os/issues) that already suggest this feature, without success.

Don't make me type in `embassy-cli auth login` when `embassy-cli package install bitcoind` fails because i'm not logged in. Just issue the password prompt.

### Describe the Use Case

Makes using the cli smoother

see above

### Describe Alternatives

can't think of any

### Anything else?

_No response_

MattDHill:
## Summary

Three CLI-only additions for 2.0.1; none changes an RPC, so all work against today's servers.

- **`start-cli completions <shell>`** (#662) — `clap_complete::generate` over `app().into_command()`, the same translated command tree the man pages come from. bash, zsh, fish, elvish, PowerShell. The deb installs only the binary, so the README and CLI reference document `eval "$(start-cli completions bash)"` rather than a packaged file. New `start-cli-completions.1` and the entry in `start-cli.1` are hand-written; Generated Artifacts will confirm the roff.
- **Log in on demand** (#1302) — `CallRemote<RpcContext> for CliContext` catches a 34 (Authorization: no key, unenrolled key), and at a terminal prints `Not logged in to <host>; enter the server password to log in.`, runs the enroll body factored out of `cli_login` (`auth::login`), and retries the call once. Not for `auth.login` itself; off a terminal the error is returned as before. The recursive future is `Box::pin`ned.
- **`binding list` bridge columns** (#3819) — `BRIDGE` / `BRIDGE SSL` show the `10.0.3.1:<port>` rows of the binding's `available` set (the rows `sdk.host.getBridgeAddress` resolves), under `package host` and `server host`. This is the discoverability slice; the `net bridge-url <pkg>` command shape stays dr-bonez's call, so the issue is left open.

Cargo.lock adds `clap_complete 4.6.1` only — pinned so clap stays at 4.6.1.

Closes #662
Closes #1302
Refs #3819

---

Carries the 2.1.0 version bump (`Cargo.toml`, `Cargo.lock`, the changelog heading, the generated lines of `start-cli.1` — cherry-picked from #3987, which this supersedes) and `projects/start-cli/release-notes/2.1.0.md`, so the release can be cut from this merge. The notes cover what is on this branch and master; #3975's `--address` change is not in them — add its bullet if it lands before the tag.


shared-libs/crates/start-core/src/net/host/binding.rs
```diff
@@ -615,8 +615,24 @@ pub fn binding<C: Context, Kind: HostApiKind>()
...
+                        "BRIDGE SSL",
+                    ]);
                     for (internal, info) in res.iter() {
+                        let bridge = |ssl: bool| {
+                            info.addresses
+                                .available
+                                .iter()
+                                .filter(|a| a.ssl == ssl)
+                                .filter_map(HostnameInfo::to_socket_addr)
+                                .find(|a| a.ip() == IpAddr::from(crate::HOST_IP))
```
helix-nine: This does not select the same row as `sdk.host.getBridgeAddress`. The SDK's `.bridge` filter requires `HostnameMetadata::{Ipv4,Ipv6}` with `gateway == "lxcbr0"`; matching the literal IP accepts any row with hostname `10.0.3.1`. Plugin rows are retained in `available`, and `PluginHostnameInfo.hostname`/`port` are unrestricted, so a plugin-provided `10.0.3.1:<lower-port>` row can win this ordered `.find()` and make the CLI print an address the SDK excludes. Filter on the lxcbr0 gateway metadata (and IPv4, matching the SDK lookup) rather than on parsed IP alone; a plugin decoy in the test would pin the distinction.

projects/start-cli/README.md
```diff
@@ -94,8 +94,13 @@ The full command tree comes from `start-core::main_api()`. Top-level groups incl
 - `registry`, `tunnel` — operate against a registry / StartTunnel server.
 - `s9pk`, `init-key`, `pubkey`, `util` — local packaging/dev tooling.
 - `echo`, `state`, `git-info` — diagnostics.
+- `completions` — a completion script for bash, zsh, fish, elvish or PowerShell.
 
-Run `start-cli <group> --help` for any group.
+Run `start-cli <group> --help` for any group. Tab completion:
+
+```sh
+eval "$(start-cli completions bash)"      # or zsh / fish / elvish / powershell
```
helix-nine: The `or ... powershell` instruction makes this pasteable line false: `eval "$(...)"` is Bourne-shell syntax, not PowerShell or Elvish activation syntax. Either label this example bash/zsh-only, as the changelog and CLI reference already do, or give each shell its own activation command.

helix-nine (changes requested):
The implementation builds and the retry behavior works in both TTY and non-TTY harnesses. One functional selector needs tightening, the README advertises one shell's activation syntax for incompatible shells, and the StartOS side of this user-visible change still needs its changelog entry: `startbox` exposes this same `start-cli` applet, this PR changes the StartOS CLI book, and the root contribution rules require the affected product's docs **and** changelog to move together. Please add the corresponding entries under `projects/start-os/CHANGELOG.md`.

Verification: `cargo check -p start-cli`; full `cargo test -p start-core --features=test` (804 passed, 3 ignored); generated-manpage test; all five completion generators plus `bash -n`; docs build; mock RPC TTY login/retry and non-TTY no-retry paths. CI is green.

shared-libs/crates/start-core/src/net/host/binding.rs
MattDHill: Done in 142db71a1 — `bridge_address` selects `HostnameMetadata::Ipv4 { gateway }` with `gateway == START9_BRIDGE_IFACE`, per `ssl`, as the SDK does; a test seeds a plugin `10.0.3.1:1` decoy ahead of the real rows and asserts 8080/8443.

projects/start-cli/README.md
MattDHill: Done in 142db71a1 — one activation line per shell (bash/zsh `eval`, fish `| source`, elvish `slurp`, PowerShell `Invoke-Expression`).

MattDHill:
StartOS changelog entries for the three changes added under 0.4.0.2 in the same commit (142db71a1).

projects/start-cli/README.md
helix-nine: One correction remains here: the combined bash/zsh line still invokes `completions bash` under zsh. I measured the literal command with zsh; it exits 127 at `command not found: complete` because clap’s bash script uses Bash’s `complete` builtin. Please split it into `eval "$(start-cli completions bash)"` for bash and `eval "$(start-cli completions zsh)"` for zsh.

MattDHill: Right — split in 30d612d09: `completions bash` for bash, `completions zsh` for zsh.

MattDHill:
Reversing myself on the StartOS changelog: the three entries are removed again in 30d612d09.

The OS changelog is read by someone deciding whether to update the OS, and none of these changes what the OS does — they change the `start-cli` applet the image bundles. On the server that applet runs as root over loopback with the authcookie, so the login prompt can never appear there; completions and the bridge column are CLI behavior whose record is `projects/start-cli/CHANGELOG.md`, which has been the applet's changelog since it went independently versioned (and is what `start-cli --version` reports inside `startbox` too). `server epp` is in the OS log because the OS gained the EPP setting; CLI-only changes — #3975, the `attach` help fix in #3922 — have never been mirrored, and this PR shouldn't start. The `cli-reference.md` edit is the CLI's reference living in the OS book, not an OS change.

helix-nine (dismissed):
Approved. You are right about the product boundary: this is independently-versioned start-cli behavior, even when the applet is bundled in `startbox`, so the start-cli changelog is the correct release record and the StartOS book remains the CLI reference location. I also verified the split zsh line against the literal documented command with `compinit`; it exits 0. The bridge regression test passed on the preceding fix commit.

dr-bonez (dismissed):

MattDHill:
Pushed the 2.1.0 bump (cherry-picked from #3987) and the release notes onto this branch; #3987 is closed as superseded.

dr-bonez (approved):

PR #3978, opened 2026-09-17
fix(start-sdk): keep the packaging skills in the guide's tree

MattDHill:
## Summary

Follow-up to #3971. The skill landed at `.claude/skills/package-service`, a repo-root path that reaches `live-docs` only when a product tag syncs the root (`live-docs-guard.yml` refuses a PR touching it) — so every future edit to the skill would wait for a release, where a guide page reaches every workspace and helix on their next sync. That is the wrong gate for packaging doctrine.

- **`projects/start-sdk/docs/skills/package-service/`** is the skill's home. #3979 put it on `live-docs` and the backport (`3fa8fee5e`) brought it to `master`; it sits beside `src/` and `package-template/` — guide content the live-docs guard admits, not built by mdBook, deployed with the rest of the book tree. The master copy's `workspace-setup.md` describes 2.0.1's `init-workspace` linking (helix's `54acd4df0`); `live-docs` keeps the neutral wording until that ships.
- **Nothing at the repo root refers to it.** The old `.claude/skills/package-service` directory is removed and not replaced by a symlink; `.claude/skills/` holds `start9-frontend` alone, and `.agents/skills` still mirrors it for Codex in-repo.
- **The Claude Code marketplace and the Codex `$skill-installer` route are gone** (`.claude-plugin/`, `.claude/.claude-plugin/`, and their paragraph in `environment-setup.md`). A skill cannot build, pack, or scaffold outside a packaging workspace, so every route around the workspace only led back to creating one — and the marketplace file was the last thing at the repo root that had to agree with the book, which is what produced the release-ordering findings on this PR. A packaging skill is reached through a packaging workspace, never installed on its own; the root and SDK `AGENTS.md` say so.
- **`init-workspace`** links a workspace's `.claude/skills` and `.agents/skills` at `start-technologies/projects/start-sdk/docs/skills` — the packaging skills alone. Test updated.
- Docs and scope files follow: `agent-context.md`, `environment-setup.md`, the root and SDK `AGENTS.md`, the start-cli 2.0.1 changelog entry. The SDK `AGENTS.md` states the routing: a skill fix to what is published goes to `live-docs` by PR; one accompanying code rides that code's release; no product tag is involved in publishing a skill.

Helix reads the book's directory directly (Start9Labs/helix#187). What rides the 2.0.1 tag is the `init-workspace` code and the sentences describing it, nothing else.

Not verified locally: the `start-core` build (`init.rs` constant and test) — CI runs it.


shared-libs/crates/start-core/src/s9pk/init.rs
```diff
@@ -70,8 +70,9 @@ const AGENTS_SYMLINK_TARGET: &str =
 const LEGACY_AGENTS_SYMLINK_TARGET: &str = "start-technologies/projects/start-sdk/docs/AGENTS.md";
 /// Path to the package template inside the cloned guide (joined onto MONOREPO_DIR).
 const TEMPLATE_SUBPATH: &str = "projects/start-sdk/docs/package-template";
-/// The fleet's agent skills inside the cloned guide (joined onto MONOREPO_DIR).
-const SKILLS_SUBPATH: &str = ".claude/skills";
+/// The packaging skills inside the cloned guide (joined onto MONOREPO_DIR). They live
+/// in the book's tree so a change to one reaches `live-docs` the way a page does.
+const SKILLS_SUBPATH: &str = "projects/start-sdk/docs/skills";
```
helix-nine, line 75: **[P1] Bootstrap this target before making start-cli depend on it**

A `start-cli/v2.0.1` tag does not publish this directory. `docs-sync-on-tag.yml` copies the tagged project's tree and the repo root, then restores every other `projects/*` subtree from `live-docs`. Today `live-docs` has no `projects/start-sdk/docs/skills`, so the start-cli release will publish the new root symlink but retain the old StartSDK tree; every newly initialized workspace then gets dangling `.claude/skills` and `.agents/skills` links. I confirmed the current shared-tree watermark predates this head, the new root entry is mode 120000, and the target is absent on `origin/live-docs`. Please make the bootstrap atomic (for example, land the target on `live-docs` before switching this constant/root entry, or change the sync arrangement so the start-cli release carries both) rather than relying on an unspecified later StartSDK release.

helix-nine (changes requested):
Requesting changes for one release-ordering issue. I verified the targeted Rust test, Prettier, the repository symlink, and a fresh isolated Claude plugin install.

MattDHill:
Right — a start-cli tag restores `projects/start-sdk/` from `live-docs`, so the target would be missing. The bootstrap is #3979: the same directory (and the `writing-instructions.md` section the skill points at, byte-identical to master's) as a `live-docs` PR, which the guard admits. Order: #3979 merges and backports to master first; then this PR is rebased to delete `.claude/skills/package-service/` and add the symlink — no rename — and merges before the 2.0.1 tag, so the constant and the root symlink never ship ahead of their target. The body says the same.

helix-nine:
That sequence resolves the finding. I’ll leave the review pending until #3979 has merged/backported and this branch is rebased, then verify the resulting diff and target on `live-docs`.

MattDHill:
Rebased on the backport of #3979 (`3fa8fee5e`): the target directory is on both `live-docs` and `master` now, so this PR is down to removing the old directory, adding the symlink, the `init-workspace` constant, and the path updates. The 2.0.1 tag can sync the root without anything dangling.

.claude/skills/package-service
```diff
@@ -0,0 +1 @@
+../../projects/start-sdk/docs/skills/package-service
```
helix-nine: **[P2] Restore the 2.0.1 workspace behavior in the canonical skill**

This symlink switches every repo/plugin consumer from the old root copy to #3979's bootstrap copy, but #3979 deliberately omitted the two descriptions of the unreleased behavior now introduced here. As a result, `references/workspace-setup.md` says workspace creation writes only `AGENTS.md`, `AGENTS.local.md`, `CLAUDE.md`, and `.startos/`, and its command description never says `.claude/skills` and `.agents/skills` are created—even though this PR makes both true. The old root copy already had those facts. Please carry them into `projects/start-sdk/docs/skills/package-service/references/workspace-setup.md` in this PR (with “guide's packaging skills” replacing “fleet's skills”), so the canonical skill and the code ship together.

helix-nine (changes requested):
The bootstrap resolves the original release-ordering finding. The rebased diff leaves one documentation gap before approval. Reverified the targeted Rust test, Prettier, target availability on both branches, repository symlink resolution, and a fresh isolated Claude plugin install.

.claude/skills/package-service
helix-nine: Fixed in `54acd4df0`: the canonical skill again names `.claude/` and `.agents/` in the consent prompt and documents both skill links in the `init-workspace` result.

helix-nine (dismissed):
The bootstrap sequencing and canonical skill documentation are both resolved. Verified the targeted `start-core` test, Prettier, target availability on `live-docs` and `master`, repository symlink resolution, and a fresh isolated Claude plugin install containing the updated setup text.

MattDHill:
Pushed `1423793de`: the root symlink is gone rather than kept. Its only real reader was the `start9` plugin, whose manifest now names the book's directory (plugin root = repo root, both skill directories listed; verified from a fresh local marketplace install). Helix will read `projects/start-sdk/docs/skills` directly (helix PR to follow), so no product tag is involved in the skill at all — the 2.0.1 tag carries only the `init-workspace` code and the docs describing it.

.claude-plugin/marketplace.json
```diff
@@ -8,7 +8,7 @@
   "plugins": [
     {
       "name": "start9",
-      "source": "./.claude",
+      "source": "./",
```
helix-nine: **[P1] Keep the plugin root away from the monorepo package.json**

With `source: './'`, Claude copies the entire monorepo as the plugin and then sees the root `package.json`/lockfile as plugin dependencies and installs them inside the plugin cache. I reproduced this from a clean `git archive` of this head: a 50 MB / 3,226-file source became a **696 MB / 52,563-file** installed plugin with a top-level `node_modules`; installing from the actual development checkout failed `ENOSPC` while copying `target/debug/incremental`. The previous `.claude` source installed only the two skills. Please keep a small plugin root that does not contain the monorepo package manifest—for example, retain the now-safe `.claude/skills/package-service` symlink (the target is already bootstrapped on `live-docs`) and keep `source: './.claude'`, or use another dedicated wrapper directory.

helix-nine (changes requested):
The direct skill paths are sound, but moving the plugin root to the monorepo root introduces a blocking install regression. I also reverified the targeted Rust test, Prettier, manifest validation, and the canonical setup text.

MattDHill:
You were right, and my earlier "verified" was wrong: a marketplace added from a local git checkout is copied too, not loaded in place — the same test on this machine had left a 68 GB copy of the working tree (`target/` included) in the plugin cache without my noticing. Fixed in `74bbbab1f`: the plugin root is `projects/start-sdk/docs/skills` itself, manifest inside it, `skills: ["./"]`. Installed through GitHub from this branch, the cache entry is 76 KB / 6 files and lists exactly `start9:package-service`; the directory carries no `package.json`, so nothing is installed into it, and the SDK `AGENTS.md` now says to keep it that way. `.claude/skills/` holds `start9-frontend` alone — the plugin ships the packaging skills, which is what it is for.

projects/start-sdk/docs/skills/.claude-plugin/plugin.json
```diff
@@ -1,11 +1,12 @@
 {
```
helix-nine: **[P1] Bootstrap the plugin manifest onto live-docs before switching the marketplace**

This direct-head install is correctly tiny, but a `start-cli/v2.0.1` sync publishes the new root marketplace file while restoring `projects/start-sdk/` from `live-docs`, and `origin/live-docs` does not contain this manifest. I reproduced that exact mixed tree: installation exits 0 and copies the five skill files into a 52 KB cache entry, but `claude plugin details start9@start9` reports **Skills (0)** because `skills: ['./']` is absent—so users get a silently empty plugin. Please land this manifest on `live-docs` first (and backport/rebase), just as #3979 bootstrapped the skill directory, or otherwise make the marketplace source discoverable with the current live tree before its root config ships.

helix-nine (changes requested):
The dedicated plugin root fixes the cache-size regression on this branch, but the release-composed `live-docs` tree still lacks its plugin manifest. I reverified both direct-head and release-state installs, manifest validation, Prettier, and the targeted Rust test.

MattDHill:
That was the same defect as the symlink and the constant — a root file paired with a book path, reaching `live-docs` on different events — and bootstrapping the manifest would only have papered over this instance. `2ab050fe8` removes the pairing instead: the marketplace entry fetches `projects/start-sdk/docs/skills` from `live-docs` by URL (`git-subdir`) and declares the skills itself (`strict: false`, `skills: ["./"]`), so there is no manifest in the book to bootstrap and the root file never depends on the tree it sits in. Reproduced your release-state case as the fix's own test: marketplace added from GitHub at this branch, plugin fetched from today's `live-docs` (no manifest there) — 68 KB / 5 files, `Skills (1) package-service`, loads as `start9:package-service`. The docs drop the `@live-docs` pin accordingly.

helix-nine:
Confirmed and fixed the remaining docs detail in `7dbeb8753`: the `git-subdir` design resolves the release-state issue. A clean GitHub install fetched `live-docs` at `f213d46fc`, installed 5 files / 52 KB, and exposed exactly `package-service`. While verifying updates, I also measured that `plugin marketplace update` leaves the installed plugin SHA unchanged; `plugin update start9@start9` advances it, so the guide now gives that command. Marketplace validation, Prettier, `git diff --check`, and the targeted `start-core` test pass.

helix-nine (dismissed):
The URL-backed `git-subdir` source removes the cross-branch manifest dependency and keeps the installed plugin small. Verified a clean install against today’s manifest-less `live-docs` tree, the update path, marketplace validation, Prettier, `git diff --check`, and the targeted `start-core` test.

MattDHill:
`4c2fad394` removes the marketplace and the Codex installer route altogether rather than keep fixing their coupling to the book. A skill cannot build, pack, or scaffold outside a packaging workspace, so every install path around the workspace only led back to creating one; `environment-setup.md` now says exactly that, and the root and SDK `AGENTS.md` state that a packaging skill is reached through a workspace, never installed on its own. With that, nothing at the repo root refers to the packaging skills, and the only things the 2.0.1 tag carries are the `init-workspace` code and its docs.

helix-nine (approved):
This removes the coupling rather than maintaining another publication route. Verified the workspace links still target the canonical `live-docs` directory, no standalone marketplace/installer references remain, Prettier and `git diff --check` pass, and the targeted `start-core` test passes.

PR #3980, opened 2026-09-17
feat(start-cli): complete an older workspace on first use

MattDHill:
## Summary

A workspace made by start-cli 2.0.0 has no `.claude/skills` / `.agents/skills` links, and nothing told a packager to re-run `init-workspace` after updating — the outdated-CLI notice only says how to update the binary, and it is printed by the old binary anyway. Two changes, before 2.0.1 is tagged:

- **Every command completes the signing workspace it runs in.** `CliContext::init` calls `complete_signing_workspace(cwd)`: one walk (`find_signing_workspace`) finds the nearest directory whose `.startos` holds the build key — renaming a legacy-named key on the way, stopping at an inaccessible ancestor, passing over a config-only workspace above or below — and there it prints the outdated-CLI notice and fills in what `init-workspace` would: the two links, if absent, and only when the guide checkout already carries `projects/start-sdk/docs/skills` (a checkout from before #3979 gets no dangling link; the sync brings the target, the next command links it). `build_key()` uses the same walk, so the two cannot drift. Nothing on this pass can fail a command. This covers `make install` (`init-key`, `s9pk select`, `package install -s`), `init-package`, `publish`, and anything else that never signs — the outdated notice had the same gap and moves with it.
- **Said in the docs.** `environment-setup.md` § Keep it current states that anything a newer `start-cli` adds to a workspace is filled in the first time any `start-cli` command runs inside it, and that re-running `init-workspace` does the same on demand; the 2.0.1 changelog entry says the same.

Tests, on the callable `CliContext::init` uses with an explicit start path: a nested cwd under an inner key-bearing workspace whose ancestor is a config-only workspace — the inner root gets the links, the ancestor and the cwd stay untouched, and a legacy-named key is found and renamed; plus the existing no-target and no-clobber cases.

Deleting the links is not an opt-out — the next command restores them — but they are inert until a skill is invoked.

Not verified locally: the `start-core` build — CI runs it (the pinned rustfmt is clean).


shared-libs/crates/start-core/src/context/cli.rs
```diff
@@ -225,6 +225,8 @@ impl CliContext {
             match candidate.try_exists() {
                 Ok(true) => {
                     crate::s9pk::init::warn_if_start_cli_outdated(&dir);
+                    // A build must not fail over a link init-workspace would report.
+                    let _ = crate::s9pk::init::link_guide_skills(&dir);
```
helix-nine: The documented `make install` path never reaches this hook. Its recipe runs `start-cli init-key`, `s9pk select`, then `package install -s`; neither `select` nor `cli_install` calls `build_key()`. A packager who updates from 2.0.0 and installs an existing artifact therefore keeps both skill links missing, contrary to the docs, changelog, and PR contract. Please run completion from general workspace discovery (or otherwise cover the install path) and add a regression test for that command path.

helix-nine (changes requested):
One blocking path remains: installation does not run the new completion hook.

Verified locally: `cargo test -p start-core s9pk::init::test --features=test`, `cargo check -p start-cli`, and the packaging mdBook build all pass.

MattDHill:
@helix-nine — confirmed, and the same gap applies to the outdated-CLI notice, which has always hung on `build_key()` too. Before I change it, the facts and the options, and I'd like your recommendation rather than another round.

**Facts**
- Every `start-cli` invocation runs `ClientConfig::load` → `find_workspace_config()` (walks up from cwd for a schema-tagged `.startos/config.yaml`) → `CliContext::init`. That is the general discovery, and it finds the *config* workspace.
- Only signing commands (`pack`, `publish`, `init-package`) call `build_key()`, which walks for `.startos/build.key.pem` — the *signing* workspace. `make install` (`init-key`, `s9pk select`, `package install -s`) never signs.
- Workspaces nest: the guide's own docs describe an outer `.startos/config.yaml` with an inner `.startos/build.key.pem`, and settings merge innermost-first. In such a layout the config walk lands on the outer directory, whose `start-technologies` (if any) may be a development checkout on `master`, while the key directory is the one `init-workspace` provisioned beside the `live-docs` checkout.

**Options**
1. Run completion (and the outdated notice) from `CliContext::init`, on every command, but keyed on the nearest `.startos/build.key.pem` — the signing workspace — never the config-only root. Covers `select`/`install`/everything; two symlinks created once, even from an unrelated command such as `server restart` run inside a workspace.
2. Same hook, keyed on the directory `find_workspace_config()` found. Simpler to wire, but wrong for the nested layout above: it would link `.agents/skills` into a `master` checkout.
3. Keep it on `build_key()` and add explicit calls in `select` and `cli_install` — covers the documented paths, misses any future one, and duplicates the walk.

I'd take 1. For the regression test: `CliContext::init` builds the runtime, so I'd test the discovery function it calls — a workspace with a key, cwd in a nested package directory, links land in the key directory and not in cwd, and a config-only ancestor is left untouched — plus keep the existing no-target and no-clobber tests. If you think the command path itself needs a binary-level test, say how you'd shape it.

Which would you do, and is there a path or a layout I've missed?

helix-nine:
I’d take option 1: centralize completion on the nearest signing workspace and invoke it during `CliContext::init`. That preserves the important association here: the checkout and skill links belong beside the key that `init-workspace` provisioned, not whichever config layer happened to resolve first. Options 2 and 3 both encode the wrong abstraction.

One correction to the call graph: `init-package` does not call `build_key()`; it uses `find_workspace_root()` and has its own explicit outdated warning. `publish` also opens the already-signed s9pk without loading the build key. Today `build_key()` is reached by `pack` plus the edit/inspect/convert paths that need the workspace key. That makes centralizing both completion and the outdated notice more important.

I would factor one signing-workspace walk and reuse it from both initialization and `build_key()`, rather than leave two walks that can drift. Preserve the current details in that helper:

- recognize/migrate `LEGACY_BUILD_KEY_FILE` as well as `BUILD_KEY_FILE`;
- stop on an inaccessible ancestor;
- choose the nearest key-bearing workspace even when a nearer/outer config-only workspace exists.

No binary-level test is needed. Shape the callable used by `CliContext::init` to accept an explicit start path, then test that function directly: nested cwd → inner key-bearing root gets the links; config-only ancestor/root remains untouched; legacy-key discovery still selects the same root. That exercises discovery plus the completion side effect without process-global `set_current_dir`, and the existing no-target/no-clobber tests cover the rest.

MattDHill:
Done as agreed, in `a7780f145`: one walk, `find_signing_workspace`, keyed on the build key (legacy name migrated, inaccessible ancestor stops it, config-only layers passed over), used by both `CliContext::init` — via `complete_signing_workspace`, which also carries the outdated-CLI notice now — and `build_key()`. Tested on the callable with an explicit start path: nested cwd → inner key-bearing root gets the links, the config-only ancestor and the cwd stay untouched, and a legacy-named key selects the same root; the no-target and no-clobber tests stand. Docs and changelog now say "any `start-cli` command" rather than naming three.

helix-nine (approved):
The revised signing-workspace walk closes the install-path gap and preserves the legacy-key and nested-workspace behavior. The existing `init-package` warning remains safe: its process-wide `Once` prevents duplicate output while preserving the config-only fallback.

Verified locally on `a7780f145`: `cargo test -p start-core s9pk::init::test --features=test` (11 passed), `cargo check -p start-cli`, and the packaging mdBook build.

PR #3985, opened 2026-09-17
feat(ci): publish the brochure and the docs to Start9 Pages

MattDHill:
## Summary

Two more of the VPS's sites now also land in the NextExplorer folders Start9 Pages serves on evelyn, ahead of moving their DNS. The VPS steps stay until then.

- **Brochure** — `deploy-brochure.yml` publishes to `marketplace.start9.com` after its rsync.
- **One command per bundle** — `npm run build:brochure` and the new `npm run build:wrt:demo` (config from the sample, mocks on, stamped) produce exactly what the workflows publish, `404.html` doubled from `index.html` included (how a static host boots the SPA on a deep link). A bundle built on a laptop and dropped into NextExplorer is the same site; the workflows are the command plus a check.
- **Docs** — new `deploy-docs-pages.yml`: `workflow_run` after every successful `Build and deploy docs`, checks out `live-docs`, runs `./build.sh` (same mdBook/mdbook-tabs pins; llms.txt is part of the build once #3986 lands, so merge that first), and publishes `projects/start-docs/docs` to `docs.start9.com` with `keep: 0.3.5.x`. It lives on `master` because a PR to `live-docs` may not carry workflow files; it references the action by `@master` for the same reason. Once the VPS is gone and a tag has carried this to `live-docs`, the publish can move into `docs-deploy.yml` itself and this workflow goes.
- **Action** — `folder` may be nested (`docs.start9.com/0.3.5.x`, how the retired Sphinx site gets seeded by hand: 293 MB that no repo builds), and a new `keep` input names entries of the previous contents to carry across the swap. They're server-side copied (`POST /api/files/copy`) into the staging folder before the renames — a copy, not a move, so the live folder holds them until the swap lands — and a run that died between its two renames is repaired at the start of the next one by putting `.old` back. A kept entry that the published tree also contains fails the run before the swap.

Verified against `nxzai/explorer:v2.2.7` locally: first publish with nothing to keep (warns, continues), nested seed, replacing publish that keeps it, the mid-swap recovery, and the collision.

The `build.sh` side of the docs move (redirect stubs and a `404.html` standing in for the VPS's nginx rules) goes to `live-docs` separately, since that's where the site builds from.


helix-nine (approved):
Verified the two bundle commands locally after a clean install/build of the workspace dependencies; both produced byte-identical `index.html`/`404.html` pairs, and the demo config retained mocks plus the stamped hash. Also ran ShellCheck on the publisher, checked the workflow/action diff, and walked the first-publish, keep collision, and interrupted-swap recovery paths.

PR #3986, opened 2026-09-17
docs(start-docs): build the whole site with build.sh

MattDHill:
## Summary

docs.start9.com is moving from the VPS to Start9 Pages on evelyn (Start9Labs/start-technologies#3985 carries the publisher), a static host with no nginx rules of its own. `./build.sh` now leaves the whole site in `docs/`, whether a workflow or a laptop ran it:

- a redirect stub for every unversioned page URL — `/start-os/page.html` → `/start-os/0.4.0.x/page.html`, fragment preserved — alongside the existing `/<book>/` stub; cross-book links and the links in the wild are written unversioned;
- `docs/404.html`, generated from `landing/404.html` with the book list filled in, whose script applies the remaining rules: `/latest/*` and any unknown path onto the `0.3.5.x` site, `/packaging-guide` → `/packaging`, the nine one-off legacy redirects from the nginx config, and extensionless page URLs (`try_files $uri.html`);
- the llms.txt generator, installing `scripts/` deps on first use, so `llms.txt` / `llms-full.txt` are part of the build rather than a CI step.

Targets `live-docs` because that is where the site builds from and the guard admits `projects/start-docs/`; the backport carries it to master. The VPS deploy is untouched — it syncs only the paths it always did, so the extra files never reach it, and its own llms step becomes a harmless repeat.

Verified locally: the build produces 558 files (171 stubs and root files, the llms files included), and the 404 script was run against representative paths (`/latest`, `/latest/x`, `/packaging-guide/foo.html`, each legacy rule with and without a trailing slash, an extensionless versioned page, a missing versioned page, an unknown top-level path, `/`).


helix-nine (approved):
Ran the complete `./build.sh` with mdBook 0.5.2 and mdbook-tabs 0.3.4: 558 files, generated per-page stubs, populated root 404 routing table, and site/book llms files. Exercised the redirect script against latest, packaging-guide, legacy, extensionless, unknown, query, and fragment paths; ShellCheck, Prettier, and diff checks pass.

PR #3993, opened 2026-09-17
ci: retire the docs VPS

MattDHill:
## Summary

`docs.start9.com` and `marketplace.start9.com` resolve to evelyn now, so the VPS halves of the deploys go:

- `deploy-brochure.yml` — SSH setup and rsync removed; the NextExplorer publish is the deploy.
- `docs-deploy.yml` — rsync, `book_versions.conf` generation and nginx reload removed; it publishes to Pages itself (`keep: 0.3.5.x`), through the action in the tree.
- `WEBSITE_DEPLOY_KEY` is no longer referenced anywhere; delete the secret once the `live-docs` copy of `docs-deploy.yml` has moved on (below).

The `docs-deploy.yml` on `master` reaches `live-docs` only with the next product tag, and until then the copy there still rsyncs to a box that is going away. So `deploy-docs-pages.yml` keeps publishing in the meantime, with two changes: it follows **every** `Build and deploy docs` run, not only successful ones — the VPS steps will start failing, and the publisher's own build is its gate — and it stands down by itself once it finds the `live-docs` copy of `docs-deploy.yml` publishing on its own. It can be deleted after that.

`versions.conf`'s header and the start-docs docs stop describing nginx maps; the routing they drove is what `build.sh` writes into the tree.


helix-nine (changes requested):
Two stale descriptions still point at the nginx mechanism this change removes:

- `AGENTS.md:150` says `versions.conf` drives “nginx routing.” This is the root authoritative coupled-change rule, so it now directs future work toward a deleted output.
- `projects/start-docs/scripts/generate-llms-txt.ts:85` repeats that `versions.conf` is shared with “nginx routing.”

Please update both to describe the generated stubs/root `404.html` (or the site routing) in this change.

Verification otherwise passed: I exercised the current-live-docs/no-match and future-live-docs/match sides of the stand-down predicate, confirmed the local publish action is already present on `live-docs`, and ran the complete docs build with mdBook 0.5.2/mdbook-tabs 0.3.4 (560 files, clean tree). CI formatting is green.

MattDHill:
Both fixed in the amended commit: root `AGENTS.md` now says `versions.conf` drives the build, the routing `build.sh` writes into the tree (the unversioned-URL stubs and `404.html`), and llms.txt; `generate-llms-txt.ts`'s doc comment says the same. Nothing else changed.

helix-nine (dismissed):
Confirmed on the amended head: both stale nginx-routing references now describe the static routing generated by `build.sh`, with no other diff from the previously reviewed commit. The targeted Prettier check is green in CI.

MattDHill:
The `Run Automated Tests` failure is the cutover, not this diff: `util::http_reader::main_test` fetches a live URL and needs range support, and the URL it used — `https://start9.com/latest/_static/css/main.css` — was a chain of nginx redirects on the VPS into the legacy docs. On Pages that path is the 404 page (the redirect is a script), so the reader gets HTML with no `Accept-Ranges`. Repointed the test at `https://docs.start9.com/llms.txt`, a file the docs build always produces and which Pages serves with `Accept-Ranges: bytes` (206 on a range request). Amended into the same commit.

PR #3997, opened 2026-09-18
fix(start-os): filter legacy certificate addresses

### LINKED ISSUE #3992: A legacy package with a `certificate` volume cannot start on a server with a WAN IP — compat requests a cert for every available address and `get_ssl_certificate` rejects the WAN IP (DockerProcedureContainer.ts:88, ssl.rs:82)
### Environment

StartOS 0.4.0.1 (user report, x86_64, upgraded from 0.3.5.1) and master at c5d366303 (dev box, x86_64, StartTunnel gateway). Area: service runtime (0.3.x compat).

### What happens

Starting a 0.3.x package that declares a `certificate` volume — Nextcloud (Legacy) here — fails immediately, before its entrypoint runs:

```
start failed Error: Not Found: unknown hostname: 64.23.194.12
   0: unknown hostname: 64.23.194.12
Location: shared-libs/crates/start-core/src/service/effects/net/ssl.rs:50
…@get-ssl-certificate
    at Socket.<anonymous> (/usr/lib/startos/init/Adapters/EffectCreator.js:99:32)
```

`64.23.194.12` is the tunnel gateway's WAN IP on the dev box; on the user's box it is his router's WAN IP (`98.97.1.166`). The service shows Stopped and its logs carry only the error above. Every start attempt repeats it.

The legacy service starts, with a certificate covering the addresses it can actually serve TLS on.

- `projects/start-os/container-runtime/src/Adapters/Systems/SystemForEmbassy/DockerProcedureContainer.ts:88-101` — for a `certificate` volume the compat runtime collects every hostname in `hostInfo.bindings[*].addresses.available` and passes them all to `getSslCertificate`, without filtering on the address's `ssl` flag.
- A gateway's WAN IP is projected into `available` as `HostnameInfo { ssl: false, public: true, hostname: <ip>, metadata: Ipv4 }` (fixture: `shared-libs/crates/start-core/src/net/host/address.rs:778`).
- `shared-libs/crates/start-core/src/service/effects/net/ssl.rs:82-101` — an IP hostname is accepted only if it lies in the container bridge /24 or equals one of a gateway's own subnet addresses. A WAN IP is neither, so the call fails with `unknown hostname`, and the compat runtime aborts the start.

So any box where a gateway reports a WAN IP (public IP on the router, or a StartTunnel gateway) cannot start any legacy package with a `certificate` volume. Filtering the compat list to `ssl: true` addresses — the ones the 0.3.x package's own TLS listener was ever meant to serve — would fix it on the runtime side; alternatively `get_ssl_certificate` could accept any address the OS itself advertises as available.

### Steps to reproduce

1. On a server whose gateway has a WAN IP (a StartTunnel gateway suffices), sideload a 0.3.5.1 `nextcloud.s9pk` (v1; `manifest.yaml` declares `cert: {type: certificate, interface-id: main}`).
2. Configure it (the compat `config` action with defaults) and start it.

### Logs & evidence

User's OS log: `2026-09-05T14:45:02.826681Z ERROR start_core::error: Unknown Error: Error: Not Found: unknown hostname: 98.97.1.166`, logged at the moment Nextcloud (Legacy) was started after the 0.3.5.1 → 0.4.0.1 upgrade; the service never got past it. Reproduced on the dev box today with the 0.3.5.1 `32.0.11` s9pk from the registry's 0.3 API.

helix-nine:
## Summary

- exclude WAN IP addresses from legacy package certificate requests while preserving secure and retained onion names
- cover mixed secure, plaintext, duplicate, onion, and WAN addresses with a focused runtime test
- document legacy-service behavior and add the StartOS changelog entry
- remove an unused generated-binding import from the changed runtime module

## Verification

- `make container-runtime-test`
- `npm --prefix projects/start-os/container-runtime run check`
- `npm --prefix projects/start-os/container-runtime run build`
- `npm --prefix projects/start-os/container-runtime test`
- `make web-format-check`
- `cd projects/start-docs && ./build.sh`
- ablation: adding an SSL WAN address to the focused test makes the previous predicate fail with that IP present in `hostnames`

Closes #3992


helix-nine:
Review round 1 found and fixed one regression in c1647269c: the `ssl`-only filter dropped retained legacy `.onion` certificate names. Migrated package TLS routes can carry `ssl: false` in host info even though the package serves TLS directly on that onion route, so the predicate now preserves `.onion` SANs while still excluding the plaintext WAN address that prevents startup. The focused test now covers this case.

Verified with:
- `make container-runtime-test`
- `npm --prefix projects/start-os/container-runtime run check`
- `npm --prefix projects/start-os/container-runtime run build`
- `make web-format-check`
- `cd projects/start-docs && ./build.sh`

helix-nine:
Review round 2 found and fixed one remaining startup failure in d80a768d0. A legacy `lan-config.ssl: true` binding publishes its WAN IP with `ssl: true`, so the round-1 predicate still passed that IP to `getSslCertificate`, which rejects it. The filter now excludes public IPv4 WAN addresses regardless of their SSL flag while retaining secure and legacy onion names; the focused test covers both plaintext and SSL WAN entries.\n\nVerified with the full container-runtime test suite (13 tests), typecheck, build, targeted test ablation, Prettier, and `git diff --check`.

dr-bonez (approved):

PR #3998, opened 2026-09-18
fix(start-tunnel): preserve published port controls on refresh

### LINKED ISSUE #3947: A PCP delete removes a manual published port that targets the same device port — `remove_peer_forward` matches by target and ignores `auto` (tunnel/forward/pcp.rs:663)
### Environment

- StartTunnel 1.3.0 (`start-tunnel --version`, `dpkg -l start-tunnel` → `1.3.0`), tag `start-tunnel/v1.3.0`; the same code is on `master` @ `81c0b5a08`.
- VPS: DigitalOcean (sfo3), Debian 13 (trixie), dedicated public IPv4 `64.23.194.12`; not behind NAT. Host firewall is the tunnel's own nftables (`ip startos`, `ip6 startos`, `inet startos_egress_guard`), no ufw.
- Client: StartOS 0.4.0.2 (`a30d84d2`) as device `10.59.88.5` on the tunnel's subnet, opening its own ports over PCP.

### What happens

A **manual** published port is deleted by the tunnel when the StartOS device releases an **automatic** PCP mapping that happens to point at the same device port.

Setup on the VPS, for a mail package whose SMTP binding StartOS publishes on device port `2525`:

| From (VPS)           | Target             | Kind                  |
| -------------------- | ------------------ | --------------------- |
| `64.23.194.12:25`    | `10.59.88.5:2525`  | manual, added by hand |
| `64.23.194.12:2525`  | `10.59.88.5:2525`  | automatic (PCP)       |

(likewise `465 → 10465` / `10465 → 10465` and `993 → 10993` / `10993 → 10993`.)

StartOS then stopped the package (a failed restore attempt: stop → `cannot restore over existing package` → start). On stop it withdrew the binding's public address and sent PCP MAP deletes (lifetime 0) for device ports `2525`, `10465`, `10993`. Afterwards the tunnel's `db dump` showed the three **manual** rows gone — `64.23.194.12:25`, `:465` and `:993` — while the automatic rows were re-created by the next PCP request when the package came back. Nothing was logged at the tunnel's default level; the device's log shows only the normal address withdrawal.

Re-adding the rules by hand restores the ports until the next time StartOS releases the mapping, which every package update, reinstall, restore, or address toggle does. For a mail server this is the whole public path (port 25 can only be reached this way, since a package cannot claim ports below 1025 on the box).

A PCP delete only removes mappings the PCP client created. A `PortForward::Dnat { auto: false, .. }` row is the operator's and must survive the device releasing, re-requesting, or losing its own automatic mapping to the same target. Lease expiry already behaves this way; the explicit delete path does not.

### Steps to reproduce

1. On a StartOS box with a StartTunnel gateway, enable the public IPv4 address on any raw-port interface — e.g. `lnd`'s peer port `9735`. The tunnel gains `64.23.194.12:9735 → <device>:9735` (auto).
2. On the tunnel, add a manual forward to the same device port on a different external port, the way a user maps a standard port: `start-tunnel port-forward add 25 <device>:9735`.
3. Disable that public address on the box (or restart the service, or reinstall it).
4. `start-tunnel db dump` → the manual `:25` row is gone; the auto `:9735` row is what got re-created (or is gone too, depending on order).

`shared-libs/crates/start-core/src/tunnel/forward/pcp.rs`:

- `remove_forward` (`:151`) → `remove_peer_forward` (`:672-703`) resolves a PCP delete by **target only**: it takes the first entry in `port_forwards` for which `peer_forward_matches` (`:663-668`) returns true, and that predicate is `PortForward::Dnat { target: t, .. } => t == target` — it never looks at `auto`.
- `port_forwards` is a `BTreeMap<SocketAddrV4, PortForward>` keyed by the **external** address (`tunnel/db.rs:266`), so `64.23.194.12:25` sorts before `64.23.194.12:2525`, and the manual row is the first match. The auto row survives that call and is later reaped by its own lease or re-created by the client — which is exactly the end state observed.
- Callers: `net/port_map/server/mod.rs:534` and `:576` (`lifetime == 0` → `backend.remove_forward(peer, internal_port)`).

The lease path is already correct by contrast: `forward/lease.rs:72` only seeds leases for `auto: true` rows and `reap_dnat` (`:141`, `:172`) removes by source key. The explicit-delete path should apply the same filter — match `Dnat { target, auto: true, .. }` (and the SNI fallback's `auto`) in `peer_forward_matches`, or skip `auto: false` rows in `remove_peer_forward`. UPnP's `remove_forward_by_source` (`:155`) is keyed by source and does not have this problem.

### Logs & evidence

- Tunnel `db dump` before (manual `:25`, `:465`, `:993` present alongside the auto rows) and after (only the auto rows), captured 2026-09-15 on the VPS above.
- StartOS journal at the time of the stop: `net_controller.rs:577: port 2525: WAN address … on gateway(s) [wg0]` lines for the withdrawal and re-add; the tunnel journal shows nothing at default verbosity.
- Related: #3886 (mail on StartTunnel; item 1 assumes manual DNAT rules for the mail ports are durable), the Stalwart package's `instructions.md` (in review, not yet pushed) tells users to re-check these rules after an update or restore because of this.

### LINKED ISSUE #3974: A device's port-map refresh re-enables a published port the operator disabled — the re-assert paths never read `enabled` (tunnel/forward/igd.rs:293, tunnel/forward/pcp.rs:420)
### Environment

StartTunnel 1.3.0 (`start-tunnel --version` → `start-tunnel 1.3.0`, `dpkg -l start-tunnel` → `1.3.0`), tag `start-tunnel/v1.3.0`; the same code is on `master` @ `6261b81f6`. Found by reading the code; the deployment it was read against is a DigitalOcean Debian 13 VPS with its public IPv4 on the NIC, no NAT, host firewall is the tunnel's own nftables. Area: published ports (PCP/UPnP re-assert vs. the operator's enable/disable).

### What happens

Disabling a published port with `start-tunnel port-forward set-enabled --enabled false` holds only until the device that asked for that external port next refreshes its mapping — for StartOS, half of its 3600 s lease, so at most 30 minutes. The refresh puts the forward back into the dataplane while the database, the UI and the CLI keep reporting it disabled. A daemon restart then honours the database again (restore skips disabled entries), so the port flips between live and dead depending on whether a restart or a refresh ran last.

Three re-assert paths ignore the flag:

- A DNAT entry (automatic or manual) on the external port a device requests: `apply_peer_forward_range` matches any existing `Dnat`, sees it is no longer in `active_forwards`, and re-installs the nft rule.
- An SNI hostname route: `persist_sni_forward` deliberately preserves the stored `enabled: false`, then registers the route in the live demux regardless.
- An SNI fallback (a DNAT converted to a shared port): `persist_fallback_forward_locked`, same shape.

The manual-table toggle in the web UI reaches the DNAT case whenever a manual forward sits on the exact external port a device refreshes — e.g. a hand-added `9735 → <device>:9735` beside a service that maps `9735` itself.

A disabled entry stays out of the dataplane until the operator enables it. A device refresh against a disabled entry should renew the lease (so it is not reaped) and leave the forward down — or refuse the mapping — but never re-arm it.

Found by reading code, not reproduced live.

- `shared-libs/crates/start-core/src/tunnel/api.rs:1518` `set_forward_enabled` — writes `enabled` and tears down: `active_forwards` removal at `:1580`, `sni.unregister` at `:1603`. It accepts any source, automatic entries included.
- `shared-libs/crates/start-core/src/tunnel/forward/igd.rs:293` — the `Some(PortForward::Dnat { .. })` arm of `apply_peer_forward_range` binds no field, so `enabled` is never consulted; `!active` → `add_forward_range` at `:298`.
- `shared-libs/crates/start-core/src/tunnel/forward/pcp.rs:644` `sni_route_fields` returns the stored `enabled` for an existing route; `persist_sni_forward` writes it back at `:390` and then calls `self.sni.register(...)` unconditionally at `:420`. `persist_fallback_forward_locked` does the same at `:537`.
- Contrast `shared-libs/crates/start-core/src/tunnel/context.rs:327` and `:354`: the startup restore skips `!enabled` entries, which is why a restart makes the flag stick again.
- Refresh cadence: `shared-libs/crates/start-core/src/net/port_map/client.rs:61` (`PCP_LIFETIME_SECONDS = 3600`) and `renew_due` at `:1083` (renew at half-life).

Fix shape: in each re-assert path, when the stored entry is `enabled: false`, stamp the lease and return success without touching the dataplane (or return the PCP/UPnP failure code, if a disabled port should read as refused to the device).

### Logs & evidence

None beyond the code; nothing is logged on either the disable or the re-assert.

helix-nine:
## Summary

- preserve manual ownership and disabled state when devices refresh matching DNAT, SNI, or IPv6 pinhole mappings
- limit PCP/UPnP deletion and lease cleanup to automatic mappings, including when a manual forward shares the same device target
- let a device repoint an automatic mapping it already holds (IPv4 forward, IPv6 pinhole, SNI route or fallback) to a different internal port, instead of conflicting until the old lease lapses; SNI ownership in the shared demux is per device, so StartWRT takes the same rule
- document the published-port behavior in the StartTunnel guide, changelog, and 1.3.1 release notes

## Testing

- `make start-core-test`
- `cargo check --tests -p start-core --features=test`
- `cargo check -p start-tunnel` (with its generated web include directory present)
- `make start-core-format-check start-tunnel-format-check`
- Prettier check on changed Markdown
- `git diff --check`

Closes #3947
Closes #3974


helix-nine:
Review round 1 found and fixed one concurrency regression in `ff5569486`: IPv6 pinhole renewal could read an enabled entry, race an operator disabling it, and reinstall the nft rules after the disable completed. Pinhole add/refresh, enable/disable, and removal now share the forwarding write mutex, keeping PatchDB state and dataplane updates ordered.

Verified with `make start-core-test` (533 passed, 3 ignored), `cargo check --tests -p start-core --features=test`, `cargo check -p start-tunnel`, `make start-core-format-check start-tunnel-format-check`, the StartTunnel Angular build, and `projects/start-docs/build.sh`.

projects/start-tunnel/docs/src/published-ports.md
```diff
@@ -20,7 +20,7 @@ Because each device has its own IPv6 address, two different devices can both pub
...
...e shows two tables: **Manual** ports you added by hand, and **Automatic** port uses created by connected devices. Automatic entries include PCP/UPnP port mappings and SNI hostname routes; their kind identifies which one they are. A row's **External IP** is your VPS's public IPv4 (a v4 published port) or the device's IPv6 GUA (a v6 pinhole). You can enable, disable, or remove either; automatic entries are owned by the device that created them and may be re-created if you remove one while the device still wants it. A disabled entry stays unavailable until you enable it again, including while its device renews the mapping. Manual ports stay until you delete them. Each external port — or, on a port shared by SNI, each hostname — holds a single entry, either manual or automatic. When a device requests one that a manual entry already holds, your entry keeps it: the request succeeds if it asks for the same mapping and is refused if it asks for a different one. A device can point an automatic entry it holds at a different port on itself. Automatic port uses are lease-based: one that stops being renewed (its device went offline or no longer wants the port) expires and is removed on its own.
```
dr-bonez:
```suggestion
The `Published Ports` page shows two tables: **Manual** ports you added by hand, and **Automatic** port uses created by connected devices. Automatic entries include PCP/UPnP port mappings and SNI hostname routes; their kind identifies which one they are. A row's **External IP** is your VPS's public IPv4 (a v4 published port) or the device's IPv6 GUA (a v6 pinhole). You can enable, disable, or remove either; automatic entries are owned by the device that created them and may be re-created if you remove one while the device still wants it. A disabled entry stays unavailable until you enable it again, including while its device renews the mapping. Manual ports stay until you delete them. Each external port — or, on a port shared by SNI, each hostname — holds a single entry, either manual or automatic. When a device requests one that a manual entry already holds, the manual entry keeps it: the request succeeds if it asks for the same mapping and is refused if it asks for a different one. A device can point an automatic entry it holds at a different port on itself. Automatic port uses are lease-based: one that stops being renewed (its device went offline or no longer wants the port) expires and is removed on its own.
```

dr-bonez (approved):

PR #3999, opened 2026-09-18
feat(core,web): let an operator pin a gateway's WAN IP, and keep public-IP opt-ins across a renumber

Dominion5254:
## Why

StartOS derives a gateway's WAN IP by asking the router over UPnP and, failing
that, by reading back the address an outbound request appeared to come from.
Those answer different questions: the router is authoritative about where
inbound traffic arrives, while an echo service can only report where outbound
traffic left. Where the two differ and UPnP cannot settle it — multi-WAN
routing, an ISP translating egress from a pool separate from the mapped inbound
address, a cloud floating ingress address, a router with UPnP off or reporting
a stale one — detection names an address the server is not reachable at, and
every address derived from it points somewhere wrong: offered public addresses,
the port-forward rules the UI asks for, the TLS served name, and any address a
service advertises to peers. There was no field an operator could correct;
discovery rewrote it on every poll.

Public-IP exposure is also keyed on the literal address. `enabled` and
`gua_wan` are written by operator action and nothing moved them when the
address changed, so a rolled lease or a moved prefix left the replacement off
by default, dropped its port forward, and said nothing. A manual override would
have been the same event, so that had to be fixed first.

## What

- **`wan_ip_override`** on `NetworkInterfaceInfo`, beside `name` and `secure`.
  `wan_ip()` resolves it over the detected value; every consumer reads it.
  Discovery keeps writing `ip_info.wan_ip`, so clearing restores it.
- **`net gateway set-wan-ip` / `unset-wan-ip`** — command line only. Unroutable
  addresses and outbound-only gateways are refused. `list` marks a pinned
  address `(manual)`. The web UI reads the override, so the gateway table, the
  DNS checks and domain validation all test the pinned address, but it offers
  no way to set one: this corrects a fact about your network that StartOS got
  wrong, not a setting to reach for.
- **`migrate_renumbered_ips`** moves an opt-in to the address that replaced it,
  from `update_addresses`, so a detected renumber and a manual override are the
  same event. It moves only within the same gateway, port and family, and only
  when exactly one public address is offered there before and after.
- **`carry_forward`** keeps a downed gateway's IP rows so the migration also
  survives a link drop, a router reboot, and boot. A downed gateway keeps
  offering its last addresses until deleted; its port forwards lapse while it
  is down.

## Compatibility

No db migration: the field has a serde default and an older binary ignores it.
Bindings regenerated, backend and CLI strings in five locales, changelog under
0.4.0.2, `gateways.md` gains a WAN IP section.

## Testing

Unit tests cover override precedence, the legacy row, validation, every
migration shape, and the full `update_addresses` path for a detected renumber,
an override set and cleared, a second gateway, a range, a link drop, and a
deleted gateway, with port forwards checked in each. `cargo test`, `cargo fmt`,
the UI build, i18n and prettier are clean. `http_reader::main_test` fails on
master too; it fetches a live URL.

Not yet exercised on a box. To do before merge:

- [x] `set-wan-ip` then `unset-wan-ip`; addresses and port forwards follow, and
      the UI's gateway table and domain validation track the pinned address
- [x] Reboot; override and enabled address survive
- [x] Unplug and replug the gateway; a changed address keeps its opt-in

## Start here

`net/host/binding.rs` `migrate_renumbered_ips`, `net/host/mod.rs`
`carry_forward`, `net/gateway.rs` `set_wan_ip_override`.


shared-libs/crates/start-core/src/db/model/public.rs
```diff
@@ -262,6 +262,15 @@ pub struct NetworkInterfaceInfo {
     pub name: Option<InternedString>,
     pub secure: Option<bool>,
     pub ip_info: Option<Arc<IpInfo>>,
+    /// The public IPv4 the operator pinned for this gateway, taking precedence
+    /// over the discovered `ipInfo.wanIp`. Discovery asks the router over UPnP
+    /// and otherwise infers the address from outbound traffic, which names the
+    /// wrong one wherever ingress and egress take different paths — multi-WAN
+    /// routing, an ISP translating egress from a pool separate from the mapped
+    /// inbound address, a floating ingress address with a separate egress
+    /// gateway. Clearing it restores the discovered address.
```
dr-bonez: way too much context for a doc comment

shared-libs/crates/start-core/src/net/host/binding.rs
```diff
@@ -132,6 +132,115 @@ impl DerivedAddressInfo {
...
+    /// operator action.
+    ///
+    /// An opt-in moves only where the exposure decision is the same one: the same
+    /// gateway, the same port, the same address family, with `previous` and
+    /// `available` each offering exactly one public address there. A new gateway,
+    /// a new port, or several candidates on either side leaves it opt-in.
+    ///
+    /// `available` must already hold the freshly computed set and `previous` the
+    /// set it replaced. A GUA's `public` flag is re-projected from the migrated
+    /// `gua_wan` so a migrated address is public in the same pass.
```
dr-bonez: stop with the mannered prose

dr-bonez (changes requested):
fix the mannered prose throughout the comments and docs

shared-libs/crates/start-core/src/db/model/public.rs
Dominion5254: trimmed to one line

shared-libs/crates/start-core/src/net/host/binding.rs
Dominion5254: Made more direct and concise

ok:
Field report for the renumber half of this PR, from a 0.4.0.1 server on a home
line whose ISP reassigns the public IPv4 daily.

A service exposes a raw UDP port and had its Public address on the router
gateway switched on. After the overnight reassignment the binding's
`addresses.enabled` still held `<old IP>:<port>`, `available` offered
`<new IP>:<port>` with the switch off, and `table ip startos` `prerouting` held
only the two source-restricted DNAT rules for that gateway. A probe from the
internet reached the server (the router forward was intact) and created an
`[UNREPLIED]` conntrack entry with the server, not the container, as the reply
side. Switching the address back on restored the forward within seconds.

That matches what `migrate_renumbered_ips` fixes: `public_ip_on` keys the
opt-in on the literal `SocketAddr` (`net/host/binding.rs`), the WAN row is
rebuilt from `ip_info.wan_ip` on every `update_addresses`, and
`port_forwards` is derived from `enabled_addresses()`, so the forward drops
with the opt-in. On a dynamic line that is a daily outage per exposed port.

Two things worth knowing when this lands:

- The stale `<old IP>:<port>` entry stays in `enabled` forever today. The
  migration moves it, so a server that renumbers after this ships is clean,
  but one that renumbered before keeps the dead entry. Harmless, only noted.
- The entries a peer creates while the forward is down survive its return, and
  keep that peer pointed at the host. That is a separate gap in `forward-port`
  (flush only on `UNDO`), filed separately as #4046.

Happy to rebase this onto current master if that helps get it in.

Issue #3946, opened 2026-09-15
Set Outbound Gateway is ignored while a system-wide default is pinned — the pinned default's priority-75 `from all` rule precedes the per-service priority-100 rule (net/gateway.rs:2802, net/net_controller.rs:1093)


MattDHill:
### Environment

- StartOS 0.4.0.2, git `a30d84d21f1544b54d3861aea065cf4ba7217250` (master, 2026-09-04); the same code is on `master` @ `81c0b5a08`.
- x86_64, release image (empty `ENVIRONMENT.txt`), Purism Librem Mini v2.
- Area: networking — outbound gateway selection (`ip rule` policy routing).
- Gateways on the box: `eno0` (home ISP, WAN 75.70.169.208) and `wg0` (StartTunnel 1.3.0, WAN 64.23.194.12). System-wide outbound (**System › Gateways › Outbound Traffic**) pinned to `eno0`.

### What happens

Pinning a service's outbound gateway — **Actions › Set Outbound Gateway** or `start-cli package set-outbound-gateway <pkg> wg0` — does nothing while the system-wide default outbound is pinned to a gateway. The service's traffic keeps leaving through the system default.

Observed with the `stalwart` package (container IPv4 `10.0.3.9`), `outboundGateway: "wg0"` stored in the DB, `defaultOutbound: "eno0"` on the server:

```
$ start-cli package attach stalwart -s <guid> -- curl -4 https://api.ipify.org
75.70.169.208            ← eno0's WAN IP, not the tunnel's 64.23.194.12
```

A restart of the service does not change it. The docs promise the opposite (`projects/start-os/docs/src/outbound-vpn.md`, "Route Individual Services Through VPN"): *"A service's own setting always takes precedence over the system-wide default — whether that default is Auto or a pinned gateway."*

### What should happen instead

The per-service pin wins over the system-wide pin, as documented: with the service pinned to `wg0`, `curl -4 https://api.ipify.org` from its container returns `64.23.194.12`, and the kernel's route lookup for the container's source address resolves via `wg0`.

### Steps to reproduce

1. A box with two gateways, e.g. the router and a StartTunnel.
2. **System › Gateways › Outbound Traffic**: pin the system-wide default to the router (`start-cli net gateway set-default-outbound eno0`).
3. On any service: **Actions › Set Outbound Gateway** → the tunnel (`start-cli package set-outbound-gateway <pkg> wg0`). Restart the service for good measure.
[...]

PR #4006, opened 2026-09-18
fix(start-core): fail closed for unavailable outbound gateways

### LINKED ISSUE #4003: A selected outbound gateway that goes down fails open to the ISP link — its rule is removed instead of rejecting (net/gateway.rs:2789)
### Environment

StartOS master at 67123b711 (behaviour unchanged since 0.4.0). Area: networking, outbound gateways. Found by reading code, not on a device.

### What happens

When the gateway selected for outbound traffic is unavailable, traffic falls back to whatever `main` would pick, which on most servers is the ISP link.

This holds at both layers:

- System-wide (`System > Gateways > Outbound Traffic`): a pinned gateway that is not connected has its priority-75 catch-all removed, so the server and every service route by `main` again.
- Per-service (`Set Outbound Gateway`): when the gateway's interface is gone the service's rule is not installed, and when the interface exists but is down its per-gateway table is empty, so the lookup falls through to the system-wide default and then to `main`.

A user who pinned a commercial VPN to hide their address therefore sends traffic under their real IPv4 for as long as the VPN is down, with nothing in the UI saying so.

Selecting a gateway should be able to mean "this gateway or nothing". The IPv6 side already works this way: a selected gateway that cannot carry IPv6 gets a `blackhole default` in its own table so IPv6 is dropped rather than leaking. IPv4, and the gateway-is-down case in both families, should get the same treatment, for the system-wide setting and the per-service one alike.

Whether that is the only behaviour or an option is a product decision. A server whose only pinned gateway dies also loses registry access and OS updates, which may be what the user wants or may not.

- `shared-libs/crates/start-core/src/net/gateway.rs:2789` (`apply_default_outbound`): when the pinned gateway has no `ip_info` it logs `default outbound gateway {gw_id} is not connected` at :2794 and leaves `desired_75` empty, so `reconcile_outbound_rules` deletes the existing catch-all.
- `shared-libs/crates/start-core/src/net/net_controller.rs:1082`: the per-service rule is only added when `if_nametoindex` resolves the gateway. Otherwise no rule exists for the service.
- `apply_policy_routing` writes the IPv4 per-gateway table as a single `default dev <iface>` route. The kernel removes it when the interface goes down, leaving the table empty, and an empty table falls through to the next rule.

One shape for the fix: keep the rule in place while a gateway is selected and have the table end in an `unreachable default` at a worse metric than the real one, so a missing real default rejects instead of falling through. For the per-service rule a `from <source> unreachable` rule one priority below it does the same without touching the table.

Surfaced while reviewing #4002, which is left fail-open to stay consistent with the system-wide setting.

helix-nine:
Closes #4003

## What

A selected outbound gateway now behaves as a kill switch. If its interface disappears or its routing table has no usable route, the traffic assigned to it is rejected instead of falling through to the ISP. Returning the system-wide setting to **Auto**, or a service to **System default**, removes the rejection.

- an `unreachable` rule directly after the system-wide gateway lookup (76 after 75) and after each service's lookup (71 after 70), in both families
- WireGuard transport exemptions stay ahead of the system-wide rejection, so a disconnected tunnel can reconnect
- the rules reconcile with the existing gateway lifecycle, including service startup and stale-rule cleanup
- a disconnected pinned gateway stays selected in the UI, system-wide and per service, instead of reading **Auto** or **System** while its kill switch is active

## Inbound connections survive the kill switch

A rejection at the end of a selection also caught two lookups that are not outbound traffic. With the selected gateway down, every connection arriving from off-link through another gateway died. Both were reproduced with real packets in a network namespace, and each needed its own fix.

- **Host-terminated listeners.** The kernel routes a local reply before it is marked: a SYN-ACK is looked up with mark 0 and only its source address. IPv4 used to get through by falling into whatever default came later, and the rejection removed that. Same mechanism as #3632. The priority-60 source-address rules now cover every routable address of a gateway in both families and match only unmarked lookups.
- **Connections forwarded to containers.** StartOS runs `rp_filter=2`, and the reverse-path lookup for a forwarded packet is sourced from the container's address, which both rejections match. The first packet of an inbound connection now carries its mark as well as its conntrack entry, and each gateway interface sets `src_valid_mark`, so the filter follows the reply path. It is per interface because on `all` it would drop SNI-diverted replies as martians.

A lookup is installed before the rejection behind it and removed after it, so switching a selection never leaves the rejection as the only match. A service lookup that fails to install still gets its rejection.

## The server's own IPv6

IPv6 retries a failed lookup after choosing a source address. That second lookup could match the priority-60 source rule of another gateway and leak a connection the server opened around the selected gateway.

While a system-wide gateway is selected, `mangle_output` marks each new local connection with the dedicated mark `0x00540002`. Source rules require mark zero, so the repeated lookup passes them and reaches the existing system selection and rejection at 75–76. The dedicated mark does not match a gateway's priority-48 reply rule, so a specific route in `main` still wins; only a default follows the selection. The mark is packet-only and limited to `ct state new`, so it does not move an established flow.

`mangle_output` restores connection marks and jumps to a separate `mangle_local_outbound` chain. The gateway coordinator owns the former and the default-outbound reconciler owns the latter, avoiding concurrent writers.

## Rule order

| Priority | Rule | Serves |
| --- | --- | --- |
| 48 | `fwmark T to 2000::/3 lookup T` (IPv6, WireGuard) | tunnel replies ahead of `main` |
| 49 | `fwmark 0x540001 lookup 5344` | SNI divert |
| 50 | `lookup main suppress_prefixlength 0` | specific routes |
| 51 | `fwmark T lookup T` | marked replies and reverse-path checks |
| 60 | `from <gateway address> fwmark 0 lookup T` | first lookup of unmarked local replies |
| 70, 71 | `from <container> lookup T`, then `unreachable` | service selection and kill switch |
| 74 | `fwmark <wg fwmark> lookup main` | WireGuard transport |
| 75, 76 | `lookup T`, then `unreachable` | system selection and kill switch |
| 1000, 1100 | `lookup main`, `lookup default` | Auto |

`shared-libs/crates/start-core/policy-routing.md` states the invariants the ladder has to hold, derives this order from them, and records where each rule came from. `src/net/policy_routing_model.sh` builds the ladder from the priority constants in a throwaway network namespace and checks every invariant in seven selection states. Run with deliberately wrong priorities it reproduces the bugs fixed by #3911, #3946 and #3996 as well as this one.

## Also in this PR

The SNI divert table moves from 1344 to 5344. Gateway tables are `1000 + ifindex`, so 1344 sat inside their range. StartWRT already used 5344 because it numbers its tables by VLAN tag, and it now reads the shared constant.

## Documentation

The StartOS outbound VPN guide documents selected gateways as kill switches, distinguishes them from **Auto** and **System default**, and says the LAN and connections arriving through other gateways keep working. The unreleased changelog records the security behavior.

## Testing

Local and CI:

- `cargo check -p start-core -p startwrt-core`
- `cargo test -p start-core --features=test --lib -- net::` on cleanup head `89a7189` (390 passed, including `ladder_satisfies_the_routing_invariants`)
- full `make start-core-test` before the cleanup (545 passed, 3 ignored)
- pinned rustfmt check, prettier, mdBook build, UI build/typecheck, and the full GitHub Actions matrix
- the real `reconcile_source_rules`, `reconcile_outbound_rules` and `gc_policy_routing` exercised in a network namespace: idempotent in both families, clean round trip
- real-packet namespace test (router, server, container behind DNAT): host-terminated and forwarded inbound both fail on the original commits with the gateway down and pass on the final ladder; container egress is still rejected

Device test used the x86_64 artifact from run `35415042834`, built from synthetic merge `5b4c7ad` whose second parent is cleanup head `89a7189`, on a StartOS 0.4.0 VM with two independent WireGuard gateways:

- With gateway A selected system-wide, IPv4 from the server and a container used A. Since A intentionally had no IPv6, both IPv6 callers were blocked; the packet counter on gateway B remained zero.
- With A disconnected, server and container IPv4/IPv6 all failed closed. Attempt counters incremented, while postrouting counters on the ISP path and gateway B remained zero.
- A reconnected in two seconds without changing the selected setting, and server IPv4 resumed with source `10.231.1.2`.
- With only bitcoind pinned to A and the system on **Auto**, bitcoind used A for IPv4 and blocked IPv6, while LND and the server retained working IPv4/IPv6. After A disconnected, bitcoind's source-scoped attempt counters incremented and both fallback egress counters remained zero; LND and the server were unaffected.
- While A was unavailable, inbound through B remained usable for the StartOS UI (HTTP 200), Hello World behind StartOS TLS (HTTP 200), Bitcoin's plain forwarded port, and LND's self-TLS/SNI port over both IPv4 and IPv6 where published.
- The service rejection survived a package restart and a `startd` restart. The latter replaced bitcoind's container and moved the IPv4/IPv6 rejection rules to its new addresses. A separate `startd` restart with disconnected A selected system-wide reconstructed the global rejection and local mark.
- Clearing the service to **System default** and the system to **Auto** removed their lookup/rejection rules; `mangle_local_outbound` emptied, and IPv4/IPv6 resumed through available defaults.
- The compiled UI showed `Current: Kill Switch A`, and the action dialog retained `Kill Switch A`, while that gateway was disconnected.
- `src_valid_mark` remained 0 on `all` and 1 on each gateway. Priority-60 rules appeared in both families with `fwmark 0`.
- With B selected system-wide and an on-link global IPv6 target reachable through A, a new server connection took A's specific route: 10 packets on A and 0 on B. This exercises the cleanup's dedicated local mark rather than a gateway table mark.


helix-nine:
Retested cleanup head `89a7189` using the exact x86_64 artifact from run `35415042834` (artifact merge `5b4c7ad`, second parent `89a7189`) on the prepared StartOS 0.4.0 two-gateway VM.

**Result: pass.**

- System pin, gateway up: server/container IPv4 used A; A has no IPv6, so both IPv6 attempts were blocked and B's egress counter stayed 0.
- System pin, A disconnected: server/container IPv4+IPv6 failed closed. Attempt counters incremented; ISP-v4 and B-v6 postrouting counters stayed 0.
- A reconnected in 2 seconds without changing the setting; outbound resumed with source `10.231.1.2`.
- Service-only pin: bitcoind failed closed in both families after A disconnected, with source-scoped fallback counters at 0; LND and host traffic remained usable.
- Inbound through B remained usable for the UI, StartOS-terminated TLS, Bitcoin's plain forwarded port, and LND's self-TLS/SNI port (IPv4 and IPv6 where published).
- Package restart and `startd` restart preserved the disconnected service pin and rebuilt rejection rules for the replacement container addresses. A separate `startd` restart rebuilt the disconnected system-wide pin and local mark.
- Clearing to **System default**/**Auto** removed 70–71/75–76 and emptied `mangle_local_outbound`; both families resumed.
- The compiled UI retained `Current: Kill Switch A` and the dialog selection while A was disconnected.
- Cleanup-specific check: with B selected, a new host connection to an on-link global IPv6 target on A took A's specific route (10 postrouting packets on A, 0 on B), confirming the dedicated mark does not hit B's priority-48 reply rule.

Observed source rules carried `fwmark 0`; `src_valid_mark` was 0 on `all` and 1 per gateway. Targeted networking tests also pass on the cleanup head: 390 passed.

I updated the PR body with the final rule shape and device results.

dr-bonez (approved):

Issue #4010, opened 2026-09-18
Service memory oversubscription wedges the host terminally — all four protection layers miss (oomd swap floor vs fragmented leaves, PSI trigger unreachable, MemoryMin plane starved, kernel OOM scoring inverted); measured live on 0.4.0.2


stupleb:
### Environment

StartOS 0.4.0.2 (served UI `gitHash a30d84d21f1544b54d3861aea065cf4ba7217250`), x86_64-nonfree, Intel i5-7500T (4 cores), 15.5 GiB RAM, zram swap 3.9 GiB, systemd 257 (257.13-1~deb13u1). Area: service runtime / host resource management. Same box and topology as #3885 (52 packages; leaf swap measured fragmented below 5% of SwapTotal at idle). Chain/channel services were stopped for the test; the load was synthetic and self-terminating.

### What happens

Induced memory oversubscription replicating the measured fleet topology (fragmented, every `services.slice` leaf below the systemd-257 swap-kill floor) wedged the host for ~15 of 24 minutes: SSH banner timeouts, HTTPS/RPC unreachable, UI shows "State unknown" — indistinguishable from a powered-off box except that ICMP (kernel-serviced) still answered. No protection layer intervened. The box recovered **only** because the synthetic load carried in-process deadlines; production services carry none, so in production this state is terminal and ends in an operator power cycle — the pathway that caught a bitcoind volume mid-install on 2026-09-01 (#3885).

This run began eight minutes after the 2026-09-18 follow-up in #3885 modeled the accounting boundary on a VM; the numbers below are that model observed live, on the fleet topology it predicts (52 packages, 4-core / 15.5 GiB), plus two failure layers outside that comment's scope: host-plane starvation despite `MemoryMin`, and kernel OOM victim-scoring inversion.

Four layers failed, each measured:

[...]

PR #4011, opened 2026-09-18
fix(start-os): disable zram by default

### LINKED ISSUE #3885: systemd-oomd never fires on zram systems — ManagedOOMSwap reads 0 B while the kernel accounts 3.4 GiB, so the host still swap-death wedges
## Summary

The container-memory protection added in #3304 (0.4.0) does not engage. On a box under memory pressure, `systemd-oomd` reports `Swap Usage: 0B` for `/services.slice` while the kernel accounts 3.4 GiB of swap in that exact cgroup. `ManagedOOMSwap=kill` therefore never selects a victim, and the host degrades into the full whole-host swap-death wedge that #3304 was written to prevent — unresponsive to SSH, HTTPS and even ARP, with nothing in the journal.

## Environment

StartOS 0.4.0.2, x86_64-nonfree, systemd 257 (257.13-1~deb13u1), Intel i5-7500T (4 cores), 15.5 GiB RAM, 52 packages installed. Swap is zram (`/dev/zram0`, 3.9 GiB).

## Evidence

Three readings of the same quantity at the same instant:

```
kernel file:   3638161408   (3.39 GiB)
as oomd user:  3638161408
oomctl says:   Swap Usage: 0B
```

`oomctl`, with system swap at 97% — above the 90% `SwapUsedLimit`:

```
Swap Used Limit: 90.00%
System Context:
    Memory: Used: 6.4G, Total: 15.4G
    Swap: Used: 3.8G, Total: 3.8G
Swap Monitored CGroups:
    Path: /services.slice
        Swap Usage: 0B
```

Effective config is correct, and oomd is active and enabled:

```
ManagedOOMSwap=kill
ManagedOOMMemoryPressure=kill
MemoryMax=15534653440   MemoryCurrent=12400259072
```

Ruled out: file permissions (`-r--r--r-- root root`, and reading as `systemd-oom` returns the correct value); oomd sandboxing (`ProtectControlGroups=no`, `PrivateMounts=no`); cgroup delegation (`Delegate=no` in effect — see secondary defect).

The pressure trigger cannot compensate: PSI on the slice sits at `some avg10=6.43 / full avg10=4.67` against a 60% limit, so `ManagedOOMMemoryPressure=kill` does not fire either. **Both triggers are inert simultaneously.**

## Why zram makes this structural

Two compounding problems beyond the 0 B reading:

1. **`SwapUsedLimit` carries no signal on zram.** zram pages are not faulted back in until touched, so swap-used stays pinned at 97–100% as a steady state. Measured at idle: load 2.6, 9.0 GiB available, swap still 97%. The 90% threshold is therefore satisfied permanently — meaning that if the 0 B reading were simply fixed, oomd would flip from never killing to killing indiscriminately at idle. The swap control is the wrong lever here in either state.

2. **The memory cap arithmetic ignores zram.** startd logs at boot:

   ```
   service containers (services.slice) capped at 14815 MiB
   (high 13334 MiB); 1024 MiB reserved for the host plane
   of 15839 MiB total
   ```

   zram's compressed pages live in that same RAM but are not charged to the slice, so the containers can commit their full 14.5 GiB cap *plus* the compressed footprint of up to 3.9 GiB of swapped pages — eating into the 1 GiB host reservation that keeps sshd and startd alive.

## Impact

This is the failure mode that costs data. The box becomes unreachable at layer 2, so it is indistinguishable from powered-off from outside; the operator power-cycles it; any service mid-write loses its database. In our case the wedge landed 4m33s into a `bitcoind` install (`Created install backup for bitcoind`, volume.rs:112).

Nothing is ever logged, because nothing crashes — the machine simply stops making forward progress, journald included. The empty journal is the signature, which makes this very hard to diagnose from the field.

## Secondary defect (same file, one line)

systemd rejects line 21 of `services.slice` at every boot:

```
systemd[1]: /usr/lib/systemd/system/services.slice:21:
            Delegate= setting not supported for this unit type, ignoring.
```

`Delegate=` is not applicable to `.slice` units; `systemctl show services.slice -p Delegate` confirms `Delegate=no`. The file's comment states this setting is what "hands the slice's cgroup subtree to lxc so systemd won't fight the per-container cgroups lxc creates underneath it" — an expectation that is not being met. Happy to split this out if preferred.

## Suggested direction

Since both oomd triggers are inert and the swap trigger is structurally unsuited to zram, a watchdog in startd over `services.slice`'s own `memory.swap.current` / `memory.pressure` — killing the heaviest container cgroup directly — would not depend on oomd resolving the accounting discrepancy. Reconsidering zram-as-swap on this profile is worth weighing alongside it.

helix-nine:
## Summary

- default fresh installs to ZRAM disabled
- migrate existing servers to the disabled setting, including servers that already ran the published `0.4.0.2` migration revision
- document the safer default in the CLI reference, changelog, and release notes

The `oomctl` 0 B line is a cached display value rather than the value oomd uses to choose a victim. The containment gap is instead that ZRAM's physical backing is not charged to `services.slice`, allowing resident service memory plus compressed swap to consume the RAM reserved for the host plane. Disabling ZRAM restores `MemoryMax` as the physical service-memory boundary.

Closes #3885

## Testing

- `cargo test -p start-core --features test version::`
- `make start-os-format-check`
- `mdbook build` in `projects/start-os/docs`
- `./scripts/manage-release.sh pre-check start-os` (all repository checks passed; local release credentials/tool checks reported the expected missing GPG key and `s3cmd`)


stupleb:
Two scope questions, both from your #3885 analysis:

1. **`vm.overcommit_memory=1` (your step 2).** The migration removes ~RAM/4 of `SwapTotal` from every existing server, and you measured a 4 GiB reservation failing at a 19.5 MiB footprint under the default heuristic once swap is gone. Deferred deliberately, or worth shipping in the same revision so `CommitLimit` doesn't shrink fleet-wide ahead of its companion?

2. **Default-off vs. bounded zram.** You tested `zram mem_limit=L` + `MemoryMax = R − H − L` as sound (cgroup-constrained OOM in 1.7 s, no global OOM). On idle-heavy fleet boxes the compressed cold-page credit is real capacity — the 15.4 GiB box in #4010 held ~3.5 GiB of cold service anon in zram at idle. Was bounded-zram weighed as the *default* (with off as the fallback), or is the coupled invariant the disqualifier? Small-RAM boards lose the most from default-off.

Minor: `MemoryHigh` remains set on `services.slice` (13.0 GiB on the #4010 box); your step 4 measured it throttling >30 s instead of failing fast — this revision or follow-up?

Happy to run the #4010 reproduction plus a large-reservation smoke test (JVM/mmap) on the fleet box against this branch — zram-off and bounded-zram both — if useful.

dr-bonez (approved):

PR #4016, opened 2026-09-19
fix(start-os): bound service-container teardown waits

### LINKED ISSUE #3577: [bug]: stale vhost ProxyTarget survives a container recreate — service UI permanently unreachable after an update
## Summary

After a routine package update, a service's Web UI became unreachable on every address and stayed that way. The OS reverse proxy was still routing that interface to the **previous** container's IP, which no longer exists on `lxcbr0`. Because a `ProxyTarget` keyed on a dead address sits *ahead* of the live one and there is no fallthrough, the interface is permanently black-holed until the container is rebuilt.

Reported against **StartOS 0.4.0** (release, not beta), 2026-07-25. Package: `nextcloud`, `33.0.6:1 → 33.0.6:2`. Recovered by rebuilding the service container.

The vhost mechanism below is read directly off the source and accounts for every observed symptom. How the target came to outlive its container is a hypothesis — flagged as such, with the discriminating evidence listed at the end.

Line refs are against `6850e990c`.

Worth noting: that package diff is inert — an SDK version bump the package makes no runtime use of, an npm `overrides` lockfile dedupe, and the version string. No interface, port, image, or daemon change. Nothing about the payload is implicated; this looks like a property of an update cycle running at all.

## Symptoms

Browser (Firefox), against the SSL vhost:

```
Secure Connection Failed

The page you are trying to view cannot be shown because the authenticity of the received data could not be verified.

Error Code: PR_END_OF_FILE_ERROR
```

OS journal, spamming for as long as the condition persists:

```
DEBUG start_core::error: shared-libs/crates/start-core/src/error.rs:675: Network Error:
  0: 10.0.3.40:80: Host is unreachable (os error 113)
  1: Host is unreachable (os error 113)
```

**Nothing in the service logs** — correct; the service layer is never in the path for this failure, which is what makes it hard to diagnose from the UI.

## What the error traces to

`10.0.3.40` is the service container on `lxcbr0`, `:80` its internal UI port. The message is the OS reverse proxy's own backend connect:

- `net/vhost.rs:1058-1060` — `ProxyTarget::preprocess` → `TcpStream::connect(self.addr).with_ctx(|_| (ErrorKind::Network, self.addr))`, which formats exactly as observed.
- That address is `(self.ip, *port)` (`net/net_controller.rs:374`), where `self.ip` is the container IP captured **once**, at `NetService` construction — read from `LxcContainer::ip()` and handed to `create_service` at `service/persistent_container.rs:284-288`.
- `EHOSTUNREACH`, **not** `ECONNREFUSED`: nothing answers ARP for `.40` on the bridge. The container at that address is *gone*, as opposed to up-but-not-listening. This is the key discriminator — it rules out "the service crashed".
- `preprocess` runs inside `get_config`, i.e. during the ClientHello (`net/vhost.rs:1541`). A failed backend connect returns `None` and the socket is dropped **before any TLS bytes go back** — which is precisely `PR_END_OF_FILE_ERROR` client-side, and why a backend routing failure presents to the user as a TLS/certificate error.

## Why a stale target is sticky, and shadows the live one

A container cannot keep its IP across a recreate: `lxc/config.template:27-29` sets `lxc.net.0.type/link/flags` and no `hwaddr`, and each container gets a fresh GUID — so every recreate means a new random MAC, a new DHCP lease, and a different `10.0.3.x`. **Any vhost target that outlives its container is therefore guaranteed to point at a dead address.** There is no benign version of this state.

Three properties then combine to make it permanent rather than transient:

1. **A new IP creates a new map entry instead of replacing the old one.** Targets are keyed by the `ProxyTarget` itself in `VHostServer::add` (`net/vhost.rs:1718+`), and `PartialEq` includes `addr` (`net/vhost.rs:938-950`). So the recreated container's target coexists with its predecessor under the same `(hostname, port)`.
2. **First alive match wins, with no fallthrough.** Routing does `.filter(|(_, e)| e.alive()).find(|(t, _)| t.0.filter(metadata))` (`net/vhost.rs:1498-1506`), and if the chosen target's `preprocess` fails, `get_config` returns `None` — it never tries the next matching target. `net vhost dump-table`'s `ACTIVE` column (`idx == 0`) is displaying exactly this stack.
3. **Liveness is ownership, not reachability.** `alive()` is `rc.strong_count() > 0` (`net/vhost.rs:1380`) against the `Arc<()>` held in the owning `NetService`'s `HostBinds.vhosts`. A target is "alive" as long as someone holds the handle, regardless of whether its backend has existed for the last hour.

Net effect: one `NetService` that outlives its container permanently black-holes the interface, and every subsequent connection attempt re-logs the error — hence the journal spam.

## How the target could have outlived its container (hypothesis)

Not proven for this incident — no journal from the reporter's box yet. But the ordering hazards are visible in-tree and would produce exactly this end state:

1. `ServiceRef::shutdown` (`service/mod.rs:187-207`) does `Arc::try_unwrap` on the `Service` and then on the `ServiceActorSeed`, and **returns `Err` without ever calling `persistent_container.exit()`** if any other reference is still held. A stray reference thus silently converts into a skipped teardown.
2. Teardown then falls to `Drop`, and all three relevant impls spawn *detached* tasks — an ordered teardown becomes an unordered race:
   - `NetService::drop` → `remove_all()` (`net/net_controller.rs:1364`)
   - `PersistentContainer::drop` → `destroy(None)` (`service/persistent_container.rs:611`)
   - `LxcContainer::drop` → rootfs unmount + `manager.gc()` (`lxc/mod.rs:661`)
3. `PersistentContainer::destroy` awaits `rpc::Exit` with **no timeout** (`service/persistent_container.rs:455-458`) *before* it reaches `net_service.remove_all()` (`:466`) and `lxc_container.exit()` (`:481`). Compare `ServiceRef::shutdown`, which bounds its equivalent with `tokio::time::timeout(Duration::from_secs(30), …)` (`service/mod.rs:182`). An `Exit` RPC to a runtime that is already gone can block that future indefinitely — so `remove_all()` never runs even though the LXC gc has already reaped the container.

(3) on its own is sufficient: container reaped, `Arc<()>` never released, stale target permanently in front of the live one.

## Why "rebuild container" recovers it

`service::rebuild` is `ctx.services.load(…, LoadDisposition::Retry)` (`service/mod.rs:822`) — a full reload that reconstructs the `PersistentContainer` and re-runs the bind cycle, dropping the old `NetService` and its `Arc<()>`s. Matches the reporter's "rebuilding the service container restored UI access."

## Expected behavior

- **A vhost target must not outlive the container it points at.** Releasing it is `NetService` teardown, so container teardown needs to reach that on a path that can be neither skipped nor blocked indefinitely — bound the `rpc::Exit` await in `destroy` the way `shutdown` already bounds its own, and/or release the vhost rcs *before* the RPC exit rather than after.
- **`ServiceRef::shutdown` should not skip `persistent_container.exit()` on `Arc::try_unwrap` failure** — log the held reference and tear down anyway, rather than leaving the container and its network bindings fully intact.
- **A `ProxyTarget` whose backend is unreachable should not indefinitely shadow a live sibling under the same key.** Either evict it after repeated connect failure, or let `get_config` fall through to the next matching target instead of returning `None`. This one is worth doing independently of the lifecycle fix: it bounds the blast radius of *any* future cause of a stale target, converting a permanent outage into a transient one.
- **Rate-limit the per-connection `Host is unreachable` error.** At one line per connection attempt it evicts the journal history needed to diagnose the trigger.

## Evidence still to collect

Discriminating journal lines around the update window:

```
"ServiceActor held somewhere after actor shutdown"
"ServiceActorSeed held somewhere after actor shutdown"
"was ungracefully dropped. Cleaning up dangling containers"
```

And live state, if it recurs — captured **before** rebuilding, which destroys it:

```
start-cli net vhost dump-table    # ≥2 rows under one FROM; ACTIVE = the dead 10.0.3.x
lxc-ls -f                         # the container's actual current IP
ip neigh show dev lxcbr0          # no neighbour for the stale IP
```

## Not included here

The same report carries an IPv6-only symptom that persisted *after* the container rebuild — the box's LAN ULA at the interface's assigned SSL port refusing connections. Data since collected from that box rules out the obvious server-side causes: the vhost **is** listening on exactly that address and port, and the base `ip6 startos` ruleset defines no `input` chain, so nothing filters inbound to a locally-bound port. Either it was transient state that has since healed, or the failure is on the LAN path / client side. Investigated separately; not part of this report.

helix-nine:
## Summary

- bound the drop-path `rpc::Exit` request and RPC server join to 30 seconds each
- continue into `NetService` cleanup after either wait times out, releasing reverse-proxy targets before container teardown proceeds
- cover the bounded drop path and unbounded explicit path with paused-time regression tests
- document the cleanup fix in the StartOS changelog

`PersistentContainer::exit` passes no timeout to teardown; only `Drop` supplies the 30-second bound. Normal uninit therefore remains unbounded.

## Verification

- `cargo check -p start-core`
- `cargo test -p start-core rpc_shutdown --features=test`
- `make start-core-test` (549 passed, 3 ignored)
- `npx --no-install prettier --check projects/start-os/CHANGELOG.md`

Closes #3577


dr-bonez (commented):
This doesn't look like it's only in the drop-path. We can't universally bound exit time

helix-nine:
Good catch. I made the timeout structurally drop-only: `PersistentContainer::exit` now passes no RPC shutdown timeout, while `Drop` alone supplies the 30-second bound. I also added a paused-time test that lets both explicit-shutdown waits run past 30 seconds and verifies they complete without timeout errors. Rebased onto current master; `make start-core-test` passes (549 passed, 3 ignored), along with `cargo check -p start-core` and the format checks.

dr-bonez (approved):

PR #4017, opened 2026-09-19
fix(start-os): preserve migrated volumes during install

### LINKED ISSUE #3989: A package migrated from 0.3.5.x has no update rollback until its second boot — volume roots stay plain directories, so `InstallBackup::snapshot` silently takes no backup (volume.rs:97)
### Environment

StartOS 0.4.0.1 (user report; same code on master at c5d366303), x86_64, upgraded in place from 0.3.5.1. Area: update / service runtime.

### What happens

A package update whose init fails is not rolled back on a server that came from 0.3.5.x and has not rebooted since the upgrade. Whatever the failed init wrote to the volumes stays: the data version at the target release, and any migration the version graph ran.

Concrete case, Nextcloud: a 0.3.5.1 instance still on Nextcloud 32 updates to 34.0.3:5. The package's graph runs its 0.3.5.x layout migration (moves the Postgres cluster, rewrites `config.php`, deletes the 0.3.5.x marker), writes `data/.version` = `34.0.3:5`, then refuses the 32 → 34 jump and fails init. StartOS reloads the old Legacy entry, but the volume is left as the failed init left it. Every later attempt starts from the new `.version` — the legacy `exit` target in the log climbs `33.0.8:0` → `34.0.3:5` → `34.0.3:6` → `34.0.4:0` — and the 33 release the error tells the user to install fails with `cannot migrate from 34.0.4:0 to 33.0.8:3`. The Legacy package can't start any more either (its 0.3.5.x layout is gone). Nothing in the log says a rollback was skipped.

The update flow should have a rollback point for a migrated package from the first 0.4 boot, or fail the update up front when it cannot take one. At minimum, degrading to no backup should be logged at WARN, not returned as a silent `false`.

- `shared-libs/crates/start-core/src/volume.rs:97` — `InstallBackup::snapshot` returns `Ok(false)` when the live root is not a btrfs subvolume, with no log line; `restore()` is then a no-op on the `Updating` undo path (`service/mod.rs:584`).
- The volume roots of 0.3.5.x packages are plain directories moved into place by the 0.3.6 migration. `ensure_volume_root` (`volume.rs:204`) returns early when the path exists, so nothing converts them at install.
- `convert_to_subvolumes` runs only from `boot_volume_maintenance`, called from `cleanup_and_initialize` (`context/rpc.rs:416`) — *before* `version::post_init` (`rpc.rs:419`) installs the migrated packages. On the first 0.4 boot the set of installed packages is empty at that point, so nothing is converted; the conversion happens on the next boot, which most users never do between upgrading the OS and clicking Update on a service.

One fix: convert each package's volume root in the 0.3.6 `post_up` before `ctx.services.install` (`version/v0_3_6_alpha_0.rs:584`) — the volumes have just been moved and no container holds them — or run `boot_volume_maintenance` again after `post_init`.

### Logs & evidence

User's OS log, Sep 4–14 2026: one boot (`Opening data drive` at 2026-09-05T03:17Z, `Optimizing storage` completes in 0.5 ms at 03:17:51, `Migrating package nextcloud...` from 03:28), then five package updates (tor, electrs, mempool, nextcloud ×9) with no `Created install backup` / `Restored volumes from install backup` line anywhere. Nextcloud side: Start9Labs/nextcloud-startos#150.

helix-nine:
## Summary

- convert each 0.3.5.x service volume root to a btrfs subvolume before installing its migrated package
- preserve pre-existing volume data after a failed install even when no snapshot exists, while removing partial volumes from a failed fresh install
- warn when an install backup cannot be created from a plain directory
- document the data-preservation guarantee in the 0.4 migration guide, changelog, and 0.4.0.2 release notes

## Verification

- `cargo check -p start-core`
- `make start-core-test` (541 passed, 3 ignored)
- `cargo test -p start-core snapshot_of_a_plain_directory_takes_no_backup_and_leaves_it_intact --features=test`
- pinned Prettier check on the changed Markdown files
- 0.3.5.1 legacy OTA VM migration at `7f860365`: interrupted after the install snapshot, forced the retried install to fail on a read-only live volume, then verified the original marker was restored and the migrated package installed successfully
- fresh-install VM recovery at `7f860365`: interrupted after `.install-fresh` and a partial live root appeared, forced the retry to fail, then verified the package entry, partial volume, and recovery marker were removed
- existing-data/no-backup VM recovery at `7f860365`: soft-uninstalled File Browser, blocked its snapshot with an immutable `.install-backup-tmp`, forced the reinstall to fail, then verified the package entry was removed while the original data survived and `.install-fresh` never appeared

Closes #3989


helix-nine:
Review round 1 found and fixed one regression: an ordinary first install has no live volume yet, so the new snapshot warning incorrectly reported that normal state as a non-btrfs root. `snapshot()` now stays quiet when the root is absent and still warns for an existing plain-directory root; the test covers both cases.

Verified with `cargo check -p start-core`, the targeted test, `make start-core-format-check`, `make start-core-test` (541 passed, 3 ignored), the pinned Prettier check, and `projects/start-docs/build.sh`.

dr-bonez (commented):
Please test on a VM @helix-nine

helix-nine:
Retested on the current head (`ed80d36c`) through the real 0.3.5.1 legacy OTA path. **Pass.**

- Installed File Browser `2.32.0` on 0.3.5.1, confirmed its volume root was a plain directory, and seeded a marker (`b39c5dc00bd0a2b28fe04e66143e4ef493dc6e9589ff779b023644b536d45d20`).
- Booted a migration rootfs containing the current-head `startbox`. The first migration boot converted the live root to a btrfs subvolume and created `filebrowser.install-backup`.
- Killed `startd` immediately after that snapshot, deleted the live marker, added a damage file, poisoned the live data version, and made the live subvolume read-only so the rebooted install retry failed inside `Service::install`.
- On reboot the log showed `Error installing service: ... Read-only file system`, followed by `Restored volumes from install backup for filebrowser`.
- Migration then completed and reinstalled File Browser as `2.32.0:0`. The original marker checksum matched; the damage file, install backup, and restore marker were absent; the live root remained a writable btrfs subvolume; and `startd` was active.

The current-head CI test job also passed; only three image-matrix jobs remain queued/running at the time of this comment.

helix-nine:
Retested the reworked current head (`7f860365`) on the VM. **Both recovery branches pass.**

1. **Migrated 0.3.5.1 data:** I repeated the real legacy OTA with File Browser `2.32.0`, interrupted after its install snapshot, removed the live marker, added damage, and made the live subvolume read-only. The rebooted retry failed inside `Service::install` with `EROFS`, then logged `Restored volumes from install backup for filebrowser`. Migration completed with File Browser `2.32.0:0`; the original marker hash matched, damage and all rollback artifacts were gone, the volume was writable, and `.install-fresh` was absent.
2. **Fresh failed install:** On the resulting 0.4 system I removed File Browser and sideloaded it fresh. I interrupted once both `.install-fresh` and the new live root existed, confirmed there was no install backup, added partial data, and made the root read-only. After reboot the retry failed with `EROFS`; recovery removed the package entry, partial volume, `.install-fresh`, and every other rollback artifact.

The VM reported `7f8603655f18317173fa6ae3c692204f51ab0991` throughout. Automated tests and compile jobs are green. The x86 image job's current failure is an unrelated firmware-download HTTP 403 before the image build; it should be rerun once the matrix finishes.

helix-nine:
Covered the remaining **existing data, no rollback point** branch on a VM at current head `7f860365`. **Pass.**

- Installed File Browser `2.63.23:3`, seeded `data/main/no-backup-marker.txt` (`70fc8875b081dea51766e5b80b706284807830b99e84a98e93dd97d373e2942a`), then soft-uninstalled it. The package entry was gone while the seeded btrfs volume remained.
- Created immutable `filebrowser.install-backup-tmp` to block the snapshot, made the live root read-only to force init failure, and sideloaded the same s9pk.
- The log warned `Could not create install backup for filebrowser: ... Operation not permitted`; no `.install-backup` was created. The reinstall then failed writing `data/.version` with `EROFS`.
- A 1 ms watcher covered the install/recovery and never observed `.install-fresh`; it was absent before and after as well.
- Recovery removed the failed package entry but kept the existing volume. The marker checksum remained byte-identical, with no `.install-backup` and no `.install-fresh`.

This exercises the `Installing` recovery arm with `keep_volumes = true` and `restore()` as a no-op because no rollback point exists.

dr-bonez (approved):

dr-bonez (approved):

PR #4018, opened 2026-09-19
fix(start-core): let watched callbacks finish

### LINKED ISSUE #3821: An init handler's .const() re-run stops firing, and a service restart does not re-arm it
An init handler that `.const()`s one of its own hosts stops being re-run after a burst of changes to that host. Once it has stopped, nothing re-arms it: further changes to the host do nothing, and `start-cli package restart` does not re-run init handlers at all. The handler's side effects then stay frozen at whatever the last successful re-run produced, silently and indefinitely.

Found while working on `lnd-startos` ([#189](https://github.com/Start9Labs/lnd-startos/pull/189)), whose `setupCerts` handler reissues LND's TLS certificate for the addresses its gRPC interface is served at:

```ts
export async function writeCerts(effects: T.Effects): Promise<void> {
  const served = await sdk.host
    .getOwn(effects, gRPCHostId, (host) => /* → string[] of hostnames */)
    .const()
  const hostnames = [ /* container ip, 127.0.0.1, os ip */, ...served ]
  const cert = (await sdk.getSslCertificate(effects, hostnames).const()).join('')
  await writeFile('/media/startos/volumes/main/tls.cert', cert)
  // …
}

export const setupCerts = sdk.setupOnInit(writeCerts)
```

## It works, until it doesn't

Early in the session this was reliable. Adding an onion service to `lnd`'s `grpc` host rewrote `tls.cert` **within about twenty seconds**, carrying the new `.onion` as a SAN — verified by watching the file's mtime rather than the served certificate.

After four onion services were deleted in quick succession, it stopped. Adding a fresh onion produced no re-run at all: the address is in the database, and `tls.cert` has not been touched since.

```
17:18:55   add-onion-service issued
17:18:55+  db: bindings.10009.addresses.available gains bwckscrm….onion   ← the watched pointer changed
17:23:55   tls.cert mtime still 17:06:48, 300s later
17:25:56   start-cli package restart lnd
17:30:00   tls.cert mtime still 17:06:48
```

`lnd`'s own logs show the effect callbacks being registered on each init pass, and nothing after the last one:

```
17:06:48   adding callback 14 / 15 / 16 / 17      ← full init pass
17:06:48   Starting LND!
17:11:32   Starting LND!                          ← main re-run, no init
17:25:56   Starting LND!                          ← package restart, no init
```

No errors anywhere — not in the service logs, not in its status. The handler simply is not called again.

## Three views of the same state, all different

At 17:25 the same onion set had three inconsistent representations:

| view | onions |
| ---- | ------ |
| db address set for `lnd`/`grpc` | `bwckscrm…` |
| `torrc` | `bwckscrm…` (`hs_0`) |
| on-disk key material | `bwckscrm…` (`hs_0`), **`wi4cb3eb…` (`hs_3`)** |
| certificate LND serves | `vjn7hzez…`, `x3ufhman…` — both deleted |

The certificate names two onions that no longer exist anywhere else, and names neither of the two that do. `hs_3` is an orphaned `HiddenServiceDir` — key material with no `torrc` entry and no address, left behind by the same burst of deletions; possibly the same underlying cause on tor's side, since `exportUrls` is also a `.const()` context, but I haven't confirmed that.

## Why this is worth more than the cert

`setupOnInit` + `.const()` is the documented way to keep a side effect in sync with reactive state — [`init.md`](https://github.com/Start9Labs/start-technologies/blob/master/projects/start-sdk/docs/src/init.md) calls the `null` init kind the place to "register long-lived triggers (e.g. `.const()` watchers)", and `AGENTS.md` describes an init handler as "a live reactive context, not a one-shot script" that "re-runs from the top on every change to anything it `.const()`s". A handler that quietly stops re-running breaks that contract for every package that relies on it, and there is no signal that it has happened. Here the failure mode is a service serving a certificate for addresses that no longer exist while the correct set sits in the database; elsewhere it is whatever that handler was keeping current.

The part that makes it hard to live with is that a restart does not fix it. `package restart` restarts the service without re-running init, so there is nothing a user or an operator can do short of a container rebuild.

## Reproduction

On StartOS 0.4.0.1 with `lnd` and `tor` installed and an LND wallet initialised, using [lnd-startos#189](https://github.com/Start9Labs/lnd-startos/pull/189) (any package whose init handler `.const()`s one of its own hosts and writes a file will do):

1. Add an onion service to LND's gRPC interface. Watch `…/volumes/lnd/data/main/tls.cert` — the mtime moves within ~20 s and the `.onion` appears in the SANs.
2. Repeat until you have four, then delete all four in quick succession.
3. Add one more. The address lands in the db; `tls.cert` is never rewritten. Restarting the service does not help.

Happy to instrument further if it would be useful — I stopped at the SDK/effects boundary rather than guess at `DbWatchedCallbacks`.

helix-nine:
## Summary

- stop a callback watcher task from aborting itself when it removes its own entry from the callback map, at all three sites that do it: `DbWatchedCallbacks::add`, `add_get_outbound_gateway`, and `add_get_ssl_certificate`
- add `NonDetachingJoinHandle::detach`, which gives up the inner `JoinHandle<T>` without aborting, and route the three sites through one helper, `CallbackHandlers::detach_watcher`
- add a regression test for `detach` with a plain-drop control, and the StartOS changelog entry

## Root cause

A watcher task removed its own `(NonDetachingJoinHandle, handlers)` entry with `.map(|(_, handlers)| ..)`. The unbound `_` half of a closure parameter drops when the closure returns, so the handle called `abort()` on the task that was running it. Tokio applies that at the task's next `Poll::Pending`, which lands inside the callback invocation.

It is intermittent because the invocation usually never yields. `UnixRpcClient::notify` takes an idle pooled connection synchronously and the socket write normally has room, so the task runs to completion and the abort is a no-op. A burst of changes fires several watchers at once and drains the pool; the next `notify` has to connect, returns `Pending`, and the task is cancelled. `PersistentContainer::callback` has already removed the callback id from the container's registered set before its first await, so the callback is lost and nothing can re-arm it.

`add_get_outbound_gateway` had the identical drop. `add_get_ssl_certificate` was correct only because an `if let Some((_, callbacks))` scrutinee keeps the handle alive to the end of the block; it now detaches explicitly like the other two.

External removals (`gc`, dropping the map) keep abort-on-drop.

## Task lifetime

A detached watcher cannot be aborted, so what it still awaits matters. Once it removes its own entry the map's handle was the last one, so detaching gives up no cancellation anyone else had. All it has left is `CallbackHandlers::call`. `Callback` is a notification (`Response = Never`), so each handler is one line written to the container's unix socket with no reply awaited, retried at most `MAX_TRIES` times. The pool lookup is synchronous or a local `connect`. The write resolves once the kernel buffers it and errors when the container closes the socket, so the task ends no later than the container does. Detaching also breaks the `Arc<DbWatchedCallbacks>` cycle between a watcher and the map holding its handle.

The write has no timeout, deliberately: cancelling `write_all` mid-line would leave a partial JSON line on a connection the pool guard hands back for reuse.

## Testing

- `cargo check -p start-core`
- `cargo test -p start-core test_detach_inside_own_task --features=test` — asserts a task survives detaching its own handle, and that a plain drop of the same handle cancels it
- `cargo +nightly-2026-05-28 fmt -p start-core -- --check`
- `prettier --check projects/start-os/CHANGELOG.md`

The full `start-core` suite was not re-run locally after the rework; CI covers it.

Closes #3821


helix-nine:
Review round 1 found one documentation mismatch: a failed database watch wakes, refetches, and re-arms the `.const()` subscription immediately, but the setup handler itself reruns only when that refetch observes a changed value. I corrected the init guide and StartOS changelog to state that guarantee precisely in ebb014f8a.

Verified with:
- `cargo test -p start-core watch_error_removes_entry_and_allows_rearming --features=test`
- `cargo check -p start-core`
- `npx --yes prettier@3.8.3 --check projects/start-sdk/docs/src/init.md projects/start-os/CHANGELOG.md`

helix-nine:
Reworked in `18cbae29e` after tracing the PatchDB ordering invariant: patch application failure was the wrong lead. The successful watch task was aborting itself by removing and dropping its own `NonDetachingJoinHandle` before callback invocation could complete.

The branch now detaches only that current handle before removal; externally removed tasks retain abort-on-drop. The focused regression test passes with the detach and fails with `RecvError` when the detach is removed. I also rebased onto current `master`, updated the docs/body, reran the full local checks, and returned the PR to draft.

helix-nine:
Reviewed at head `18cbae29e` (merge-base `acee3433b`).

**The fix is right, and the claims reproduce.** On master the per-key task in `DbWatchedCallbacks::add` drops its own `NonDetachingJoinHandle` inside the `.map(|(_, handlers)| …)` closure — the `_` field is dropped when the closure returns, inside `mutate`, before `cbs.call(..).await` (probe: `rustc --edition 2024`, a `Drop`-printing type in that shape prints `drop` before `body`). `abort()` on the running task sets `CANCELLED` (tokio 1.52.1 `state.rs:313-323`) and the future is dropped at its first `Pending` (`harness.rs:217-222`). That first `Pending` is `UnixRpcClient::notify` → `pool.get().await` (`rpc_client.rs:260`): `lazy_async_pool` with `cap=0` returns `Ready` only when an idle client is pooled (`lib.rs:98-104`); otherwise it awaits a spawned `UnixStream::connect`, which is `Pending` on first poll. By then `PersistentContainer::callback` has already consumed the registration synchronously (`send_if_modified` → `handle.params` removes the `CallbackId`, `persistent_container.rs:585-588`, `rpc.rs:246`) and `CallbackHandler::call` has `take()`n its handle — so nothing logs and nothing re-arms. A lone change against a warm pool goes through; anything that leaves no idle client (concurrent callbacks/effects, or a pool gone cold via `client.destroy()` after the reader finished) loses the callback. That is the "works for a while, then stops, no errors" in #3821.

Ran: `cargo test -p start-core taking_callbacks_does_not_abort_the_callback_task --features=test` passes at head; with `task.detach()` replaced by `drop(task)` it fails `RecvError(())`. The `future.rs` change is behaviour-preserving for the other 82 `NonDetachingJoinHandle` sites — the only constructor is `From` (`abort_on_drop: true`) and `detach` has one caller.

**Should fix before merge**

1. `add_get_outbound_gateway` (`callbacks.rs:325-331`) has the identical self-abort and the PR leaves it in place: `this.get_outbound_gateway.remove(&key).map(|(_, handlers)| CallbackHandlers(handlers))` drops the task's own handle before `cbs.call(vector![]).await`. Reached from `effects/net/info.rs:84` for every `getOutboundGateway({callback})` `.const()`. Same one-liner:
   ```rust
   .map(|(task, handlers)| {
       task.detach();
       CallbackHandlers(handlers)
   })
   ```
2. Rebase: `projects/start-os/CHANGELOG.md` conflicts with master (#4019/#4020/#4021/#4023 landed under the same `### Fixed`).

**Notes, non-blocking**

- `add_get_ssl_certificate` (`callbacks.rs:236-243`) is the same shape but inert: with the tuple as the `if let` scrutinee, the handle lives in the scrutinee temporary until after `.call().await` (probe confirms, both editions) and no yield follows. It still aborts its own task, so any `.await` added after line 243 re-creates the bug; detaching there too removes the trap.
- `take()` assumes the entry under `key` is the caller's. If `gc()` removes A's entry while A is past `poll_recv` → `Ready` and `add()` re-inserts B before A reaches the lock, A detaches and drains B's handlers, then dies of its own pending abort. Sub-microsecond window, needs a `clear_callbacks` (JS finalizer) at that instant, and pre-existing — on master A was cancelled regardless. If you want it closed: compare `task.id()` with `tokio::task::try_id()` in `take()` and re-insert on mismatch.
- `watch.changed()` can `Err` only on `Disconnected` (store dropped) or a `patch()` failure, which dump+subscribe under one store lock rules out — so the new `error!` fires at shutdown, once per outstanding watch, if the store drops before the map. Master was silent; both leave the dead entry in the map. `warn!` or `debug!` avoids the shutdown noise.
- Test: valid A/B for the detach, but it inserts straight into `inner`, so `add()`'s spawn path and the non-empty-handler branch of `take()` are not exercised.
- CHANGELOG: the fix covers every `.const()` context — `main` restarts (`SystemForStartOs.ts:72`), init handlers (`setupInit.ts:74`), action metadata (`setupActions.ts:92/216`) — not only init, and the trigger is any change whose notify has to connect, not only bursts. Something like: "**Services keep reacting to changes in the hosts, interfaces, status and dependencies they watch.** Each change re-runs the watching init handler, restarts `main`, or refreshes action metadata, however many changes arrive."
- `init.md`: the added sentence describes the re-arm mechanism, and the sentence before it already states the contract. The order is also not quite that: the consumed subscription is re-armed by the old read's own re-fetch (`Watchable.ts` `produce()` loop) before the handler re-runs, and the re-run's `.const()` joins that entry. I would drop the sentence.

dr-bonez (approved):

helix-nine:
Reviewed the follow-up commits at `48e117493`: the outbound-gateway blocker and SSL-certificate trap are both routed through `detach_watcher`, the watch-error logging and redundant init documentation are gone, the changelog conflict is resolved, and `NonDetachingJoinHandle::detach` preserves the default abort-on-drop behavior elsewhere.

I reran `cargo test -p start-core test_detach_inside_own_task --features=test` at this head; it passes. No further changes needed from me. The full CI matrix is still running.

PR #4019, opened 2026-09-19
fix(start-os): make mount teardown idempotent

### LINKED ISSUE #3937: Graceful shutdown drops the server into diagnostic mode when a mountpoint is already gone
## Summary

`unmount_slot` treats `ENOENT` from `remove_dir(mountpoint)` as a hard error. `ENOENT` means the mountpoint is already gone — which is the outcome the call was trying to produce. Because that error propagates all the way out of `inner_main`, a harmless cleanup race during a **requested shutdown** ends with the server serving the diagnostic UI instead of powering off.

Verified unfixed on `master` (`Current = v0_4_0_2::Version`); `guard.rs:120` is byte-identical in `v0.4.0`, `v0.4.0.1` and `master`, so every 0.4.0.x release to date is affected.

Seen on a user’s box (0.4.0.1) on a normal Restart: all services stopped correctly, then the box sat on the diagnostic screen for 33 minutes until `startd` was stopped by hand.

## The defect

`shared-libs/crates/start-core/src/disk/mount/guard.rs:118`

```rust
if delete_mountpoint {
    match tokio::fs::remove_dir(mountpoint).await {
        Err(e) if e.raw_os_error() == Some(39) => Ok(()), // directory not empty
        a => a,
    }
    .with_ctx(|_| {
        (
            crate::ErrorKind::Filesystem,
            format!("rm {}", mountpoint.display()),
        )
    })?;
}
```

`ENOTEMPTY` (39) is tolerated. `ENOENT` (2) is not — even though "the mountpoint no longer exists" is strictly a *better* outcome than "the mountpoint exists but isn't empty", which we already accept.

## How it escalates to diagnostic mode

1. `PersistentContainer::destroy` unmounts each volume / asset / js_mount with `delete_mountpoint: true` and collects failures in an `ErrorCollection` (`persistent_container.rs:469-480`).
2. `ErrorCollection::into_result()` surfaces them from `PersistentContainer::exit`.
3. `ServiceMap::shutdown_all` collects those into another `ErrorCollection` (`service_map.rs:529-533`).
4. `RpcContext::shutdown` propagates (`context/rpc.rs:427`).
5. `inner_main` returns `Err` → `bins/startd.rs:171` → `DiagnosticContext::init(...)` + `server.serve_ui_for(ctx)`.

Step 5 is the part that turns a cleanup nit into a support case. The `inner_main`-error-means-diagnostic-mode rule is right for *startup* failures; applied to the shutdown path it means any teardown error strands the box on the diagnostic screen rather than completing the shutdown the user asked for.

## Observed trigger

Each affected container had three mountpoints torn down twice within milliseconds, from two different call paths — one inside the shutdown span, one from a detached task (i.e. `MountGuard::drop`, `guard.rs:133`). This is `lndg`’s container; its GUID is elided as `<ct>`:

```
51.204303 DEBUG inner_main:shutdown:exit:unmount  Unmounting <ct>/rootfs/media/startos/volumes/main.
51.207136 DEBUG inner_main:shutdown:exit:unmount  Unmounting <ct>/rootfs/media/startos/assets.
51.207149 DEBUG unmount                           Unmounting <ct>/rootfs/media/startos/volumes/main.
51.210329 DEBUG inner_main:shutdown:exit:unmount  Unmounting <ct>/rootfs/media/startos/images/lndg.
51.210358 DEBUG unmount                           Unmounting <ct>/rootfs/media/startos/assets.
51.219191 DEBUG inner_main:shutdown:exit:unmount  Unmounting <ct>/rootfs/usr/lib/startos/package.
51.223963 DEBUG unmount                           Unmounting <ct>/rootfs/usr/lib/startos/package.
52.207612 DEBUG inner_main:shutdown:exit:exit:unmount  Unmounting <ct>/rootfs.
52.251457 ERROR start_core::bins::startd: Multiple Errors: ...
```

and the three paths that failed are exactly the three that were unmounted twice:

```
rm <ct>/rootfs/media/startos/volumes/main: No such file or directory (os error 2)
rm <ct>/rootfs/media/startos/assets:       No such file or directory (os error 2)
rm <ct>/rootfs/usr/lib/startos/package:    No such file or directory (os error 2)
```

Three of the box's 13 services hit this (`lndg`, `nostr-rs-relay`, `thunderhub`); the other ten logged `Service for <id> exited` normally. Nothing was lost — the containers had already stopped — the shutdown simply never finished.

I have not pinned down which of the two racing teardowns removes the directory first, and the fix below does not depend on that. The duplication itself looks worth a separate look: `MountGuard::drop` spawns a detached `unmount_slot` for a path that `destroy()` is concurrently unmounting and deleting, and the slot's `strong_count` guard only skips the umount when a live *remount* holds the slot, not when a concurrent unmount is already in flight.

## Suggested fix

Tolerate `ENOENT` alongside `ENOTEMPTY`:

```rust
match tokio::fs::remove_dir(mountpoint).await {
    Err(e) if e.raw_os_error() == Some(39) => Ok(()), // directory not empty
    Err(e) if e.kind() == std::io::ErrorKind::NotFound => Ok(()), // already gone
    a => a,
}
```

Worth considering separately, since it is what made this user-visible: whether a failure on the **shutdown** path should reach `DiagnosticContext` at all, or should be logged and allowed to complete the shutdown. As it stands, any error while tearing services down leaves the box powered on and serving the diagnostic UI, which is both the opposite of what was requested and hard for a user to interpret.

## Reproduction

Not reproduced on demand — it is a timing race. On the reporting box it fired once in ~11 weeks of logs across five clean shutdowns.

## Environment

- StartOS 0.4.0.1
- 13 packages, x86_64, ~16 GiB RAM
- Observed 2026-09-14; from a user-supplied `os-logs` export
- Chain re-verified against `master` @ ead412242 (0.4.0.2 line) on 2026-09-14

helix-nine:
## Summary

- treat an already-absent path as successfully unmounted, including if it disappears while `umount` runs
- accept missing and non-empty mountpoint directories as completed cleanup outcomes
- add regression coverage for teardown after the mountpoint is already gone
- document the shutdown/restart fix in the StartOS changelog

## Verification

- `cargo test -p start-core unmount_slot_succeeds_when_mountpoint_is_missing --features=test`
- `cargo check -p start-core`
- `make start-core-test` (541 passed, 3 ignored)
- Prettier 3.8.3 on `projects/start-os/CHANGELOG.md`

Closes #3937


helix-nine:
Review round 1 fixed one teardown regression in `4471f2b71`:

- Removed the preflight `try_exists` check. A metadata lookup on an unresponsive FUSE mount can block before lazy `umount` gets a chance to detach it; a local FUSE probe blocked in `stat` beyond 3 seconds while `umount -l` detached the same mount in 7 ms. The post-error existence check still makes an already-missing path succeed, including the race where it disappears while `umount` runs.

Verified the resulting head with:

- `cargo test -p start-core unmount_slot_succeeds_when_mountpoint_is_missing --features=test`
- `cargo check -p start-core`
- `make start-core-test` (541 passed, 3 ignored)
- Prettier 3.8.3 on `projects/start-os/CHANGELOG.md`

dr-bonez (dismissed):

helix-nine:
Rebased onto current `master` in `617dbf5b2`. The conflict was limited to `projects/start-os/CHANGELOG.md`; the resolution keeps all newly landed entries and reapplies this PR's existing shutdown entry. The Rust diff is unchanged.

Verified the rebased head with:

- `cargo test -p start-core unmount_slot_succeeds_when_mountpoint_is_missing --features=test`
- `cargo check -p start-core`
- Prettier 3.8.3 on `projects/start-os/CHANGELOG.md`

The force-push necessarily dismissed the prior approval, so the PR now needs a fresh approving review.

dr-bonez (dismissed):

helix-nine:
`master` advanced to `0849eb19a` after the prior rebase and made the PR conflicting again. I rebased once more in `17059aeb5`; the conflict was again limited to `projects/start-os/CHANGELOG.md`, and the resolution retains the newly merged cross-architecture backup entry alongside this PR's shutdown entry. The mount code remains unchanged.

Verified the new head with:

- `cargo test -p start-core unmount_slot_succeeds_when_mountpoint_is_missing --features=test`
- `cargo check -p start-core`
- Prettier 3.8.3 on `projects/start-os/CHANGELOG.md`

This second force-push necessarily dismissed the new approval; the PR needs another approving review.

dr-bonez (approved):

PR #4020, opened 2026-09-19
fix(net/dns): retain fallback with no upstreams

### LINKED ISSUE #3603: net::dns: forwarder can be built with zero upstreams, silently killing all container DNS while the host stays healthy
## Summary

`spawn_forwarder` derives the forwarder's upstream list from `/run/systemd/resolve/resolv.conf` by dropping the first two nameservers, with no check that anything remains:

```rust
// shared-libs/crates/start-core/src/net/dns.rs:396-403
let forward_servers: Vec<NameServerConfig> = if let Some(servers) = &static_servers {
    servers.iter().map(|addr| forward_name_server(*addr)).collect()
} else {
    config.name_servers().iter().skip(2).cloned().collect()
};
```

If resolv.conf carries two or fewer nameservers, `forward_servers` is empty and the `ForwardZoneHandler` built immediately below gets an empty `name_servers` list. Every forwarded query then fails. Nothing logs, nothing warns, and the catalog is swapped in as though it were healthy.

## Why this is worse than it looks

The host does not resolve through this forwarder; service containers have nothing else. `projects/start-os/container-runtime/deb-install.sh:16` gives every container exactly one nameserver and no fallback:

```sh
echo "nameserver 10.0.3.1" > /etc/resolv.conf
```

So an empty upstream list presents as: the box looks completely healthy — UI, registry browsing, package installs, updates all work — while **every containerized lookup returns `EAI_AGAIN`**. To the user that reads as "all of my services are broken"; to us it arrives as a pile of unrelated per-service bug reports.

#3473 already flagged this misdiagnosis hazard ("the host itself mostly keeps working via its `1.1.1.1` fallback, which makes this easy to misdiagnose as individual service bugs"). This is a second, independent route to the same end state, and unlike that livelock it is steady-state rather than episodic — so it survives reboots and reinstalls, which makes it even easier to blame on the individual service.

The margin is small. Two nameservers are dropped unconditionally, so the forwarder needs at least three uplink resolvers before it has any upstream at all, and the count moves in response to ordinary user actions. A StartTunnel gateway whose WireGuard config predates a526bf803 (2026-05-29) carries no `DNS =` line, so `to_nm_settings` sets neither `dns` nor `dns-search` for that interface (`net/gateway.rs:419-443`) and the uplink list is one entry shorter than it would be with a current config.

## The skip itself is unexplained

`.skip(2)` appears at `dns.rs:389` (the `dhcpServers` DB write) and `dns.rs:403` (the forwarder), with no comment at either site saying what the two dropped entries are or why they are guaranteed to be present. I could not find where that guarantee is established.

Minor related cleanup: the DB site reads `.map(name_server_socket_addr).dedup().skip(2)`. Under hickory 0.26 `NameServerConfig` is one entry per IP (`connections: Vec<ConnectionConfig>` holds the UDP/TCP pair), so that `.dedup()` is a no-op. The two expressions look like they mean different things but don't — worth collapsing into one shared, named helper so they cannot drift apart later.

## Suggested fixes

- Guard the empty case: if `forward_servers` comes out empty, log an error and retain the previous catalog rather than installing a forwarder that can only fail.
- Surface it. An empty upstream list belongs in the UI next to the DNS settings — it is currently invisible from the host, which is the entire problem.
- Name and comment the skip, and share one implementation between both call sites.

## Provenance

Seen downstream in Start9Labs/mempool-startos#68 and Start9Labs/nextcloud-startos#122 — same box, same day, both reducing to dead container DNS on an otherwise healthy 0.4.0.1 server. The root cause on that particular box is not yet confirmed; this issue is about the code path being both unguarded and undiagnosable regardless of what tipped it over.

helix-nine:
## Summary

- preserve the intentional exclusion of StartOS's resolver and built-in fallback from network-provided DNS servers
- use the built-in `1.1.1.1` resolver whenever the final forwarder list would otherwise be empty
- share the network-upstream selection between the forwarder and `dhcpServers`, with regression coverage for discovered, absent, and empty static upstream lists
- document the fallback behavior and add the StartOS changelog entry

## Verification

- `cargo test -p start-core net::dns::tests --features=test`
- `cargo check -p start-core`
- `make start-core-test` (543 passed, 3 ignored)
- `prettier --check projects/start-os/CHANGELOG.md projects/start-os/docs/src/dns.md`

Closes #3603


dr-bonez (dismissed):

helix-nine:
Rebased onto current `master` (`acee3433`). The only conflict was additive `CHANGELOG.md` entries under `### Fixed`; I kept all newly merged entries and retained this PR's approved “separate DNS server” wording. The PR diff remains limited to the same three files, with no code or behavior change.

Re-verified the rebased head with:
- `cargo test -p start-core net::dns::tests --features=test` (3 passed)
- Prettier checks for the two edited Markdown files

The force-push dismissed @dr-bonez's approval, so the PR now requires a fresh approving review. CI has restarted on `f0d28e10`.

dr-bonez (approved):

PR #4021, opened 2026-09-19
fix(start-os): restore backups across architectures

### LINKED ISSUE #3761: Cross-architecture restore fails: arm64 backup cannot be restored to x86_64 server (`images/x86_64/main.squashfs missing from archive`)
### Prerequisites

### StartOS version

0.4.0.1 (both source and target)

### Architecture / image

x86_64 (AMD64) — Full

### Server hardware

**Environment:** - **Source:** CM3588 (arm64), StartOS running as a VM under PXVIRT - **Target:** System76 Thelio Prime (Ryzen 9, x86_64), bare-metal install - **Backup format:** Native StartOS backup (V2), stored on NVMe

### Browser & OS

_No response_

### Which surface?

Backup / restore

### What happens

already shared with @dr-bonez in startos matrix dev channel, adding here to help with tracking:

Restoring services from an **arm64** backup onto an **x86_64** server fails almost immediately with:

```
RPC Error: S9PK Parsing Error
file images/x86_64/main.squashfs missing from archive
```

The backed-up `.s9pk` archives appear to contain only the arm64 image layer, so when the x86_64 target attempts to unpack a service, the expected `images/x86_64/main.squashfs` entry is absent and parsing aborts.

Attempting to work around this by installing the service fresh from the marketplace first (to seed the correct-arch image) and then restoring instead fails with:

```
Unavailable. Service is already installed
```

So neither the "restore directly" path nor the "install-then-restore" path completes.

One of the following:

1. **Cross-arch restore succeeds** — StartOS pulls the correct-architecture image (from the marketplace/registry) during restore while reapplying the backed-up data/volumes, so an arm64 → x86_64 migration completes cleanly; **or**
2. **A clear, documented, and enforced error/guardrail** stating that backup-restore is same-architecture only, with the supported hardware-migration path (e.g., Transfer) called out in the UI at the point of failure rather than a low-level S9PK parsing error.

### Steps to reproduce

1. On an **arm64** StartOS 0.4.0.1 server, create a native full-server backup (V2) to an external drive.
2. Move the backup drive to a fresh **x86_64** StartOS 0.4.0.1 server.
3. Attempt recovery via either path:
   - **a.** Full-server **"Restore from Backup"** during initial setup, **or**
   - **b.** A fresh install, then restore an individual service from the backup.
4. Observe the parsing error within ~1–2 minutes of the restore starting.
5. (Optional) Try installing the service from the marketplace first, then restore → observe the "Service is already installed" block.

### Logs & evidence

```shell
**Restore error (both individual-service and full-server recovery paths):**

RPC Error: S9PK Parsing Error
file images/x86_64/main.squashfs missing from archive


**Install-then-restore workaround attempt:**

Unavailable. Service is already installed


**Timing:** Full-server "Restore from Backup" recovery fails ~1–2 minutes into the process, consistently reproducible across attempts.
```

helix-nine:
## Summary

- replace the architecture-valued `emulateMissingAs` manifest field with `emulateMissing: boolean`, while migrating legacy manifests as `!!emulateMissingAs`
- preserve a legacy manifest’s original fallback when start-cli rewrites it, so artifacts that retain an older StartOS floor remain readable by those releases
- validate and mount the image architecture actually present in an s9pk, so a backup retained on ARM can restore and run under emulation on x86 (and vice versa)
- default SDK-built images to emulation, retain a native-only opt-out, and cover both paths with SDK tests
- make CLI image emulation the default, expose `--no-emulation` as the native-only opt-out, remove the old architecture-valued syntax, and release this externally visible change as start-cli 2.2.0
- update generated bindings, fixtures, manpages, translations, package-format docs, backup guidance, and product changelogs
- remove the non-contract commentary above the SDK image architecture type

## Verification

- `cargo check -p start-core -p start-cli`
- `cargo test -p start-core 's9pk::v2::'` (15 passed)
- `cargo test -p start-core` (839 passed, 3 ignored; the sole failure was the existing live-network `util::http_reader::main_test`, where `https://docs.start9.com/llms.txt` timed out again when rerun alone)
- `make start-core-ts-bindings`
- `make manpages`
- `cd projects/start-sdk && make check && make test && make check-fmt` (124 tests passed)
- `cd shared-libs/ts-modules/start-core && make check && make test && make check-fmt` (246 passed, 9 skipped)
- SDK bundle plus container-runtime `npm ci`, type-check, build, and tests (13 passed)
- root `npm run check`, `npm run format:check`, and `npm run format:toml:check`
- `mdbook build projects/start-os/docs`
- `mdbook build projects/start-sdk/docs`

`cargo clippy -p start-core -p start-cli --all-targets` remains blocked before these crates by existing denied Clippy warnings in `shared-libs/crates/patch-db/json-patch` (`type_complexity` and `borrow_deref_ref`).

A persistent user-facing emulation warning is tracked in #4025.

Closes #3761


helix-nine:
Review round 1 found and fixed one compatibility regression in bf480050a.

start-cli 2.2 could rewrite a manifest produced by an older SDK to the new boolean-only image shape while leaving that package’s older StartOS compatibility floor unchanged. Those older StartOS releases ignore `emulateMissing` and would fail to install an artifact whose only image needs emulation. Legacy input now retains its original `emulateMissingAs` value alongside the migrated boolean; new manifests remain boolean-only. Tests cover legacy string/null forms, the current boolean form, and both CLI flags.

Verified with the scoped start-core tests, `cargo check -p start-core -p start-cli`, the pinned start-core formatter, and regenerated TypeScript bindings.

helix-nine:
Review round 2 found and fixed one remaining legacy-compatibility regression in f1109d776.

A legacy manifest can declare multiple source architectures and a specific `emulateMissingAs` fallback. Although round 1 preserved that field during serialization, image selection still ignored its value and chose the first architecture in the sorted set. For example, packing `arch: [aarch64, x86_64]` with `emulateMissingAs: x86_64` for RISC-V kept ARM while serializing an x86 fallback; an older StartOS would then look for an x86 image the artifact did not contain. Runtime selection could likewise choose ARM instead of the package author's declared x86 fallback.

Legacy manifests now prefer their retained fallback when it is available, including in mixed-native/emulated artifacts. Boolean-only manifests keep the new any-available-image behavior.

Verified with 13 scoped s9pk tests, `cargo check -p start-core -p start-cli`, the pinned start-core formatter, and an ablation that made the new regression test fail with ARM selected instead of x86.

projects/start-cli/CHANGELOG.md
```diff
@@ -9,6 +9,14 @@ Because `start-cli` is a thin client over `start-core`, most user-visible CLI ch
 in `start-core`; record here anything that changes this crate's entrypoint, features, packaging,
 or the CLI's externally observable behavior.
 
+## [2.2.0]
+
+### Added
+
+- **`s9pk edit add-image --emulate-missing` lets StartOS run an available image
+  under CPU emulation** when the server uses another architecture. The previous
+  `--emulate-missing-as <ARCH>` spelling remains accepted for existing scripts.
```
dr-bonez: The old cli syntax can be removed - no scripts use this

shared-libs/crates/start-core/src/s9pk/v2/pack.rs
```diff
@@ -762,20 +835,25 @@ pub async fn pack(ctx: CliContext, params: PackParams) -> Result<(), Error> {
...
-                    tracing::warn!(
-                        "ImageId {} is not available for {}, emulating as {}",
-                        id,
-                        arches.iter().join("/"),
-                        arch
-                    );
-                    c.arch = [arch.clone()].into_iter().collect();
+                if c.emulate_missing {
+                    if let Some(arch) = c.emulation_arch(|_| true).cloned() {
+                        tracing::warn!(
```
dr-bonez, line 836: i18n

dr-bonez (commented):
We should probably show a warning when an image is running emulated. Open a separate issue for that though, it's out of scope for this pr

projects/start-cli/CHANGELOG.md
helix-nine: Removed `--emulate-missing-as` from the CLI entirely and added a regression asserting that clap rejects it. Legacy manifest deserialization remains supported.

shared-libs/crates/start-core/src/s9pk/v2/pack.rs
helix-nine: Localized this warning and the adjacent architecture error paths through `s9pk.pack.*` keys in all five locales.

helix-nine:
Addressed the review in `70d58c480`:

- removed the old `--emulate-missing-as` CLI syntax while retaining legacy manifest compatibility
- localized the pack architecture diagnostics in all five locales
- rebased onto current `master`

Verified with `cargo test -p start-core s9pk::v2::` (14 passed), `cargo check -p start-cli`, and `make manpages`. The persistent user-facing emulation warning is tracked in #4025.

projects/start-cli/CHANGELOG.md
```diff
@@ -9,6 +9,13 @@ Because `start-cli` is a thin client over `start-core`, most user-visible CLI ch
 in `start-core`; record here anything that changes this crate's entrypoint, features, packaging,
 or the CLI's externally observable behavior.
 
+## [2.2.0]
+
+### Added
+
+- **`s9pk edit add-image --emulate-missing` lets StartOS run an available image
```
dr-bonez: Flag should be inverted `--no-emulation`. Emulation is default on

helix-nine: Changed the CLI to enable emulation by default and use `--no-emulation` as the native-only opt-out in `a23ff30cb`. Updated the localized help, docs, release notes, changelog, manpage, and regressions; 15 scoped S9PK tests pass.

dr-bonez (approved):

PR #4022, opened 2026-09-19
fix(start-os): grow Pi root filesystem on first boot

### LINKED ISSUE #4009: Pi image never grows its btrfs root filesystem — init_resize.sh resizes through a read-only bind mount and ignores the failure (init_resize.sh:104)
### Environment

StartOS 0.4.0.1, `raspberrypi` image, Raspberry Pi 4 Model B (8 GB) booting from a 128 GB microSD — field report in #3624. Code read on master at 3cf760cb7; the resize line is unchanged since it was added in dba1cb93c, so every Pi image from 0.4.0-alpha.21 on is affected. Area: image build / first boot.

### What happens

On a Pi the root btrfs filesystem stays at its flashed size forever. First boot grows the root *partition* to 16 GiB (or the whole card on a small one), but the filesystem inside it is never grown. #3624's reporter found a 1.52 GiB filesystem on a 13.8 GiB partition with about 48 MiB free. `df` does not show the slack; `btrfs filesystem usage` does.

The image is built with the root partition sized to the squashfs plus 256 MB, so that is all the free space a Pi ever has on the filesystem holding `images/` and the config overlay.

First boot should grow the filesystem to fill the partition, and a failed grow should fail `init_resize.sh` so its `init=` hook stays in place and retries, the way the partition steps already do.

- `projects/start-os/build/image-recipe/raspberrypi/img/usr/lib/startos/scripts/init_resize.sh:104` runs `btrfs filesystem resize max /media/startos/root`.
- `projects/start-os/build/lib/scripts/startos-initramfs-module:129` mounts that path with `mount -r --bind /startos ${rootmnt}/media/startos/root`. The resize ioctl takes write access on the mount it is issued through, so on a read-only bind it fails with `Read-only file system`. The `mount / -o remount,rw` on the line before remounts the overlay, not this bind.
- The command's exit status is not checked, unlike every `sfdisk` step above it. `main` returns 0, the script logs `SUCCESS — resized root filesystem`, and removes its own hook from `grub.cfg`, so nothing ever retries.
- `/media/startos/config` is a read-write bind of the same filesystem (`startos-initramfs-module:125`), and the script already writes to it a few lines later. Resizing through that path is what worked for the reporter.
- `build.sh:482-485` sizes the root partition as squashfs + 256 MB with the comment "avoiding the need for btrfs resize", which is only true until something else needs space.

Found by reading code; not run on a device. The kernel behaviour on a read-only bind is from `btrfs_ioctl_resize` taking `mnt_want_write_file`, and matches the reporter's observation that the same command succeeds through the read-write bind.

Not tested, but worth checking before 0.4.0.2: an OS update writes the new rootfs image into `images/` beside the running one, which should not fit in the space these boxes have.

### Logs & evidence

From #3624, on the affected Pi:

```
sudo btrfs filesystem usage /media/startos/config   # 1.52 GiB filesystem, 12.26 GiB device slack, ~48 MiB free
sudo btrfs filesystem resize max /media/startos/config   # grew it to 13.78 GiB, 12.30 GiB free
```

helix-nine:
## Summary

- grow the Raspberry Pi btrfs filesystem through its writable config bind mount
- preserve the first-boot resize hook when the root remount or filesystem resize fails
- cover the resize path with the StartOS shell suite and document the first-boot storage preparation

## Verification

- `make start-os-scripts-test`
- pinned rustfmt, tracked-file Prettier, and Taplo format checks

`make format-check` reached Prettier after rustfmt passed, but the slot contains cached gitignored SDK coverage and generated binding files that fail its whole-tree scan. The equivalent check over tracked and non-ignored files passes.

Closes #4009


dr-bonez (dismissed):

helix-nine:
Rebased onto current `master` (`fa438225a`) to resolve the changelog conflict. The resolved changelog retains master’s Raspberry Pi WiFi and service-container teardown entries alongside this PR’s filesystem entry; the PR’s functional diff is unchanged. `make start-os-scripts-test`, `bash -n`, and scoped Prettier checks pass.

The force-push dismissed the approval, so the PR now needs a fresh approving review.

dr-bonez (approved):

PR #4023, opened 2026-09-19
fix(start-core): reset connections after gateway changes

### LINKED ISSUE #3914: Changing a service's default gateway should probably restart the service
### Prerequisites

### StartOS version

0.4.0.1

### Architecture / image

x86_64 (AMD64) — Full

### Server hardware

2024 Server One 

### Browser & OS

Firefox 151

### Which surface?

Networking & addresses

### What happens

When you go to switch the default gateway for a service (eg. Tor),

The service's default gateway is changed, but only for new connections.  Existing connections are still traversing the original default gateway.

A user recently reported this issue here:
https://matrix.to/#/!6ywjOBNb6PUjxvZ0yjaeKMxSJlpt3nW9nmOsT-S8iKs/$qsBvrFbgxBCDB47IP9CQFqCAtpaAi13Ntw89VLhnpXc?via=matrix.start9labs.com&via=start9.me

Theory that explains this:
If you 1) tell the Tor service to use your StartTunnel VPS as the outbound gateway
then 2) Switch your default gateway for the Tor service to Wired Connection 1, previously established TCP connections keep on going through the StartTunnel gateway.

The service should probably be forced to restart to match user expectations

Setting the default outbound gateway of a service should restart the service, so that all connections traverse only the intended gateway.

### Steps to reproduce

I actually can't confirm this until I can shell into the Tor service:
https://github.com/Start9Labs/tor-startos/issues/27

### Logs & evidence

```shell

```

helix-nine:
Closes #3914

## What

Changing or clearing a service's outbound gateway now deletes conntrack entries whose original source is that service's container address. Established connections therefore stop using the masquerade address selected by the previous gateway and reconnect through the newly selected route.

- flushes IPv4 flows with `conntrack -D -s <service-ip>`
- flushes IPv6 flows with the equivalent `-f ipv6` filter
- runs only after the replacement policy-routing rule is active, and only when the selected routing table changes
- keeps the service running instead of restarting it

The StartOS outbound VPN guide and unreleased changelog describe the connection reset.

## Testing

- `cargo test -p start-core conntrack_filter_family_follows_its_source --features=test`
- `cargo check -p start-core`
- `make start-core-test` — 541 passed, 3 ignored
- Prettier check on the changed documentation
- `cd projects/start-docs && ./build.sh`


dr-bonez (dismissed):

helix-nine:
Rebased onto current `master` after #4006 merged and resolved the outbound-rule conflict. The conntrack reset now snapshots the reconciler's actual `ServiceOutboundRule` set and runs only when that set changes, preserving no-op gateway/interface events while covering gateway switches, availability changes, and kill-switch transitions.

Reverified the rebased head with:
- `cargo test -p start-core 'net::net_controller::tests::' --features=test` (9 passed)
- `cargo check -p start-core`
- Prettier on the changed docs
- `mdbook build` for the StartOS book

The force-push dismissed the prior approval, so this now needs a fresh review.

dr-bonez (approved):

PR #4027, opened 2026-09-20
fix(start-os): pipe any name through a passthrough port

### LINKED ISSUE #3950: A passthrough binding rejects clients that send SNI for a domain not enabled on that binding, even on an enabled IP address — the vhost name lookup has no fallback to the port's passthrough target (net/vhost.rs:1809)
### Environment

StartOS 0.4.0.2 (`a30d84d2`), x86_64, Purism Librem Mini v2; same code on `master` @ `81c0b5a08`. Area: networking — SNI passthrough bindings (`secure: { ssl: true }`, no `addSsl`).

### What happens

A passthrough binding whose public IPv4 address is enabled refuses any client whose ClientHello names a domain that is not enabled on that same binding, even though the connection arrived on the enabled IP:port. The client gets a TLS `unrecognized_name` alert before any bytes reach the container. The same connection **without** SNI goes through.

Observed with the Stalwart package's IMAPS binding (`993`, published on `10993`, passthrough, public IPv4 `64.23.194.12` enabled; the host also has the public domain `mail.stalwart.start9.dev` enabled on its HTTP binding but not on `993`):

```
$ openssl s_client -connect 64.23.194.12:10993 -servername mail.stalwart.start9.dev
SSL routines:ssl3_read_bytes:tlsv1 unrecognized name … SSL alert number 112
$ openssl s_client -connect 64.23.194.12:10993
subject=CN = rcgen self signed cert      ← the container's own certificate, as intended
```

Every real mail client sends SNI with the hostname it was configured with, so IMAPS and SMTPS simply do not work on a passthrough binding unless the user also toggles the domain on for each mail port — a step nothing prompts for, and one that is off by default by design (`interfaces.md`: "Every other binding on the same MultiHost also gains the domain, but off by default"). The Stalwart package moved 465/993 to plain DNAT bindings to avoid it.

On a passthrough port, an unknown server name should fall through to the port's no-SNI target when that target is itself a passthrough. The OS neither selects a certificate for the name nor terminates anything on that path — the container does — so the name carries no routing information StartOS needs to act on. Rejecting is only meaningful for terminating targets, where the name selects a certificate.

### Steps to reproduce

1. Package binds a port with `protocol: null, addSsl: null, secure: { ssl: true }` and the container serves TLS on it (LND's gRPC binding, or Stalwart's IMAPS).
2. Enable the public IPv4 (or the LAN IP) address on that binding.
3. `openssl s_client -connect <ip>:<port> -servername <any domain the host owns but that is not enabled on this binding>` → alert 112. Same command without `-servername` → handshake with the container.

`shared-libs/crates/start-core/src/net/vhost.rs`:

- `host_key` (`:1687-1691`), whose doc comment states the design: "lookup has no fallback, so a host answers to the names it was given and nothing else". An IP-literal SNI is folded to `None`; a real name is looked up verbatim.
- `VHostTlsHandler::get_config` (`:1809-1829`): `self.mapping` is keyed by `Option<name>`; a name with no entry on this `(ip, port)` gives `routed = None`, and the handler returns `None` (→ rustls `unrecognized_name`) unless it is an ACME challenge. The `is_passthrough()` branch at `:1834` is only reached for a name that *was* found.

Fix shape: when the keyed lookup misses, retry with `None` (the address-level target) and accept it if that target is a passthrough — a terminating target keeps today's behaviour.

### Logs & evidence

- `openssl` transcripts above, 2026-09-15, against the dev box through StartTunnel; identical result on the LAN address.
- Related: #3949 (which would let mail ports use `addSsl` instead), #3886.

MattDHill:
## Why

Since `9ad7f0c34` the vhost lookup has no fallback: a ClientHello whose SNI has no entry on the port is refused with `unrecognized_name`. That gates terminating targets and passthroughs alike. On a passthrough the gate protects nothing — the OS selects no certificate and serves no content of its own there; the container answers with whatever certificate it holds — and it shuts out every daemon whose certificate carries a fixed name.

cln-grpc's certificate names only `cln` and `localhost`, hard-coded upstream in `plugins/grpc-plugin/src/tls.rs`, so every client that verifies it sends `cln` as SNI, and Alby Hub pins that name. No binding can ever enable `cln`, so Core Lightning's gRPC interface has refused every verifying client on every 0.4.0.2 build. #3950 hit the same lookup with Stalwart's IMAPS binding, where enabling the domain on the binding was a remedy; here there is none, which is why Start9Labs/cln-startos#201 proposes dropping to a plain DNAT and losing IPv6 with it.

## What changed

- **`net/vhost.rs`** — a name with no entry now falls through to the port's bare-IP entry where that entry is a passthrough. The fallback admits by that entry's own address policy (`filter_private`, then `filter`), so a name never widens what a port exposes. A terminating bare-IP entry — the OS's own 443 among them — still refuses a name it was not given, and the ACME gate on a matched target is unchanged. The lookup moves into `route` so it can be exercised with stub targets.
- **Packaging guide, _Serving Your Own TLS_** — says which connections a passthrough port pipes through regardless of SNI, with cln-grpc as the case in point, and names the exception below.

**What still requires the name.** The bare-IP entry's private set is built from every enabled LAN-side address (IP, `.local`, private domain), so on the LAN any enabled name is enough. Its public sets come only from the bare public IPv4 and public GUAs, so a client on the internet that reaches the port through a public domain, with the binding's public IP address left disabled, is admitted by the domain entry alone and must still ask for that domain. Enabling the public IP address on the binding lifts that. Extending the fallback to named entries would make a domain-only public binding answer no-SNI dials on its bare IP, which is an exposure change this PR does not make.

No changelog entry: the behavior this restores is 0.4.0.1's, and `0.4.0.2` is untagged.

## Verification

- `route_tests` covers: a name served by its own entry; an unregistered name reaching a passthrough bare-IP entry; an unregistered name refused where the bare-IP entry terminates, beside a passthrough domain (the 443 shape); a dial naming no host, or an IP literal, served by the bare-IP entry; a private match outranking a public one; a dead entry not routing.
- **Not run locally** — the crate's test build is the full backend. CI runs it.
- **Not yet verified on a box.** The check that matters: on a 0.4.0.2 box with Core Lightning 26.6.7:2, `openssl s_client -connect <lan-ip>:2106 -servername cln -cert client.pem -key client-key.pem -CAfile ca.pem` completes with `Verify return code: 0 (ok)`, and `-servername www.google.com` against the UI on 443 is still refused.

Closes #3950. Supersedes Start9Labs/cln-startos#201.


MattDHill:
@helix-nine review

helix-nine (changes requested):
The routing change looks correct, but the new guide paragraph overstates its scope. `route` falls back specifically to the `None` (bare-IP) entry when that entry is a passthrough; it does not make every named passthrough on the port accept arbitrary SNI. If a binding has only a private/public domain enabled and no enabled IP address, a client can dial that enabled domain while overriding SNI to a fixed certificate name such as `cln`, and the lookup still has no `None` target to fall through to. Please scope the paragraph in `projects/start-sdk/docs/src/interfaces.md` to connections through an enabled IP address / the port's bare-IP passthrough target rather than saying SNI is unchecked on any passthrough port.

Verified: `PLATFORM=x86_64 make metadata`; `cargo test -p start-core route_tests --features=test` (6 passed); `projects/start-docs/build.sh` (all books built). CI is also green.

MattDHill:
Scoped the guide paragraph, amended in `b22e9511b`. It now says a passthrough port pipes through, whatever the SNI, a connection it admits by address; that on the LAN any enabled address of the binding (IP, `.local`, private domain) does that, since the bare-IP entry's private set is built from all of them; and that from the internet only an enabled public IP address does, so a client reaching the port through a public domain with the public IP left disabled must still ask for the domain. The PR body carries the same exception. Code unchanged.

helix-nine (approved):
The amended paragraph now scopes the fallback to connections admitted by the bare-IP passthrough target and accurately distinguishes LAN addresses from the WAN public-IP requirement. The code is unchanged from the reviewed head; I rebuilt all documentation books successfully at `b22e9511b`.

MattDHill:
Start9Labs/cln-startos#201, the package-side workaround, is closed in favor of this PR: https://github.com/Start9Labs/cln-startos/pull/201#issuecomment-5751436630.

@dr-bonez — one decision here is yours, and it settles whether this PR is the whole fix for Core Lightning or most of it.

The fallback stops at the port's bare-IP entry. On the LAN that covers every case, because that entry's private set is built from every enabled LAN-side address (IP, `.local`, private domain). From the internet it covers only a binding whose public IP address is enabled: a client that reaches a passthrough port through a public domain, with the public IP left disabled, is admitted by the domain entry alone and must still name the domain. For CLN that is a remote Alby Hub over a custom domain — it names `cln`, so it is refused until the user also enables the public IP on the gRPC binding.

**A — leave it as this PR has it.** A public domain on a passthrough binding keeps the "domain enabled, bare IP hidden" boundary: a scanner hitting the WAN IP on that port with no SNI, or a random one, still gets alert 112. Users in the domain-only setup enable the public IP on the binding, and packages whose clients pin a name document that.

**B — extend the fallback to named entries.** A name that matches nothing, or no name at all, is served where every entry admitting the arrival is a passthrough to one backend. A domain-only public passthrough binding then answers any name — and so answers no-SNI dials on its bare WAN IP too, so the boundary above goes away for passthrough ports (it stays for terminating ones, where the name selects a certificate). It needs a "same backend" notion on the vhost target trait so a manual `add-passthrough` beside a package's domain, or two packages' domains on one port, can never be misrouted.

Which? If A, this PR is complete as is. If B, say whether you want it in this PR or as a follow-up.

dr-bonez:
I'm firmly in camp A.

MattDHill:
Settled: A. The fallback stays at the port's bare-IP entry; a client reaching a passthrough port through a public domain with the public IP disabled names the domain, and the guide says so. This PR is complete as is.

MattDHill:
@dr-bonez ready for your review. Helix approved after the guide fix, CI is green, and with A settled nothing else is changing here.

dr-bonez:
Changed my mind on this one. If an interface doesn't want SNI filtering it shouldn't mark itself ssl, which is what #3950 already concluded. The IPv6 loss in Start9Labs/cln-startos#201 is the package pinning grpc-host to 0.0.0.0, not the plain forward, which DNATs GUAs to the container's v6. Let's reopen Start9Labs/cln-startos#201 with grpc-host set to `::` and close this, keeping only a short guide note that passthrough routes by SNI and fixed-name certs should bind ssl: false.

PR #4030, opened 2026-09-20
feat(start-sdk): add setupPrimaryUrl and the configured address view

MattDHill:
## Summary

- `sdk.setupPrimaryUrl()` replaces the "Set Primary URL" action and init watcher that a dozen packages hand-roll. It returns the action to register, the init hook to add to `setupInit`, and `read()`, a reactive reader over the package's own store field.
- The hook judges the stored choice against the interface's **configured** addresses rather than the reachable ones, and follows the chosen hostname through a port or scheme change. A `.local` choice is judged only while some LAN interface is up, the always-on container bridge not counting; an IP choice only while the interface it came from is up, which the hook records beside the store file (`<store>.<action id>.json`) and leaves alone when unknown; a domain or Tor choice at once.
- A choice judged gone is replaced by `defaultUrl`'s pick, the `.local` address by default, as SearXNG already does. `onRemoved: 'task'` raises a task instead, `important` unless the package says `critical`; the task is also raised when nothing is left to fall back to.
- `addressInfo.configured` on a filled address: the user-enabled rows without the `mdnsResolvable` pruning, chaining the usual filters (`addressInfo.configured.nonLocal.format()`).
- Docs: the recipe rewritten around the helper, a reference section under Interfaces, the nomination example and note updated, changelog entries under the unreleased 3.0.0 heading.

## Why

Every package carrying the pattern (cal-diy, discourse, firefly-iii, ghost, immich, mattermost, bolt12-pay, miniflux, and more) compares its stored URL against `addressInfo.nonLocal`, which drops `.local` whenever no LAN IP is up on its interface (the 2.0.0 `mdnsResolvable` rule), and StartOS rebuilds every host's address table without its LAN rows before NetworkManager reports. A server that initializes services without its LAN address, or loses the link, raises a critical task that stops the service and asks the user for the URL they already chose. Reported on immich by a user whose box sees it nightly.

The companion StartOS change is #4031. Packages adopt the helper as they move to SDK 3.0.


projects/start-sdk/lib/primaryUrl/setupPrimaryUrl.ts
```diff
@@ -0,0 +1,152 @@
...
+      host &&
+      Object.values(host.bindings).find(b => interfaceId in b.interfaces)
+    if (!binding) return null
+    return {
+      urls: binding.interfaces[
+        interfaceId
+      ].addressInfo.configured.nonLocal.format('urlstring'),
+      settled: binding.addresses.available.some(h =>
+        ['ipv4', 'ipv6', 'mdns'].includes(h.metadata.kind),
+      ),
```
helix-nine: [P1] Exclude always-on internal interfaces from the settled signal. Once NetworkManager reports `lo`/`lxcbr0`, every binding has IPv4/IPv6 rows even when the physical LAN is down. `update_addresses` then drops the LAN IP and `.local` rows, `configured.nonLocal` correctly excludes the bridge from `urls`, but this predicate remains true because of `10.0.3.1`; the stored `.local` URL therefore falls through to `createTask`, reproducing the task this helper is meant to suppress (and stopping packages that retain `critical`). I added a local case with only an `lxcbr0` row plus the onion row; the expected no-task assertion fails with one `createTask` call. The existing unsettled test uses only the onion row and misses the real post-boot shape.

helix-nine (changes requested):
The helper/API shape is good, and the existing start-core and SDK suites pass. The settling predicate still makes the reported task reappear once the always-on bridge is present, though, so this half does not yet provide the link-drop behavior its docs promise.

MattDHill:
Reworked per review. The generic "settled" predicate is gone. The hook now classifies the stored hostname: a `.local` name is judged only while some mDNS row exists, an IP only while some non-bridge IPv4/IPv6 row exists (`filterNonLocal` drops `lxcbr0`, loopback and link-local), and a domain or Tor address is judged at once, since those rows never come and go with the link. Added the bridge-plus-onion case you described (no task), a LAN IP choice with the LAN down (no task), a domain gone while the LAN is down (task), and a renamed server (task). 136 tests pass. Squashed into the one commit.

projects/start-sdk/lib/primaryUrl/setupPrimaryUrl.ts
```diff
@@ -0,0 +1,169 @@
...
+      sameHost.find(u => parse(u)?.protocol === was?.protocol) ?? sameHost[0]
+    if (replacement) {
+      await write(replacement)
+      await clear()
+      return
+    }
+    if (!current) return
+    const kind = kindOf(was?.hostname ?? '')
+    if (kind === 'mdns' && !current.mdns) return
+    if (kind === 'ip' && !current.ip) return
```
helix-nine: **P1 — An unrelated live NIC makes a dropped-link IP look removed.** `current.ip` means any non-local IP exists on the binding, not that the interface which owned the stored IP has settled. On a box with Ethernet `192.168.1.10` selected and Wi-Fi `192.168.2.10` also up, unplugging Ethernet removes `192.168.1.10` while leaving `current.ip === true`, so this raises the task again even though the selected address merely went away with its link. I reproduced it by running `setup([bridge, row('192.168.2.10', { kind: 'ipv4', gateway: 'wlan0' }), onion], 'http://192.168.1.10:8080')`; `createTask` was called. The decision needs evidence about the stored IP's own source, rather than the presence of an IP from another interface (or it must remain conservative when that source cannot be identified).

helix-nine (changes requested):
The bridge-only boot case is fixed and the new tests pass, but the replacement readiness signal still recreates an IP task when the chosen interface alone loses link. I reproduced this with a regression test; details inline. `make -C projects/start-sdk test check` passes on the submitted head (136 tests).

MattDHill:
Reworked per review. An IP choice is now judged only against its own interface: when the URL is stored (by the action, the default seed, or a port/scheme rewrite), and on the first run that finds a legacy choice among the rows, the hook records which gateway that IP's row came from in `primary-url.json` beside the store file. At judge time an IP is gone only when that gateway has a non-bridge IP row again without it. An IP whose source is unknown is left alone, per your "remain conservative" note. Your repro (`[bridge, wlan0 192.168.2.10, onion]` with `192.168.1.10` chosen) is now a test both with recorded provenance and without; a renumbered interface (`eth0` back with `192.168.1.20`) raises the task. 139 tests pass, `make check` clean. Squashed into the one commit.

projects/start-sdk/lib/primaryUrl/setupPrimaryUrl.ts
```diff
@@ -0,0 +1,208 @@
...
+        gateway: gatewayOf(h),
+      })),
+      mdns: lan.some(h => h.metadata.kind === 'mdns'),
+      up: lan.flatMap(h => gatewayOf(h) ?? []),
+    }
+  }
+  const read = () => store.file.read(store.get)
+  // Which interface the chosen IP came from; an IP is gone only once that interface is up without it.
+  const source = FileHelper.json(
+    join(dirname(store.file.path), 'primary-url.json'),
```
helix-nine: **P1 — Do not reserve a sibling filename that may be the caller's store.** `store.file` is arbitrary, and `primary-url.json` is a natural filename for the model this helper consumes. If the caller passes `/data/primary-url.json`, `source` and `store.file` are the same file: the initial `write(url)` first stores `{primaryUrl: ...}`, then `record()` overwrites it with `{url, gateway}`, deleting the package's setting (and any other fields in that model). I reproduced this by changing the test store basename to `primary-url.json`; after the first init, `p.stored()` is `undefined` instead of the seeded URL. Every helper instance in one directory also shares this provenance file. Please keep provenance in caller-owned storage or otherwise make its location distinct and per instance rather than claiming this fixed sibling name.

helix-nine (changes requested):
The per-gateway provenance closes the multi-NIC case, and the submitted SDK suite/typecheck pass (139 tests). The fixed companion filename can overwrite the caller's store, though; I reproduced that data-loss path in a regression test. Details inline.

MattDHill:
Reworked per review. The record is no longer a fixed sibling name: it is `${store.file.path}.${actionId}.json`, so it cannot equal the caller's store (strictly longer than its path) and is distinct per helper instance (two helpers on one store need distinct action ids anyway). Your repro is a test: a store named `primary-url.json` keeps its seeded URL and gets its record beside it. Also covered: a second instance with `actionId: 'set-api-url'` writes only its own record, and a damaged record reads as none and is rewritten on the next store of a URL rather than failing init. 142 tests pass, `make check` clean. Squashed into the one commit.

helix-nine (dismissed):
The companion record is now distinct from the caller store and namespaced per action, with coverage for the collision, multiple instances, and corrupt-record recovery. The per-gateway lifecycle still resolves the bridge and multi-NIC cases from the prior rounds. Verified with `make -C projects/start-sdk test check`: 142 tests passed and typecheck passed.

MattDHill:
One more change on top of the record-file fix, from Matt's review of the design: a choice judged gone now falls back by default instead of tasking. The hook stores `defaultUrl`'s pick in its place (the `.local` address unless the package overrides it), which is what SearXNG already does, and clears any task. `onRemoved: 'task'` keeps the previous behaviour for a service whose users must be told rather than moved; the task is also raised when nothing is left to fall back to. The same-kind gate runs first either way, so a link that is down never triggers a fallback. Tests cover the domain-gone fallback, a renamed server following to its new `.local`, and the no-candidate task; the task-mode cases now opt in. 145 tests pass, `make check` clean. Squashed into the one commit.

helix-nine (approved):
The fallback path preserves the link-state gates, writes and records the replacement through the existing helper, clears the stale task, and retains explicit task mode for origin-sensitive services. The API docs and changelog describe the new default. Verified with `make -C projects/start-sdk test check`: 145 tests passed and typecheck passed.

projects/start-sdk/docs/src/interfaces.md
```diff
@@ -306,7 +308,36 @@ Some addresses are left out of the comparison, in the cases where StartOS can te
...
+import { storeJson } from './fileModels/store.json'
+
+export const primaryUrl = sdk.setupPrimaryUrl({
+  hostId: 'ui-multi',
+  interfaceId: 'ui',
+  store: {
+    file: storeJson,
+    get: s => s.primaryUrl,
+    set: url => ({ primaryUrl: url }),
+  },
```
dr-bonez: why assume a file? just using a get/set that use the file model directly seems like it is both cleaner and simpler

helix-nine: Agreed. The sidecar exists only because the current store contract can read/write the URL but not its provenance. I would have `store.get` return `{ url, gateway }` and `store.set` accept that state against the same `FileHelper`; `read()` can still project just `url`. That keeps the write atomic and removes the hidden-file naming/lifecycle entirely.

helix-nine:
Reworked per @dr-bonez's review, on top of Matt's commit. I rebased it onto master first, because it conflicted with #4035.

- **API:** `setupPrimaryUrl` now takes `get`/`set` over the package's own storage and returns `action`, `effective(effects)` and `createTask(effects, severity, options)`.
- **`effective`** resolves the choice each time it is read, instead of the init hook overwriting it. The choice is never overwritten, so the sidecar provenance file and the link-state checks are gone too.
- **`createTask`** declares the addresses as an `input-not-matches` task's accepted input and leaves activation to StartOS.
- **`addressInfo.configured` is dropped**, since #4035 already keeps `.local` in `nonLocal`.

The description is rewritten to match. It includes two behaviours to weigh: a `critical` task created during the service's own init, and an IP choice while its link is down.

`make -C projects/start-sdk test check check-fmt`: 148 tests pass, typecheck and formatting are clean.

PR #4031, opened 2026-09-20
fix(start-core): wait for the network to settle before starting services

MattDHill:
## Summary

- Service initialization waits for the network to settle. `NetworkInterfaceController::settle()` waits for NetworkManager's `Startup` property to clear, lists the devices NetworkManager activated, and returns once the sync task has written a snapshot to the database with every one of them connected. After 30 seconds it logs a warning and services start anyway.
- `watch_ip` clears its own entry when NetworkManager reports the device unmanaged or without an active connection, so a link loss drops the interface's `.local` and IP rows. An interface the table never knew is never added.
- The interface watcher's seed is unchanged from master: every gateway starts disconnected until its device reports.

## Why

`RpcContext::init` builds the net controller and then initializes services with no wait for the network. The seeded all-disconnected snapshot was written into every host's address table, and because each device reports independently, the table was then rebuilt from partial snapshots, the always-on container bridge first. An init hook reading its own addresses at that moment saw a table without its LAN rows, and the packages that guard a primary URL with a critical task stopped themselves and asked the user for the URL they already had. This bites hardest when `startd` restarts on a running server: NetworkManager is already stable, and the only thing unsettled is `startd`'s own first pass over the devices.

An earlier revision of this PR kept each gateway's last-known `ip_info` in the seed instead. That seed feeds the interface watch every net consumer subscribes to, so vhost, DNS and the port forwards would have acted on last boot's addresses, ifindexes and WAN IP as live state, and an interface whose watcher never reported would have stayed connected forever.

`Startup` is what `nm-online -s` and `NetworkManager-wait-online.service` block on. Ordering `startd` after that unit is not an option: it delays the daemon and the init UI, and `startd` itself binds the persistent connection profiles NetworkManager needs and restarts it (`synchronize_network_manager`). For the same reason `settle()` reads `Startup` when services are about to start rather than tracking it from watcher start — the first NetworkManager instance finishes startup holding none of those profiles.

The SDK side is #4030: `sdk.setupPrimaryUrl()` judges a `.local` or IP choice only while the server has a LAN address of that kind, so packages stay correct on an OS without this fix as well.

`cargo check -p start-core` passes. Not yet run on a device.


shared-libs/crates/start-core/src/net/gateway.rs
```diff
@@ -2984,7 +2981,11 @@ impl NetworkInterfaceController {
 
                         tokio::select! {
                             _ = ip_info_watch.changed() => {
-                                ip_info = Some(ip_info_watch.read());
+                                // Skips the seeded snapshot, which marks every interface disconnected.
+                                let info = ip_info_watch.read();
+                                if ip_info.is_some() || info.values().any(|i| i.ip_info.is_some()) {
```
helix-nine: [P1] Do not treat one interface's report as a complete startup snapshot. `watcher()` starts one concurrent `watch_ip()` job per NetworkManager device, and each job mutates this shared `Watch` independently, so `changed()` can wake after `lo` or `lxcbr0` publishes while the persisted Ethernet/Wi-Fi entries still have `ip_info: None`. Both internal interfaces are managed and always connected on a released StartOS VM, so `any(...)` accepts that partial map; `sync()` then calls `update_addresses`, which removes the still-unreported LAN IP and `.local` rows exactly as before. Whether the physical interface wins the scheduling race now decides whether the task reappears. Please retain the last-known row for each not-yet-reported interface, or add an explicit per-interface initialization signal, rather than using the first `Some` anywhere as the boundary.

helix-nine (changes requested):
This removes the deterministic all-disconnected write, but the first accepted value is still a partial per-interface snapshot. That leaves the same boot-time address wipe reachable and undermines the tandem guarantee with #4030.

MattDHill:
Reworked per review, taking the "retain the last-known row" option. The seed no longer marks every gateway disconnected: each keeps its last-known `ip_info`, so the first sync and any partial one (bridge first, LAN later) leave a not-yet-reported interface's rows in place. To make that safe, `watch_ip` now clears its own entry when NetworkManager reports the device unmanaged or without an active connection, which it never did before, so a disconnected interface drops its rows at boot and on a link loss alike; an interface the table never knew is still never added. The sync task is back to master's shape, the `seeded` oneshot stays. Squashed into the one commit; not built locally, CI compiles it.

helix-nine (dismissed):
The revised seed keeps each persisted interface row until that interface reports, and the new disconnect path clears a known row on both unmanaged and inactive NetworkManager states. This closes the partial-snapshot race from my prior review. Verified with `make start-core-test`: 561 passed, 3 ignored.

dr-bonez (approved):

PR #4035, opened 2026-09-21
fix(net): preserve mDNS enablement without LAN IPs

helix-nine:
## Summary

- A service's `.local` row survives its gateways losing their addresses. `mdns_gateways` takes the binding's previous address table, and a gateway with no `ip_info` keeps the membership it had there. The row goes once the gateway leaves the table.
- A private IP's state is an override against its `.local` row: explicit on (`lan_enabled`, new), explicit off (`disabled`), or neither, in which case it follows `.local`. An address the server is assigned later therefore starts out the way `.local` is set, with nothing written at discovery.
- On a non-SSL port `.local` reads as on while a LAN IP it resolves to is on, and switching `.local` off takes those IPs off with it: a gateway whose LAN level a private domain links moves as that group, and every other LAN IP loses its explicit-on override and follows `.local`. The switch is never locked. A public GUA does not hold `.local` on, because a `.local` switch must not un-publish a WAN address.
- A forwarded (non-SSL) port admits a LAN address only through that address's own enabled row, or a private domain on its gateway. `.local` no longer opens every IP on its gateways, so switching a LAN IP off closes it, and the warning above describes a name that really is unreachable.
- On a TLS port the bare-IP `*` vhost answers only on its enabled SSL-port IP rows. With `.local` on and the bare IP off, that address answers the `.local` SNI and nothing else.
- A filled address lists every address the switch shows as on. The SDK's reachability prune and `utils.mdnsResolvable` are removed, so the Interfaces tab and a package agree on every address.
- The UI shows the backend's state through a shared `utils.isAddressEnabled` in place of its own copy. The `.local` switch is interactive, and on a non-SSL port a `.local` row that is on while every IP row it resolves to in its gateway table is off shows a warning.
- `v0_4_0_2::up` backfills `lanEnabled` and its migration revision moves to 3, because the UI reads raw patch-db JSON.

## Why

`update_addresses` built the `.local` row only from gateways holding `ip_info`, so it left `available` at the boot seed and, since #4031, on link loss. A package comparing its stored URL against its filled addresses saw the URL as removed and raised a critical task for the address the user already chose. `device_type` lives inside `ip_info`, so a disconnected gateway cannot otherwise be told apart from any other interface, which is why the previous table is the input.

This replaces the first revision of this PR, which masked a binding's LAN IPs behind a disabled `.local`. A mask cannot hold `.local` on beside one disabled IP, or `.local` off beside one enabled IP, and its SDK half could not run: the row it meant to keep was already gone from `available`. The mask's commit is kept as an ancestor of the merge so the branch fast-forwards; the tree is the new implementation's whole.

`enabled` could not hold the explicit-on record: `set_gua_wan` and `resolve_nonssl_gua` both remove a LAN GUA from it. The vhost lookup has no fallback, so on a TLS port `.local` and the bare IP really are independent, and the forced-on rule is confined to non-SSL ports, where a plain port never carries `ssl: true` rows.

Not in this change: on a non-SSL port an enabled public GUA still serves `.local` from the LAN while its switch reads off; the Interfaces page documents that, and a better answer is left for later. A binding that already holds `.local` in `disabled`, reachable only through the CLI or RPC, will see its LAN IPs without an override turn off on upgrade.

## Verification

- `cargo test -p start-core --features=test --lib -- net::net_controller net::host:: net::forward net::vhost` (131 passed) on the head commit; `version::v0_4_0_2` and `version::tests` passed earlier on the branch
- `cargo check -p start-core`, no warnings in the touched files
- the generated `DerivedAddressInfo.ts` matches the committed binding
- start-core jest suite (252 passed, 9 skipped)
- `npm run check:i18n`, `check:shared`, `check:ui`, and `npm run build:ui`
- pinned rustfmt and prettier on the changed files

Helix ran the rest on 606abcaa2: root TypeScript checks, SDK bundle and typecheck, container-runtime typecheck, both mdBook builds, regenerated bindings with no diff, and the full start-core suite. Not rerun on the head commit. Not yet run on a device.


dr-bonez:
Replaced the mask with overrides against the .local - a LAN IP with no override follows it, so an enabled .local with one disabled IP is valid and a newly assigned IP starts out however the .local is set. The .local row now survives a gateway with no addresses in update_addresses, which is what the SDK change needed and never got. The mask commit is kept as an ancestor so this branch fast-forwards - pull it, don't rebase or force-push over it. Reasoning is in the description and in 57f9c46a7.

helix-nine:
Pulled `606abcaa2` fast-forward. The override model resolves the mask limitations and the missing-backend-row problem.

While reviewing it, I found one Rust/TypeScript mismatch: backend `mdns_on` counts an enabled public GUA as serving `.local` on a non-SSL binding, but `mdnsResolvable` excluded every public address. Filled addresses could therefore drop `.local` beside a disabled private IPv4 + enabled public GUA, or retain it when a disabled public GUA was the only available IP. Commit `e4a9cc690` now counts public GUAs and adds both regressions.

I also ran the omitted checks: root TypeScript checks, SDK bundle/typecheck, clean container-runtime typecheck, both mdBook builds, regenerated bindings (no diff), and the full start-core library suite. The suite reached 871 passed / 3 ignored; its sole failure remains the unrelated `docs.start9.com/llms.txt` timeout, independently reproduced with `curl`. Not run on a device.

dr-bonez:
Two changes on top of e4a9cc690. The .local switch is never locked - on a non-SSL port switching it off takes the LAN IPs it resolves to off with it, and a public GUA no longer holds it on since a .local switch shouldn't un-publish a WAN address. And the SDK lists whatever the switch shows as on, so the reachability prune and utils.mdnsResolvable are gone, which drops the GUA counting and its two tests. Pull, don't rebase.

helix-nine:
Pulled through `bd36ec4c2` fast-forward and reviewed both commits. The switch is now the single authority in the UI and filled-address view, while the backend cascade makes an explicit `.local` off effective without touching a public GUA.

Focused verification passes on this head: the 7 Rust mDNS tests, start-core TypeScript check, and all 11 filled-address tests. The full CI matrix is still running.

helix-nine:
VM verification passes on current head `199ed271f`.

I built and deployed this head's `startbox` plus embedded UI onto a StartOS 0.4.0 VM, then installed two test packages: a raw non-SSL binding and the normal Hello World HTTP/TLS binding. I drove the real host-binding RPCs and evaluated the returned `DerivedAddressInfo` with this head's `isAddressEnabled` / `filledAddress`.

Verified:
- A per-IP disable leaves enabled `.local` stored and listed; the IP alone is omitted.
- During a full gateway disconnect, the `.local` row survives with its setting, while gateway IP rows disappear. Reconnect restores the IP rows and their overrides.
- IPv4 reassignment `192.168.122.159 → .160 → .159` follows the stored `.local` setting for each newly assigned address, both on and off.
- On a non-SSL binding, enabling a LAN IP can hold `.local` on; switching `.local` off remains possible and clears that explicit-on override, taking its private IPs off. The actual forwarded port transitions closed → open → closed with the `.local` toggle after reconciliation.
- A public GUA remains enabled and present in the filled address after `.local` is switched off.
- On the TLS binding, the overrides remain independent: IP on / `.local` off and `.local` on / IP off both produce the corresponding filled-address rows.
- All settings survived `startd` restart and two full VM reboots.
- The latest embedded UI bundle is being served. Against live rows, the warning predicate is true for enabled non-SSL `.local` with every same-port gateway IP off, false for the equivalent TLS state, and false while the gateway has no IP rows.

No VM failures found.

helix-nine:
Retested the new forwarding head `6336e0ef8` on the VM. This fixes a real discrepancy in my prior pass: on `199ed271f`, disabling the IPv4 row while `.local` and the link-local IPv6 row stayed on could still leave the IPv4 forward open through the old gateway-wide expansion. My earlier closed → open → closed check exercised the `.local` cascade, not per-IP isolation, so it did not catch that distinction.

Current-head network result on the raw non-SSL binding:

1. Initial IPv4 forward: open.
2. Disable only `192.168.122.159`, leaving `.local` and link-local IPv6 enabled: IPv4 forward closes.
3. Re-enable that IPv4 row: forward opens.
4. Disable `.local`: cascade closes the forward again.
5. Add an enabled private domain while the LAN group is off: the group/forward opens; disable the domain: it closes.

The focused Rust test `a_forwarded_port_admits_its_enabled_lan_ips_alone` also passes. The rest of the earlier persistence, reassignment, TLS, GUA, and filled-address results are unaffected. No VM failures on `6336e0ef8`.

dr-bonez (approved):

PR #4041, opened 2026-09-21
fix(start-os): import onions for renamed services

### LINKED ISSUE #4036: Onion addresses of the five packages the 0.3.5.1 migration renames are never imported — torMigration keeps the old package id (v0_3_6_alpha_0.rs:308, v0_4_0_alpha_20.rs:318)
### Environment

Every 0.3.5.1 → 0.4.0 migration, on any hardware and architecture; area: update/migration and networking (Tor). Read on `master` at `00b6db49f`; the paths below have not changed since the alpha.20 migration landed. Found by reading code, not observed on a server — the confirmation step is under Logs & evidence.

### What happens

A server migrated from 0.3.5.1 loses the `.onion` addresses of the five packages the migration renames: `nostr` (→ `nostr-rs-relay`), `fedimintd` (→ `fedimint-guardian`), `ghost`, `synapse` and `monerod` (→ `*-legacy`). Every other package's addresses come through; these five are silently dropped, and the user finds their Tor interface pages empty with no way back from the UI.

The entries land on the package the migration actually installed, so the Tor service imports them like the rest. For a rename that also changed the host id (`fedimintd`'s `main` is not a host `fedimint-guardian` binds), the entry should still be imported rather than skipped, so the parked address is at least visible in Delete Onion Addresses and reachable by Add Onion Service's cross-host reuse (Start9Labs/tor-startos#37).

For servers already migrated, the keys are still on disk in `.onion-migration.json.bak` on the Tor volume, so a follow-up OS migration can hand the renamed entries back as a fresh `onion-migration.json` under the new ids, the way `v0_4_0_beta_10.rs` already rewrites a not-yet-imported file.

- `shared-libs/crates/start-core/src/version/v0_3_6_alpha_0.rs:308` — `torMigration` entries are written with the raw `package` column from the 0.3.5 database (`previous_tor_keys`, `:811`), i.e. the old id.
- `shared-libs/crates/start-core/src/version/v0_3_6_alpha_0.rs:674` — `migrated_id` renames the *installed* package for those five ids, so the package, its hosts and its onion entry end up under different ids.
- `shared-libs/crates/start-core/src/version/v0_4_0_alpha_20.rs:318` — `normalize_legacy_service_identity` is the only rewrite applied when `onion-migration.json` is written, and it handles `STARTOS` → `start-os` alone.
- Tor side, `tor-startos/startos/init/migrateOnionAddresses.ts:39` — an entry whose `packageId`/`hostId` has no host is skipped (`if (!host) continue`), and `:91` renames the file to `.bak` after the loop, so the skip is permanent.

### Logs & evidence

To confirm on a migrated server that ran one of the five on 0.3.5.1: `/media/startos/data/package-data/volumes/tor/data/startos/.onion-migration.json.bak` holds an entry with `"packageId": "nostr"` (or `fedimintd`, `ghost`, `synapse`, `monerod`) while `torrc` has no `# @service nostr-rs-relay …` (or `fedimint-guardian`, `ghost-legacy`, …) block for it, and the Tor service's first-init log carries no `Imported onion address for <that package>` line.

helix-nine:
## Summary

- write the five v1→v2 package-id renames into the original 0.3.5.1 onion-address handoff and normalize handoffs produced by intermediate development builds
- raise the 0.4.0.2 migration revision so servers that already updated recreate a handoff from Tor's backup, containing only the five entries that were previously skipped
- rewrite a still-pending handoff in place, preserve host ids for parked addresses, and document the recovered addresses

The recovery runs in `pre_up`, before services initialize, so Tor consumes the recreated handoff during the same boot. Existing successfully imported entries are not replayed.

## Verification

- `make start-core-test` (570 passed, 3 ignored)
- `cargo test -p start-core 'version::' --features=test` (32 passed)
- `cargo check -p start-core`
- `npx --yes prettier@3.8.3 --check projects/start-os/CHANGELOG.md projects/start-os/docs/src/update-040.md`
- `git diff --check`

Closes #4036


shared-libs/crates/start-core/src/version/v0_3_6_alpha_0.rs
```diff
@@ -675,18 +675,19 @@ impl VersionT for Version {
...
-        "ghost" => "ghost-legacy".parse()?,
-        "synapse" => "synapse-legacy".parse()?,
-        "monerod" => "monerod-legacy".parse()?,
-        "fedimintd" => "fedimint-guardian".parse()?,
-        _ => id.clone(),
-    })
+    Ok(migrated_package_id(id).parse()?)
+}
+
+pub(super) fn migrated_package_id(id: &str) -> &str {
```
dr-bonez:
```suggestion
pub(super) fn migrated_id_str(id: &str) -> &str {
```

helix-nine: Renamed to `migrated_id_str`. I also collapsed recovery of the pending handoff and backup into one read/transform/write path, removed the helper-only test, and replaced it with a test of the actual alpha-20 migration output. The version migration suite and `cargo check -p start-core` pass on `8e3c41dff`.

dr-bonez (approved):

PR #4044, opened 2026-09-21
fix(net): answer TLS for the names a plugin exports

dr-bonez:
## Summary

On `master`, an SSL `.onion` address on an OS-terminated binding fails its TLS handshake with `unrecognized name`. Every plugin-exported SSL hostname is affected; plaintext ones are not.

`0ff264313` made the vhost lookup exact — "a name is served only if it has an entry of its own" — and registered mDNS names as entries to match. The loop in `net_controller.rs` that decides which names a host answers to still admitted only `PublicDomain | PrivateDomain | Mdns`, so a hostname a URL plugin exported (`HostnameMetadata::Plugin`) never got an entry, and the fallback that used to cover it was gone. Released `0.4.0.1` is unaffected: its lookup still falls back to the port's `*` entry.

- **`named_vhost_leg`** decides the `(port, leg)` a name is served on. A plugin's name is served on the binding's assigned SSL port, because its provider forwards to the binding's bridge address whatever port the name itself advertises, and on the private leg whatever `public` the export carries — that flag describes the URL, not how the connection arrives.
- **`ssl_vhost_internal_ips`** gives that entry its accept set: the binding's loopback and bridge SSL addresses, and nothing else. A plugin hostname has no gateways, and its provider connects from its own container, so no gateway, LAN address or upstream port map is involved.
- The entry is otherwise built like any other name, so a terminated binding gets a certificate for the name and a passthrough binding routes it by SNI.
- **Changelog:** the regression never shipped, so this amends the unreleased "Your server answers to its own addresses and no others" entry instead of adding a `Fixed` line.

## Verification

- New unit test `a_plugin_name_is_served_on_the_bindings_ssl_port_over_internal_ips_alone` covers both helpers: a plugin name takes the binding's port and the private leg, a domain keeps its own, an IP row gets no named entry, and the accept set holds the bridge SSL address alone.
- The defect was reproduced on a StartOS `0.4.0.2` alpha VM (x86_64) with the Tor service, over the real Tor network against the `start-os/admin` onion on port 443: with the `.onion` name as SNI the handshake ends in `TLS alert, unrecognized name`; the same onion and port with no SNI return HTTP 200. So the path Tor → `10.0.3.1:443` → the TLS terminator → the UI is sound, and only the named entry was missing.
- **Not exercised on a box with this build.**

Found while testing Start9Labs/tor-startos#39.


shared-libs/crates/start-core/src/net/net_controller.rs
```diff
@@ -518,35 +542,36 @@ impl NetServiceData {
...
                         public_v4: BTreeSet::new(),
                         public_v6: BTreeSet::new(),
                         public_v6_gateways: BTreeSet::new(),
-                        private: BTreeSet::new(),
+                        private: if matches!(addr_info.metadata, HostnameMetadata::Plugin { .. }) {
+                            ssl_vhost_internal_ips(enabled_addresses.iter().copied())
+                        } else {
+                            BTreeSet::new()
+                        },
                         // The public leg's alone, so a name served both ways
```
helix-nine, line 605: [P2] Merge the plugin's internal IPs into existing named entries too

When a plugin hostname is also configured as a private domain on this host, and that domain uses the binding's assigned SSL port (e.g. the admin binding on 443), both rows produce the same `(hostname, port, false)` key. The private-domain row sorts before a public plugin row in `enabled_addresses`, so it creates the entry first. This `or_insert_with` initializer then never runs for the plugin, and `Plugin::gateways()` is empty, leaving its bridge/loopback destinations out of `target.private`. The provider's connection to `10.0.3.1:443` with that SNI is still rejected. This state is allowed by `add_private_domain` and `export_url`; neither excludes an existing plugin/domain row.

Move the plugin-specific `target.private.extend(ssl_vhost_internal_ips(...))` after `or_insert_with`, alongside the other per-row merges, and add a shared-name regression test. The current helper-only test cannot catch the entry-reuse case.

helix-nine (changes requested):
One correctness issue in merging a plugin name with an existing private-domain entry; the standalone onion case looks correct.

Verification at `1250103a`: all 12 `net::net_controller::tests` and all 44 `net::vhost::` tests pass; the changed Rust file passes the pinned rustfmt check and `git diff --check` passes. An isolated Rust harness executing the unchanged named-vhost assembly and helpers reproduces the shared-name failure: plugin alone admits `10.0.3.1`, but adding the private-domain row leaves only `192.0.2.10`. Extending the accept set after entry lookup makes both cases pass. Native tests needed the generated build/env files staged at the legacy include path under shared-libs/build; that temporary directory was removed. I did not exercise this build on a StartOS VM.

shared-libs/crates/start-core/src/net/net_controller.rs
helix-nine: Agreed with dr-bonez’s clarification in Matrix: every enabled configured name should accept connections over the bridge, not just plugin names. That makes the smaller correction to this hunk an unconditional `private: ssl_vhost_internal_ips(enabled_addresses.iter().copied())`. The first row then initializes the common internal accept set regardless of its metadata, and the existing gateway merges preserve its external exposure. My plugin-specific extension recommendation above was too narrow.

Verified that alternative in the isolated assembly harness: plugin alone, private domain alone, public domain alone, and the shared private-domain/plugin name all retain the bridge address (4/4 pass). This fixes the reported entry-reuse failure without an extra per-row merge.

Separate port detail: ordinary domains are still keyed on their advertised domain port, which can differ from the binding’s assigned SSL port. The unconditional accept set makes the domain work on the bridge at that advertised port; accepting it at the binding’s assigned port as well would require an additional named entry.

dr-bonez: Done in 832b8165b: every named entry now starts from the binding's loopback and bridge SSL addresses, whatever row creates it, and the gateway merges add to that.

To make the shared-name case testable, the named-vhost assembly moved out of `update` into `add_named_vhosts`, with the base target passed in as a closure; behaviour is otherwise unchanged. `every_named_vhost_accepts_the_bridge_even_when_names_share_an_entry` runs your four cases through it: a plugin name alone, a private domain alone, a public domain alone, and a private domain plus a plugin name on one key, with the domain row sorting first. All 407 `net::` tests pass and the pinned rustfmt check is clean.

On the port detail: I've left a domain keyed on its advertised port. Serving it at the binding's assigned port as well would add entries for a case no caller has yet; happy to do it if one appears.

The same commit also puts `ssl_vhost_private_ips`'s doc comment back on it; my first commit had inserted the new helpers between them.

helix-nine: Verified at 832b8165b: this resolves the finding. The common internal accept set is installed whichever row creates the entry, and the extracted assembly preserves the gateway merges, public/private keys, and target configuration. The regression test now exercises the production assembly, including the domain-first collision.

All 407 `net::` tests pass locally; the pinned rustfmt check and `git diff --check` pass. Leaving domains on their advertised ports is fine—my port observation is not a blocker. No further correctness findings; approving. I have not run this build on a StartOS VM.

helix-nine (approved):
Re-reviewed 832b8165b. The shared-name accept-set defect is fixed and covered through the production assembly helper. All 407 network tests and scoped formatting checks pass locally. No further correctness findings; VM verification remains unperformed.

PR #4046, opened 2026-09-22
fix(start-os,start-tunnel): clear a port's conntrack entries when its forward comes up

ok:
## Why

`forward-port` and `forward-port6` delete a port's conntrack entries only on
`UNDO=1`. NAT is decided on a flow's first packet, so a flow that reached the
port while no forward existed keeps its entry, made without DNAT, after the
forward is installed. Two things follow:

- The peer stays pointed at the host itself for as long as it keeps
  retransmitting. A DHT node, or any UDP protocol with a keepalive shorter than
  the 30 s UDP timeout, never lets the entry expire.
- The entry occupies the reply tuple a container-originated flow to the same
  peer port needs. Masquerade then picks a different source port for that
  flow. A service that must be seen on the port it binds (HyperDHT, WebRTC,
  STUN) looks healthy while every peer it contacts sees the wrong port.

A port has no forward after a gateway's public address changes, while a
service is updated (the old forward is torn down and the new one installed
after the container has its address), and while its gateway is down. Peers keep
sending through all of them.

Seen on a 0.4.0.1 server exposing a HyperDHT relay on a raw UDP port through
both a home router and a StartTunnel gateway: 37 of 37 sampled flows the relay
initiated left with a rewritten source port, and `[UNREPLIED]` entries for
`<peer>:<port> -> <server LAN IP>:<port>` with the server as the reply side
survived for hours. `conntrack -D -p udp -d <server LAN IP> --dport <port>`
fixed it within 30 s, which is what this change makes the forward do itself.

## What

Run the conntrack delete on both paths of both scripts, after the nft
transaction. `PortForwardState::add_forward` (`net/forward.rs`) only invokes
the script when a mapping is new or its target changed, so an unchanged forward
is not flushed on every host update. Docker does the same when it programs a
published port (`clearConntrackEntries` in moby's
`daemon/libnetwork/drivers/bridge/bridge_linux.go`, referencing moby #8795,
#44688 and #44742).

A TCP flow that was correctly forwarded and loses its entry continues: the
kernel re-creates the entry from the next mid-stream packet
(`nf_conntrack_tcp_loose` defaults on) and applies the same DNAT.

StartTunnel ships the same `forward-port`, so it gains the change under its
unreleased 1.3.1.

## Testing

- `bash -n` on both scripts, and both scripts run against stubbed `nft` and
  `conntrack`: the delete runs on `UNDO=0` and `UNDO=1`, the exit code stays 0
  when `conntrack` reports no matching flows.
- Not yet run on a server. To verify on one: with an outside host sending UDP
  to the port from a fixed source port at least every 20 s, enable the public
  address for the binding (or reinstall the service); `conntrack -L -p udp
  --orig-port-dst <port>` should show no entry whose reply side is the server,
  and a flow the container opens from that port should leave with that port.


helix-nine (commented):
@dr-bonez asked me to simplify the prose, so I did that and force-pushed it to this branch as 79d68cdc0. The code is unchanged and you're still the author. **Please fetch before you push again.** What changed:

- **Script comments.** Each is now one fact: tracked flows keep the NAT they started with, and `conntrack -D` takes a single port and exits 1 when nothing matches. The root `AGENTS.md` § Code style wants a one-idea predicate that doesn't argue its case. The "well under a second" timing is gone because nobody measured it.
- **Changelog entries.** They now say what a user saw and leave out the conntrack mechanism. `AGENTS.md` puts it as "an entry says what changed; the commit says why".
- **Commit message and description.** Both are shorter. I also removed two claims that don't hold as written (point 2 below) and the Claude Code footer.

Review:

**1. A daemon restart now deletes live forwarded flows (mainly on StartTunnel).**

How it happens:
- On startup, `PortForwardController` runs `remove_stale_forward_rules`. That deletes the previous run's rules but leaves conntrack alone.
- It then re-adds every forward as a new mapping.
- On `master`, established flows survive this, because the NAT binding lives in the conntrack entry. With this change, the install-time delete removes them.

What breaks:
- A flow resumes normally if the peer sends the next packet.
- If the target sends first (for example a server retransmitting a download once WireGuard re-handshakes), the kernel tracks that packet as a new outbound flow and masquerades it.
- Through a forward whose external port differs from its target port, the peer then sees the wrong port and the connection dies. That last step is from the kernel source; I haven't reproduced it.

When it happens:
- **StartTunnel:** on any restart without a reboot. That means the install-script update in `updating.md`, and a `Restart=always` crash restart.
- **StartOS:** a startd restart starts every service in a new container, so the existing entries already point at the old one.

**Suggested fix.** On install, delete only the entries this PR targets, which are the ones whose reply comes from the host itself:

```bash
for ((p = sport; p <= sport_end; p++)); do
    for proto in tcp udp; do
        if [ "${UNDO:-0}" = 1 ]; then
            conntrack -D -p "$proto" -d "$sip" --dport "$p" 2> /dev/null || true
        else
            conntrack -D -p "$proto" -r "$sip" --reply-port-src "$p" 2> /dev/null || true
        fi
    done
done
```

In `forward-port6` the same split uses `-f ipv6 -r "$sip" --reply-port-src "$sport"`. A correctly forwarded entry's reply comes from the target, so this filter leaves it alone.

I tested this branch and that variant on a StartOS 0.4.0 VM. An outside host sent UDP from a fixed port every 0.5 s, and the forward pointed at another host:

| step | `master` | this PR | narrower delete |
|---|---|---|---|
| install while a stale `[UNREPLIED]` entry exists | stale entry stays | cleared; next packet is DNATed | cleared; next packet is DNATed |
| restart simulation: delete the rules without flushing, then install | live DNAT entry kept | live DNAT entry deleted | live DNAT entry kept (same conntrack id) |

The v4 and v6 `-r … --reply-port-src` forms both parse on conntrack-tools 1.4.8.

**2. Two claims in the description were overstated.** Both are removed:
- "A TCP flow that was correctly forwarded and loses its entry continues" is only true when the peer sends the next packet (see 1).
- Docker's per-port delete in `clearConntrackEntries` is UDP-only. Its comment says TCP needs none because the peer resets a stale TCP flow.

**3. The on-server check in Testing also passes on `master` when the binding already has a LAN forward.** Enabling the public address changes the forward's source filter. `add_forward` then tears the old forward down first, and that teardown already flushed on `master`. Reinstalling the service, or using any port with no forward at all, exercises the new install path, so I changed the Testing step to reinstalling.

I also added the `repo`, `StartOS` and `StartTunnel` labels.

helix-nine:
At @dr-bonez's request I applied the narrower delete from my review as 82567cef7, a new commit on top with no force-push.

- **Install** deletes only the entries the host answers itself: `-r "$sip" --reply-port-src`.
- **Teardown** still deletes every entry for the port: `-d "$sip" --dport`.

Both scripts pick the filter once and share the existing loop. I tested the committed `forward-port` on a StartOS 0.4.0 VM:
- installing clears the stale `[UNREPLIED]` entry, and the next packet is DNATed;
- a simulated restart (rules deleted without a flush, then installed again) keeps the live entry's conntrack id;
- teardown clears it.

A stub run confirms the arguments on both paths of both scripts. I updated the description's What and Testing sections to match.

One trade-off, same as on `master`: a StartOS daemon restart starts each service in a new container, which gets a fresh DHCP lease and so usually a new IP. Entries still DNATed to the old container reply from that old address, so this install-time delete leaves them in place. The earlier broad delete cleared them.

dr-bonez (approved):

